PR #2, opened 2025-03-20
collision check nconmax and disableflags

thowell:
- check `nconmax` and `disableflags` in `collision`
- add tests for `collision` and `disableflags` options


adenzler-nvidia (approved):
LGTM - thanks!

PR #3, opened 2025-03-20
check nv in put_model if dense

thowell:
check `nv` in `put_model` if jacobian="dense"

---
change the `n_humanoids.xml` solver to `CG`. 

the default solver option `Newton` will currently produce a shared memory error. for example:
```
Warning: Failed to configure kernel dynamic shared memory for this device, tried to configure _update_gradient__locals___cholesky_8922edc4_cuda_kernel_backward kernel for 106368 bytes, but maximum available is 101376
```
as a result, the humanoids will fall through the floor when simulated using the `Newton` solver. changing the solver to `CG` resolves this issue.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -13,33 +13,53 @@
 # limitations under the License.
 # ==============================================================================
 
+from typing import Optional
+
 import mujoco
 import numpy as np
 import warp as wp
+from packaging import version
 
 from . import support
 from . import types
 
 
 def put_model(mjm: mujoco.MjModel) -> types.Model:
-  if mjm.neq > 0:
-    raise NotImplementedError("Equality constraints are unsupported.")
+  # check supported features
+  for field, field_types, field_str in (
+    (mjm.actuator_trntype, types.TrnType, "Actuator transmission type"),
+    (mjm.actuator_dyntype, types.DynType, "Actuator dynamics type"),
+    (mjm.actuator_gaintype, types.GainType, "Gain type"),
+    (mjm.actuator_biastype, types.BiasType, "Bias type"),
+    (mjm.eq_type, types.EqType, "Equality constraint types"),
+    (mjm.sensor_type, types.SensorType, "Sensor types"),
+  ):
+    unsupported = ~np.isin(field, list(field_types))
+    if unsupported.any():
+      raise NotImplementedError(f"{field_str} {field[unsupported]} not supported.")
 
-  if mjm.nsensor > 0:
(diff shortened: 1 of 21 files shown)


adenzler-nvidia:
One option to avoid the error I'm currently using is to add enable_backwards = False to that cholesky kernel. But I'm ok with the change until we have a way to block the cholesky.

erikfrey:
There's a few things going on here:

1) We know we need to fix one of the remaining cholesky solves to use proper subtree tile sizes.
2) Even with that, we know there is some max subtree tile size after which our dense algos will fail.

Even with subtree tile sizes, if someone wants to simulate a 100-body whip we should probably raise an error in `io.py` and tell them to switch to sparse.

So my recommendation would be:

* Leave `n_humanoids.xml` as is
* Change `io.py` to throw an error if dense and `nv` > some max tile size

Then later when we have time:

* Fix the remaining cholesky to use subtree tile size instead of `nv` tile size
* With the fix, change `io.py` only throw an error if dense and `max subtree nv` > some max tile size

adenzler-nvidia:
My understanding was that the cholesky in the solver is not really tilable due to fill-in, but I really don't know the solver well enough. We should be able to do a blocked cholesky there though, but I haven't looked into it in detail.

erikfrey:
Ah yes, there is fill-in and it's dynamic per step.  Yes I wonder if there's some way of just doing a dense batched cholesky for arrays that are larger than the shared memory max?  I'm pretty sure jax cholesky on GPU just calls into cuSOLVER and that supports plenty large sizes.

thowell:
@erikfrey the pr is updated:
- undo change to n_humanoids.xml
- in `put_model` check `nv` if dense
- add test to io_test.py

mujoco_warp/_src/io.py
```diff
@@ -41,6 +41,11 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   if mjm.opt.solver not in set(types.SolverType):
     raise NotImplementedError(f"Solver: {mjm.opt.solver} is unsupported.")
 
+  # TODO(team): max dofs for dense
```
erikfrey: Hmm, I would guess this should say something like:

```
# TODO(team): remove this check after changing Newton gradient update cholesky to use smaller tile sizes
```

erikfrey (approved):
Looks great, just one nit

mujoco_warp/_src/io.py
thowell: todo comment updated

PR #5, opened 2025-03-20
solver warmstart

thowell:
the existing solver implementation initializes `qacc` with `qacc_warmstart`. 

this pr:
- if the warmstart `disableflags` option is not set, `solve` compares the costs associated with `qacc_smooth` and `qacc_warmstart` in order to determine the initialization for `qacc`
- if the warmstart `disableflags` option is set, `qacc` is initialized with `qacc_smooth`

benchmarking:
```
mjwarp-testspeed --function=solve --mjcf=humanoid/humanoid.xml --batch_size=8192 --iterations=1 --ls_iterations=5 --solver=newton
```

this pr:
```

 Total JIT time: 0.15 s
 Total simulation time: 8.48 s
 Total steps per second: 966,092
 Total realtime factor: 4,830.46 x
 Total time per step: 1035.10 ns
```

main:
```

 Total JIT time: 0.14 s
 Total simulation time: 7.91 s
 Total steps per second: 1,036,164
 Total realtime factor: 5,180.82 x
 Total time per step: 965.10 ns
```

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -346,7 +346,9 @@ def _constraint(mjm: mujoco.MjModel, nworld: int, njmax: int) -> types.Constrain
   efc.worldid = wp.zeros((njmax,), dtype=wp.int32)
 
   efc.Jaref = wp.empty(shape=(njmax,), dtype=wp.float32)
+  efc.Jaref_warmstart = wp.empty(shape=(njmax,), dtype=wp.float32)
   efc.Ma = wp.empty(shape=(nworld, mjm.nv), dtype=wp.float32)
+  efc.Ma_warmstart = wp.empty(shape=(nworld, mjm.nv), dtype=wp.float32)
   efc.grad = wp.empty(shape=(nworld, mjm.nv), dtype=wp.float32)
   efc.grad_dot = wp.empty(shape=(nworld,), dtype=wp.float32)
   efc.Mgrad = wp.empty(shape=(nworld, mjm.nv), dtype=wp.float32)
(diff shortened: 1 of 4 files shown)


erikfrey:
This is an interesting scenario!  It seems to suggest we should not follow the exact same behavior here as MuJoCo because the cost of the thread divergence is higher than the cost of the occasional extra solver step.  That's my read at least.

I would suggest holding off on this and revisiting after we implement logging and measuring solver statistics, because then we can get a really good intuition on how early returning relates to step perf and the tradeoffs for this kind of thread divergence.

erikfrey:
@thowell let's close this for now unless you're planning to come back to it soon?

thowell:
@erikfrey the pr is updated with a simplified warmstart option for the solver

erikfrey:
@thowell maybe in this same PR we should do some testing and compare solver iterations for MuJoCo Warp vs. MuJoCo C.

My expectation is that we are close to the number of solver iterations compared to MJC, let's say at most 1 more solver iteration for constraints that converge quickly, maybe a a few more for solutions that take a lot of iterations to converge quickly.  And that warmstart helps as much in MJWarp as it does in MJC.  Do you want to do that analysis over a few scenes?

Maybe as part of that analysis we should add printing solver stats to mjwarp-testspeed - WDYT?

erikfrey:
@thowell what's the status of this PR?

thowell:
@erikfrey the changes in this pr are now quite minimal, essentially just an option to initialize the solve `qacc` with `qacc_smooth` instead of `qacc_warmstart`. my preference would be to merge this pr as-is and then investigate solver iterations for difference scenarios in the future.

erikfrey (approved):

PR #6, opened 2025-03-20
Precompute valid geom collision pairs and implement contact include/exclude

thowell:
this pr precomputes valid geom collision pairs with the function `collision_driver.geom_pair`. additional filtering for contact exclude is included with this function. the resulting list of geom pairs is computed in `put_model` and is available as the field `nxn_geom_pair` in `Model`.

benchmarking:
```
mjwarp-testspeed --function=collision --mjcf=collision.xml --batch_size=8192
```

this pr:
```

 Total JIT time: 0.01 s
 Total simulation time: 0.12 s
 Total steps per second: 67,267,046
 Total realtime factor: 134,534.09 x
 Total time per step: 14.87 ns
```

main:
```

 Total JIT time: 0.01 s
 Total simulation time: 0.12 s
 Total steps per second: 65,986,259
 Total realtime factor: 131,972.52 x
 Total time per step: 15.15 ns
```


mujoco_warp/_src/collision_driver.py
```diff
@@ -24,6 +26,52 @@
...
+      and (weldid2 != 0)
+      and ((weldid1 == weld_parentid2) or (weldid2 == weld_parentid1))
+    )
+    mask = (contype1 & conaffinity2) or (contype2 & conaffinity1)
+    exclude = (bodyid1 << 16) + (bodyid2) in exclude_signature
+
+    if mask and (not self_collision) and (not parent_child_collision) and (not exclude):
+      pairs.append([geom1, geom2])
+
+  return np.array(pairs)
```
adenzler-nvidia: I would have expected this function in io.py - or to be running on the Model/Data structs we have in warp. Is this a pattern you're copying from MJX?

mujoco_warp/_src/collision_driver.py
```diff
@@ -462,20 +510,8 @@ def nxn_broadphase(m: Model, d: Data):
...
-    )
-    geom2 = (
-      elementid
-      + geom1
-      + 1
-      - m.ngeom * (m.ngeom - 1) // 2
-      + (m.ngeom - geom1) * ((m.ngeom - geom1) - 1) // 2
-    )
+    geom1 = m.collision_geom_pair[elementid][0]
+    geom2 = m.collision_geom_pair[elementid][1]
```
adenzler-nvidia: Thanks - I think it makes sense to just have the lookup here. Minor nitpick - My intuition tells me that it's probably better to load the vec2 into a variable and then split from that variable, that way the compiler can lower this to a wider load instruction. Not sure if this is going to help performance here, but it certainly won't hurt.

mujoco_warp/_src/collision_driver.py
```diff
@@ -24,6 +26,52 @@
...
+      - 2
+      - int(wp.sqrt(float(-8 * i + 4 * m.ngeom * (m.ngeom - 1) - 7)) / 2.0 - 0.5)
+    )
+    geom2 = (
+      i
+      + geom1
+      + 1
+      - m.ngeom * (m.ngeom - 1) // 2
+      + (m.ngeom - geom1) * ((m.ngeom - geom1) - 1) // 2
+    )
```
adenzler-nvidia: If this isn't going to run during sim - wouldn't it make more sense to just have the double loop to enumerate the pairs? Would make this easier much easier to understand.

mujoco_warp/_src/types.py
```diff
@@ -488,6 +488,7 @@ class Model:
     actuator_gear: scale length and transmitted force        (nu, 6)
     exclude_signature: body1 << 16 + body2                   (nexclude,)
     actuator_affine_bias_gain: affine bias/gain present
+    collision_geom_pair: valid collision pair geom ids       (<= ngeom * (ngeom - 1) // 2,)
```
adenzler-nvidia: I think it would be useful to somehow encode in the name or at least have a comment telling that this is only for the n2 broadphase.

mujoco_warp/_src/collision_driver.py
thowell: done

mujoco_warp/_src/types.py
thowell: `collision_geom_pair` -> `nxn_geom_pair`

mujoco_warp/_src/collision_driver.py
thowell: modified to use the indices from `numpy.triu_indices`

mujoco_warp/_src/collision_driver.py
thowell: mjx has a similar function `geom_pairs` in collision_driver.py https://github.com/google-deepmind/mujoco/blob/dc96ed6bdb67dfb9cc277977c99340dd71ea8d2d/mjx/mujoco/mjx/_src/collision_driver.py#L121

erikfrey:
Hmm!  This is an interesting design quesiton, I think there's a few pros and cons.  I'll list them out here later today and then we can discuss.

erikfrey:
OK!  Very interesting!

Pros of precomputing valid geom pairs:
* We can skip ever even considering invalid geom pairs, e.g. scenes with many visual geoms will be a bit faster because broadphase never even visits that pair
* `<pair>` include and exclude is a bit easier to insert/delete statically than insert/delete on the fly.  MuJoCo efficiently merges broadphase results and pair contacts by ensuring both are sorted, then it can do a linear merge, in order to ensure we don't have duplicate contacts.  If we do this dynamically we'll have to futz with it a bit to ensure it's fast.

Cons:
* We're reifying an array of geom pairs that is ngeom^2.  Pragmatically this is probably not a big deal, even at 10,000 geoms that's only like 500MB or so on device.  So this is not an enormous con.
* If user changes contype or conaffinity, we must now re-JIT the physics step, whereas the existing version has no such requirement.

I think I could go either way.  Curious what you all think.

erikfrey:
@thowell @adenzler-nvidia WDYT?

thowell:
i think the identified cons are minor compared to the pros for the precomputed geom pairs approach.

- based on the estimate of ~500MB for 10,000 geoms, the memory requirements for the precomputed geom pairs seem reasonable
- if the model is changed (eg, contype/conaffinity modified) i think it is reasonable to expect that the physics step should be recompiled

adenzler-nvidia:
I would like to do a deep-dive and see how the performance profile changes with this, but that's more for my own education..

I'm with Taylor on this one. Given that we have the (somewhat arbitrary) threshold that switches to SAP broadphase for large geom counts, I think it's ok to have this list as it will not be very large in many cases.

erikfrey:
OK!  Then I'll list two more design requests before starting to look carefully at the code:

1) Can we also add support for contact/pair in this PR?  This will help us fully address #115 and seems related enough, but let me know if you feel otherwise @thowell 

2) Can we reuse this array in the sweep and prune broadphase?  Pair and exclude are just tricky enough that we ought to reuse the same logic in both approaches if possible.

thowell:
@erikfrey yes will add support for contact pair to this pr

btaba:
OOC why is the broadphase happening on geom pairs rather than body pairs? In MJ we do broadphase on body pairs, midphase on geom pairs arising from body pairs, etc. I did this in the CUDA impl as well, but certainly a static geom-pair list is a good improvement already since it will avoid inter-body geom pairs.

erikfrey:
@btaba just for simplicity - you're right that in C MuJoCo, `mj_broadphase` does SAP at the body level and then compares bounding spheres for midphase - it's not clear that following these steps will be performant for GPU.  We already have a SAP broadphase but my understanding is sorting is currently a bottleneck, in practice we've found that n^2 bounding sphere test is performant for at least `ngeom` <= 100 - haven't tested higher yet.

btaba:
@erikfrey Thanks for context; I wasn't necessarily mentioning SAP in particular, just mentioning broadphase on bodies, midphase on geoms, rather than broadphase on geoms, no midphase. The full version I think would involve a CPU callback [something like this](https://github.com/btaba/mujoco/blob/06f5d7c0b0cfed27007cf19a2d45736edfd2e212/mjx/mujoco/mjx/_src/cuda/engine_collision_driver.cu.cc#L513-L514), but it might still be performant esp. if different parts of the mj_step graph can run in parallel (multiple streams), and it may scale better in the number of bodies/geoms

Just some food for thought

adenzler-nvidia:
It's always hard to have an intuition about performance, but my intuition is that you need a lot of potential geom pairs to justify the need for a midphase. Most of the time it's better to expose as much parallelism as possible, especially looking at potential future bigger GPUs. But of course only measuring this will tell us where the tradeoff is.

There are still optimizations on the table with n2 because in theory you only need to load echo geom once and can cache it in shared memory - that likely gives another nice boost.

Running with multiple streams is cool as well but I would only go there once we have a more or less complete port ready. These things are super hard to debug, and if you have good GPU utilization it likely isn't really helping a lot with performance.

thowell:
@erikfrey support for contact pairs has been added

mujoco_warp/_src/collision_driver.py
```diff
@@ -13,17 +13,71 @@
...
 from .collision_primitive import primitive_narrowphase
 from .types import MJ_MINVAL
 from .types import Data
 from .types import DisableBit
 from .types import Model
-from .types import vec5
 from .warp_util import event_scope
 
 
+def geom_pair(m: mujoco.MjModel) -> Tuple[np.array, np.array]:
```
erikfrey: Although we have logic like this in C MuJoCo's collision driver, it's being used quite differently in Warp.  I'd recommend putting this in `io.py` since it's part of model creation.

mujoco_warp/_src/collision_driver.py
```diff
@@ -13,17 +13,71 @@
...
+  filterparent = not (m.opt.disableflags & DisableBit.FILTERPARENT.value)
+  exclude_signature = set(m.exclude_signature)
+  predefined_pairs = [(g1, g2) for g1, g2 in zip(m.pair_geom1, m.pair_geom2)]
+
+  tri = np.triu_indices(m.ngeom)
+  ntri = tri[0].size
+
+  pairs = []
+  pairids = []
+  for i in range(ntri):
```
erikfrey: nit: consider instead:

```py
for geom1, geom2 in zip(*tri):
  ...
```

side benefit is you redefined `i` below which now you can avoid

mujoco_warp/_src/collision_driver.py
```diff
@@ -13,17 +13,71 @@
...
 from .types import DisableBit
 from .types import Model
-from .types import vec5
 from .warp_util import event_scope
 
 
+def geom_pair(m: mujoco.MjModel) -> Tuple[np.array, np.array]:
+  filterparent = not (m.opt.disableflags & DisableBit.FILTERPARENT.value)
+  exclude_signature = set(m.exclude_signature)
+  predefined_pairs = [(g1, g2) for g1, g2 in zip(m.pair_geom1, m.pair_geom2)]
```
erikfrey: I may have been tempted to do:

```py
predefined_pairs = {(m.pair_geom1[i], m.pair_geom2[i]): i for i in range(m.npair)}
```

then down below:

```py
pairid = predefined_pairs.get((geom1, geom2), -1)
pairid = predefined_pairs.get((geom2, geom1), pairid)
```

But feel free to ignore if you think it's overkill.

erikfrey (changes requested):
OK, this is looking good.  Besides a few nits, we just need to see if we can also plug this into SAP and then we're all set to merge.

mujoco_warp/_src/collision_driver.py
thowell: done

mujoco_warp/_src/collision_driver.py
thowell: done

mujoco_warp/_src/collision_driver.py
thowell: done

thowell:
@erikfrey `io.geom_pair` now generates `ngeom * (ngeom - 1) / 2` pair ids and geom pairs:
- `id == -2`: pair should be skipped during broadphase
- `id == -1`: pair is evaluated during broadphase and contact parameters are mixed
- `id > -1`: pair is evaluated during broadphase and contact parameters are specified via contact pair

`nxn_broadphase` and `sap_broadphase` both utilize `nxn_pairid` to filter geom pairs

erikfrey (approved):
OK - this ended up in a great place.  Thanks for you patience on the back and forth on this one.

PR #9, opened 2025-03-21
Avoid out-of-bounds accesses if nconmax and njmax are exceeded.

adenzler-nvidia:
We should probably find a way to correct the count once and avoid all the min() calls for nefc, but I couldn't find a good place for it. This at least fixes all the errors.


adenzler-nvidia:
ok going back on this, after the discussion in PR 11 becase we only need these checks for the ones with launch dimensions that can exceed this. Will update.

adenzler-nvidia:
closing this as all of this seems to be guarded by launch config. I was panicking because I had a similar issue in WIP code that had a different launch config.

PR #11, opened 2025-03-21
Fix for OOB access in zero_jv

adenzler-nvidia:


thowell:
can `efcid >= min(d.nefc[0], d.njmax)` be simplified to `efcid >= d.nefc[0]` since `_zero_jv` is launched with `njmax`?

mujoco_warp/_src/solver.py
```diff
@@ -742,6 +742,9 @@ def _linesearch(m: types.Model, d: types.Data):
   def _zero_jv(d: types.Data):
     efcid = wp.tid()
 
+    if efcid >= min(d.nefc[0], d.njmax):
+      return
+
     if wp.static(m.opt.iterations) > 1:
       if d.efc.done[d.efc.worldid[efcid]]:
```
erikfrey, line 749: Ohhh, OK.  Good catch.  It's because of this line isn't it:

```
if d.efc.done[d.efc.worldid[efcid]]:
```

`efc.worldid` will give you an invalid index for `efcid > d.nefc[0]`.

@adenzler-nvidia should we move your check into this if block?

More generally speaking, @thowell you're pretty sure we can't just zero out the entire array each time without impacting perf?

adenzler-nvidia: Yeah I didn't look at the launch config, this can just be a nefc[0] check.

adenzler-nvidia: should be fixed now.

I'd vote for leaving as-is w.r.t having a kernel vs zeroing the entire array - we're one bad njmax config away from this showing up on a profile.

erikfrey: OK!

PR #12, opened 2025-03-21
dof frictionloss

thowell:
adds support for dof frictionloss


mujoco_warp/_src/solver.py
```diff
@@ -101,18 +101,31 @@ def _efc_kernel(d: types.Data):
...
-      d.efc.force[efcid] = -1.0 * efc_D * Jaref
+    # cost = 0.5 * sum(efc_D * Jaref * Jaref * active)
+    cost = 0.5 * efc_D * Jaref * Jaref
 
-      # cost = 0.5 * sum(efc_D * Jaref * Jaref * active))
-      wp.atomic_add(d.efc.cost, worldid, 0.5 * efc_D * Jaref * Jaref)
+    # efc_force = -efc_D * Jaref * active
+    efc_force = -1.0 * efc_D * Jaref
+
+    if wp.static(m.dof_frictionloss_adr.size) > 0 and (not dsbl_floss):
```
adenzler-nvidia: could you make the disable_floss parts wp.static as well? Then we can avoid compiling that code alltogether.

mujoco_warp/_src/solver.py
```diff
@@ -101,18 +101,31 @@ def _efc_kernel(d: types.Data):
...
 
-    if active:
-      # efc_force = -efc_D * Jaref * active
-      d.efc.force[efcid] = -1.0 * efc_D * Jaref
+    # cost = 0.5 * sum(efc_D * Jaref * Jaref * active)
+    cost = 0.5 * efc_D * Jaref * Jaref
 
-      # cost = 0.5 * sum(efc_D * Jaref * Jaref * active))
-      wp.atomic_add(d.efc.cost, worldid, 0.5 * efc_D * Jaref * Jaref)
+    # efc_force = -efc_D * Jaref * active
```
adenzler-nvidia: are these comments outdated or does this need an `if active`?

mujoco_warp/_src/solver.py
```diff
@@ -101,18 +101,31 @@ def _efc_kernel(d: types.Data):
...
+      floss_cost = float(linear_neg) * (-0.5 * r * f * f - f * Jaref)
+      floss_cost += float(linear_pos) * (-0.5 * r * f * f + f * Jaref)
     else:
-      d.efc.force[efcid] = 0.0
+      floss_cost = 0.0
+      floss_force = 0.0
+
+    d.efc.active[efcid] = int(active)
+    d.efc.force[efcid] = float(active) * efc_force + floss_force
+    wp.atomic_add(d.efc.cost, worldid, float(active) * cost + floss_cost)
```
adenzler-nvidia: Oh I think I see what you did. Did you benchmark the choice of multiplying with 0 vs if-guarding the calculations and memory writes?

I think this warrants some specializations. For example, I would avoid calculation efc_force alltogether if the constraint is not active. I would also avoid the atomic add if not active, but that one needs to be benchmarked for sure.

mujoco_warp/_src/solver.py
```diff
@@ -376,11 +412,18 @@ def _init_p0(p0: wp.array(dtype=wp.vec3), d: types.Data):
...
-    # TODO(team): active and conditionally active constraints:
-    if d.efc.Jaref[efcid] >= 0.0:
+    Jaref = d.efc.Jaref[efcid]
+    if (Jaref >= 0.0) and (efcid >= d.ne[0] + d.nf[0]):
       return
 
     quad = d.efc.quad[efcid]
+    jv = d.efc.jv[efcid]
+
+    if wp.static(m.dof_frictionloss_adr.size) > 0 and (not dsbl_floss):
```
adenzler-nvidia: same thing with wp.static here - the entire if condition can be a wp.static.

adenzler-nvidia (commented):
Left some comments. I'm not 100% it's not already doing the right thing where I suggested to use wp.static, so let's verify.

mujoco_warp/_src/solver.py

mujoco_warp/_src/solver.py
thowell: removed these comments

mujoco_warp/_src/solver.py

mujoco_warp/_src/solver.py
thowell: added a todo to benchmark different options for active

thowell:
added a check for `tendon_frictionloss` in `put_model`

mujoco_warp/_src/solver.py
```diff
@@ -101,18 +103,39 @@ def _efc_kernel(d: types.Data):
...
+      floss_cost += linear_posf * (-0.5 * rf + Jaref)
+
+      cost = floss_cost
+      efc_force = floss_force
     else:
-      d.efc.force[efcid] = 0.0
+      cost = 0.0
+      efc_force = 0.0
+
+    # TODO(team): benchmark options for active
```
erikfrey: what does this TODO mean?

mujoco_warp/_src/solver.py
```diff
@@ -101,18 +103,39 @@ def _efc_kernel(d: types.Data):
...
-    active = int(Jaref < 0.0)
-    d.efc.active[efcid] = active
+    active = (Jaref < 0.0) or (efcid < d.ne[0] + d.nf[0])
+
+    if wp.static(m.dof_frictionloss_adr.size > 0 and (not DSBL_FLOSS)):
+      f = d.efc.frictionloss[efcid]
+      r = _safe_div(1.0, efc_D)
+      rf = r * f
+      f_pos = f > 0.0
+      linear_neg = (Jaref <= -rf) and f_pos
```
erikfrey: This looks like a JAX-y no-if version of the code.  I'm thinking Warp can let us do something a bit more readable:

```py
if Jaref <= -rf:
  cost = -0.5 * rf * f - f * Jaref
  efc_force = f
elif Jaref >= rf:
  ...
```

Benchmark it though to make sure it doesn't slow down the step!

mujoco_warp/_src/solver.py
```diff
@@ -101,18 +103,39 @@ def _efc_kernel(d: types.Data):
     Jaref = d.efc.Jaref[efcid]
     efc_D = d.efc.D[efcid]
 
-    # TODO(team): active and conditionally active constraints
-    active = int(Jaref < 0.0)
-    d.efc.active[efcid] = active
+    active = (Jaref < 0.0) or (efcid < d.ne[0] + d.nf[0])
```
erikfrey: This looks like the JAX-y no-if version of this code, I'm wondering if Warp can let us do:

```py
ne = d.ne[0]
nf = d.nf[0]
if efcid < ne:
  # equality constraints
  pass
elif efcid < ne + nf:
  # friction
  ...
else:
  # contact
  active = Jaref < 0.0
  ...
```

Benchmark it though to make sure it doesn't slow down the step!

erikfrey (changes requested):
Couple comments to start, looking good though!

mujoco_warp/_src/solver.py
thowell: this todo is referring to benchmarking the performance of multiplying by `active` versus `active` in an if statement.

the code is updated to if statements and this todo has been removed.

mujoco_warp/_src/solver.py
thowell: updated the code to use if statements

mujoco_warp/_src/solver.py
thowell: updated the code to use if statements

thowell:
changed `active` type to bool

mujoco_warp/_src/solver.py
thowell: implemented if statements (and removed the todo)

mujoco_warp/_src/solver.py
```diff
@@ -101,18 +103,50 @@ def _efc_kernel(d: types.Data):
...
+      # friction
+      active = True
+    else:
+      # limits, contact
+      active = Jaref < 0.0
+
+    cost = 0.0
+    force = 0.0
+
+    if wp.static(m.dof_frictionloss_adr.size > 0 and (not DSBL_FLOSS)):
```
erikfrey: Let's try to simplify this further.  Unless we have a compelling benchmark that shows a slowdown, we should really make code as dumb as possible.  To that end:

* I'm not sure we need this `dof_frictionloss_adr` in Warp.  These static variables surprise users if they change `dof_frictionloss` down the line and don't recalculate `dof_frictionloss_adr`.  And we already know we're handling frictionloss when `efcid >= ne && efcid < ne + nf`
* Unless we have a benchmark that shows a significant slowdown, I think it would be fine to move the logic for each constraint type into the first if block.

So:

```py
if efcid < ne:
  # equality (not implemented yet)
  pass
elif wp.static(not DSBL_FLOSS) and efcid < ne + nf:
  # friction loss
  floss = d.efc.frictionloss[efcid]
  rfloss = _safe_div(floss, efc_D)
  if Jaref <= -rfloss:
    # linear negative
    wp.atomic_add(d.efc.cost, worldid, -0.5 * rfloss - Jaref)
    d.efc.force[efcid] = floss
    d.efc.active[efcid] = False
  elif Jaref >= rfloss:
    # linear positive
    wp.atomic_add(d.efc.cost, worldid, -0.5 * rfloss + Jaref)
    d.efc.force[efcid] = -floss
    d.efc.active[efcid] = False
  else:
    # quadratic
    wp.atomic_add(d.efc.cost, worldid, 0.5 * efc_D * Jaref * Jaref)
    d.efc.force[efcid] = - efc_D * Jaref
    d.efc.active[efcid] = True
else:
  # limits, contact
  ...
```

Basically let's be really dumb and spell it out and then only optimize if we see the dumb version is a problem.

mujoco_warp/_src/solver.py
```diff
@@ -376,11 +426,18 @@ def _init_p0(p0: wp.array(dtype=wp.vec3), d: types.Data):
...
+    if (Jaref >= 0.0) and (efcid >= d.ne[0] + d.nf[0]):
       return
 
     quad = d.efc.quad[efcid]
+    jv = d.efc.jv[efcid]
+
+    if wp.static(m.dof_frictionloss_adr.size > 0 and (not DSBL_FLOSS)):
+      floss = d.efc.frictionloss[efcid]
+      efc_D = d.efc.D[efcid]
+      quad = _quad_floss(quad, floss, efc_D, Jaref, jv)
```
erikfrey: it seems like you're calculating this _quad_floss a lot - is it possible to just store this once in `efc.quad` when calling _init_quad?

mujoco_warp/_src/solver_test.py
```diff
@@ -325,6 +325,33 @@ def cost(m, d, qacc):
         "efc_force2",
       )
 
+  def test_frictionloss(self):
```
erikfrey, line 359: this is decent, but also consider adding back frictionloss to `constraints.xml` as it is in MJX - that might test more edge cases.

erikfrey (changes requested):
Getting there - everything should be made as simple as possible but not simpler.

mujoco_warp/_src/solver_test.py
thowell: dof frictionloss is restored to `constraints.xml`

thowell: `test_frictionloss` now tests the first two keyframes from `constraints.xml`.

note: the tests do not pass with the third keyframe with or without frictionloss. added a todo for testing keyframe 2.

mujoco_warp/_src/solver.py
thowell: this functionality has been moved to `_init_quad`

mujoco_warp/_src/solver.py
thowell: `dof_frictionloss_adr` has been removed

erikfrey (approved):
Looks great, let's merge after resolving conflicts.

PR #13, opened 2025-03-21
set random seed in test_forward to avoid test flakyness for test_fwd_acceleration

adenzler-nvidia:
I need someone to confirm this is the right thing to do. I had frequent failures before that were barely above the tolerance - none anymore.

The differences look like they are originating in the cholesky factorization. I compared in/outputs and my current best guess is that we are hitting numerical instabilities that make mujoco C and mjwarp diverge because of the float/double difference. But I'm happy to be proven wrong. Is there a way to use mujoco C with floats without recompiling?

### DIFF
--- a/mujoco_warp/_src/forward_test.py
+++ b/mujoco_warp/_src/forward_test.py
@@ -44,6 +44,7 @@ class ForwardTest(absltest.TestCase):
     mjm.opt.jacobian = is_sparse
     mjd = mujoco.MjData(mjm)
     mujoco.mj_resetDataKeyframe(mjm, mjd, 1)  # reset to stand_on_left_leg
+    np.random.seed(42)
     mjd.qvel = np.random.uniform(low=-0.01, high=0.01, size=mjd.qvel.shape)
     mjd.ctrl = np.random.normal(scale=1, size=mjd.ctrl.shape)
     mujoco.mj_forward(mjm, mjd)
--- a/mujoco_warp/_src/forward_test.py
+++ b/mujoco_warp/_src/forward_test.py
@@ -44,7 +44,6 @@ class ForwardTest(absltest.TestCase):
     mjm.opt.jacobian = is_sparse
     mjd = mujoco.MjData(mjm)
     mujoco.mj_resetDataKeyframe(mjm, mjd, 1)  # reset to stand_on_left_leg
-    np.random.seed(42)
     mjd.qvel = np.random.uniform(low=-0.01, high=0.01, size=mjd.qvel.shape)
     mjd.ctrl = np.random.normal(scale=1, size=mjd.ctrl.shape)
     mujoco.mj_forward(mjm, mjd)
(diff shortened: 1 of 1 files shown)


erikfrey:
Yeah good question!  These are tricky tests to get right.  To your question about MuJoCo, nope - only way to run it in 32 bit is to recompile it.  That said, that might make the divergence worse, not better.

The real lever we have here is `_TOLERANCE` and we want to find the magic window inside of which it catches real numerical bugs and doesn't false positive and flake.

If you look at MJX's [forward_test.py](https://github.com/google-deepmind/mujoco/blob/main/mjx/mujoco/mjx/_src/forward_test.py#L28) you'll see its tolerance is actually lower than our test, but I see no reason that MJWarp should not be able to have just as tight a tolerance.  I suspect the real culprit here is either:

- an actual bug, OR
- something we're doing in the loading of the fixture that creates a large force or jerk of some kind in step 0 of that loaded keyframe OR
- we shouldn't be calling `factor_m` when all we are testing is `fwd_acceleration`, we should rely on the `qLD` that came from CPU

Anyway this is a pretty long winded answer, I know you just want the flakiness to go away        :-)      if you don't want to look into this let me know and one of us can take it.

adenzler-nvidia:
Thanks - I think I chased this long enough without luck.. so if you have ideas feel free.

To me it really looked like some creeping very small differences in qLD that are amplified by the fact that we are somehow solving a system this is not very stable numerically. But I don't know enough about the algorithms yet. I wasn't able to pinpoint a clear location where things deviate significantly.

It would be useful of course to just use the factorization from mujoco-C - which I think we can't do as-is right now because in the dense version the factorizations are not exactly the same. But if we assume the factorizations are correct, we can just only test sparse and take qLD from mujoco. Let me try that on monday.

erikfrey:
OK I played with this, here's what I found:

* I can definitely replicate the flakes on CPU and also my 4090
* If I remove `mjwarp.factor_m` the flakes go away
* I can lower `_TOLERANCE` to `1e-5` and still no flakes

So I think my recommendation is to remove `mjwarp.factor_m` from the test - it's not what we want to test here anyway.

Probably we added `factor_m` to this test back before we were doing factorization in `io.py` - you're right that `qLD` is the wrong format, but we use CPU numpy cholesky (rather than Warp cholesky) in `io.py` which is 64 bit so it produces a better factorization than GPU cholesky.  Probably this is why the test flakes go away if we rely on that factorization instead.

Let me know if you can repro the fix on your end and are satisfied with my suggestion to just remove `factor_m` from the test?

adenzler-nvidia:
Thanks for looking into this - I agree we can just remove the factor_m call. I did lower the tolerance in forward_test as well but that seems to break other tests using cholesky factorization. So we can either lower just for that specific test or just call it a day.

Otherwise no more flakyness, thanks again. I updated the PR to remove the seed again and remove the factor_m call.

erikfrey (approved):
LGTM!

PR #14, opened 2025-03-21
fix fwd_actuation dense

thowell:


mujoco_warp/_src/forward.py
```diff
@@ -627,9 +627,7 @@ def qfrc_actuator_kernel(
     for i in range(len(qderiv_tileadr)):
       beg = qderiv_tileadr[i]
       end = (
-        m.actuator_moment_tileadr.shape[0]
-        if i == len(qderiv_tileadr) - 1
-        else qderiv_tileadr[i + 1]
+        m.qLD_tile.shape[0] if i == len(qderiv_tileadr) - 1 else qderiv_tileadr[i + 1]
```
erikfrey, line 630: I'm not entirely following this change... it looks like you did `s/actuator_moment_tileadr/qLD_tile` - but why?

My first impression is this change is incorrect because it deviates from our usual pattern of going through each beg, end in the array with the final end just being the shape of the array.  But I'm probably missing something.

erikfrey (changes requested):
Not quite sure what this is fixing!

mujoco_warp/_src/forward.py
thowell: `m.qLD_tile.shape[0]` is the number of dof islands. it is incorrect to use `m.actuator_moment_tileadr.shape[0]` here because the number of addresses can be less than the number of islands in cases where there are islands that have the same number of dofs.

adenzler-nvidia: I'm also not following - but this is already too long ago for me to just remember everything. Is the existing unit test broken, or how did you find this issue?

I kind of see your point but then I had to duplicate things exactly for the reason that there could be islands with the same number of dofs, but different numbers of actuators. Otherwise we could just take the addresses from the qLD data.

But I need to go back and look at some code on Monday for sure.

thowell: here is the reference implementation: https://github.com/google-deepmind/mujoco_warp/blob/370942d2a5e5639c71e39f927ff3aa0f1232f30b/mujoco_warp/_src/forward.py#L382

this pr updates fwd_actuation to match the reference implementation pattern

adenzler-nvidia: oooh I see it - for some reason I was thinking that this is in the integration code at the time I wrote that last comment.

I do think it makes sense for both implementations to have the same code there. And I think it checks out, we do not have an actuator specific value there and the actual number of tiles is the same. We just have different groupings because not all of the tiles with the same number of dofs have the same number of actuators, which means we can't process them together.

If you don't mind I'm holding off with approving until monday because it's late here and I feel like my brain is not at 100% capacity anymore, so I shouldn't be approving MRs anymore. But if @erikfrey agrees, he can approve as well.

thowell: the issue can be visualized by simulating the leap hand (https://github.com/google-deepmind/mujoco_menagerie/tree/main/leap_hand) with jacobian="dense"

erikfrey (changes requested):
Whew OK, I think I get it too.  Thanks @thowell for bearing with us as we remind ourselves about what's going on here. Couple of thoughts:

* crafting the tiling arrays is getting just gnarly enough that we should come through and clean them up soon
* I have a sneaking suspicious we will decide soon that we want `Model` and `Data` to become generic classes, not `wp.struct`
* When we do, we might more easily express tile groupings as lists or tuples of a struct that just has tile size (int) and tile ids (wp.array)
* Then some of this stuff will be a little cleaner to write and reason about

so something like:

```
class TileSet:
  size: tuple[int, ...]
  adr: wp.array(dtype=int)

class Model:
  qLD_tileset: tuple[TileSet, ...]
```

Something like that.  In the meantime, just so one of us doesn't mess this up again, can you add a simple test that goes from red to green as a result of your change?  Then we'll merge.

thowell:
updated forward_test.py and added actuation.xml.

note: I think there is an additional bug with fwd_actuation (dense). for example, simplifying actuation.xml by removing the third body + third joint + associated actuator from each branch will result in a test failure for qfrc_actuator fwd_actuation_test.

```
<mujoco>
  <worldbody>
    <body pos="0 0 0">
      <joint name="hinge0" type="hinge" axis="1 0 0"/>
      <geom type="sphere" size=".1"/>
      <body>
        <joint name="hinge1" type="hinge" axis="1 0 0"/>
        <geom type="sphere" size=".1"/>
        <!-- <body>
          <joint name="hinge2" type="hinge" axis="1 0 0"/>
          <geom type="sphere" size=".1"/>
        </body> -->
      </body>
    </body>
    <body pos="1 0 0">
      <joint name="hinge3" type="hinge" axis="1 0 0"/>
      <geom type="sphere" size=".1"/>
      <body>
        <joint name="hinge4" type="hinge" axis="1 0 0"/>
        <geom type="sphere" size=".1"/>
        <!-- <body>
          <joint name="hinge5" type="hinge" axis="1 0 0"/>
          <geom type="sphere" size=".1"/>
        </body> -->
      </body>
    </body>
  </worldbody>
  <actuator>
    <motor joint="hinge0"/>
    <motor joint="hinge1"/>
    <!-- <motor joint="hinge2"/> -->
    <motor joint="hinge3"/>
    <motor joint="hinge4"/>
    <!-- <motor joint="hinge5"/> -->
  </actuator>
</mujoco>
```

adenzler-nvidia (approved):
Thanks for the test - approving.

adenzler-nvidia:
could you open an issue for the other bug (and assign to yourself if you're looking into it, otherwise I can also assign someone else)?

PR #15, opened 2025-03-21
Update `.gitignore`

hartikainen:
I played around with the code and noticed some files and folders (specifically `.venv` and `MUJOCO_LOG.TXT`) that could be git-ignored.

Here's the default python `.gitignore`, based on the [github's templates](https://github.com/github/gitignore), that I typically use in my python projects.


erikfrey (approved):
Thanks Kristian!

PR #16, opened 2025-03-21
island option

thowell:
an argument `island` is added to `put_model`. `island` determines if tile operations use multiple tiles (`island=True`) or one tile (`island=False`). additionally, the flag `island` is added to the viewer.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -21,7 +21,7 @@ from . import support
 from . import types
 
 
-def put_model(mjm: mujoco.MjModel) -> types.Model:
+def put_model(mjm: mujoco.MjModel, island: bool = True) -> types.Model:
   if mjm.neq > 0:
     raise NotImplementedError("Equality constraints are unsupported.")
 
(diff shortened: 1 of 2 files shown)


erikfrey:
What's the use case for allowing users to change the tile behavior?

thowell:
cuda graph capture failure can occur for scenes where individual islands have different numbers of dofs. 

for example, this xml
```
<mujoco>
  <worldbody>
    <body pos="-1 0 0">
      <joint type="slide" axis="1 0 0"/>
      <geom type="sphere" size=".1"/>
    </body>
    <body pos="1 0 0">
      <joint type="slide" axis="1 0 0"/>
      <joint type="slide" axis="0 1 0"/>
      <geom type="sphere" size=".1"/>
    </body>
  </worldbody>
</mujoco>
```
produces
```
raise RuntimeError(f"CUDA graph capture failed. {runtime.get_error_string()}")
RuntimeError: CUDA graph capture failed. Warp CUDA error 900: operation not permitted when stream is capturing (in function cuda_load_module, /builds/omniverse/warp/warp/native/warp.cu:3417)
```

when loaded in the viewer.

the scene loads in the viewer when `island=False`.

Warp 1.7.0.dev20250321
CUDA Toolkit 12.8, Driver 12.2

adenzler-nvidia:
I do think we need a way to switch between one island and multiple, there is probably a corner in that space where it doesn't make sense to split things. That being said - I'm surprised that xml above doesn't work. I thought the entire point of that tiling was to exploit block-sparsity and batch by size to just run a kernel for each size?

erikfrey:
Hmm, this model works OK for me, this command succeeds:

```
mjwarp-testspeed --function=step --mjcf=taylor_example.xml --batch_size=8192 --iterations=2 --ls_iterations=10 --solver=newton
```

thowell:
@erikfrey after upgrading cuda and upgrading the driver the test xml above works

thowell:
@adenzler-nvidia @erikfrey should i close this pr or do we want to incorporate this feature?

erikfrey:
Let’s close it for now but can always come back if we think we need this flag.

PR #18, opened 2025-03-21
Fix ruff error.

erikfrey:


adenzler-nvidia:
Thanks - should we start blocking MRs that don't pass formatting?

PR #20, opened 2025-03-22
Optimize _JTDAJ

sivahari:
To improve the efficiency of the `_JTDAJ` kernel, we have reduced the number of launched CTAs. Profiling indicated that the original kernel, launched with `dim_x = d.njmax` blocks, resulted in the execution of many blocks with no effective work. The optimized kernel now launches a number of blocks proportional to the GPU's SM count.

While running `mjwarp-testspeed --function=step --mjcf=humanoid/humanoid.xml --batch_size=8192` on an RTX A5000 at 1695MHz, the optimization reduced the kernel time spent on this kernel from 33.6% to 1.4%. These numbers are obtained from the CUDA GPU kernel summary page from the nsys reports.

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -310,19 +310,22 @@ def _update_gradient(m: types.Model, d: types.Data):
     else:
       wp.launch(_copy_lower_triangle, dim=(d.nworld, m.dof_tri_row.size), inputs=[m, d])
 
-    # Optimization: launching _JTDAJ with limited number of blocks. Profiling
-    # suggests that only a fraction of blocks out of the original d.njmax blocks
-    # do the actual work. It aims to minimize #CTAs with no effective work. It
-    # launches with #blocks that's proportional to the number of SMs on the GPU.
-    # We can now query the SM count:
+    # Optimization: launching _JTDAJ with limited number of blocks on a GPU.
+    # Profiling suggests that only a fraction of blocks out of the original
+    # d.njmax blocks do the actual work. It aims to minimize #CTAs with no
+    # effective work. It launches with #blocks that's proportional to the number
+    # of SMs on the GPU. We can now query the SM count:
     # https://github.com/NVIDIA/warp/commit/f3814e7e5459e5fd13032cf0fddb3daddd510f30
-    sm_count = wp.get_device().sm_count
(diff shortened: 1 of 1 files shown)


adenzler-nvidia:
I did play with max_count sms quickly. Using the existing implementation, I can get down to about 1% of total runtime using max_count == 20 * sm_count. I feel like I'm overfitting to the benchmark here though. @sivahari  did you do a parameter sweep on the number of blocks? My intuition tells me that your solution is probably sensitive to that as well, so you can likely get it lower as well. On my a6000 we're in the 5us average range though, so in many cases we're just looking at kernel launch overhead. Worst-case is sitting at roughly 30us.

Conceptually I like this solution more than using max_count because it avoids even launching the warp logical threads that won't be doing any work - while using max_count I'm assuming we still have to go all the way through the grid-stride loop to check whether all of the other threads need to be doing any work.

erikfrey:
Very cool, taking notes over here!  :notebook:  I will leave this one to you two to sort out.

Separately though I wonder if we should be treating other kernel launches the same way!

adenzler-nvidia:
We do have a warp language improvement in flight as well that aims to improve this for all kernels - but the explicit coarsening is still a bit faster so we'll likely use that technique where it makes sense/becomes dominant.

sivahari:
Thanks for the question. I swept the multiplication factor after the message and found little difference by going from 6 to 20 in my tests. For other kernels, increasing the factor might help minimize the tail effects as the time per wave will reduce. The optimal factor will depend on the kernel's occupancy (which depends on register and shared memory usage) and how many blocks can simultaneously run on the SM. For this one, I think the factor can be left as it is with a comment in the code summarizing this discussion, which I added in my latest commit.

adenzler-nvidia (approved):
Looks good, thanks for sweeping the parameter and for all the explanations!

erikfrey (changes requested):
Looks like we need some kind of special casing for CPU?  Otherwise this looks ready to merge.

sivahari:
Added a check before applying this optimization using ` if wp.get_device().is_cuda`. Thanks for that suggestion. 

I mistakenly clicked the close button. Reopened it.

adenzler-nvidia:
Thanks - we're almost there, seems to be another formatting issue.

PR #22, opened 2025-03-22
Testspeed fixes and improvements.

erikfrey:
* Fix bug where event tracing failed if trace had multiple nested substacks.
* Change testspeed behavior to defer to model params unless overridden by commandline params.
* Remove `--base_path` testpeed param and instead use this behavior: 1) try default path 2) try resource path
* Remove double JIT from benchmark invocation, seems to be no longer needed.
* Add a few docstrings.


adenzler-nvidia (approved):
Looks good, thank you!

PR #27, opened 2025-03-25
Added preliminary implementation of Gjk

Kenny-Vilella:
Implementation was made by @eric-heiden  and @nvtw, and I made the integration. Only ported the sparse version.
It seems to not fully work.
When I made some columns of sphere collapse on a plane, after a while the spheres will all roll in the same direction.
But since it is still an alpha, it is maybe better to merge it in its current state to let people have better collision support and try to fix the issue in parallel.


gavrielstate:
Can we put the GJK solver code into a separate directory so that we can move it into the public Newton repo when ready so it can be used by other Newton solvers?

adenzler-nvidia:
@erikfrey (and Kyle, cannot tag you so far) - I see Kyle is assigned to GJK. What are your plans with this implementation? Are you planning on iterating on that or will the be another (convex mesh only?) implementation?

kbayes:
@adenzler-nvidia, I have an GJK/EPA implementation that's used in MuJoCo CPU that's extremely robust and fast with no dependencies, that I'm going to try porting over. Moreover I'm going to merge in whatever good parts from the CUDA version that's been floating around.

adenzler-nvidia:
Hi Kyle - looking forward to that GJK implementation, I heard it's quite famous :-) A few questions:

* What's your estimation on a timeline? Would it make sense to submit this PR in the mean time?
* How does the implementation differ from that CUDA code - was that not a port of MuJoCo C as well?
* Does that GJK only cover mesh-mesh or does it also include support for some other collision types?

erikfrey:
Hey all, recap of our discussion this morning:

- We won't have multiple GJKs, @kbayes will improve or replace parts of this one
- We should definitely get this in and not block on Kyle - he'll probably start out by porting over some correctness tests that he developed for his CPU GJK
- Yep, his GJK/EPA supports geom types besides meshes

That said, even if want to get this merged, there's an important design question before we proceed:

In MuJoCo we let the collision driver decide what geom type pairs should be handled by GJK/EPA rather than a user flag that switches it on for everything.  I think the former is preferable for three reasons:

1) One less flag for the users
2) I probably want to use GJK/EPA for mesh and primitive for sphere<>sphere, having a single `enableGjk` param would seem to disallow that
3) We'll probably get fewer edge cases and numerical instability issues if we minimize what GJK/EPA is used for

So before we proceed my recommendation would be for someone to have a go at removing the `enableGjk` flag and instead explicitly calling it for pair types as we do in MuJoCo, e.g. ellipsoid<>sphere, ellipsoid<>capsule, ... cylinder<>capsule, cylinder<>ellipsoid, ... mesh<>mesh.  Basically any reference to `mj_Convex` in the function map here:

https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_collision_driver.c#L41

@gavrielstate Let's make a directory in here called `collision` so we'd have:

```
_src/collision/convex.py
_src/collision/driver.py
_src/collision/primitive.py
...
```

Can do that in either this or a separate PR.

Kenny-Vilella:
@erikfrey Thank you for the recap.

I have other things to do this week, but I should be able to get back to it on Monday.
If it is actively blocking other dev works, then someone should take over.

adenzler-nvidia:
@Kenny-Vilella - this has not been properly reviewed yet anyway, right? Then we can focus on reviewing the actual functionality of the code now, and decide which parts of the code organization and settings items we can put into additional issues/tasks.

kbayes:
Thanks @erikfrey for the recap from Monday's meeting.  Here's some extra info:

-  I'm on vacation until the end of the week, and plan to import a strip down GJK on Monday when I get back. Please feel free to submit whatever you like in the meantime (it'll give me something to benchmark off of).
- For some quick models to test against and benchmark, see the directory [test/engine/testdata/collision_convex](https://github.com/google-deepmind/mujoco/tree/main/test/engine/testdata/collision_convex) in MuJoCo.
- The C implementation GJK/EPA in MuJoCo and CUDA GJK take different approaches. I pulled out some of the good parts of the CUDA implementation for some speedups in the C version but that's about as much as they have in common.
- The C implementation is nice if you want support for geom distances, and that's low hanging fruit to support on GPU. See [here](https://mujoco.readthedocs.io/en/latest/XMLreference.html#sensor-fromto) for sensors in MuJoCo.
- Any convex geom with a support function can be supported in GJK/EPA. Moreover, capsule and sphere collisions with other geoms are extremely cheap (1 - 2 iterations) as GJK can recover penetration depth.

gavrielstate:
> Let's make a directory in here called collision so we'd have:
> 
> ```
> _src/collision/convex.py
> _src/collision/driver.py
> _src/collision/primitive.py
> ...
> ```

That works, and tracks with our discussion of eventually moving to a geometry library.

adenzler-nvidia:
Thanks Kyle - we definitely don't want to hurry this or put you under any kind of stress that would impact the quality of your work. So let's review this and then let you decide how to handle combining the two versions or flipping a switch depending on what works best for you to develop things.

mujoco_warp/_src/collision_primitive.py
```diff
@@ -45,6 +46,13 @@ def _geom(
   geom.rot = rot
   geom.size = m.geom_size[gid]
   geom.normal = wp.vec3(rot[0, 2], rot[1, 2], rot[2, 2])  # plane
+  dataid = m.geom_dataid[gid]
+  if dataid >= 0:
+    geom.vertadr = m.mesh_vertadr[dataid]
+    geom.vertnum = m.mesh_vertnum[dataid]
+  else:
+    geom.vertadr = 0
+    geom.vertnum = 0
```
adenzler-nvidia: is the 0 here an existing convention? I would rather have some thing like -1 or even INT_MAX because 0 can be a valid address.

mujoco_warp/_src/types.py
```diff
@@ -164,11 +164,17 @@ class GeomType(enum.IntEnum):
...
+  ELLIPSOID = mujoco.mjtGeom.mjGEOM_ELLIPSOID
+  CYLINDER = mujoco.mjtGeom.mjGEOM_CYLINDER
   BOX = mujoco.mjtGeom.mjGEOM_BOX
-  # unsupported: HFIELD, ELLIPSOID, CYLINDER, MESH,
+  MESH = mujoco.mjtGeom.mjGEOM_MESH
+  # unsupported: HFIELD,
   # NGEOMTYPES, ARROW*, LINE, SKIN, LABEL, NONE
 
 
+NUM_GEOM_TYPES = 8
```
adenzler-nvidia, line 201: this can be len(GeomType)

mujoco_warp/_src/collision_convex.py
```diff
@@ -15,12 +15,806 @@
...
+
+vec6 = wp.types.vector(6, dtype=int)
+VECI1 = vec6(0, 0, 0, 1, 1, 2)
+VECI2 = vec6(1, 2, 3, 2, 3, 3)
+
+
+@wp.func
+def sign(x: float):
+  # XXX we have to match the sign function from CUDA here
+  return wp.where(x < 0.0, -1.0, 1.0)
```
adenzler-nvidia: https://nvidia.github.io/warp/modules/differentiability.html#custom-native-functions - this could be useful here. But maybe first check the assembly whether it does not already map to the right function.

adenzler-nvidia (commented):
added a few comments, not finished yet though.

mujoco_warp/_src/collision_convex.py
adenzler-nvidia: I see there is wp.sign() that should do this.

mujoco_warp/_src/collision_convex.py
```diff
@@ -15,12 +15,806 @@
...
+  GeomType.ELLIPSOID.value,
+  GeomType.CYLINDER.value,
+  GeomType.MESH.value,
+}
+
+
+@wp.func
+def gjk_normalize(a: wp.vec3):
+  norm = wp.length(a)
+  if norm > 1e-8 and norm < 1e12:
```
adenzler-nvidia: these numbers also look a bit arbitrary, are they in any way related to Mj_Minval and maxval?

mujoco_warp/_src/collision_convex.py
```diff
@@ -15,12 +15,806 @@
...
+  elif wp.abs(normal[1]) < wp.abs(normal[2]):
+    dir = wp.vec3(
+      -normal[1] * normal[0], 1.0 - normal[1] * normal[1], -normal[1] * normal[2]
+    )
+  else:
+    dir = wp.vec3(
+      -normal[2] * normal[0], -normal[2] * normal[1], 1.0 - normal[2] * normal[2]
+    )
+  dir, _ = gjk_normalize(dir)
+  return dir
```
adenzler-nvidia: probably makes sense to put this into math.py?

mujoco_warp/_src/collision_convex.py
```diff
@@ -15,12 +15,806 @@
...
+  count: int,
+  prevCount: int,
+  dists: vecc3,
+  tris: mat2c3,
+  p: matc3,
+):
+  # Expand the polytope greedily.
+  for j in range(count):
+    bestIndex = int(0)
+    dd = float(dists[0])
```
adenzler-nvidia: why does this need a cast to float? Isn't this already a float?

mujoco_warp/_src/collision_convex.py
```diff
@@ -15,12 +15,806 @@
...
+    dists[bestIndex] = 2e30
+
+    parentIndex = bestIndex // 3
+    childIndex = bestIndex % 3
+    # fill in the new triangle at the next index
+    tris[TRIS_DIM + j * 3 + 0] = tris[parentIndex * 3 + childIndex]
+    tris[TRIS_DIM + j * 3 + 1] = tris[parentIndex * 3 + ((childIndex + 1) % 3)]
+    tris[TRIS_DIM + j * 3 + 2] = p[parentIndex]
+
+  for r in range(EPS_BEST_COUNT * 3):
```
adenzler-nvidia: can you make sure EPS_BEST_COUNT * 3 is statically evaluated?

mujoco_warp/_src/collision_convex.py
```diff
@@ -15,12 +15,806 @@
...
+    # TODO(btaba): split get_multiple_contacts into a separate kernel.
+    count, points = _get_multiple_contacts(m, info1, info2, depth, normal)
+
+    cid = wp.atomic_add(d.ncon, 0, count)
+    for i in range(count):
+      d.contact.dist[cid + i] = -depth
+      d.contact.geom[cid + i] = geoms
+      d.contact.frame[cid + i] = make_frame(normal)
+      d.contact.pos[cid + i] = points[i]
+      d.contact.worldid[cid + i] = worldid
```
adenzler-nvidia, line 757: could you use write_contact here?

mujoco_warp/_src/collision_convex.py
```diff
@@ -15,12 +15,806 @@
...
+  ):
+    normal = input_normal
+
+    # Get the support. If less than 0, objects are not intersecting.
+    depth, _simplex = _gjk_support(info1, info2, type1, type2, normal, m.mesh_vert)
+
+    if depth < -depth_extension:
+      # Objects are not intersecting, and we do not obtain the closest points as
+      # specified by depth_extension.
+      return wp.nan, wp.vec3(wp.nan, wp.nan, wp.nan)
```
adenzler-nvidia, line 312: I'm not sure if it's a good idea to use nan to encode invalid, as it can pollute all the calculations downstream.. But probably fine for now.

adenzler-nvidia (commented):
some more comments. Honestly, this is quite difficult to review so I mostly focused on style/coding issues and did not try to understand all the functionality in detail so far.

mujoco_warp/_src/types.py
Kenny-Vilella: Nope, because we have unsupported types and the enum follow mujoco value.
So if we use the len(GeomType), one type would not be supported.

mujoco_warp/_src/collision_primitive.py
Kenny-Vilella: I think -1 should work.
It is now changed.

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Changed to use wp.sign.
And anyway it was only to make the implementation as close as possible from the origin one, to reduce potential issues.

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: These numbers comes from the original implementation.
Not sure how significant they are, I don't feel very confident to change them to Mj_Minval and maxval.

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Moved it along with gjk_normalize

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Added wp.static

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Removed the cast

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: We could but it will be less efficient, we currently do only one atomic_add for count contact.
If we use write_contact we will do count atomic_add.

If you want to keep the same architecture, I can create a write_contactS function that is more suitable.

Kenny-Vilella:
@gavrielstate  @erikfrey  It will probably better to create the collision folder in a different PR.

I can see that @nvtw has several PRs to add more primitive collision, so let's wait that all the PRs are in before changing the code structure.

Kenny-Vilella:
Removed the `enableGjk` param and launch the Gjk only for the pairs specified in Mujoco.
Also updated the test accordingly.

mujoco_warp/_src/types.py
adenzler-nvidia: I see, thanks for the reminder.

mujoco_warp/_src/collision_convex.py
adenzler-nvidia: write_contactS is probably a good idea, but doesn't have to be done in this PR. Does this respect the contact margin? That was the point of the helper function :)

mujoco_warp/_src/collision_convex.py
```diff
@@ -13,14 +13,767 @@
...
+    dd = dists[0]
+    for i in range(1, 3 * prevCount):
+      if dists[i] < dd:
+        dd = dists[i]
+        bestIndex = i
+
+    dists[bestIndex] = 2e30
+
+    parentIndex = bestIndex // 3
+    childIndex = bestIndex % 3
```
adenzler-nvidia, line 171: kind of random question - would it make sense to over-allocate a bit and have 1 element of padding, with the benefit of having quicker arithmetic here? Division/modulo by 4 should be faster than 3.

Not sure if it will make a difference in the overall picture. But would be cool to know. Can also be a follow-up experiment.

mujoco_warp/_src/collision_convex.py
```diff
@@ -13,14 +13,767 @@
...
+    #    directions are found. This can be modified to the extremes in the
+    #    direction of eigenvectors of the variance of points of each polygon. If
+    #    they do not intersect, the closest points of both polygons are found.
+    if depth < -depth_extension:
+      return
+
+    dir = orthonormal(normal)
+    dir2 = wp.cross(normal, dir)
+
+    angle = MULTI_TILT_ANGLE * wp.pi / 180.0
```
adenzler-nvidia: could you make this a wp.static?

mujoco_warp/_src/collision_convex.py
```diff
@@ -13,14 +13,767 @@
...
+    c = wp.cos(angle)
+    s = wp.sin(angle)
+    t = 1.0 - c
```
adenzler-nvidia: I guess all of these can be static.

mujoco_warp/_src/collision_convex.py
```diff
@@ -13,14 +13,767 @@
...
+
+    contact_points = mat3c()
+
+    # Obtain points on the polygon determined by the support and tilt angle,
+    # in the basis of the contact frame.
+    v1count = int(0)
+    v2count = int(0)
+    # return 0, contact_points
+    for i in range(wp.static(MULTI_POLYGON_COUNT)):
+      angle = 2.0 * float(i) * wp.pi / float(MULTI_POLYGON_COUNT)
```
adenzler-nvidia: wp.pi / MULTI_POLYGON_COUNT can also be a wp.static

mujoco_warp/_src/collision_convex.py
```diff
@@ -13,14 +13,767 @@
...
+    geoms = d.collision_pair[tid]
+
+    if tid >= d.ncollision[0]:
+      return
+
+    # Check if we generated max contacts for this env.
+    # TODO(btaba): move max_contact_points_per_env culling to a point later
+    # in the pipline, where we can do a sort on penetration depth per env.
+    if d.ncon[0] > d.nconmax:
+      return
```
adenzler-nvidia, line 720: could you move the two returns above the worldid/geoms lookup?

mujoco_warp/_src/collision_convex.py
adenzler-nvidia: ah never mind, I see it above now.

adenzler-nvidia (approved):
Approving - thank you very much. Feel free to merge after addressing the comments.

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: The whole question is to get good benchmark to test this properly.
I will try to see tomorrow, If I can get something representative.
But that is a good idea.

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Done

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Done

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Not sure what you mean, so I put 
2.0 * wp.pi / float(MULTI_POLYGON_COUNT)
outside the loop and made it a static

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: Done

mujoco_warp/_src/collision_convex.py
Kenny-Vilella: I didn't get measurable perf diff using a simple anymal scene, but I'm not quite sure I modified the code properly everywhere.
We could keep this is mind to try again after the algorithm has been consolidated.

adenzler-nvidia: yeah let's not make it complex unless we need to. But keeping this in mind for the future. Thanks for checking!

PR #28, opened 2025-03-25
update readme testspeed benchmarking path

thowell:
update readme testspeed benchmarking path after #22


erikfrey (approved):
Good catch thank you!

PR #30, opened 2025-03-26
check actuator_trntype in put_model

thowell:


adenzler-nvidia (approved):
Looks good - didn't double-check the unsupported ones from the test but I trust you there.

PR #31, opened 2025-03-26
check actuator_dyntype in put_model

thowell:


mujoco_warp/_src/io.py
```diff
@@ -24,6 +24,12 @@
 
 
 def put_model(mjm: mujoco.MjModel) -> types.Model:
+  dyntype_supported = np.isin(mjm.actuator_dyntype, list(types.DynType))
+  if (~dyntype_supported).any():
+    raise NotImplementedError(
+      f"Actuator dynamics type {mjm.actuator_dyntype[~dyntype_supported]} not supported."
+    )
```
adenzler-nvidia: I'm sure you experimented with this, but if there's a way to make this snippet shorter that would be appreciated given that there's a bunch of them in sequence.

Otherwise all the checks could also go into a separate function but that kind of defeats the purpose as it won't be more legible..

adenzler-nvidia (approved):
Thank you!

mujoco_warp/_src/io.py
thowell: #30 is merged and now the checks for `actuator_trntype` and `actuator_dyntype` are performed in a loop. if we like this pattern i will update the other relevant prs to utilize the pattern.

adenzler-nvidia: I like it! thanks a lot. Will merge this one.

PR #32, opened 2025-03-26
check actuator_gaintype in put_model

thowell:


adenzler-nvidia (approved):
Approving under the condition that this will be updated to the nice loop pattern that already exists. Thanks!

PR #34, opened 2025-03-26
check geom_type in put_model

thowell:


mujoco_warp/_src/io.py
```diff
@@ -24,6 +24,16 @@
 
 
 def put_model(mjm: mujoco.MjModel) -> types.Model:
+  geom_type_render = mjm.geom_type >= mujoco.mjtGeom.mjNGEOMTYPES.value
+  geom_type_supported = np.isin(mjm.geom_type, list(types.GeomType))
+  geom_type_unsupported = ~(geom_type_supported | geom_type_render)
+  if geom_type_unsupported.any():
+    raise NotImplementedError(
+      f"Geom type {mjm.geom_type[geom_type_unsupported]} is unsupported."
+    )
```
adenzler-nvidia: side questions - how are we dealing with render geoms right now? We didn't care, right?

I think that's one of the questions that needs a discussion in the overall newton collab, as that's likely going to differ between implementations.

No action required.

mujoco_warp/_src/io_test.py
```diff
@@ -90,6 +90,27 @@ def test_tendon(self):
...
+        <asset>
+          <hfield name="hfield" nrow="1" ncol="1" size="1 1 1 1"/>
+          <mesh name="mesh" vertex="1 0 0  0 1 0  0 0 1  1 1 0  1 0 1  0 1 1  1 1 1  0 0 0"/>
+        </asset>
+        <worldbody>
+          <geom type="hfield" hfield="hfield"/>             
+          <geom type="ellipsoid" size="1 1 1"/>
+          <geom type="cylinder" size="1 1"/>
+          <geom type="mesh" mesh="mesh"/>
+        </worldbody>
```
adenzler-nvidia, line 131: should this include some render geoms as well?

mujoco_warp/_src/io.py
thowell: this check in `put_model` will not raise an error if the mjcf has render geoms

thowell: should we change the check to raise an error if the mjcf has render geoms?

mujoco_warp/_src/io_test.py
thowell: if we decide to raise an error for an mjcf with render geoms we should add render geoms to this test mjcf

mujoco_warp/_src/io.py
adenzler-nvidia: I don't think we should error in that case, as a lot of MJCF's will have render geoms and we would be limiting ourselves a lot. But we'll see whether all the downstream consumers will use the render geom concept for their visualization, so we might need way to disable any code related to render geoms.

thowell: based on the documentation: https://mujoco.readthedocs.io/en/latest/APIreference/APItypes.html#mjtgeom render geoms are not included in `mjModel` so we don't need a check for them in `put_model`.

mujoco_warp/_src/io_test.py
thowell: this mjcf does not need to include render geoms since the tested function `put_model` is not checking for render geoms as they are not part of `mjModel`.

adenzler-nvidia (approved):
Thanks!

adenzler-nvidia:
Thanks - could you run the linter again? Looks like some checks failed.

PR #37, opened 2025-03-26
check and test supported equality constraints by type

thowell:


adenzler-nvidia (approved):
Thanks - please also unify with the loop pattern.

PR #41, opened 2025-03-26
Add box-box collision

camevor:
Implements box-box collision function based on brute-force SAT test as in mjx. Uses tile reductions for parallel computation of the support across the 21 axes considered.


thowell:
thanks for this contribution @camevor!

please format collision_box.py (`ruff format`)

thowell (approved):

PR #46, opened 2025-03-27
add elliptic friction cone constraint: condim=1, condim=3

### LINKED ISSUE #68: Elliptic cone friction

thowell:


thowell:
support for condim=1 is added for elliptic cones

mujoco_warp/_src/types.py
```diff
@@ -132,10 +132,11 @@ class ConeType(enum.IntEnum):
 
   Members:
     PYRAMIDAL: pyramidal
+    ELLIPTIC: elliptic
   """
 
   PYRAMIDAL = mujoco.mjtCone.mjCONE_PYRAMIDAL
-  # unsupported: ELLIPTIC
+  ELLIPTIC = mujoco.mjtCone.mjCONE_ELLIPTIC
```
adenzler-nvidia, line 139: Does this need a README update?

mujoco_warp/_src/constraint.py
```diff
@@ -210,6 +210,102 @@ def _efc_contact_pyramidal(
...
+    return
+
+  includemargin = d.contact.includemargin[conid]
+  pos = d.contact.dist[conid] - includemargin
+
+  if pos < 0.0:
+    efcid = wp.atomic_add(d.nefc, 0, 1)
+    worldid = d.contact.worldid[conid]
+    d.efc.worldid[efcid] = worldid
+    d.contact.efc_address[conid, dimid] = efcid
```
adenzler-nvidia, line 266: I suppose there are other places that set efc_address that also need updating? Or where are actually reading this?

mujoco_warp/_src/constraint.py
```diff
@@ -210,6 +210,102 @@ def _efc_contact_pyramidal(
...
+    return
+
+  # TODO(team): condim=6
+  if condim == 6:
+    return
+
+  includemargin = d.contact.includemargin[conid]
+  pos = d.contact.dist[conid] - includemargin
+
+  if pos < 0.0:
```
adenzler-nvidia: would it make sense to follow the same wording as the contact constraints with the "active" variable?

mujoco_warp/_src/constraint.py
```diff
@@ -210,6 +210,102 @@ def _efc_contact_pyramidal(
...
+
+    invweight = m.body_invweight0[body1, 0] + m.body_invweight0[body2, 0]
+
+    if dimid > 0:
+      solreffriction = d.contact.solreffriction[conid]
+
+      if not ((solreffriction[0] != 0.0) or (solreffriction[1] != 0.0)):
+        _solref = d.contact.solref[conid]
+        solreffriction[0] += _solref[0]
+        solreffriction[1] += _solref[1]
```
adenzler-nvidia: This is definitely me having no idea, but wouldn't it make more sense to do this kind of calculation in get_contact_solver_params_kernel?

adenzler-nvidia (commented):
Looking good, just a few questions.

mujoco_warp/_src/types.py
thowell: done

mujoco_warp/_src/constraint.py
thowell: done

mujoco_warp/_src/constraint.py
thowell: `contact.efc_address` should also be set in `put_data`. `contact.efc_address` will be used by the engine solver for elliptic friction cones.

mujoco_warp/_src/constraint.py
thowell: for elliptic cone contacts, the first constraint (normal direction) will use `solref` and the remaining constraints (non-normal directions) will use `solreffriction` .

added a comment and updated the implementation to better match the mujoco reference https://github.com/google-deepmind/mujoco/blob/40393f460a4a378ac26dbf0416217e90063d84b0/src/engine/engine_core_constraint.c#L1518

i think that it makes sense to perform this calculation here because it constraint-level logic whereas `get_contact_solver_params_kernel` operates on contacts.

mujoco_warp/_src/constraint.py
adenzler-nvidia: I see, let's follow the lead of MJX/MJC.

mujoco_warp/_src/constraint.py
adenzler-nvidia: Thanks for the comment. Your argument makes sense from an API point of view - although I would rather make sure we're not accessing too much memory or adding too much divergence. But let's follow the lead by MJX/MJC right now and change it if there is a reason (i.e. benchmark) showing that this is needed.

mujoco_warp/_src/constraint.py
```diff
@@ -210,6 +210,101 @@ def _efc_contact_pyramidal(
...
+
+    geom = d.contact.geom[conid]
+    body1 = m.geom_bodyid[geom[0]]
+    body2 = m.geom_bodyid[geom[1]]
+
+    cpos = d.contact.pos[conid]
+    frame = d.contact.frame[conid]
+
+    Jqvel = float(0.0)
+    for i in range(m.nv):
```
adenzler-nvidia, line 277: something to consider/try in the future, might be beneficial to launch nconmax * nv threads and fill the jacobian per-element?

No action required.

mujoco_warp/_src/constraint.py
```diff
@@ -210,6 +210,101 @@ def _efc_contact_pyramidal(
...
+    pos_aref = pos
+
+    if dimid > 0:
+      solreffriction = d.contact.solreffriction[conid]
+
+      # non-normal directions use solreffriction (if non-zero)
+      if solreffriction[0] or solreffriction[1]:
+        ref = solreffriction
+
+      invweight = invweight / m.opt.impratio
```
adenzler-nvidia, line 301: might be beneficial to store impration as a reciprocal because multiplication is much faster than division on a GPU.

No action needed, but feel free to experiment.

adenzler-nvidia (approved):
Thank you! Looking good, a few comments but more something to think about than anything urgent to fix.

mujoco_warp/_src/constraint.py
thowell: updated `put_data` and `get_data_into`

mujoco_warp/_src/constraint.py
thowell: added a todo

mujoco_warp/_src/constraint.py
thowell: added a todo

thowell:
@adenzler-nvidia please merge this pr if you approve the changes made since your approval. thanks!

adenzler-nvidia:
Thanks - merging.

PR #48, opened 2025-03-28
add sensors

thowell:
add sensors:

- joint position
- joint velocity
- actuator force

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -392,7 +392,6 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   # sensors
   m.sensor_type = wp.array(mjm.sensor_type, dtype=wp.int32, ndim=1)
   m.sensor_datatype = wp.array(mjm.sensor_datatype, dtype=wp.int32, ndim=1)
-  m.sensor_needstage = wp.array(mjm.sensor_needstage, dtype=wp.int32, ndim=1)
   m.sensor_objtype = wp.array(mjm.sensor_objtype, dtype=wp.int32, ndim=1)
   m.sensor_objid = wp.array(mjm.sensor_objid, dtype=wp.int32, ndim=1)
   m.sensor_reftype = wp.array(mjm.sensor_reftype, dtype=wp.int32, ndim=1)
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -37,8 +37,10 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     if unsupported.any():
       raise NotImplementedError(f"{field_str} {field[unsupported]} not supported.")
 
-  if mjm.nsensor > 0:
-    raise NotImplementedError("Sensors are unsupported.")
+  # TODO(team): check sensors by type
+
+  if mjm.sensor_cutoff.any():
+    raise NotImplementedError("Sensor cutoff is unsupported.")
 
   if mjm.ntendon > 0:
     raise NotImplementedError("Tendons are unsupported.")
@@ -77,6 +79,8 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   m.nsite = mjm.nsite
   m.nmocap = mjm.nmocap
   m.nM = mjm.nM
+  m.nsensor = mjm.nsensor
+  m.nsensordata = mjm.nsensordata
   m.nlsp = mjm.opt.ls_iterations  # TODO(team): how to set nlsp?
   m.nexclude = mjm.nexclude
   m.opt.timestep = mjm.opt.timestep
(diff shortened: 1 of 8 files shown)


mujoco_warp/_src/io.py
```diff
@@ -787,6 +821,10 @@ def tile(x):
   d.collision_pair = wp.empty(nconmax, dtype=wp.vec2i, ndim=1)
   d.collision_worldid = wp.empty(nconmax, dtype=wp.int32, ndim=1)
   d.ncollision = wp.zeros(1, dtype=wp.int32, ndim=1)
+
+  # sensors
+  d.sensordata = wp.array(tile(mjd.sensordata), dtype=wp.float32, ndim=2)
```
adenzler-nvidia, line 860: does it make sense to copy the sensordata from CPU?

mujoco_warp/_src/io.py
```diff
@@ -385,6 +389,33 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     or np.any(mjm.actuator_gaintype == types.GainType.AFFINE.value)
   )
 
+  # sensors
+  m.sensor_type = wp.array(mjm.sensor_type, dtype=wp.int32, ndim=1)
+  m.sensor_datatype = wp.array(mjm.sensor_datatype, dtype=wp.int32, ndim=1)
+  m.sensor_needstage = wp.array(mjm.sensor_needstage, dtype=wp.int32, ndim=1)
```
adenzler-nvidia: do we really need needsstage? Looks like a JAX leftover, but I'm happy to be corrected.

mujoco_warp/_src/sensor.py
```diff
@@ -0,0 +1,101 @@
...
+  @kernel
+  def _sensor_pos(m: Model, d: Data):
+    worldid, posid = wp.tid()
+    posadr = m.sensor_pos_adr[posid]
+    sensortype = m.sensor_type[posadr]
+    objid = m.sensor_objid[posadr]
+    adr = m.sensor_adr[posadr]
+
+    if sensortype == int(SensorType.JOINTPOS.value):
+      d.sensordata[worldid, adr] = _joint_pos(m, d, worldid, objid)
```
adenzler-nvidia, line 44: will be interesting to see whether it makes sense to have 1 kernel for all sensors or we should split it. 

No action required for now.

mujoco_warp/_src/io.py
```diff
@@ -385,6 +389,33 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
...
+  m.sensor_objtype = wp.array(mjm.sensor_objtype, dtype=wp.int32, ndim=1)
+  m.sensor_objid = wp.array(mjm.sensor_objid, dtype=wp.int32, ndim=1)
+  m.sensor_reftype = wp.array(mjm.sensor_reftype, dtype=wp.int32, ndim=1)
+  m.sensor_refid = wp.array(mjm.sensor_refid, dtype=wp.int32, ndim=1)
+  m.sensor_dim = wp.array(mjm.sensor_dim, dtype=wp.int32, ndim=1)
+  m.sensor_adr = wp.array(mjm.sensor_adr, dtype=wp.int32, ndim=1)
```
adenzler-nvidia, line 421: is sensor_adr really needed if you have the ones below for each stage?

mujoco_warp/_src/io_test.py
```diff
@@ -71,23 +71,7 @@ def test_equality(self):
...
-        </worldbody>   
-        <sensor>
-          <jointpos joint="slide"/>                      
-        </sensor> 
-      </mujoco>
-    """)
-
-    with self.assertRaises(NotImplementedError):
-      mjwarp.put_model(mjm)
+  # TODO(team): sensors
```
adenzler-nvidia, line 75: would be great to have test coverage for the missing sensors here.

mujoco_warp/_src/sensor_test.py
```diff
@@ -0,0 +1,84 @@
...
+    mujoco.mj_forward(mjm, mjd)
+
+    m = mjwarp.put_model(mjm)
+    d = mjwarp.put_data(mjm, mjd)
+
+    d.sensordata.zero_()
+
+    mjwarp.sensor_pos(m, d)
+    mjwarp.sensor_vel(m, d)
+    mjwarp.sensor_acc(m, d)
```
adenzler-nvidia, line 77: wouldn't it make sense to do a step and compare afterwards? I know that's probably technically already an integration test, but otherwise this mostly tests the io function?

mujoco_warp/_src/types.py
```diff
@@ -379,6 +386,8 @@ class Model:
     nexclude: number of excluded geom pairs                  ()
     nmocap: number of mocap bodies                           ()
     nM: number of non-zeros in sparse inertia matrix         ()
+    nsensor: number of sensors                               ()
+    nsensordata: number of elements in sensor data vector    ()
```
adenzler-nvidia, line 411: is there any value in having these for all sensors combined if we're splitting addresses and counts per stage anyway?

mujoco_warp/_src/io.py
thowell: i think the expectation for `put_data` is that fields from `MjData` will be copied to device

mujoco_warp/_src/io.py
thowell: `sensor_needstage` is an `mjModel` field https://mujoco.readthedocs.io/en/latest/APIreference/APItypes.html#mjmodel

removed `sensor_needstage` since it is currently unused.

mujoco_warp/_src/io.py
thowell: `sensor_pos_adr`, `sensor_vel_adr`, and `sensor_acc_adr` index into `sensor_adr`.

mujoco_warp/_src/io_test.py
thowell: agreed.

will leave adding this test coverage to a future pr.

mujoco_warp/_src/sensor_test.py
thowell: calling `mj_forward` is sufficient for sensors. after zeroing `sensordata`, calling the sensor functions will correctly populate `sensordata` for the current state.

adenzler-nvidia: I see, thanks. Just make sure the sensordata is not zero by accident after the forward step, but I think you got that covered.

adenzler-nvidia (approved):
Approving - feel free to merge after resolving the conflicts.

mujoco_warp/_src/types.py
thowell: we probably don't need some of these dimensions as fields in `Model` if the dimension can be determined from an array size.

thowell: it could also make sense to have `Model` contain all `mjModel` fields and contain additional mujoco warp specific fields

PR #49, opened 2025-03-31
Small set of follow-up optimizations for _JTDAJ

adenzler-nvidia:
* value-gating the atomic add for 0 values
* move nefc lookup out of the loop
* make dim_X and nblocks_per_block static to enable unrolling the loop

Difficult to tell the effect of individual parts here, but mostly better codegen and overall some perf gains (humanoid.xml):

before:
```

 Total JIT time: 0.66 s
 Total simulation time: 4.46 s
 Total steps per second: 1,834,768
 Total realtime factor: 9,173.84 x
 Total time per step: 545.03 ns
```

after: 
```

 Total JIT time: 0.66 s
 Total simulation time: 4.35 s
 Total steps per second: 1,884,647
 Total realtime factor: 9,423.23 x
 Total time per step: 530.60 ns
```

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -251,10 +251,12 @@ def _update_gradient(m: types.Model, d: types.Data):
     # TODO(team): static m?
     efcid_temp, elementid = wp.tid()
 
+    nefc = d.nefc[0]
+
     for i in range(nblocks_perblock):
       efcid = efcid_temp + i * dim_x
 
-      if efcid >= d.nefc[0]:
+      if efcid >= nefc:
         return
 
       worldid = d.efc.worldid[efcid]
@@ -272,11 +274,13 @@ def _update_gradient(m: types.Model, d: types.Data):
         continue
 
       # TODO(team): sparse efc_J
-      wp.atomic_add(
-        d.efc.h[worldid, dofi],
-        dofj,
-        d.efc.J[efcid, dofi] * d.efc.J[efcid, dofj] * efc_D,
-      )
+      value = d.efc.J[efcid, dofi] * d.efc.J[efcid, dofj] * efc_D
+      if value != 0.0:
+        wp.atomic_add(
+          d.efc.h[worldid, dofi],
+          dofj,
+          value
+        )
 
   @kernel
   def _cholesky(d: types.Data):
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -253,7 +253,7 @@ def _update_gradient(m: types.Model, d: types.Data):
   # of SMs on the GPU. We can now query the SM count:
   # https://github.com/NVIDIA/warp/commit/f3814e7e5459e5fd13032cf0fddb3daddd510f30
 
-  # AD: make dim_x and nblocks_perblock static arguments for _JTDAJ to allow unrolling the loop
+  # make dim_x and nblocks_perblock static arguments for _JTDAJ to allow unrolling the loop
   if wp.get_device().is_cuda:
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -246,15 +246,38 @@ def _copy_lower_triangle(m: types.Model, d: types.Data):
...
       d.efc.h[worldid, rowid, colid] = d.qM[worldid, rowid, colid]
 
+  # Optimization: launching _JTDAJ with limited number of blocks on a GPU.
+  # Profiling suggests that only a fraction of blocks out of the original
+  # d.njmax blocks do the actual work. It aims to minimize #CTAs with no
+  # effective work. It launches with #blocks that's proportional to the number
+  # of SMs on the GPU. We can now query the SM count:
+  # https://github.com/NVIDIA/warp/commit/f3814e7e5459e5fd13032cf0fddb3daddd510f30
+
+  # AD: make dim_x and nblocks_perblock static arguments for _JTDAJ to allow unrolling the loop
```
erikfrey: nit, would you mind using `TODO(adenzler-nvidia): foo bar baz` - this does automagical things with some of our tooling

mujoco_warp/_src/solver.py
```diff
@@ -246,15 +246,38 @@ def _copy_lower_triangle(m: types.Model, d: types.Data):
...
+  if wp.get_device().is_cuda:
+    sm_count = wp.get_device().sm_count
+
+    # Here we assume one block has 256 threads. We use a factor of 6, which
+    # can be change in future to fine-tune the perf. The optimal factor will
+    # depend on the kernel's occupancy, which determines how many blocks can
+    # simultaneously run on the SM. TODO: This factor can be tuned further.
+    dim_x = int((sm_count * 6 * 256) / m.dof_tri_row.size)
+  else:
+    dim_x = d.njmax  # fall back
```
erikfrey: nit: maybe let's say "fall back for CPU" for the less initiated future reader of this code

mujoco_warp/_src/solver.py
adenzler-nvidia: Thanks - this is not really a TODO though. I fell back into the pattern we have here internally where you add your initials to some comments to track down the author quickly. But git blame can do the same thing, so I just removed the initials.

mujoco_warp/_src/solver.py
adenzler-nvidia: done, makes sense.

mujoco_warp/_src/solver.py
```diff
@@ -246,15 +246,38 @@ def _copy_lower_triangle(m: types.Model, d: types.Data):
...
+    # simultaneously run on the SM. TODO: This factor can be tuned further.
+    dim_x = int((sm_count * 6 * 256) / m.dof_tri_row.size)
+  else:
+    dim_x = d.njmax  # fall back for CPU
+
+  nblocks_perblock = int((d.njmax + dim_x - 1) / dim_x)
+
   @kernel
-  def _JTDAJ(m: types.Model, d: types.Data, nblocks_perblock: int, dim_x: int):
+  def _JTDAJ(m: types.Model, d: types.Data):
```
yuvaltassa, line 271: Why is this called `_JTDAJ` rather than `_JTDJ`? We are computing $J^T \cdot D \cdot J$ are we not?

The reason I ask is that in the dual solvers (which I don't think are implemented yet?) we have $A = J \cdot D^{-1} \cdot J^T$, so A means something and this could be confusing to the reader.

adenzler-nvidia: Deferring this to @thowell who I think did the original implementation.

PR #82, opened 2025-04-01
Camlight

Andrew-Luo1:
Adds around 0.3% overhead to step. Currently a straight-forward implementation of [mjx's version](https://github.com/google-deepmind/mujoco/blob/0abf184d49b7e3c45fb403a2dc0f6b38cd388090/mjx/mujoco/mjx/_src/smooth.py#L209), with a copy-paste for the lights. Tried factoring out commonalities between cameras and lights to a wp.func but the # of args required made it not worth it.

On my RTX4060Ti:

```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --event_trace=True

 Total JIT time: 12.31 s
 Total simulation time: 9.85 s
 Total steps per second: 831,960
 Total realtime factor: 4,159.80 x
 Total time per step: 1201.98 ns

Event trace:

step: 1200.44
  forward: 1196.80
    fwd_position: 191.79
      kinematics: 46.50
      com_pos: 14.86
      camlight: 3.34
      crb: 44.25
      factor_m: 29.07
      collision: 14.69
      make_constraint: 30.38
      transmission: 7.47
    fwd_velocity: 53.37
      com_vel: 19.71
      passive: 1.75
      rne: 19.73
    fwd_actuation: 14.95
    fwd_acceleration: 27.55
      xfrc_accumulate: 8.96
      solve_m: 16.58
    solve: 908.27
      mul_m: 15.53
      _linesearch: [ 74.49, 43.05, 29.85, 26.91, 25.57, 25.01, 24.63, 24.35, 24.18, 24.09 ]
        mul_m: [ 16.16, 3.34, 1.85, 1.66, 1.62, 1.63, 1.63, 1.63, 1.63, 1.63 ]
  euler: 3.24
```

Passes correctness tests on pendula.xml, which has all camera and light types.

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -267,7 +267,7 @@ def camlight(m: Model, d: Data):
   wp.launch(cam_local_to_global, dim=(d.nworld, m.ncam), inputs=[m, d])
 
   @kernel
-  def fn(m: Model, d: Data):
+  def cam_fn(m: Model, d: Data):
     worldid, camid = wp.tid()
     is_target_cam = (m.cam_mode[camid] == wp.static(CamLightType.TARGETBODY.value)) or (
       m.cam_mode[camid] == wp.static(CamLightType.TARGETBODYCOM.value)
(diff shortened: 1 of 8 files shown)


mujoco_warp/_src/smooth.py
```diff
@@ -247,6 +248,103 @@ def cdof(m: Model, d: Data):
   wp.launch(cdof, dim=(d.nworld, m.njnt), inputs=[m, d])
 
 
+@event_scope
+def camlight(m: Model, d: Data):
+  """Computes camera and light positions and orientations."""
+
```
adenzler-nvidia, line 254: could you add an early out if there are no lights and/or no cameras to skip this function?

Andrew-Luo1: Thanks for the review! To make the edits, I had two questions about Warp performance.

First, does the presence of a @kernel result in any overhead if it's never launched? For example, does
```python
def foo():
  @kernel
  def bar():
    pass
  print("hello world")
```
End up doing the same machine instructions as if def bar wasn't there?

Second, with the launch API, would this:
```python
if m.ncam:
  wp.launch(cam_fn, dim=(d.nworld, m.ncam), inputs=[m, d])
  wp.launch(cam_local_to_global, dim=(d.nworld, m.ncam), inputs=[m, d])
if m.nlight:
  wp.launch(light_local_to_global, dim=(d.nworld, m.nlight), inputs=[m, d])
  wp.launch(light_fn, dim=(d.nworld, m.nlight), inputs=[m, d])

```

be faster than the current implementation?
```python
wp.launch(cam_fn, dim=(d.nworld, m.ncam), inputs=[m, d])
wp.launch(cam_local_to_global, dim=(d.nworld, m.ncam), inputs=[m, d])
wp.launch(light_local_to_global, dim=(d.nworld, m.nlight), inputs=[m, d])
wp.launch(light_fn, dim=(d.nworld, m.nlight), inputs=[m, d])
```

Or is it correct to assume that if one of the dimensions such as m.ncam is zero, wp.launch is simply ignored, so there's no speedup?

adenzler-nvidia: defining the kernel but not doing a wp.launch does not incur any performance overhead. Depending on where you define it, it will be compiled but there's not runtime penalty.

And you are right about the launch being ignored if any dimension is zero. So the two code snippets are equivalent. However, I think for readability it's much better to have an explicit if statement there. Shows the intent much better.

mujoco_warp/_src/smooth.py
```diff
@@ -247,6 +248,102 @@ def cdof(m: Model, d: Data):
...
+    d.cam_xmat[worldid, camid] = math.quat_to_mat(
+      math.mul_quat(xquat, m.cam_quat[camid])
+    )
+
+  @kernel
+  def cam_fn(m: Model, d: Data):
+    worldid, camid = wp.tid()
+    is_target_cam = (m.cam_mode[camid] == wp.static(CamLightType.TARGETBODY.value)) or (
+      m.cam_mode[camid] == wp.static(CamLightType.TARGETBODYCOM.value)
+    )
```
adenzler-nvidia, line 272: Would it make sense to precompute the list of target cams upfront and then only process that list here?

I don't know too much about camera usage, is it common to have a lot of target cameras? What are some usual ratios of target cams vs other cams?

mujoco_warp/_src/smooth.py
```diff
@@ -247,6 +248,102 @@ def cdof(m: Model, d: Data):
...
+    d.light_xdir[worldid, lightid] = math.rot_vec_quat(m.light_dir[lightid], xquat)
+
+  @kernel
+  def light_fn(m: Model, d: Data):
+    worldid, lightid = wp.tid()
+    is_target_light = (
+      m.light_mode[lightid] == wp.static(CamLightType.TARGETBODY.value)
+    ) or (m.light_mode[lightid] == wp.static(CamLightType.TARGETBODYCOM.value))
+    invalid_target = is_target_light and (m.light_targetbodyid[lightid] < 0)
+    if invalid_target:
```
adenzler-nvidia, line 321: same comment as above here.

adenzler-nvidia (approved):
Approving - one question/suggestion but not strictly required and could be a follow-up PR if we decide to do this.

PR #83, opened 2025-04-01
fixed tendon

thowell:


mujoco_warp/_src/smooth.py
```diff
@@ -781,3 +781,37 @@ def factor_solve_i(m, d, M, L, D, x, y):
...
+
+      wrap_objid = m.wrap_objid[wrap_jnt_adr]
+      prm = m.wrap_prm[wrap_jnt_adr]
+
+      # add to length
+      L = prm * d.qpos[worldid, m.jnt_qposadr[wrap_objid]]
+      wp.atomic_add(d.ten_length[worldid], tendon_jnt_adr, L)
+
+      # add to moment
+      d.ten_J[worldid, tendon_jnt_adr, m.jnt_dofadr[wrap_jnt_adr]] = prm
```
adenzler-nvidia, line 914: would it be an option to run this kernel per tendon and loop inside, avoiding the atomic add?

adenzler-nvidia (commented):
Am I missing this, or is the actual integration of the data you compute here into the solver planned in a follow-up PR?

mujoco_warp/_src/smooth.py
```diff
@@ -781,3 +781,37 @@ def factor_solve_i(m, d, M, L, D, x, y):
...
     _factor_solve_i_dense(m, d, M, x, y)
+
+
+def tendon(m: Model, d: Data):
+  """Computes tendon lengths and moments."""
+
+  if not m.ntendon:
+    return
+
+  d.ten_length.zero_()
```
erikfrey, line 892: let's use the kernel zeros, these don't work with JAX

mujoco_warp/_src/types.py
```diff
@@ -492,6 +505,13 @@ class Model:
     actuator_gear: scale length and transmitted force        (nu, 6)
     exclude_signature: body1 << 16 + body2                   (nexclude,)
     actuator_affine_bias_gain: affine bias/gain present
+    tendon_adr: address of first object in tendon's path     (ntendon,)
+    tendon_jnt_adr: joint tendon address                     (<=ntendon,)
```
erikfrey: from the code this looks like it's size `(nwrap,)` ?

mujoco_warp/_src/types.py
```diff
@@ -806,3 +835,7 @@ class Data:
   collision_pair: wp.array(dtype=wp.vec2i, ndim=1)
   collision_worldid: wp.array(dtype=wp.int32, ndim=1)
   ncollision: wp.array(dtype=wp.int32, ndim=1)
+
+  # tendon
+  ten_length: wp.array(dtype=wp.float32, ndim=2)
```
erikfrey, line 941: these fields should also be added to docstring

mujoco_warp/_src/smooth.py
erikfrey: oh, ignore this comment.  I forgot it's copy that doesn't work - `zero_` is fine.

mujoco_warp/_src/smooth.py
thowell: added a todo to compare atomic add and loop approaches

mujoco_warp/_src/types.py
thowell: i think this should be `(<=nwrap,)` since in general joints comprise a subset of the wrap objects

mujoco_warp/_src/types.py
thowell: done

thowell:
@adenzler-nvidia this pr adds functionality for computing tendon lengths and the tendon Jacobian for fixed tendons. future prs will add additional tendon functionality, for example, via transmission for actuators.

PR #97, opened 2025-04-02
Add basic contact force test

nvtw:


mujoco_warp/_src/support.py
```diff
@@ -200,3 +204,79 @@ def mat33_from_rows(a: wp.vec3, b: wp.vec3, c: wp.vec3):
...
+
+  if m.opt.cone == ConeType_PYRAMIDAL:
+    force = _decode_pyramid(
+      d.efc.force, efc_address, d.contact.friction[contact_id], condim
+    )
+  # elif m.opt.cone == ConeType.ELLIPTIC.value:
+  #   for i in range(condim):
+  #     force[i] = d.efc_force[efc_address + i]
+  # else:
+  #    wp.abort(f'Unknown cone type: {m.opt.cone}')
```
adenzler-nvidia: can you make this a todo comment?

And not sure if we need the abort - we should check for unknown cone type in io.py.

mujoco_warp/_src/support.py
```diff
@@ -200,3 +204,79 @@ def mat33_from_rows(a: wp.vec3, b: wp.vec3, c: wp.vec3):
...
+  if to_world_frame:
+    # Transform both top and bottom parts of spatial vector by the full contact frame matrix
+    t = wp.spatial_top(force) @ d.contact.frame[contact_id]
+    b = wp.spatial_bottom(force) @ d.contact.frame[contact_id]
+    force = wp.spatial_vector(t, b)
+
+  if efc_address >= 0:
+    valid_contact = 1.0
+  else:
+    valid_contact = 0.0
```
adenzler-nvidia: wouldn't it make much more sense to skip all the calculations for any contact that is not active?

mujoco_warp/_src/types.py
```diff
@@ -138,6 +138,10 @@ class ConeType(enum.IntEnum):
   # unsupported: ELLIPTIC
 
 
+# Define again to use inside warp kernels as they do not suport enum.IntEnum
+ConeType_PYRAMIDAL = int(mujoco.mjtCone.mjCONE_PYRAMIDAL)
```
adenzler-nvidia: That shouldn't be needed, we have plenty of other places where we use enums in kernels. You're likely just missing a int(enumtype).value.

mujoco_warp/_src/support_test.py
```diff
@@ -85,6 +100,91 @@ def test_make_put_data(self):
...
+          force,
+          wp.array(
+            [
+              j,
+            ],
+            dtype=int,
+          ),
+          True,
+        ],
+      )
```
adenzler-nvidia, line 176: is that really how the formatter puts it?

mujoco_warp/_src/support.py
```diff
@@ -200,3 +204,79 @@ def mat33_from_rows(a: wp.vec3, b: wp.vec3, c: wp.vec3):
...
+  to_world_frame: bool,
+):
+  tid = wp.tid()
+
+  contact_id = contact_ids[tid]
+
+  if contact_id >= d.ncon[0]:
+    return
+
+  force[tid] = contact_force(m, d, contact_id, to_world_frame)
```
adenzler-nvidia, line 298: where is this kernel going to be used except for the test? Is that missing still missing/part of another PR?

adenzler-nvidia (commented):
Thanks, added some comments.

mujoco_warp/_src/support.py
nvtw: Milad asked for contact force support. But at the moment the kernel is only used in that test.

mujoco_warp/_src/support_test.py
nvtw: Yes

mujoco_warp/_src/support.py
adenzler-nvidia: is this adapted from contact_force in mjx or contact_force_dim?

Either way, might make sense to at least add some comments for whoever implements touch sensors to indicate that this functionality is already present.

mujoco_warp/_src/support.py
```diff
@@ -24,6 +24,7 @@
 from .types import array3df
 from .warp_util import event_scope
 from .warp_util import kernel
+from .types import vec5
```
erikfrey: Please fix import order.

mujoco_warp/_src/support.py
```diff
@@ -200,3 +203,81 @@ def mat33_from_rows(a: wp.vec3, b: wp.vec3, c: wp.vec3):
...
+) -> wp.spatial_vector:
+  """Converts pyramid representation to contact force."""
+  force = wp.spatial_vector()
+
+  if condim == 1:
+    force[0] = pyramid[efc_address]
+    return force
+
+  # force_normal = sum(pyramid0_i + pyramid1_i)
+  normal_force = float(0.0)
```
erikfrey: looks like these two loops are compatible, no?  could you do:

```py
for i in range(condim - 1):
  dir1 = pyramid[2 * i + efc_address]
  dir2 = pyramid[2 * i + efc_address + 1]
  force[0] += dir1 + dir2
  force[i + 1] = (dir1 - dir2) * mu[i]
```

mujoco_warp/_src/support.py
```diff
@@ -200,3 +203,81 @@ def mat33_from_rows(a: wp.vec3, b: wp.vec3, c: wp.vec3):
...
+
+  if efc_address >= 0:
+    condim = d.contact.dim[contact_id]
+
+    # TODO: add support for elliptical cone type
+    if m.opt.cone == int(mujoco.mjtCone.mjCONE_PYRAMIDAL.value):
+      force = _decode_pyramid(
+        d.efc.force, efc_address, d.contact.friction[contact_id], condim
+      )
+    # elif m.opt.cone == ConeType.ELLIPTIC.value:
```
erikfrey: nit: better to remove commented out code

mujoco_warp/_src/support.py
erikfrey: +1

mujoco_warp/_src/support.py
erikfrey: +1, since you've gone this far why not define a `contact_force` function mirroring the `mj_contactForce` public API function, and add to `__init__.py`:

```py
from ._src.support import contact_force as contact_force
```

After doing that, you can call that function in your tests instead of doing the kernel launch.

mujoco_warp/_src/support_test.py
```diff
@@ -25,13 +25,28 @@
 
 from . import test_util
 
+from .support import contact_force_kernel
```
erikfrey, line 27: Please fix import order.

mujoco_warp/_src/support_test.py
```diff
@@ -85,6 +100,91 @@ def test_make_put_data(self):
...
+
+  def test_contact_force(self):
+    mjm, mjd, m, d = _load_from_string(self._CONTACTS)
+
+    # map MJX contacts to MJ ones
+    def _find(g):
+      val = (g == mjd.contact.geom).sum(axis=1)
+      return np.where(val == 2)[0][0]
+
+    contact_id_map = {i: _find(mjd.contact.geom[i]) for i in range(mjd.ncon)}
```
erikfrey: MJWarp contacts align with MuJoCo contacts, so this index search should be unnecessary.

erikfrey:
Looks like one last unresolved comment, then fix conflicts and failing test and we should be good to go.

mujoco_warp/_src/support.py
```diff
@@ -226,3 +229,70 @@ def mat33_from_rows(a: wp.vec3, b: wp.vec3, c: wp.vec3):
...
+  """Extract 6D force:torque for one contact, in contact frame by default."""
+  efc_address = d.contact.efc_address[
+    contact_id, 0
+  ]  # 0 in second dimension to get the normal force
+  force = wp.spatial_vector(0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
+
+  if efc_address >= 0:
+    condim = d.contact.dim[contact_id]
+
+    # TODO: add support for elliptical cone type
```
adenzler-nvidia: one final thing - could adapt the todo comment to the same format we have everywhere else in the repo. Apparently this makes some magic things with Google tooling.

erikfrey (approved):
Looks great, thank you!

PR #102, opened 2025-04-03
RK4 integrator

Andrew-Luo1:
## Notes
1. This PR currently doesn't set the time in the intermediate steps, [unlike in cmujoco](https://github.com/google-deepmind/mujoco/blob/0abf184d49b7e3c45fb403a2dc0f6b38cd388090/src/engine/engine_forward.c#L908), because it doesn't seem to do anything. I can add this back in if it does.
2. Due to kernel loading, the new test (test_rungekutta4) is quite slow, taking about 9 seconds whereas the other tests in forward_test.py are instant.

#### Euler (default)
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=1024 --event_trace=True

Summary for 1024 parallel rollouts

 Total JIT time: 0.45 s
 Total simulation time: 2.72 s
 Total steps per second: 377,154
 Total realtime factor: 1,885.77 x
 Total time per step: 2651.44 ns

Event trace:

step: 2639.15
  forward: 2627.89
    fwd_position: 292.25
      kinematics: 75.39
      com_pos: 26.55
      crb: 51.29
      factor_m: 25.87
      collision: 26.13
      make_constraint: 71.69
      transmission: 6.21
    fwd_velocity: 109.43
      com_vel: 35.62
      passive: 6.49
      rne: 43.24
    fwd_actuation: 14.06
    fwd_acceleration: 40.29
      xfrc_accumulate: 15.13
      solve_m: 18.08
    solve: 2164.86
      mul_m: 16.29
      _linesearch: [ 206.28, 170.60, 152.00, 137.90, 126.04, 120.00, 114.55, 109.53, 106.25, 104.79 ]
        mul_m: [ 16.27, 9.63, 9.26, 8.99, 8.56, 8.61, 8.43, 8.07, 7.88, 7.86 ]
  euler: 7.95
```
#### RK4

Subtracting the intermediate forward calls, RK4 takes 40.8 ns.

```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=1024 --event_trace=True

Summary for 1024 parallel rollouts

 Total JIT time: 1.48 s
 Total simulation time: 10.02 s
 Total steps per second: 102,183
 Total realtime factor: 510.92 x
 Total time per step: 9786.35 ns

Event trace:

step: 9763.19
  forward: 2429.37
    fwd_position: 223.03
      kinematics: 74.48
      com_pos: 26.41
      crb: 50.79
      factor_m: 25.81
      collision: 16.00
      make_constraint: 14.37
      transmission: 6.51
    fwd_velocity: 108.68
      com_vel: 35.31
      passive: 6.47
      rne: 42.91
    fwd_actuation: 13.86
    fwd_acceleration: 44.82
      xfrc_accumulate: 15.31
      solve_m: 22.42
    solve: 2032.10
      mul_m: 16.36
      _linesearch: [ 116.55, 116.04, 115.32, 114.88, 114.49, 114.27, 113.81, 113.68, 113.44, 113.24 ]
        mul_m: [ 16.16, 16.21, 16.07, 16.02, 15.98, 16.05, 15.94, 15.98, 15.97, 15.93 ]
  rungekutta4: 7330.62
    forward: [ 2432.05, 2424.35, 2433.46 ]
      fwd_position: [ 223.76, 224.40, 224.25 ]
        kinematics: [ 74.57, 74.31, 74.81 ]
        com_pos: [ 26.52, 26.24, 26.49 ]
        crb: [ 50.44, 50.79, 50.45 ]
        factor_m: [ 25.83, 25.95, 25.88 ]
        collision: [ 16.12, 16.76, 16.16 ]
        make_constraint: [ 14.95, 14.96, 14.85 ]
        transmission: [ 6.57, 6.34, 6.82 ]
      fwd_velocity: [ 109.25, 108.98, 109.76 ]
        com_vel: [ 35.32, 35.28, 35.34 ]
        passive: [ 6.55, 6.61, 6.59 ]
        rne: [ 42.97, 42.93, 43.07 ]
      fwd_actuation: [ 13.96, 13.88, 14.01 ]
      fwd_acceleration: [ 44.31, 44.48, 44.84 ]
        xfrc_accumulate: [ 14.90, 15.11, 14.92 ]
        solve_m: [ 22.38, 22.35, 22.40 ]
      solve: [ 2034.24, 2025.97, 2034.05 ]
        mul_m: [ 16.29, 16.36, 16.29 ]
        _linesearch: [ 116.36, 116.14, 115.41, 115.00, 114.56, 114.19, 113.68, 113.60, 113.50, 113.25, 116.25, 115.72, 114.78, 114.49, 114.11, 113.80, 113.54, 113.40, 113.24, 113.07, 116.80, 116.11, 115.69, 115.01, 114.41, 114.06, 113.78, 113.43, 113.29, 113.10 ]
          mul_m: [ 16.16, 16.16, 16.08, 16.04, 16.00, 15.98, 15.95, 15.99, 16.02, 15.91, 16.17, 16.18, 16.02, 15.93, 15.99, 16.00, 15.98, 16.03, 15.95, 15.96, 16.17, 16.16, 16.09, 16.13, 15.95, 15.97, 15.96, 15.97, 15.96, 15.88 ]
```

### DIFF
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -21,6 +21,7 @@ from ._src.collision_driver import sap_broadphase as sap_broadphase
 from ._src.collision_primitive import primitive_narrowphase as primitive_narrowphase
 from ._src.constraint import make_constraint as make_constraint
 from ._src.forward import euler as euler
+from ._src.forward import rungekutta4 as rungekutta4
 from ._src.forward import forward as forward
 from ._src.forward import fwd_acceleration as fwd_acceleration
 from ._src.forward import fwd_actuation as fwd_actuation
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -21,13 +21,13 @@ from ._src.collision_driver import sap_broadphase as sap_broadphase
 from ._src.collision_primitive import primitive_narrowphase as primitive_narrowphase
 from ._src.constraint import make_constraint as make_constraint
 from ._src.forward import euler as euler
-from ._src.forward import rungekutta4 as rungekutta4
 from ._src.forward import forward as forward
 from ._src.forward import fwd_acceleration as fwd_acceleration
 from ._src.forward import fwd_actuation as fwd_actuation
 from ._src.forward import fwd_position as fwd_position
 from ._src.forward import fwd_velocity as fwd_velocity
 from ._src.forward import implicit as implicit
+from ._src.forward import rungekutta4 as rungekutta4
 from ._src.forward import step as step
 from ._src.io import get_data_into as get_data_into
 from ._src.io import make_data as make_data
@@ -37,8 +37,8 @@ from ._src.passive import passive as passive
 from ._src.sensor import sensor_acc as sensor_acc
 from ._src.sensor import sensor_pos as sensor_pos
 from ._src.sensor import sensor_vel as sensor_vel
-from ._src.smooth import com_pos as com_pos
 from ._src.smooth import camlight as camlight
+from ._src.smooth import com_pos as com_pos
 from ._src.smooth import com_vel as com_vel
 from ._src.smooth import crb as crb
 from ._src.smooth import factor_m as factor_m
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -41,67 +41,73 @@ from .warp_util import kernel_copy
 import numpy as np
 
 # RK4 tableau
-_RK4_A = np.array([
+_RK4_A = np.array(
+  [
     [0.5, 0.0, 0.0],
     [0.0, 0.5, 0.0],
     [0.0, 0.0, 1.0],
-])
+  ]
+)
 _RK4_B = np.array([1.0 / 6.0, 1.0 / 3.0, 1.0 / 3.0, 1.0 / 6.0])
 
 
 @wp.func
-def _integrate_pos(worldId: int, jntid: int, 
-                   m: Model, qpos_out: array2df, 
-                   qpos_in: array2df, qvel_in: array2df):
+def _integrate_pos(
+  worldId: int,
+  jntid: int,
+  m: Model,
+  qpos_out: array2df,
+  qpos_in: array2df,
+  qvel_in: array2df,
+):
+  jnt_type = m.jnt_type[jntid]
+  qpos_adr = m.jnt_qposadr[jntid]
+  dof_adr = m.jnt_dofadr[jntid]
+  qpos = qpos_in[worldId]
+  qvel = qvel_in[worldId]
 
-    jnt_type = m.jnt_type[jntid]
-    qpos_adr = m.jnt_qposadr[jntid]
-    dof_adr = m.jnt_dofadr[jntid]
-    qpos = qpos_in[worldId]
-    qvel = qvel_in[worldId]
+  if jnt_type == wp.static(JointType.FREE.value):
+    qpos_pos = wp.vec3(qpos[qpos_adr], qpos[qpos_adr + 1], qpos[qpos_adr + 2])
+    qvel_lin = wp.vec3(qvel[dof_adr], qvel[dof_adr + 1], qvel[dof_adr + 2])
 
-    if jnt_type == wp.static(JointType.FREE.value):
-      qpos_pos = wp.vec3(qpos[qpos_adr], qpos[qpos_adr + 1], qpos[qpos_adr + 2])
-      qvel_lin = wp.vec3(qvel[dof_adr], qvel[dof_adr + 1], qvel[dof_adr + 2])
+    qpos_new = qpos_pos + m.opt.timestep * qvel_lin
 
-      qpos_new = qpos_pos + m.opt.timestep * qvel_lin
+    qpos_quat = wp.quat(
+      qpos[qpos_adr + 3],
+      qpos[qpos_adr + 4],
+      qpos[qpos_adr + 5],
+      qpos[qpos_adr + 6],
+    )
+    qvel_ang = wp.vec3(qvel[dof_adr + 3], qvel[dof_adr + 4], qvel[dof_adr + 5])
 
-      qpos_quat = wp.quat(
-        qpos[qpos_adr + 3],
-        qpos[qpos_adr + 4],
-        qpos[qpos_adr + 5],
-        qpos[qpos_adr + 6],
-      )
-      qvel_ang = wp.vec3(qvel[dof_adr + 3], qvel[dof_adr + 4], qvel[dof_adr + 5])
+    qpos_quat_new = math.quat_integrate(qpos_quat, qvel_ang, m.opt.timestep)
 
-      qpos_quat_new = math.quat_integrate(qpos_quat, qvel_ang, m.opt.timestep)
+    qpos_out[worldId, qpos_adr] = qpos_new[0]
+    qpos_out[worldId, qpos_adr + 1] = qpos_new[1]
+    qpos_out[worldId, qpos_adr + 2] = qpos_new[2]
+    qpos_out[worldId, qpos_adr + 3] = qpos_quat_new[0]
+    qpos_out[worldId, qpos_adr + 4] = qpos_quat_new[1]
+    qpos_out[worldId, qpos_adr + 5] = qpos_quat_new[2]
+    qpos_out[worldId, qpos_adr + 6] = qpos_quat_new[3]
 
-      qpos_out[worldId, qpos_adr] = qpos_new[0]
-      qpos_out[worldId, qpos_adr + 1] = qpos_new[1]
-      qpos_out[worldId, qpos_adr + 2] = qpos_new[2]
-      qpos_out[worldId, qpos_adr + 3] = qpos_quat_new[0]
-      qpos_out[worldId, qpos_adr + 4] = qpos_quat_new[1]
-      qpos_out[worldId, qpos_adr + 5] = qpos_quat_new[2]
-      qpos_out[worldId, qpos_adr + 6] = qpos_quat_new[3]
+  elif jnt_type == wp.static(JointType.BALL.value):  # ball joint
+    qpos_quat = wp.quat(
+      qpos[qpos_adr],
+      qpos[qpos_adr + 1],
+      qpos[qpos_adr + 2],
+      qpos[qpos_adr + 3],
+    )
+    qvel_ang = wp.vec3(qvel[dof_adr], qvel[dof_adr + 1], qvel[dof_adr + 2])
 
-    elif jnt_type == wp.static(JointType.BALL.value):  # ball joint
-      qpos_quat = wp.quat(
-        qpos[qpos_adr],
-        qpos[qpos_adr + 1],
-        qpos[qpos_adr + 2],
-        qpos[qpos_adr + 3],
-      )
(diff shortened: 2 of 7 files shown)


mujoco_warp/_src/forward.py
```diff
@@ -242,6 +264,81 @@ def eulerdamp(
...
+      worldId, tid = wp.tid()
+      d.qvel_rk[worldId, tid] += b * d.qvel[worldId, tid]
+      d.qacc_rk[worldId, tid] += b * d.qacc[worldId, tid]
+
+    wp.launch(_qvel_acc, dim=(d.nworld, m.nv), inputs=[d, b])
+
+    @kernel
+    def _act_dot(d: Data, b: float):
+      worldId, tid = wp.tid()
+      d.act_dot_rk[worldId, tid] += b * d.act_dot[worldId, tid]
```
adenzler-nvidia, line 291: small nitpick, could you move the launch of the first kernel below the definition of the second one? That way we only need to compile once. Same thing also applies to perturb_state.

mujoco_warp/_src/forward.py
```diff
@@ -242,6 +264,81 @@ def eulerdamp(
...
+      # Temp storage, since soon overwritten
+      d.qvel[worldId, tid] = d.qvel[worldId, tid] * a
+
+    wp.launch(_dqvel, dim=(d.nworld, m.nv), inputs=[d, a])
+
+    @kernel
+    def _qpos(m: Model, d: Data):
+      """Integrate joint positions"""
+      worldId, jntId = wp.tid()
+      _integrate_pos(worldId, jntId, m, d.qpos, d.qpos_t0, d.qvel)
```
adenzler-nvidia: I think it would make sense to just do the qvel scaling in _integrate_pos to avoid launching a kernel just for this and touching the memory twice.

you can add an if clause with a wp.static which is like a template that then generates 2 versions of the kernel, one with scaling and one without.

mujoco_warp/_src/forward.py
```diff
@@ -242,6 +264,81 @@ def eulerdamp(
...
+  rk_accumulate(d, B[0])
+  for i in range(3):
+    a, b = float(A[i, i]), B[i + 1]
+    perturb_state(m, d, a)
+    forward(m, d)
+    rk_accumulate(d, b)
+
+  kernel_copy(d.qpos, d.qpos_t0)
+  kernel_copy(d.qvel, d.qvel_t0)
+  kernel_copy(d.act, d.act_t0)
```
adenzler-nvidia, line 328: I'm not sure I understand this, so I need an explanation.

What you are doing is:
copy qpos/qvel/act to the t0 versions before integration.
do the integration
copy the t0 version back into the original arrays.

I can't find a place where you actually change the t0 versions, wouldn't it make more sense to just use temporary memory for the integration? Then you only have to do the copy at the beginning, and can skip the one at the end?

mujoco_warp/_src/forward_test.py
```diff
@@ -144,6 +144,43 @@ def test_disable_eulerdamp(self):
 
     np.testing.assert_allclose(d.qvel.numpy()[0], 1 + mjm.opt.timestep)
 
+  def test_rungekutta4(self):
+    # `forward` kernel dispatch is very slow
```
adenzler-nvidia: could you elaborate on this? Isn't this just compilation?

adenzler-nvidia (commented):
Thank you, some comments.

mujoco_warp/_src/forward.py
Andrew-Luo1: Sounds good. Out of curiousity, why does weaving the launches between kernel definitions, rather than putting them all together, result in multiple compilations? Is there a place in the docs where I can read more about this?

adenzler-nvidia: the overview on code generation covers this quite well: https://nvidia.github.io/warp/codegen.html#overview

mujoco_warp/_src/forward.py
Andrew-Luo1: Integrating qvel scaling into _integrate_pos seems like it would require a for loop over up to 6 values. For example, a thread updating a ball joint would need to first scale the 6 corresponding qvels. Would this be expected faster than the current approach of two separate kernels?

Not sure I understand the wp.static recommendation - which function would you recommend this for?

mujoco_warp/_src/forward.py
Andrew-Luo1: For some notation, let f be `def forward`. y is the state, in our case, `qpos`, `qvel` and `act`. The `k_i` are intermediate slope values, in our case, `qvel`, `qacc`, `act_dot`. These result from f. The rk4 integration is:

$$
\begin{aligned}
k_1 &= f(t_n, y_n), \\
k_2 &= f\left(t_n + \frac{h}{2}, y_n + \frac{h}{2} \, k_1\right), \\
k_3 &= f\left(t_n + \frac{h}{2}, y_n + \frac{h}{2} \, k_2\right), \\
k_4 &= f(t_n + h, y_n + h \, k_3).
\end{aligned}
$$

$$
y_{n+1} = y_n + \frac{h}{6}\left(k_1 + 2k_2 + 2k_3 + k_4\right).
$$

What this function is doing is:
1. copy qpos/qvel/act to the t0 versions before integration.
2. calculate $\frac{1}{6}\left(k_1 + 2k_2 + 2k_3 + k_4\right)$ by perturbing $y$ and calling `forward`
3. calculate $y_{n+1}$ by calling `_advance` on $y_n$, which is the $t_0$ versions

The $t_0$ versions must be saved at the beginning, since y gets overriden during the intermediate forward calls. Please let me know if this clarifies

mujoco_warp/_src/forward_test.py
Andrew-Luo1: Yes, will change the comment

mujoco_warp/_src/forward.py
Andrew-Luo1: Ah nevermind I see, it can just be an arg in _integrate_pos!

Andrew-Luo1:
RK4-only time 40.8 ns -> 36.77

mujoco_warp/_src/forward.py
adenzler-nvidia: Thanks - the forward call is sneaky, I missed the side effects of that one. Makes sense. Thanks for bearing with me.

adenzler-nvidia: While we're here - do we really need the entire forward step? Or are there parts that could be skipped?

adenzler-nvidia:
Could you also add a README change stating that RK4 is now supported?

mujoco_warp/_src/forward.py
Andrew-Luo1: cmujoco [skips sensors and energy computations](https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_forward.c#L911) (seems like a ~1% speedup), while MJX keeps the entire forward step. This optimization can be added down the line, but as far as I can tell, sensors and energy aren't implemented yet in Newton.

adenzler-nvidia: agreed, let's work in this if it becomes a problem. RK4 isn't going to be our fastest version anyway.

adenzler-nvidia (approved):
Approving, thank you!

mujoco_warp/_src/forward.py
```diff
@@ -39,6 +39,81 @@
 from .warp_util import event_scope
 from .warp_util import kernel
 from .warp_util import kernel_copy
+import numpy as np
+
+# RK4 tableau
+_RK4_A = np.array(
```
erikfrey: Given you're just converting these to floats anyway, you can avoid the `np.array(...` bit, which also means you don't need to import numpy.

mujoco_warp/_src/forward.py
```diff
@@ -39,6 +39,81 @@
...
+      qpos[qpos_adr + 5],
+      qpos[qpos_adr + 6],
+    )
+    qvel_ang = (
+      wp.vec3(qvel[dof_adr + 3], qvel[dof_adr + 4], qvel[dof_adr + 5]) * qvel_scale
+    )
+
+    qpos_quat_new = math.quat_integrate(qpos_quat, qvel_ang, m.opt.timestep)
+
+    qpos_out[worldId, qpos_adr] = qpos_new[0]
```
erikfrey: nit: perhaps can save some indexing by doing the same as the other variables above:

```py
qvel = qvel_in[worldId]
qpos_o = qpos_out[worldId]
...
qpos_o[qpos_adr] = qpos_new[0]
...
```

mujoco_warp/_src/forward.py
```diff
@@ -39,6 +39,81 @@
...
+    qpos_out[worldId, qpos_adr + 1] = qpos_new[1]
+    qpos_out[worldId, qpos_adr + 2] = qpos_new[2]
+    qpos_out[worldId, qpos_adr + 3] = qpos_quat_new[0]
+    qpos_out[worldId, qpos_adr + 4] = qpos_quat_new[1]
+    qpos_out[worldId, qpos_adr + 5] = qpos_quat_new[2]
+    qpos_out[worldId, qpos_adr + 6] = qpos_quat_new[3]
+
+  elif jnt_type == wp.static(JointType.BALL.value):  # ball joint
```
erikfrey: nit: you can delete the `# ball joint` comment, this was probably here back before we had enums.  Now it's superfluous.

mujoco_warp/_src/types.py
```diff
@@ -885,6 +893,12 @@ class Data:
   njmax: int
   rne_cacc: wp.array(dtype=wp.spatial_vector, ndim=2)
   rne_cfrc: wp.array(dtype=wp.spatial_vector, ndim=2)
+  qpos_t0: wp.array(dtype=wp.float32, ndim=2)
```
erikfrey, line 927: These arrays are not a lot of memory, but it's also not a little.

Can you confirm that this PR doesn't slow down benchmarks for non-rk4 scenes?  You can check humanoid.xml in this PR vs HEAD

If it does slow down, we should do like `is_sparse` and allocate these with shape `(0, 0)` if `opt.integrator != RK4`

erikfrey (approved):
Very nice, just a few nits.

mujoco_warp/_src/types.py
Andrew-Luo1: Does not appear to impact speed.

Comparing against commit 5ee45ec516dd3e6cb377db6ac52ab765ec858133, which is the origin/main that this PR is currently based on.

Command: 
`mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --event_trace=True`
Averaging over 3 runs.

Before commits: step: 1201.35
After: 1200.38

erikfrey: OK!

mujoco_warp/__init__.py
```diff
@@ -21,6 +21,7 @@
 from ._src.collision_primitive import primitive_narrowphase as primitive_narrowphase
 from ._src.constraint import make_constraint as make_constraint
 from ._src.forward import euler as euler
+from ._src.forward import rungekutta4 as rungekutta4
```
erikfrey: Please fix import order (see ruff error below) then we will merge.

PR #103, opened 2025-04-03
Implement sphere vs capsule collision

nvtw:


adenzler-nvidia (approved):
looks good, could you run ruff check --fix to fix the linter error? Then we can merge.

PR #105, opened 2025-04-03
Implement analytical sphere vs cylinder collisions

nvtw:
Note that this is not a direct port of mjx but an analytical implementation. There was a TODO for having an analytical implementation in the mjx code, so directly went for that one.

### DIFF
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -123,24 +123,49 @@ class CollisionTest(parameterized.TestCase):
           </worldbody>
         </mujoco>
         """
-  _SPHERE_CYLINDER = """
+  _SPHERE_CYLINDER_CORNER = """
         <mujoco>
           <worldbody>
             <body>
               <joint type="slide" axis="1 0 0"/>
               <joint type="slide" axis="0 1 0"/>
               <joint type="slide" axis="0 0 1"/>
-              <geom size="0.1" type="sphere"/>
+              <geom size="0.1" type="sphere" pos=".33 0 0"/>
             </body>
             <body>
               <geom size="0.15 0.2" type="cylinder" euler="30 45 0"/>
             </body>
           </worldbody>
-          <keyframe>
-            <key name="corner" qpos=".33 0 0"/>
-            <key name="cap" qpos=".26 -.14 .1"/>
-            <key name="side" qpos="0 -.26 0"/>
-          </keyframe>
+        </mujoco>
+        """
+  _SPHERE_CYLINDER_CAP = """
+        <mujoco>
+          <worldbody>
+            <body>
+              <joint type="slide" axis="1 0 0"/>
+              <joint type="slide" axis="0 1 0"/>
+              <joint type="slide" axis="0 0 1"/>
+              <geom size="0.1" type="sphere" pos=".26 -.14 .1"/>
+            </body>
+            <body>
+              <geom size="0.15 0.2" type="cylinder" euler="30 45 0"/>
+            </body>
+          </worldbody>
+        </mujoco>
+        """
+  _SPHERE_CYLINDER_SIDE = """
+        <mujoco>
+          <worldbody>
+            <body>
+              <joint type="slide" axis="1 0 0"/>
+              <joint type="slide" axis="0 1 0"/>
+              <joint type="slide" axis="0 0 1"/>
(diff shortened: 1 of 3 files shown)


adenzler-nvidia:
Is this analytical version based on the one in MJC or did you do that from scratch?

mujoco_warp/_src/collision_primitive.py
```diff
@@ -232,6 +232,44 @@ def plane_box(
       break
 
 
+@wp.func
+def sphere_cylinder(
```
erikfrey, line 297: This is surprisingly simple compared to C MuJoCo's Sphere<>Cylinder:

https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_collision_primitive.c#L324

Does this handle all three cases that are explicitly laid out in C MuJoCo: side, cap, corner?

If so, cool!  Perhaps C MuJoCo ought to use this code!  cc @yuvaltassa

If not, please fix.

mujoco_warp/_src/collision_primitive.py
```diff
@@ -232,6 +232,44 @@ def plane_box(
...
+
+  # Project and clamp to cylinder length
+  proj = wp.clamp(wp.dot(delta, axis), -cylinder.size[1], cylinder.size[1])
+  closest = cylinder.pos + axis * proj
+
+  radial = sphere.pos - closest
+  radial_len = wp.length(radial)
+
+  # Handle degenerate case when sphere center is on cylinder axis
+  if radial_len < 1e-6:
```
erikfrey: nit: where did you get this `1e-6`?  If it's just "any epsilon" please use `types.MJ_MINVAL`

mujoco_warp/_src/collision_primitive.py
```diff
@@ -232,6 +232,44 @@ def plane_box(
...
+    if wp.abs(axis[1]) < 0.7:
```
erikfrey: similar question, where does this `0.7` come from?

erikfrey (changes requested):
Very intriguing!  Curious to know where this sphere<>cylinder algo came from.

mujoco_warp/_src/collision_primitive.py
yuvaltassa: Ya I think this code is missing some cases. The point of side/cap/corner is that the normals are always aligned correctly.

In particular `proj = wp.clamp(wp.dot(delta, axis), -cylinder.size[1], cylinder.size[1])` is what you want for cap and corner but not for side. In the current implementation for the side collision case `closest` will not actually be the closest point to the sphere center.

The function in engine_collision_primitive is well tested. Happy to write simple test model for you covering all the branches if this would help.

mujoco_warp/_src/collision_driver_test.py
```diff
@@ -101,6 +101,20 @@ class PrimitiveTest(parameterized.TestCase):
           </worldbody>
         </mujoco>
         """,
+    "sphere_cylinder": """
```
yuvaltassa: There is no collision here 😛 

Here, I made you a model that covers all the branches:
```xml
<mujoco>
  <worldbody>
    <body>
      <joint type="slide" axis="1 0 0"/>
      <joint type="slide" axis="0 1 0"/>
      <joint type="slide" axis="0 0 1"/>
      <geom size="0.1" type="sphere"/>
    </body>
    <body>
      <geom size="0.15 0.2" type="cylinder" euler="30 45 0"/>
    </body>
  </worldbody>
  <keyframe>
    <key name="corner" qpos=".33 0 0"/>
    <key name="cap" qpos=".26 -.14 .1"/>
    <key name="side" qpos="0 -.26 0"/>
  </keyframe>
</mujoco>
```

and then in `test_primitives` you want to loop over the keyframes like 

```py
  def test_primitives(self, name):
    """Tests collision primitive functions."""
    mjm = mujoco.MjModel.from_xml_string(self._MJCFS[name])
    mjd = mujoco.MjData(mjm)
    for i in range(max(mjm.nkey, 1)):
      mujoco.mj_resetDataKeyframe(mjm, mjd, i)
      mujoco.mj_forward(mjm, mjd)
      ...
```
Note that this will work fine for the models without the keyframes, `mj_resetDataKeyframe` just resets mjData if the keyframe is invalid (hence the `max(mjm.nkey, 1)`)

yuvaltassa (approved):
nice

erikfrey (approved):
:clap:

PR #110, opened 2025-04-03
change rne cacc world body initialization and test rne with and without gravity

thowell:
- for case without gravity, set only the world body cacc to zero instead of zeroing the entire array
- test `rne` with and without gravity


adenzler-nvidia (approved):
Looks good, thank you!

PR #113, opened 2025-04-04
subtree_vel

thowell:
Implementation of `mj_subtreeVel`. This routine is called by subtreelinvel and subtreeangmom sensors.


erikfrey (approved):
Looks great, thank you.

erikfrey (approved):

PR #116, opened 2025-04-04
n*max should scale with nworld

btaba:


mujoco_warp/_src/io.py
```diff
@@ -506,11 +506,11 @@ def make_data(
   # TODO(team): move to Model?
   if nconmax == -1:
     # TODO(team): heuristic for nconmax
-    nconmax = 512
+    nconmax = nworld * 10
```
adenzler-nvidia: it should scale, but 10 contacts per world might be a bit on the small side?

mujoco_warp/_src/support_test.py
```diff
@@ -72,19 +72,6 @@ def test_xfrc_accumulated(self):
...
-    md = mjwarp.make_data(mjm)
-
-    # same number of fields
-    self.assertEqual(len(d.__dict__), len(md.__dict__))
-
-    # test shapes for all arrays
-    for attr, val in md.__dict__.items():
-      if isinstance(val, wp.array):
-        self.assertEqual(val.shape, getattr(d, attr).shape)
-
```
adenzler-nvidia, line 88: Thanks for moving this one.

mujoco_warp/_src/io.py
btaba: Hi @adenzler-nvidia , thanks for the review! What would you suggest as a better constant here? 
Perhaps we might develop a heuristic in io.put* based on the scene

erikfrey: cc @yuvaltassa - we were discussing this PR this morning and he had some good ideas on what kind of a heuristic to put here.  @btaba if Yuval doesn't comment, please bug him in person

adenzler-nvidia: Did we have a resolution on this one? Otherwise let's just use a slightly higher number for now (maybe 25 contacts?) and let users adjust based on their needs.

That being said, do we have an reporting on whether we exceed nconmax? This should be a separate task but we need to put some diagnostics in place for sure.

btaba: Agreed @adenzler-nvidia. Still waiting on @erikfrey and @yuvaltassa for a short term resolution to this PR

btaba: Ok updated based on offline conversation with @erikfrey

erikfrey (approved):
OK, we'll come back to this later with something clever-er.

PR #118, opened 2025-04-05
Tendon transmission

thowell:


mujoco_warp/_src/smooth.py
```diff
@@ -673,8 +673,14 @@ def _transmission(
...
+    elif trntype == wp.static(TrnType.TENDON.value):
+      trnid = m.actuator_trnid[actid, 0]
+      gear0 = gear[0]
+      length[worldid, actid] = d.ten_length[worldid, trnid] * gear0
+      for i in range(m.nv):
+        moment[worldid, actid, i] = d.ten_J[worldid, trnid, i] * gear0
     else:
-      # TODO handle site, tendon transmission types
+      # TODO(team): site, slidercrank, body
       wp.printf("unhandled transmission type %d\n", trntype)
```
erikfrey, line 754: You know, I'm not sure what the best practice is here.

On the one hand, this should never happen as `put_model` should fail if you try to put, say, a `slidercrank` transmission

On the other hand, who knows, maybe someone will override `put_model` somehow accidentally end up with a `Model` with `slidercrank` transmission.

But still I'm not sure if we should just `printf`?  In practice, this will spam stdout with one line per `nworld` x num physics steps, right?

Is there some good best practice for runtime asserts in Warp?  Maybe @adenzler-nvidia knows?

mujoco_warp/_src/smooth.py
```diff
@@ -673,8 +673,14 @@ def _transmission(
         moment[worldid, actid, vadr] = gear[0]
       else:
         wp.printf("unrecognized joint type")
+    elif trntype == wp.static(TrnType.TENDON.value):
+      trnid = m.actuator_trnid[actid, 0]
+      gear0 = gear[0]
+      length[worldid, actid] = d.ten_length[worldid, trnid] * gear0
+      for i in range(m.nv):
```
erikfrey: Hmmm - could you loop over `wrap_jnt_adr` instead?

erikfrey (approved):
Generally looks good - one discussion question and one technical q.

mujoco_warp/_src/smooth.py

thowell: fixed the fixed tendon case

mujoco_warp/_src/smooth.py
erikfrey: @adenzler-nvidia is out this week so let's not block, I'll make a separate issue for this.

erikfrey (approved):
Looks great, merge after fixing conflicts.

mujoco_warp/_src/smooth.py
adenzler-nvidia: Good question. Of course these things are best handled at compile time. Warp supports assert() statements but for this the kernel has to be compiled in debug mode. So they make sense if used similar to assertions in other programming languages, but not really as an error reporting mechanism.

I don't think there is a very good way to handle this - and I don't know off the top of my head what the recommended approach in warp is. As any type of error handling inside the kernel will add overhead, especially if there are function calls and prints involved, best is to just avoid it and sanitize things. One option is to return an error code from the kernel that you can read out at the end of the step, but that is probably not a good idea if CUDA graphs are involved.

PR #120, opened 2025-04-06
add subtreelinvel and subtreeangmom sensors

thowell:
Adds subtreelinvel and subtreeangmom sensors.


mujoco_warp/_src/io.py
```diff
@@ -458,6 +458,10 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     ndim=1,
   )
 
+  m.sensor_subtreevel = np.isin(
```
erikfrey: I think you know already my concerns with these precomputed fields - you should be able to get the same exact perf (since we do cudagraph capture) even if you move this code inline into `sensor_vel` - can you move it there?

erikfrey (approved):
Looks great, just one nit

mujoco_warp/_src/io.py

PR #112, opened 2025-04-04
rne_postconstraint

thowell:
implementation of `mj_rnePostConstraint`. this routine is called by accelerometer, force, and torque sensors.

future prs will add support for contacts and equality constraints.

PR #121, opened 2025-04-06
add accelerometer, force, and torque sensors

thowell:
add accelerometer, force, and torque sensors

depends on #112


mujoco_warp/_src/smooth.py
```diff
@@ -615,6 +615,11 @@ def qfrc_bias(m: Model, d: Data):
   wp.launch(qfrc_bias, dim=[d.nworld, m.nv], inputs=[m, d])
 
 
+def rne_postconstraint(m: Model, d: Data):
```
erikfrey: this is old right - do you need to bring this up to date with HEAD?

mujoco_warp/_src/sensor_test.py
```diff
@@ -43,22 +43,36 @@ def test_sensor(self):
     """Test sensors."""
     mjm = mujoco.MjModel.from_xml_string("""
       <mujoco>
+        <option gravity="-1 -1 -1"/>
         <worldbody>
-          <body>
+          <body pos="0 0 0">
```
erikfrey: nit: pos="0 0 0" is default, no?  should this be deleted or were you trying to document some expectation here?

mujoco_warp/_src/sensor.py
```diff
@@ -92,10 +132,33 @@ def _sensor_acc(m: Model, d: Data):
...
+      d.sensordata[worldid, adr + 2] = torque[2]
+    elif sensortype == int(SensorType.ACTUATORFRC.value):
       d.sensordata[worldid, adr] = _actuator_force(m, d, worldid, objid)
 
   if (m.sensor_acc_adr.size == 0) or (m.opt.disableflags & DisableBit.SENSOR):
     return
 
+  if wp.static(
+    np.isin(
+      m.sensor_type.numpy(),
```
erikfrey, line 437: it looks like this is a cuda device array?  if so we should avoid calling `numpy` on it as it would incur a device to host transfer.  if you'd like to use it this way you should do `wp.array(..., device='cpu')` in `io.py`

However you use this field in a kernel, so it needs to be on device.  This is a scenario where it's probably OK to just have a bool param as you had before.  We will clean this up further in #148

erikfrey (approved):
Looks good, just a few nits.

mujoco_warp/_src/sensor.py
erikfrey: We discussed this offline - it's possible that actually this works just fine because `wp.static` ensures the device->host transfer is only there during graph capture but not during actual stepping, but @thowell will confirm.

we also need to make sure this is JAX friendly.

mujoco_warp/_src/smooth.py
thowell: updated

mujoco_warp/_src/sensor_test.py
thowell: probably wanted to match style among the three bodies with each one having a specified position:
`pos="0 0 0"`, `pos="1 0 0"`, `pos="2 0 0"`

this body `pos` is no longer the default after the merge

PR #122, opened 2025-04-06
ten_velocity

thowell:
compute `ten_velocity` in `fwd_velocity`


mujoco_warp/_src/forward.py
```diff
@@ -496,6 +496,20 @@ def _actuator_velocity(
...
+    # TODO(team): sparse version
+
+    d.ten_velocity.zero_()
+
+    @kernel
+    def _tendon_velocity(d: Data):
+      worldid, tenid, dofid = wp.tid()
+      moment = d.ten_J[worldid, tenid]
+      qvel = d.qvel[worldid]
+      wp.atomic_add(d.ten_velocity[worldid], tenid, moment[dofid] * qvel[dofid])
```
adenzler-nvidia: Not sure if the effect of this would be visible in the overall performance, but given that this only loads 1d arrays, I think we could get away with using tile_reduce here instead of atomic adds. With 1D arrays we can handle a pretty big nv.

What do you think?

adenzler-nvidia (approved):
approving - one question but can be a follow up.

mujoco_warp/_src/forward.py
thowell: the implementation is updated to utilize tile operations with `block_dim=32`

adenzler-nvidia: Thanks - looks good!

PR #124, opened 2025-04-06
limit ball joint constraint

thowell:


mujoco_warp/_src/constraint.py
```diff
@@ -139,6 +141,58 @@ def _efc_limit_slide_hinge(
...
+  m: types.Model,
+  d: types.Data,
+  refsafe: bool,
+):
+  worldid, jntlimitedid = wp.tid()
+  jntid = m.jnt_limited_ball_adr[jntlimitedid]
+  qposadr = m.jnt_qposadr[jntid]
+
+  qpos = d.qpos[worldid]
+  quat = wp.quat(
```
Kenny-Vilella: Very minor, but maybe jnt_quat for clarity

mujoco_warp/_src/constraint.py
```diff
@@ -115,6 +116,7 @@ def _efc_limit_slide_hinge(
   active = pos < 0
 
   if active:
+    wp.atomic_add(d.nl, 0, 1)
```
Kenny-Vilella: Out of curiosity, except for testing, do we plan to use this information anywhere?
Because it will probably slow down the code, so wonder if it is worth it or if could do without it.

Kenny-Vilella (commented):
It looks good, I will make some test locally and then approve the PR.

Kenny-Vilella (approved):
All tests seems to work as expected.

mujoco_warp/_src/constraint.py
thowell: done

mujoco_warp/_src/constraint.py
thowell: the number of active limit constraints `nl` is utilized by the solver for elliptic friction cones https://github.com/google-deepmind/mujoco/blob/606f00f8024a96874ac9221f07e1114ed81e1f3c/mjx/mujoco/mjx/_src/solver.py#L178

thowell: the code is updated to call `atomic_add` with `d.nl` in the limit constraint kernels and then update `d.nefc` after the limit constraint kernel launches. with this approach it is not necessary to have two `atomic_add` calls for each active constraint.

mujoco_warp/_src/constraint.py
```diff
@@ -139,6 +142,58 @@ def _efc_limit_slide_hinge(
     )
 
 
+@wp.kernel
+def _efc_limit_ball(
+  m: types.Model,
+  d: types.Data,
+  refsafe: bool,
```
erikfrey: it kind of looks like you could just create `refsafe` directly inside of `_update_efc_row` and then you don't need to pass it around.

could you even do `refsafe = wp.static(m.opt.disableflags & ...` so that it's zero cost at runtime?

erikfrey (approved):
Looks great, just one nit and then we can get this in.

mujoco_warp/_src/constraint.py
thowell: the status of refsafe is now checked in `_update_efc_row`.

`wp.static` does not seem to work here so a todo is added for now.

PR #125, opened 2025-04-06
limit tendon constraint

thowell:
add limit tendon constraint


mujoco_warp/_src/constraint.py
```diff
@@ -139,6 +140,52 @@ def _efc_limit_slide_hinge(
...
+  active = pos < 0
+
+  if active:
+    wp.atomic_add(d.nl, 0, 1)
+    efcid = wp.atomic_add(d.nefc, 0, 1)
+    d.efc.worldid[efcid] = worldid
+
+    Jqvel = float(0.0)
+    for i in range(m.nv):
+      if dist_min < dist_max:
```
Kenny-Vilella: It looks that this if/else branch could be simplified by:
 J = wp.sign(dist_max - dist_min) * d.ten_J[worldid, tenid, i]

mujoco_warp/_src/constraint_test.py
```diff
@@ -101,6 +101,23 @@ def test_constraints(self, cone):
       _assert_eq(d.efc.pos.numpy()[: mjd.nefc], mjd.efc_pos, "efc_pos")
       _assert_eq(d.efc.margin.numpy()[: mjd.nefc], mjd.efc_margin, "efc_margin")
 
+  def test_limit_tendon(self):
+    """Test limit tendon constraints."""
+    for keyframe in range(-1, 1):
```
Kenny-Vilella, line 130: Out of curiosity, why starting by -1?

Kenny-Vilella (approved):
It looks good, just one minor change.

mujoco_warp/_src/constraint_test.py
thowell: -1 will reset to the nominal state where the limit should not be active

mujoco_warp/_src/constraint.py
thowell: updated to use `float(dist_min < dist_max) * 2.0 - 1.0` instead the if-else statement

thowell:
added fixed tendon case to `_efc_limit_tendon`

adenzler-nvidia:
@thowell anything blocking this? Or can we merge this?

thowell:
@adenzler-nvidia the pr is updated and should be ready to merge

adenzler-nvidia (approved):
Approving and merging. Thanks!

PR #126, opened 2025-04-06
add framelinacc and frameangacc sensors

thowell:
add framelineacc and frameangacc sensors

depends on #112


erikfrey (approved):
Looks good.

PR #127, opened 2025-04-07
CI: move continue-on-error to job metadata to avoid false negatives

adenzler-nvidia:
Compared to before, this change should now give you a red icon if any of the stages failed. Before linter failures would not show up on the PR overview because a successful pytest run (which happens afterwards) would give this a green icon.


Kenny-Vilella (approved):

adenzler-nvidia:
closing in favor if #128

PR #128, opened 2025-04-07
Improve CI: Separate job into multiple ones

Kenny-Vilella:


.github/workflows/ci.yml
```diff
@@ -22,18 +22,51 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install mujoco --pre -f https://py.mujoco.org/
+          pip install warp-lang --pre --upgrade -f https://pypi.nvidia.com/warp-lang/
+          pip install pytest
+          pip install ruff
```
adenzler-nvidia: I guess ruff does not need to be installed here now.

Kenny-Vilella: Yeap you are right

Kenny-Vilella:
@erikfrey  What do you think about separating the CI steps into multiple jobs?
 This is to make sure we do not merge PR with formatting or linting issue.
 
 There is also the choice to keep it as one job:
 https://github.com/google-deepmind/mujoco_warp/pull/127

erikfrey (approved):
Nice, this looks good to me, thanks.

PR #129, opened 2025-04-07
Sphere-box collision

### LINKED ISSUE #76: Sphere <> Box
Sphere <> box collision function.

camevor:
Based on MuJoCo C.

resolves #76


mujoco_warp/_src/collision_primitive.py
```diff
@@ -468,6 +470,57 @@ def plane_cylinder(
...
+  contact_pos = box.pos + box.rot @ pos
+  write_contact(
+    d,
+    contact_dist,
+    contact_pos,
+    make_frame(contact_normal),
+    margin,
+    geom_indices,
+    worldid,
+  )
```
adenzler-nvidia, line 858: wondering if we should have another function here that does not do the margin check twice? I think in other functions we started inlining the code again..

camevor: Good point, the margin check comes right before the atomic add in `write_contact`, too. `write_contact` is inlined for gjk in `collision_convex.py`, which can add up to four contacts at once. Maybe we can remove the margin check from `write_contact`, but keep it for pairs generating a single contact for the atomic add with the `nconmax` overflow check? Might also warrant a separate PR to do the whole bunch.

adenzler-nvidia: The inital point of having the function was to make sure we do not forget the margin check. So we would go back on that decision, which I think is fair for cases like this where you can avoid a lot of computation.

My preferred solution would be to have 2 functions, with write_contact doing the margin check and then calling the one with the overflow check. We could then add an assertion to the function not checking the margin, and run the tests in debug mode.

But for sure this is something that should be done in a separate MR, if at all. Then we can discuss there how we want to go about it.

mujoco_warp/_src/collision_primitive.py
```diff
@@ -23,6 +23,7 @@
 from .types import GeomType
 from .types import Model
 
+_TINY_VAL = 1e-6
```
thowell: can the code instead utilize `MJ_MINVAL` and then this variable can be removed?

camevor: Not really, as `MJ_MINVAL` is `1e-15`, whereas this value is much more generous. It's used in several places. I could inline it as a magic number (as seen in other places), or put a single definition in types, though.

erikfrey: Just so I understand - does `sphere_box` break somehow if you use `MJ_MINVAL` instead of `TINY_VAL`?  I fetched your branch and tried it and the unit tests still pass.  e.g. I changed the function to this line:

```py
  if dist <= MJ_MINVAL:
```

And the sphere box unit tests still pass.

erikfrey (approved):
This looks good to go, just a question about `MJ_MINVAL`

erikfrey (approved):
Looks great - thanks @camevor !

mujoco_warp/_src/collision_primitive.py
camevor: Sorry, I got these values mixed up somehow. Thanks for both of your comments!
Changed it to `MJ_MINVAL` (as in the mjc implementation).

Issue #50, opened 2025-03-31
Remove @wp.struct from Model, Data


erikfrey:
Defining Model and Data as `wp.struct` was convenient to start but presents three problems:

* Passing large structs to kernel launches may eventually overflow the max parameter size, which is 32kb
* It may also inhibit backwards compatibility with older versions of CUDA, before 12.8 it was 4kb
* The JAX<>Warp interop layer understands arrays but not structs.  In order to pass JAX arrays cleanly to Warp, we must be able to handle individual fields separately.

This change will make the kernel definitions more verbose as they'll now take as input the list of arrays defined as Model and Data fields, rather than just a Model and Data struct.

adenzler-nvidia:
I just measured, we're now at 7KB for both model and data structs. Looking forward to this change, let me know if you need some more hands with the kernel argument porting. Although I suppose an AI should do that work for us.

adenzler-nvidia:
Looks like #108 is blocked by this as well. Currently there is no support for storing a wp.context.device in a wp.struct, but we need to maintain that information somehow. Not urgent though.

PR #130, opened 2025-04-07
Device Management in Multi-GPU systems

### LINKED ISSUE #108: Device Management in Multi-GPU Systems
we are quite lax at the moment, rarely specifying the device we actually want to have our data allocated on and kernels running. That's likely going to break things as soon as someone is using a system with multiple GPUs.

We need to make sure all relevant data and computation is happening on the device the user specified. This means we need:
1. a proper API to capture which device this should run on
2. all our code hardened such that it defaults to the device being set from the user.
3. adequate testing such that we're not breaking this all the time.

For now let's focus on using 1 GPU in a multi-GPU system, we can worry about potential multi-GPU usecases where we partition the simulation at a later stage.

adenzler-nvidia:
WIP - DO NOT MERGE! Depends on #50 to work.

Closes #108 

This only takes care of allocating all data and running the kernels on 1 device. This does not handle potential multi-GPU setups. The strategy is to specify the target device in the put_model call and then use that device for all CUDA operations.


adenzler-nvidia:
Will take this on again this week. Might come in a different PR, let's see.

erikfrey:
Agreed this seems dependent on #50, which is dependent on #169         :-)

Working to zip this up!

adenzler-nvidia:
closing in favor of #182 182

PR #117, opened 2025-04-05
Equality joint

Andrew-Luo1:
Testing speed on RTX4060Ti

**Nominal:**
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --event_trace=True

```
Summary for 8192 parallel rollouts

 Total JIT time: 0.44 s
 Total simulation time: 9.86 s
 Total steps per second: 831,236
 Total realtime factor: 4,156.18 x
 Total time per step: 1203.03 ns

Event trace:

step: 1201.50
  forward: 1197.87
    fwd_position: 189.39
      kinematics: 46.27
      com_pos: 15.11
      crb: 44.53
      factor_m: 28.46
      collision: 14.60
      make_constraint: 32.00
      transmission: 7.30
    fwd_velocity: 53.23
      com_vel: 19.74
      passive: 1.68
      rne: 19.58
    fwd_actuation: 15.17
    fwd_acceleration: 27.54
      xfrc_accumulate: 8.86
      solve_m: 16.75
    solve: 911.68
      mul_m: 15.52
      _linesearch: [ 75.63, 43.66, 29.96, 26.95, 25.62, 25.06, 24.64, 24.36, 24.18, 24.11 ]
        mul_m: [ 16.05, 3.44, 1.85, 1.65, 1.62, 1.63, 1.63, 1.62, 1.63, 1.63 ]
  euler: 3.23
```

**Adding an equality constraint to humanoid.xml:**
```xml
<equality>
  <joint   name="joint"    joint1="elbow_left"    joint2="elbow_right"/>
</equality>
```

```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --event_trace
=True
Summary for 8192 parallel rollouts

 Total JIT time: 23.15 s
 Total simulation time: 9.87 s
 Total steps per second: 829,820
 Total realtime factor: 4,149.10 x
 Total time per step: 1205.08 ns

Event trace:

step: 1203.52
  forward: 1199.89
    fwd_position: 189.64
      kinematics: 46.36
      com_pos: 15.14
      crb: 44.60
      factor_m: 28.49
      collision: 14.63
      make_constraint: 32.03
      transmission: 7.30
    fwd_velocity: 53.30
      com_vel: 19.77
      passive: 1.68
      rne: 19.62
    fwd_actuation: 15.15
    fwd_acceleration: 27.56
      xfrc_accumulate: 8.86
[...]

PR #137, opened 2025-04-09
Aloha pot test scene.

erikfrey:
Aloha and pot test scene for feature prioritization and benchmarking.  The scene has been simplified (e.g. textures ripped out) and compressed to ~3.5MB so that we can keep it in git without too much guilt.

This scene has 26 bodies, 23 dofs, 204 geoms.  Many of the geoms are meshes with ~100 to ~1000 verts.

The scene turns on elliptic cone friction and sets a high impratio.  This scene will be a good test of broadphase, GJK, and solver early return.

The scene almost already works - it fails to load because of the equality constraints, but those will land in #117.  After that, the scene appears to simulate but there are probably some physics bugs to iron out.

image


adenzler-nvidia (approved):
Thank you - approving this without looking at every detail but it simulates for me and looks ok in the viewer.

hartikainen:
> compressed to ~3.5MB so that we can keep it in git without too much guilt

Would it make sense to use [`git-lfs`](https://git-lfs.com/) or something similar for storing binary files? 🙂

erikfrey:
That's a cool suggestion!  If we end up with many more large scenes we can definitely try this out.

PR #144, opened 2025-04-10
add support for elliptic cones to contact_force and update test_contact_force

thowell:


mujoco_warp/_src/support.py
```diff
@@ -257,25 +258,25 @@ def contact_force(
   m: Model, d: Data, contact_id: int, to_world_frame: bool = False
 ) -> wp.spatial_vector:
   """Extract 6D force:torque for one contact, in contact frame by default."""
-  efc_address = d.contact.efc_address[
-    contact_id, 0
-  ]  # 0 in second dimension to get the normal force
   force = wp.spatial_vector(0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
-
-  if efc_address >= 0:
```
nvtw, line 265: Is this efc address check not required? MJX has `return force * (efc_address >= 0)`

nvtw:
I approve. But I don't see a button that I could press to let that github know

mujoco_warp/_src/support.py
thowell: restored `efc_address` check and added `contact_id` check

erikfrey (approved):
Nice, looks good.

PR #148, opened 2025-04-11
[WIP] API changes to address multiple issues.

erikfrey:
Here are some proposed API changes, right now only applied to `kinematics` as a preview - once we work out the kinks we'll roll this out to the rest of the codebase.

Summary of changes:

* Moving kernel functions to the module level as my understanding is this is better for the warp cache (is this still the case?)
* Give kernel functions explicit in/out params, even if the params are duplicated (e.g. `xpos_in`, `xpos_out`).  This is helpful for:
  * Moving towards making MJWarp differentiable
  * Easier integration into JAX that wants control of providing separate in/out arrays.
* Changing `Model` and `Data` from `wp.struct` to `dataclasses.dataclass`
  * Should address the kernel parameter size issue we're hitting (4kb for older CUDA versions, 32kb for CUDA 12.8)
  * One ergonomic improvement is we can simplify tree-nested arrays, e.g. the tree bodyids is now a simple tuple field and easier to process.

The downside of this change is that the kernel signatures are much more verbose, and explicitly writing out the kernel parameters is slow and could be a vector for bugs.  I wonder if we should consider helper classes or some kind of code automation to make it easier to write bug-free kernels.

Would love any and all thoughts on these changes, very experimental still and I want everyone to be on board before we apply this change to the rest of the repo.

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -24,6 +24,8 @@ from .types import Data
 from .types import DisableBit
 from .types import JointType
 from .types import Model
+from .types import NewData
+from .types import NewModel
 from .types import TrnType
 from .types import array2df
 from .types import array3df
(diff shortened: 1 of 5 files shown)


mujoco_warp/_src/smooth.py
```diff
@@ -149,6 +151,195 @@ def site_local_to_global(m: Model, d: Data):
...
+  xquat_in: wp.array2d(dtype=wp.quat),
+  xmat_in: wp.array2d(dtype=wp.mat33),
+  # Data out
+  xpos_out: wp.array2d(dtype=wp.vec3),
+  xquat_out: wp.array2d(dtype=wp.quat),
+  xmat_out: wp.array2d(dtype=wp.mat33),
+  xipos_out: wp.array2d(dtype=wp.vec3),
+  ximat_out: wp.array2d(dtype=wp.mat33),
+  xanchor_out: wp.array2d(dtype=wp.vec3),
+  xaxis_out: wp.array2d(dtype=wp.vec3),
```
adenzler-nvidia, line 200: I like the structure with the comments, and also the in/out postfixes.

What I wonder (on a lower, CUDA-level) is that now we have argument arrays that are aliasing. We should check how warp is passing those arguments and how the compiler deals with them, because we cannot directly mark them const here.

OTOH, we had aliasing issues before, those are coming in as as you have 2 pointers of the same type. @mmacklin can probably comment best here.

mujoco_warp/_src/smooth.py
```diff
@@ -149,6 +151,195 @@ def site_local_to_global(m: Model, d: Data):
...
+      inputs=[m.geom_bodyid, m.geom_pos, m.geom_quat, d.xpos, d.xquat],
+      outputs=[d.geom_xpos, d.geom_xmat],
+    )
+
+  if m.nsite:
+    wp.launch(
+      kinematics_local_to_global,
+      dim=(d.nworld, m.nsite),
+      inputs=[m.site_bodyid, m.site_pos, m.site_quat, d.xpos, d.xquat],
+      outputs=[d.site_xpos, d.site_xmat],
```
adenzler-nvidia, line 339: it's a bit late for that, but we might extend the line length a bit to have a few more arrays fit on one line?

mujoco_warp/_src/smooth.py
```diff
@@ -149,6 +151,195 @@ def site_local_to_global(m: Model, d: Data):
     wp.launch(site_local_to_global, dim=(d.nworld, m.nsite), inputs=[m, d])
 
 
+@kernel
```
adenzler-nvidia, line 154: I think the real benefit from moving the kernel out of the functions comes once you start grouping all of them at the top - at which point also the @kernel is not necessary anymore and we could move to the standard wp.kernel approach again. 

@eric-heiden is that correct? Or are there other benefits of this I'm missing?

eric-heiden: Yes, you can just use `@wp.kernel` then to group these kernels together in the Python module (default behavior) instead of having `@kernel` create a new module for every Python function in which the kernels were implemented locally before.

eric-heiden:
I think these changes are very useful. By explicitly using the `outputs` argument to `wp.launch` we will be able to debug gradients more easily in the future as well, e.g. via [`tape.visualize()`](https://nvidia.github.io/warp/modules/differentiability.html#visualizing-computation-graphs). Another advantage is that some of these kernels can be reused more easily since users aren't required to create these massive `Model` and `Data` structs anymore but just allocate the necessary arrays.

mujoco_warp/_src/smooth.py
mmacklin: Alain, I guess your comment is specifically about whether or not we can inform the compiler that arrays do not alias e.g.: using `__restrict__` keywords?

Currently Warp doesn't take advantage of these, since arrays can in general represent arbitrary locations / slices / etc. We could consider an extension to the argument type definition, e.g.: `wp.array(noalias=True)` or some such.

I also wonder if from an ergonomic point of view it would be useful to mark arrays as read only / write only, e.g.: `wp.array(mode="r")`, `wp.array(mode="rw")`, etc. Let me know if you have any thoughts.

adenzler-nvidia: Yeah I was thinking about __restrict__ - need to spend more time looking at warp internals. I do like the readonly/noalias hints, but we should check how much we actually gain using these. Otherwise it's just unnecessary complexity.

erikfrey:
OK!  Thanks all - proceeding with this design, hope to update the PR soon with more of this implemented.

mujoco_warp/_src/smooth.py
yuvaltassa: As an aside, I've been playing on and off with `restrict` in MuJoCo for a few months (only clang), and am finding huge differences across architectures. On x86 I can barely see a difference, only in rare cases. On ARM it's like magic, not rare to see 2x speed gains. Godbolt makes it clear that the compiler takes the hint on both archs, but somehow ARM assembly makes better use of it. Better branch prediction? 🤷🏻‍♂️

I wonder where your GPUs fall on that spectrum?

adenzler-nvidia: There is this old blogpost about it: https://developer.nvidia.com/blog/cuda-pro-tip-optimize-pointer-aliasing/

I actually might stand corrected here - probably this is not a problem anymore. I posted the snippet from the blogpost into godbolt and the generated code is the same. So either the compiler assumes restrict by default nowadays or it can figure things out on their own much better.

My advice in CUDA land is usually to add restrict everywhere were it applies just to be safe. That being said, I never encountered a codebase where I came in, added a few restricts and it magically got x times faster.

PR #151, opened 2025-04-14
update collision driver

thowell:
- `nxn_broadphase` and `sap_broadphase` utilize `_sphere_filter` to check for sphere/plane-sphere/plane collisions
- functionality from routines `find_first_greater_than` and `find_right_most_index_int` is replaced with `_binary_search`
- sort now utilizes `wp.utils.segmented_sort_pairs` and a todo is added for tile sort
- i think sorting the geom indices is sufficient and that it is not necessary to create an array of sorted geom elements themselves so `sap_geom_sort` is removed along with `reorder_bounds_sphere_kernel` and supporting functions


nvtw:
tile_sort was merged around two weeks ago. It might not be in warp 1.7 but it will be in warp 1.8. And the pre-release warp we're using should already have it, so we could switch to tile_sort.
I have not looked at it in detail but tile_sort might also allow us to do some kernel fusion.

nvtw:
I did not check every line but I tested the change and it seems to work fine. I do approve the MR but I don't see a button I could press to make that official.

nvtw (approved):

thowell:
@nvtw @adenzler-nvidia should we update the implementation to utilize tile sort in this pr or a future pr?

nvtw:
Let's add tile_sort in a future MR as this MR seems to fix some issues that block my convex vs plane collision MR.

erikfrey (approved):
Nice.

PR #152, opened 2025-04-15
Ensure tests are deterministic.

btaba:


thowell (approved):

erikfrey:
Do we want this?  If a test is flakey, it could be the result of an actual bug.  We've seen this before, e.g. JAX changes something about its float precision and we get flakey tests.  Setting a seed could hide real bugs.

PR #153, opened 2025-04-16
add humanoid and large auto-generated terrain

Andrew-Luo1:
image

- Collisions-only version of Apollo humanoid.
- Procedurally generated landscape. Generates a NxM tile of sub-landscapes: flat, pyramid, inverted pyramid
- Framework can be easily extended to include heightfield patches once those are released.

**Acknowledgements**

Isaaclab
- Config design
- Pyramid and inverted pyramid logic

[Unitree Mujoco](https://github.com/unitreerobotics/unitree_mujoco/tree/main/terrain_tool)
- Terrain generation tool base class

**Issues**

Even with nconmax set to 512*1024 in put_data, n_worlds = 1 and a limited world size, collisions with pyramids don't look correct. (This doesn't happen in cmujoco).

image

image


btaba:
This LGTM modulo the bug in MjWarp which is perhaps independent of this PR. Can you create a bug with repro?

btaba (approved):

btaba (changes requested):
need to see about the LICENSE, hold on

btaba (changes requested):
Ok @Andrew-Luo1 I think from a licensing perspective ,it's easier to check this in without the terrain_generator code (for now). Feel free to make your own repo though (at that point it might make more sense to use MjSpec and not necessarily derive from unitree's codepath)

Andrew-Luo1:
Made new PR without terrain generation code

PR #160, opened 2025-04-17
Implement WELD equality constraint

camevor:
Implements the `WELD` equality constraint, based on the existing equality constraints.

### DIFF
--- a/mujoco_warp/_src/constraint.py
+++ b/mujoco_warp/_src/constraint.py
@@ -21,8 +21,6 @@ from . import math
 from . import types
 from .warp_util import event_scope
 
-wp.set_module_options({"enable_backward": False})
-
 
 @wp.func
 def _update_efc_row(
--- a/mujoco_warp/_src/constraint.py
+++ b/mujoco_warp/_src/constraint.py
@@ -21,6 +21,7 @@ from . import math
 from . import types
 from .warp_util import event_scope
 
+wp.set_module_options({"enable_backward": False})
 
 @wp.func
 def _update_efc_row(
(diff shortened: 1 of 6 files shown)


mujoco_warp/_src/constraint.py
```diff
@@ -124,7 +126,8 @@ def _efc_equality_connect(
 
   efcid = wp.atomic_add(d.nefc, 0, 3)
   wp.atomic_add(d.ne, 0, 3)
-  d.efc.worldid[efcid] = worldid
```
camevor, line 127: Small fix for the `CONNECT` constraint.

thowell: nice fix!

mujoco_warp/_src/constraint.py
```diff
@@ -268,6 +269,116 @@ def _efc_friction(
...
+
+  data = m.eq_data[i_eq]
+  anchor1 = wp.vec3(data[0], data[1], data[2])
+  anchor2 = wp.vec3(data[3], data[4], data[5])
+  relpose = wp.quat(data[6], data[7], data[8], data[9])
+  torquescale = data[10]
+
+  if is_site:
+    # body1id stores the index of site_bodyid.
+    body1id = m.site_bodyid[m.eq_obj1id[i_eq]]
```
thowell: can we reuse `obj1id` and `obj2id` instead of reading from global memory here and below?

mujoco_warp/_src/constraint.py
```diff
@@ -268,6 +269,116 @@ def _efc_friction(
...
+
+  for i in range(3):
+    _update_efc_row(
+      m,
+      d,
+      efcid + i,
+      cpos[i],
+      pos_imp,
+      invweight_t,
+      m.eq_solref[i_eq],
```
thowell: would it be more efficient to read `eq_solref` and `eq_solimp` from global memory and write into local variables and then utilize these local variables in both loops?

mujoco_warp/_src/types.py
```diff
@@ -345,6 +347,26 @@ class vec11f(wp.types.vector(length=11, dtype=wp.float32)):
 array3df = wp.array3d(dtype=wp.float32)
 
 
+class ObjType(enum.IntEnum):
```
thowell: it looks like this is a duplicate, please remove

mujoco_warp/_src/constraint.py
camevor: Agreed, that could be more efficient.

mujoco_warp/_src/types.py
camevor: Thanks for spotting this!

mujoco_warp/_src/constraint.py
camevor: Yes, changed it to re-use the loaded value.

thowell (approved):
lgtm, thanks!

PR #164, opened 2025-04-22
rearrange geom types in GJK/EPA to be in order

kbayes:
Fixes bug where certain geoms won't collide in GJK/EPA (e.g. box-mesh collisions).

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -131,20 +131,20 @@ def _gjk_support(
 
 
 convex_collision_functions = {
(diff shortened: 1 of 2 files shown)


yuvaltassa:
The logic is that the first value is <= than the second one, right? Maybe add a test for this?

kbayes:
@yuvaltassa yes, there's already a test.

yuvaltassa:
Ah ya sorry. GH diff is still confusing for me...

adenzler-nvidia (approved):
Thank you! Is there an issue for the bug? If so, please link and close.

PR #165, opened 2025-04-22
kernel_copy -> wp.copy

thowell:
- `warp_util.kernel_copy` is replaced with `wp.copy`
- `jax_test.py` (based on `contrib/jax_unroll.py`) is added to test that `mujoco_warp.step` is compatible with JAX
- `pyproject.toml` optional dependencies are updated and included cpu and cuda versions
- minor updates to `pyproject.toml` information


mujoco_warp/_src/jax_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+import jax
+import warp as wp
+from jax import numpy as jp
+from warp.jax_experimental.ffi import jax_callable
+
+import mujoco_warp as mjwarp
+from mujoco_warp._src.test_util import fixture
+
+if jax.default_backend() == "gpu":
```
erikfrey: does this run when you run pytest?

Can we structure this as an absltest?

Also, can you leave a comment somewhere so that it's really clear that this is temporary and that it will be removed after we land MJX:Warp?

pyproject.toml
```diff
@@ -27,22 +27,35 @@ classifiers = [
...
     "etils[epath]",
     "mujoco",
-    "pytest",
+    "numpy",
     "warp-lang",
 ]
 
+[project.optional-dependencies]
+dev = [
+    "absl-py",
```
erikfrey: We ship the tests as part of the dist, so probably we should keep absl as part of the core dependencies.

Obviously no outside user should be doing `from mujoco_warp import smooth_test` but they could... so might as well leave the dep for now.  Anyway it's quite a small dep.

pyproject.toml
```diff
@@ -27,22 +27,35 @@ classifiers = [
...
 dependencies = [
-    "absl-py",
     "etils[epath]",
     "mujoco",
-    "pytest",
+    "numpy",
     "warp-lang",
 ]
 
+[project.optional-dependencies]
```
erikfrey, line 37: let's also leave a not or a TODO: that this is temporary and will be undone after we land MJX:Warp

erikfrey (approved):
Looks great, just a few nits.

mujoco_warp/_src/jax_test.py
thowell: yes this will run when we run pytest

the test is now structured as an absltest

added a todo noting that this test is temporary and will be removed after MJX:Warp lands

pyproject.toml
thowell: done

pyproject.toml
thowell: done

PR #166, opened 2025-04-22
rne_postconstraint with connect and weld equality constraints

thowell:


mujoco_warp/_src/smooth.py
```diff
@@ -660,6 +661,93 @@ def _cfrc_ext(m: Model, d: Data):
...
+  def _cfrc_ext_equality(m: Model, d: Data):
+    eqid = wp.tid()
+
+    ne_connect = d.ne_connect[0]
+    ne_weld = d.ne_weld[0]
+
+    if eqid >= ne_connect // 3 + ne_weld // 6:
+      return
+
+    is_connect = eqid < ne_connect // 3
```
adenzler-nvidia: Division is expensive, so it probably makes sense to store ne_connect in a temp variable here.

adenzler-nvidia (approved):
Looks good, 1 small comment.

mujoco_warp/_src/smooth.py

PR #169, opened 2025-04-23
Kernel analyzer

erikfrey:
Kernel Analyzer checks warp kernel definitions and body to ensure correctness and conformity.  It comes with both a CLI (which can be used within github CI) and also a vscode plugin for automatic kernel issue highlighting.

Example output:

```bash
$ python contrib/kernel_analyzer/kernel_analyzer/cli.py --files contrib/kernel_analyzer/test_file.py --types mujoco_warp/_src/types.py 
contrib/kernel_analyzer/test_file.py:36:"hi" default not allowed.
contrib/kernel_analyzer/test_file.py:37:"args" varargs not allowed.
contrib/kernel_analyzer/test_file.py:38:"kwargs" kwargs not allowed.
contrib/kernel_analyzer/test_file.py:33:"m" should be preceded by comment: # In:
contrib/kernel_analyzer/test_file.py:33:"m" type mismatch, expected int, float, bool, wp.array.
...
```

Example vscode integration:

image


adenzler-nvidia:
Thanks - did not have a look yet but coming here from **https://github.com/google-deepmind/mujoco_warp/issues/108**. We might want to add the explicit device argument check as well here.

adenzler-nvidia:
Heads-up: I can't comment too much on the implementation as a lot of that stuff is unfamiliar to me.

I think this makes a lot of sense, we can add this in CI and enforce that we're not breaking things left and right. I also wouldn't be too concerned about only supporting vscode as long as we have the command line version as well.

It probably just needs a lot of testing to catch all of the corner cases, but we can fix these on-the-go. Maybe a pragma/decorator to explicitly allow exemptions is needed. I'd say let's start using it and see what breaks/how cumbersome it is.

contrib/kernel_analyzer/README.md
```diff
@@ -0,0 +1,55 @@
...
+# VSCode plugin
+
+Enjoy kernel analysis directly within vscode.
+
+## Installing kernel analyzer
+
+1. Inside vscode, navigate to `contrib/kernel_analyzer/`
+2. Right click on `kernel-analyzer-{version}.vsix` file
+3. Select "Install Extension VSIX"
+
```
btaba, line 34: 4. pip install -r requirements.txt
5. Add python path and absolute path to types.py to the extension settings

btaba: And if there are any issues, to check the vscode output, look for kernelAnalyzer

contrib/kernel_analyzer/kernel_analyzer/ast_analyzer.py
```diff
@@ -0,0 +1,268 @@
...
+    m = field_pattern.match(line)
+    if not class_name or not m or in_docstring:
+      continue
+
+    ret.setdefault(class_name, []).append((m.group(1), m.group(2)))
+
+  return ret
+
+
+def analyze(source: str, filename: str, type_source: str) -> List[Issue]:
```
btaba, line 126: much improved, really nice code!

contrib/kernel_analyzer/kernel_analyzer/cli.py
```diff
@@ -0,0 +1,95 @@
...
+
+import sys
+from pathlib import Path
+
+import ast_analyzer
+from absl import app
+from absl import flags
+from absl import logging
+
+FLAGS = flags.FLAGS
```
btaba: nit: unused

contrib/kernel_analyzer/client/src/extension.ts
```diff
@@ -0,0 +1,84 @@
+import * as path from 'path';
```
btaba, line 16: does this need a licenses?

contrib/kernel_analyzer/somefile.py
```diff
@@ -0,0 +1,83 @@
+import mujoco
```
btaba: is this file needed?

btaba (changes requested):
LGTM minus some small stuff

contrib/kernel_analyzer/README.md
btaba: Or replace 4 with optional deps in pyproject.toml

contrib/kernel_analyzer/kernel_analyzer/ast_analyzer.py
erikfrey: Cheerio.

erikfrey:
OK @btaba PTAL

btaba (approved):

PR #178, opened 2025-04-24
Activate `collision_test`

camevor:
The `collision_driver_test.test_collision` test case misses a collision step. This causes it to pass regardless of the collision pipeline. This change runs the collision pipeline and temporarily disables the `boxmesh` test fixture that now fails. It also removes a duplicate statement.


kbayes:
I noticed that the tests always pass when I did my PR. Thanks for taking care of this.

kbayes (approved):

PR #181, opened 2025-04-24
framelinvel and frameangvel sensors

thowell:
- add framelinvel and frameangvel sensors
- fix test_sensor: set keyframe


mujoco_warp/_src/sensor_test.py
```diff
@@ -136,6 +136,14 @@ def test_sensor(self):
           <jointvel joint="slide"/>
           <actuatorvel actuator="slide"/>
           <ballangvel joint="ballquat"/>
+          <framelinvel objtype="body" objname="body9"/>
```
yuvaltassa, line 139: Should we have some sensors with a reference frame object? (shallow reading makes me think this is supported)

PR #185, opened 2025-04-24
add camprojection sensor

thowell:


mujoco_warp/_src/sensor.py
```diff
@@ -27,6 +28,108 @@
...
+    1.0,
+    0.0,
+    0.0,
+    0.0,
+    0.0,
+    0.0,
+  )
+
+  # projection matrix (3 x 4): product of all 4 matrices
+  proj = image @ focal @ rotation @ translation
```
adenzler-nvidia, line 149: is there no way to do these calculations without allocating all the data in matrices?

thowell: directly computing the matrix `proj` should be possible. added a todo.

adenzler-nvidia (approved):

PR #186, opened 2025-04-24
sensor cutoff

thowell:


mujoco_warp/_src/sensor.py
```diff
@@ -225,18 +263,20 @@ def _sensor_pos(m: Model, d: Data):
...
-      quat = _frame_quat(m, d, worldid, objid, objtype, refid)
-      d.sensordata[worldid, adr + 0] = quat[0]
-      d.sensordata[worldid, adr + 1] = quat[1]
-      d.sensordata[worldid, adr + 2] = quat[2]
-      d.sensordata[worldid, adr + 3] = quat[3]
+      frame_quat = _frame_quat(m, d, worldid, objid, objtype, refid)
+      d.sensordata[worldid, adr + 0] = frame_quat[0]
+      d.sensordata[worldid, adr + 1] = frame_quat[1]
+      d.sensordata[worldid, adr + 2] = frame_quat[2]
+      d.sensordata[worldid, adr + 3] = frame_quat[3]
```
adenzler-nvidia, line 270: does frame_quat not support cutoff?

adenzler-nvidia (approved):
Thanks, 1 question. Looks good otherwise.

mujoco_warp/_src/sensor.py
thowell: quaternion datatype (https://mujoco.readthedocs.io/en/latest/APIreference/APItypes.html#mjtdatatype) is not supported by the mujoco refererence for sensor cutoff (https://github.com/google-deepmind/mujoco/blob/fb30e0e00c4e5a45e66e32e73f73f038a5000ee9/src/engine/engine_sensor.c#L40)

thowell: additionally, axis datatype is not supported by the mujoco reference for sensor cutoff either

PR #189, opened 2025-04-25
updates for gjk

thowell:


mujoco_warp/_src/collision_convex.py
```diff
@@ -61,75 +58,76 @@
...
-  local_dir = wp.transpose(info.rot) @ dir
-  if typeGeom == int(GeomType.SPHERE.value):
-    support_pt = info.pos + info.size[0] * dir
-  elif typeGeom == int(GeomType.BOX.value):
-    res = wp.cw_mul(wp.sign(local_dir), info.size)
-    support_pt = info.rot @ res + info.pos
-  elif typeGeom == int(GeomType.CAPSULE.value):
-    res = local_dir * info.size[0]
+  local_dir = wp.transpose(geom.rot) @ dir
+  if geom_type == int(GeomType.SPHERE.value):
```
hartikainen, line 70:
```suggestion
  if geom_type == GeomType.SPHERE:
```

adenzler-nvidia: Did that work? I would be surprised, we had to do the .value way all over the codebase due to limitations with enums in warp.

thowell:
@adenzler-nvidia this pr should just be stylistic updates

mujoco_warp/_src/collision_convex.py
hartikainen: I didn't actually try this in the warp code base but just made the comment while browsing the PRs/commits :slightly_smiling_face: It should be the typical way of comparing enums in python though. I don't want to add any unnecessary noise here so feel free to ignore this if it doesn't work for whatever reason.

thowell: @hartikainen i think the suggested change is not currently compatible with `enum.IntEnum` and @wp.kernel or @wp.func

PR #190, opened 2025-04-25
sensor_subtree_vel and sensor_rne_postconstraint

thowell:
the current logic for calling `subtree_vel` or `rne_postconstraint` by sensor functions is not compatible with cuda graph capture.

this pr add fields `sensor_subtree_vel` and `sensor_rne_postconstraint` to `Model` that determine if `subtree_vel` or `rne_postconstraint`, respectively, are required by sensors. these fields are compute in `io.put_model`.

this update makes the relevant sensors compatible with cuda graph capture.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -602,6 +602,20 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     ndim=1,
   )
 
+  m.sensor_subtree_vel = np.isin(
+    mjm.sensor_type,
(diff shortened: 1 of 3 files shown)


adenzler-nvidia (approved):
Thanks - looks good. I guess that one is on me, I should have caught this during review. Wondering if we need a test config that uses graph capture.

thowell:
@adenzler-nvidia it would probably be useful to have a test for cuda graph capture. opened an issue to test cuda graph capture: https://github.com/google-deepmind/mujoco_warp/issues/196.

PR #191, opened 2025-04-27
inertia box fluid model

thowell:
adds inertia box fluid model
https://mujoco.readthedocs.io/en/latest/computation/fluid.html#inertia-model
https://github.com/google-deepmind/mujoco/blob/d3bc0544d39c0933067e493f4c1c7ebfd15ae9f3/src/engine/engine_passive.c#L527


mujoco_warp/_src/passive.py
```diff
@@ -100,9 +104,112 @@ def _damper_passive(m: Model, d: Data):
...
+
+  if m.opt.wind[0] or m.opt.wind[1] or m.opt.wind[2]:
+    # subtract translational component from body velocity
+    lvel_force -= rotT @ m.opt.wind
+
+  lfrc_torque = wp.vec3(0.0)
+  lfrc_force = wp.vec3(0.0)
+
+  has_viscosity = m.opt.viscosity > 0.0
+  has_density = m.opt.density > 0.0
```
adenzler-nvidia: These could be wrapped in wp.static?

mujoco_warp/_src/passive.py
```diff
@@ -100,9 +104,112 @@ def _damper_passive(m: Model, d: Data):
...
+  torque = wp.spatial_top(cvel)
+  force = wp.spatial_bottom(cvel)
+  subtree_com = d.subtree_com[worldid, m.body_rootid[bodyid]]
+  dif = pos - subtree_com
+  force -= wp.cross(dif, torque)
+
+  lvel_torque = rotT @ torque
+  lvel_force = rotT @ force
+
+  if m.opt.wind[0] or m.opt.wind[1] or m.opt.wind[2]:
```
adenzler-nvidia: not sure if there is a big benefit to this, but could be wp.max(m.opt.wind) > 0.0.

Or I guess this information is even statically available, so maybe wrap in wp.static? actually, you're not running that code right now anyway if all of them are zero, are you going to reuse this function in a different kernel?

mujoco_warp/_src/passive.py
```diff
@@ -100,9 +104,112 @@ def _damper_passive(m: Model, d: Data):
...
+    # subtract translational component from body velocity
+    lvel_force -= rotT @ m.opt.wind
+
+  lfrc_torque = wp.vec3(0.0)
+  lfrc_force = wp.vec3(0.0)
+
+  has_viscosity = m.opt.viscosity > 0.0
+  has_density = m.opt.density > 0.0
+
+  if has_viscosity or has_density:
```
adenzler-nvidia: So this can be wp.static as well

mujoco_warp/_src/passive.py
```diff
@@ -100,9 +104,112 @@ def _damper_passive(m: Model, d: Data):
...
+    # force
+    lfrc_force -= wp.vec3(
+      0.5 * m.opt.density * box1 * box2 * wp.abs(lvel_force[0]) * lvel_force[0],
+      0.5 * m.opt.density * box0 * box2 * wp.abs(lvel_force[1]) * lvel_force[1],
+      0.5 * m.opt.density * box0 * box1 * wp.abs(lvel_force[2]) * lvel_force[2],
+    )
+
+    # torque
+    scl = m.opt.density / 64.0
+    box0_pow4 = box0 * box0 * box0 * box0
```
adenzler-nvidia: We have wp.pow for this - wp.pow(box0, 4)

mujoco_warp/_src/passive.py
```diff
@@ -100,9 +104,112 @@ def _damper_passive(m: Model, d: Data):
...
+  if has_density:
+    # force
+    lfrc_force -= wp.vec3(
+      0.5 * m.opt.density * box1 * box2 * wp.abs(lvel_force[0]) * lvel_force[0],
+      0.5 * m.opt.density * box0 * box2 * wp.abs(lvel_force[1]) * lvel_force[1],
+      0.5 * m.opt.density * box0 * box1 * wp.abs(lvel_force[2]) * lvel_force[2],
+    )
+
+    # torque
+    scl = m.opt.density / 64.0
```
adenzler-nvidia: interesting in the source of that number - but either way this can also be static.

mujoco_warp/_src/passive.py
```diff
@@ -100,9 +104,112 @@ def _damper_passive(m: Model, d: Data):
...
+    box0 = wp.sqrt(wp.max(MJ_MINVAL, inertia[1] + inertia[2] - inertia[0]) * scl)
+    box1 = wp.sqrt(wp.max(MJ_MINVAL, inertia[0] + inertia[2] - inertia[1]) * scl)
+    box2 = wp.sqrt(wp.max(MJ_MINVAL, inertia[0] + inertia[1] - inertia[2]) * scl)
+
+  if has_viscosity:
+    # diameter of sphere approximation
+    diam = (box0 + box1 + box2) / 3.0
+
+    # angular viscosity
+    lfrc_torque = lvel_torque * -wp.pi * diam * diam * diam * m.opt.viscosity
```
adenzler-nvidia: wp.pi * m.opt.viscosity could also be static in here.

mujoco_warp/_src/passive.py
```diff
@@ -100,9 +104,112 @@ def _damper_passive(m: Model, d: Data):
...
+  wp.launch(_box_fluid, dim=(d.nworld, m.nbody), inputs=[m, d])
+
+  # TODO(team): ellipsoid fluid model
+
+  support.apply_ft(m, d, d.fluid_applied, d.qfrc_fluid)
+
+  @kernel
+  def _qfrc_passive_fluid(m: Model, d: Data):
+    worldid, dofid = wp.tid()
+    d.qfrc_passive[worldid, dofid] += d.qfrc_fluid[worldid, dofid]
```
adenzler-nvidia: would it make sense to add the fluid contribution to the existing passive force accumulation and avoid launching a separate kernel for this?

adenzler-nvidia (commented):
Thanks - some comments, mostly about wp.static usage.

mujoco_warp/_src/support.py
```diff
@@ -198,25 +198,58 @@ def xfrc_accumulate_kernel(
   out[worldid, dofid] += accumul
 
 
-@event_scope
-def xfrc_accumulate(m: Model, d: Data, qfrc: wp.array2d(dtype=float)):
+def apply_ft(m: Model, d: Data, ft: wp.array2d(dtype=wp.spatial_vector), qfrc: wp.array2d(dtype=float)):
+  @wp.kernel
```
btaba: OOC why is this kernel nested?

thowell: updated

mujoco_warp/_src/passive.py
thowell: done

mujoco_warp/_src/passive.py
thowell: done

mujoco_warp/_src/passive.py
thowell: wrapped in `wp.static`

mujoco_warp/_src/passive.py
thowell: done

mujoco_warp/_src/passive.py
thowell: done

mujoco_warp/_src/passive.py
thowell: done

mujoco_warp/_src/passive.py
thowell: done

mujoco_warp/_src/passive.py
thowell: the functionality from `_inertia_box_fluid_model` is moved into `box_fluid`

mujoco_warp/_src/io_test.py
```diff
@@ -227,22 +227,6 @@ def test_get_data_into_m(self):
     np.testing.assert_allclose(mjd.qLD, mjd_ref.qLD)
     np.testing.assert_allclose(mjd.qM, mjd_ref.qM)
 
-  def test_option_physical_constants(self):
```
erikfrey, line 230: maybe we should still check for ellipsoid fluid and fail?

mujoco_warp/_src/passive.py
```diff
@@ -142,51 +128,189 @@ def _gravity_force(
     wp.atomic_add(qfrc_gravcomp_out[worldid], dofid, wp.dot(jac, force))
 
 
+def _box_fluid(opt: Option):
+  opt_wind = opt.wind
+  opt_viscosity = opt.viscosity
+  opt_density = opt.density
+
+  @nested_kernel
```
erikfrey: This gets a bit subjective, but I think the bar might be higher for using a nested kernel.  For tiles we have no choice, but for here, we can pass the opts in as params.

If it led to a significant performance difference I would understand, but this kernel looks pretty cheap.

In particular, my understanding is that the performance hit comes from thread divergence.  e.g. some threads have viscosity and others don't.  I don't think that's the case here, no?

erikfrey (approved):
This is lovely code and nicely written, just a few nits

mujoco_warp/_src/io_test.py
thowell: added check for ellipsoid fluid model in `io.put_model` and added a corresponding test in `io_test.py`

mujoco_warp/_src/passive.py
thowell: changed `@nested_kernel` -> `@wp.kernel`

thowell: after discussion with @erikfrey, the code is updated to have option settings be kernel inputs instead of kernel closure constants

thowell: as a result, `wp.static` calls are removed since these option settings are not constants.

potential performance implications for kernel closure and `wp.static` in this context is something that could be investigated in the future

PR #192, opened 2025-04-27
add gravcomp

thowell:
add gravity compensation

https://mujoco.readthedocs.io/en/latest/XMLreference.html#body-gravcomp

https://github.com/google-deepmind/mujoco/blob/d3bc0544d39c0933067e493f4c1c7ebfd15ae9f3/src/engine/engine_passive.c#L381


adenzler-nvidia (approved):
Looking good, thanks!

PR #195, opened 2025-04-28
Model batching v2

adenzler-nvidia:
replaces #109 - annoying not to keep the discussion there..

This now does the modulo approach with helpers.


adenzler-nvidia:
Plan after offline discussion @erikfrey:
- let's forget about the modulo indexing and just have all the batched fields be size nworlds, and have the user do repeat data for now. But only expands fields to have different data for different model batches.
- use the kernel analyzer to enforce correctness around what field can be expanded and what can't.
- wai with this until #148 is merged.

adenzler-nvidia:
PR is updated to remove the helper function and just use either stride 0 or an array that has nworld as first dimension, with users responsible for filling that array correctly.

@erikfrey I consider this ready for review - I mostly need input on what fields should be batchable.

@btaba - do you by any chance have an example learning workflow around where I could test-drive this?

My main worry right now is that I don't see a good way uf unittesting this, because we are just replacing arrays at runtime. We would either have to validate shapes at runtime (expensive) or add another layer/help that is mandatory.

adenzler-nvidia:
closing in favor of #231

PR #200, opened 2025-04-29
add pytest option for cpu

thowell:
```
pytest --cpu
```


conftest.py
```diff
@@ -0,0 +1,12 @@
+import warp as wp
```
erikfrey, line 16: probably overkill, but let's go ahead and put the apache header

erikfrey (approved):
Very cool thanks, just one nit.

conftest.py
thowell: done

PR #202, opened 2025-04-29
test cuda graph capture

thowell:
add test for cuda graph capture

a pytest option for `wp.config.verify_cuda`, similar to #200, can be added in a future pr


adenzler-nvidia (approved):
Thanks - that's a good start but I guess to be completely sure we would have to run models that exercise all possible branches?

No action required.

thowell:
@adenzler-nvidia `test_graph_capture` now tests more environments from test_data and a todo is added to test more environments

PR #208, opened 2025-04-30
Sequential Box<>Box

camevor:
This is a reimplementation of box<>box, based on the mjc code rather than the mjx code. It is sequential and does not make use of tiles, but closer to mjc and may lead to lower thread divergence.


mujoco_warp/_src/collision_primitive.py
```diff
@@ -1718,6 +1730,497 @@ def capsule_box(
     )
 
 
+@wp.func
+def compute_rotmore(face_idx: int) -> wp.mat33:
```
erikfrey: nit: let's prepend the name with an underscore as we probably don't expect others to use this outside this module, no?

erikfrey (approved):
This looks good!  Just one nit and we'll merge.

mujoco_warp/_src/collision_primitive.py
camevor: Good point, changed the name to `_compute_rotmore`.

PR #210, opened 2025-05-02
update for sensors

thowell:
- add missing camera case to sensors
- fixes for sensors with incorrect reference frame
- add sensors to test_sensor mjcf


erikfrey (approved):
Nice, thanks!

PR #212, opened 2025-05-02
Specify data files a different way

shi-eric:
I was having problems installing the `mujoco_warp` dependency in our pipelines because the files in the `test_data` directory was not being included for some reason.

This change should fix things (tested on my fork).


erikfrey (approved):
Thanks Eric.

PR #214, opened 2025-05-03
add rangefinder sensor

thowell:
add rangefinder sensor

mujoco reference: https://github.com/google-deepmind/mujoco/blob/436b5a8e1f35b6acc5c64944b58a4834e8d38f0c/src/engine/engine_sensor.c#L262


mujoco_warp/_src/io.py
```diff
@@ -248,6 +248,16 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     nxn_geom_pair.append((geom1, geom2))
     nxn_pairid.append(pairid)
 
+  # rangefinder
+  is_rangefinder = mjm.sensor_type == mujoco.mjtSensor.mjSENS_RANGEFINDER
+  sensor_rangefinder_adr = np.nonzero(is_rangefinder)[0]
+  rangefinder_sensor_adr = -1 * np.ones(mjm.nsensor)
+  rangefinder_count = 0
+  for i, rf in enumerate(is_rangefinder):
```
erikfrey: maybe a bit simpler:

```py
rangefinder_sensor_adr = np.full(mjm.nsensor, -1)
rangefinder_sensor_adr[sensor_rangefinder_adr] = np.arange(len(sensor_rangefinder_adr))
```

mujoco_warp/_src/ray.py
```diff
@@ -694,19 +696,39 @@ def ray(
...
   closest_hit_geom_id = wp.zeros((d.nworld, nrays), dtype=int)
-  num_threads = 64
 
   # Create default geomgroup if None is provided
   has_geomgroup = geomgroup is not None
   if geomgroup is None:
     geomgroup = vec6(0, 0, 0, 0, 0, 0)
 
+  if bodyexclude is None:
+    bodyexclude = wp.zeros(nrays, dtype=int)
```
erikfrey: is this JAX-safe?  I thought we can't allocate arrays during cuda graph capture in JAX - no?

erikfrey (changes requested):
One nit, and one question about whether this is JAX compatible

mujoco_warp/_src/io.py

mujoco_warp/_src/ray.py
thowell: not sure that `ray.ray` will work as is with cuda graph capture and be compatible with JAX.

the rangefinder sensor utilizes `ray._ray` and the `Model` field `sensor_rangefinder_bodyid` for the input to `bodyexclude` and should work with cuda graph capture.

erikfrey:
@thowell I see more changes to `ray.ray` in #277 - do you plan to make `ray.ray` JAX compatible in that PR, or this one?  Let me know and I'll review/approve accordingly.

mujoco_warp/_src/ray.py
thowell: #277 is updated to add fields `ray_dist` and `ray_geomid` to `Data` and `ray.ray` utilizes these fields instead of allocating arrays.

PR #216, opened 2025-05-05
Feature/71 height fields

vreutskyy:
This pull request introduces support for heightfield geometries in the collision detection system of the `mujoco_warp` library. The main changes include adding heightfield-specific data structures, functions, and logic to handle heightfield collisions, as well as extending the existing collision detection pipeline to integrate this new geometry type.

### Heightfield Geometry Support

* **Heightfield Type Definition**: Added `HFIELD` as a new geometry type in the `GeomType` enum and updated the `Model` and `Data` classes to include heightfield-specific attributes such as `hfield_size`, `hfield_data`, and `hfield_nrow/ncol`. (`mujoco_warp/_src/types.py`, [[1]](diffhunk://#diff-1d6fcc67d5d5e6b8102baf702153f06ede3ccc22f36b8faee4311c5d2faaec4aR206) [[2]](diffhunk://#diff-1d6fcc67d5d5e6b8102baf702153f06ede3ccc22f36b8faee4311c5d2faaec4aR670-R674) [[3]](diffhunk://#diff-1d6fcc67d5d5e6b8102baf702153f06ede3ccc22f36b8faee4311c5d2faaec4aR896-R900) [[4]](diffhunk://#diff-1d6fcc67d5d5e6b8102baf702153f06ede3ccc22f36b8faee4311c5d2faaec4aR1244)
* **Heightfield Collision Handling**: Introduced the `HFPrism` class and functions `get_hfield_overlap_range` and `get_hfield_triangle_prism` to calculate overlap regions and triangular prisms for heightfield collision detection. (`mujoco_warp/_src/collision_hfield.py`, [mujoco_warp/_src/collision_hfield.pyR1-R159](diffhunk://#diff-14ba3b46ec2e725f2e9acea359dfa41bffd519df263a7f2989a417111b58cdd5R1-R159))

### Collision Pipeline Enhancements

* **Support for Heightfield in `_gjk_support_geom`**: Updated `_gjk_support_geom` to compute support points for heightfield geometries by iterating over the vertices of the triangular prism. (`mujoco_warp/_src/collision_convex.py`, [mujoco_warp/_src/collision_convex.pyR108-R116](diffhunk://#diff-00567e0b505e1dc9e95fd9f45f0d67f383fb19493e3c12ea2ade524ae137136bR108-R116))
* **Integration into Broadphase and Narrowphase**: Modified `_add_geom_pair` to add collision pairs for all potentially colliding heightfield triangles. (`mujoco_warp/_src/collision_driver.py`, [[1]](diffhunk://#diff-076094eda96c2773df6bbb938fcc984a05c39bcef3b49f9bbb681e5d3a4c7d7bL63-R65) [[2]](diffhunk://#diff-076094eda96c2773df6bbb938fcc984a05c39bcef3b49f9bbb681e5d3a4c7d7bR74-R109) [[3]](diffhunk://#diff-00567e0b505e1dc9e95fd9f45f0d67f383fb19493e3c12ea2ade524ae137136bL766-R784)

### Data and Model Updates

* **Heightfield Data in Model Initialization**: Updated `put_model` to populate heightfield-related attributes from the Mujoco model, and adjusted `make_data` to initialize heightfield-specific data arrays. (`mujoco_warp/_src/io.py`, [[1]](diffhunk://#diff-b9105758e9cec712e2848dff0a662d165e220ae83a8edf66303eb9ad7fc04709R448-R453) [[2]](diffhunk://#diff-b9105758e9cec712e2848dff0a662d165e220ae83a8edf66303eb9ad7fc04709R857) [[3]](diffhunk://#diff-b9105758e9cec712e2848dff0a662d165e220ae83a8edf66303eb9ad7fc04709R1158)

### General Codebase Adjustments

* **Heightfield-Specific Logic in `_geom`**: Enhanced the `_geom` function to include heightfield triangle prism vertices when the geometry type is `HFIELD`. (`mujoco_warp/_src/collision_primitive.py`, [[1]](diffhunk://#diff-17fbba1395b3cada368ec634d4601830454a39512182be2e47716f5134c72c0eR50) [[2]](diffhunk://#diff-17fbba1395b3cada368ec634d4601830454a39512182be2e47716f5134c72c0eR66-R69)
* **Imports and Dependencies**: Added necessary imports for heightfield-related functionality in relevant modules. (`mujoco_warp/_src/collision_driver.py`, [[1]](diffhunk://#diff-076094eda96c2773df6bbb938fcc984a05c39bcef3b49f9bbb681e5d3a4c7d7bR23-R30); `mujoco_warp/_src/collision_primitive.py`, [[2]](diffhunk://#diff-17fbba1395b3cada368ec634d4601830454a39512182be2e47716f5134c72c0eR27-R28)

These changes collectively enable efficient and accurate collision detection for heightfield geometries, expanding the library's capabilities for handling complex terrain and surface interactions.


mujoco_warp/_src/types.py
```diff
@@ -350,6 +351,10 @@ class WrapType(enum.IntEnum):
   # unsupported: PULLEY, SPHERE, CYLINDER
 
 
+class vec4f(wp.types.vector(length=4, dtype=wp.float32)):
```
Kenny-Vilella: vec4f is natively supported, not sure that this is needed.
Using wp.vec4f directly should be possible.

mujoco_warp/_src/types.py
```diff
@@ -786,6 +797,7 @@ class Model:
   nsensordata: int
   nlsp: int  # warp only
   npair: int
+  nhfield: int
```
Kenny-Vilella, line 853: You don't have nhfielddata?

mujoco_warp/test_data/hfield/hfield.xml
```diff
@@ -0,0 +1,86 @@
...
+      <geom type="sphere" size="0.1" rgba="0.8 0.2 0.2 1"/>
+    </body>
+    <!---->
+
+    <body pos="0 0 0.6">
+      <freejoint/>
+      <geom type="box" size="0.1 0.1 0.1" rgba="0.8 0.2 0.2 1"/>
+    </body>
+
+    <!-- Boxes ->
```
Kenny-Vilella, line 65: Is this file needs some cleanup?
Or it is just temporary for your testing?
BTW, is it used in the test?

mujoco_warp/_src/io_test.py
```diff
@@ -86,8 +86,9 @@ def test_geom_type(self):
 
     # TODO(team): sdf
 
-    with self.assertRaises(NotImplementedError):
-      mjwarp.put_model(mjm)
+    # seems to fail coz all above's implemented now
```
Kenny-Vilella: Then you can probably delete the test.

Nate711:
Do you have fps performance benchmarks for these hfields? I'm interested in fps for a simple robot against a plane and a simple robot against an hfield. Thanks!

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,195 @@
...
+  return min_i, min_j, max_i, max_j
+
+
+@wp.func
+def get_hfield_triangle_prism(m: Model, hfield_geom: int, tri_index: int) -> wp.mat33:
+  """Returns the vertices of a triangular prism for a heightfield triangle.
+
+  Args:
+      m: Model containing geometry data
+      hfield_geom: Index of the height field geometry
```
Kenny-Vilella: hfield_geom doesn't seem to be the correct name.
Maybe something like hfieldid?

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,195 @@
...
+  # Grid coordinates (i, j) for triangle corners
+  i0 = col
+  j0 = row
+  i1 = i0 + 1
+  j1 = j0 + 1
+
+  # Convert grid coordinates to local space x, y coordinates
+  x0 = wp.float32(i0) * x_scale - size[0]
+  y0 = wp.float32(j0) * y_scale - size[1]
+  x1 = wp.float32(i1) * x_scale - size[0]
```
Kenny-Vilella, line 155: Very minor:
I guess this should be more optimized
x1 = x0 + x_scale
y1 = y0 + y_scale

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,195 @@
...
+
+  # Get height values at corners from hfield_data
+  base_addr = m.hfield_adr[dataid]
+  z00 = m.hfield_data[base_addr + j0 * ncol + i0]
+  z01 = m.hfield_data[base_addr + j1 * ncol + i0]
+  z10 = m.hfield_data[base_addr + j0 * ncol + i1]
+  z11 = m.hfield_data[base_addr + j1 * ncol + i1]
+
+  # Scale heights to range [0, 1] and then to [z_bottom, z_top]
+  z_range = size[2]
```
Kenny-Vilella: Not sure to understand this.
In hfield_data we have the raw data right?
So when do we scale to [0, 1]?

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,195 @@
...
+  z00 = z00 * z_range
+  z01 = z01 * z_range
+  z10 = z10 * z_range
+  z11 = z11 * z_range
+
+  # Set bottom z-value
+  z_bottom = -size[3]
```
Kenny-Vilella, line 173: Why minus here?

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,195 @@
...
+  local_y = wp.dot(wp.vec3(hf_mat[0, 1], hf_mat[1, 1], hf_mat[2, 1]), rel_pos)
+  local_z = wp.dot(wp.vec3(hf_mat[0, 2], hf_mat[1, 2], hf_mat[2, 2]), rel_pos)
+  local_pos = wp.vec3(local_x, local_y, local_z)
+
+  # Get bounding radius of other geometry (including margin)
+  other_rbound = m.geom_rbound[other_geom]
+  other_margin = m.geom_margin[other_geom]
+  bound_radius = other_rbound + other_margin
+
+  # Calculate grid coordinates that bound the other geometry
```
Kenny-Vilella: Is this comment correct?

mujoco_warp/_src/collision_driver.py
```diff
@@ -73,9 +71,43 @@ def _add_geom_pair(m: Model, d: Data, geom1: int, geom2: int, worldid: int, nxni
   else:
     pair = wp.vec2i(geom1, geom2)
 
-  d.collision_pair[pairid] = pair
-  d.collision_pairid[pairid] = m.nxn_pairid[nxnid]
-  d.collision_worldid[pairid] = worldid
+  # For height field, add a collision pair for every
+  # triangle that can potentially collide
```
Kenny-Vilella: We will probably need to check performance here.

Adding a collision pair for every vertex may lead to large number of collision pairs, which would slow down the code and requires large amount of memory.

mujoco_warp/_src/types.py
vreutskyy: What do you mean?

mujoco_warp/test_data/hfield/hfield.xml
vreutskyy: I used it for development. I see other xml files also have some stuff commented out. What's wrong with mine? No, it's not used in UTs

mujoco_warp/_src/io_test.py
vreutskyy: I can. Are you sure no one will miss it?

mujoco_warp/_src/collision_hfield.py
vreutskyy: Is there some naming convention I can read?

mujoco_warp/_src/collision_hfield.py
vreutskyy: I understand the logic, but isn't it more confusing that it is now?

mujoco_warp/_src/collision_hfield.py
vreutskyy: https://mujoco.readthedocs.io/en/stable/XMLreference.html#asset-hfield

mujoco_warp/_src/collision_hfield.py
vreutskyy: What's wrong with it?

mujoco_warp/_src/collision_driver.py
vreutskyy: For each triangle, not each vertex. I create 2 pairs/triangles for each HF cell. If you don't have a HF with absurdly small cells, a body of an average size will collide with up to 4 cells - 8 triangles. Each pair is just 2 integers (plus I added an integer for tri index), and they are already pre-allocated, it's not like I allocate them when I add a pair. For the performance, for GPU it doesn't matter if it's 8 or 800 - they are all computed simultaneously. And, after all, what are the alternatives? Would like to hear them )

mujoco_warp/_src/collision_hfield.py
Kenny-Vilella: It seems to me that what you calculate is the resolution of your hfield.

mujoco_warp/_src/collision_hfield.py
Kenny-Vilella: Not to my knowledge.
Following the naming for similar properties is your best bet.

mujoco_warp/_src/collision_hfield.py
Kenny-Vilella: Both are equivalent to me. But if you think the original version is better, it's fine for me.

mujoco_warp/_src/types.py
Kenny-Vilella: In Line 670 you mention that the size of hfield_data is (nhfielddata,).
You should probably add the nhfielddata property here, even if it is not used.

Kenny-Vilella:
There are some unexpected contact points that are generated.
But it doesn't seem to be caused by the change in the PR.
I also see them when playing with a non-flat mesh.

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,409 @@
...
+      # Each grid cell contains two triangles
+      base_idx = ((j * (ncol - 1)) + i) * 2
+
+      # Add both triangles from this cell
+      for t in range(2):
+        if i == 0 and j == 0 and t == 0:
+          # We reuse the initial pair for the 1st triangle
+          new_pairid = pairid
+        else:
+          # For the rest we create a new pait
```
Kenny-Vilella: pait -> pair

mujoco_warp/_src/collision_primitive.py
```diff
@@ -85,6 +94,12 @@ def _geom(
   if geom_type[gid] == int(GeomType.MESH.value):
     geom.vert = mesh_vert
 
+  # If geom is HFIELD triangle, compute triangle prism verts
+  if hftri_index > -1 and geom_type[gid] == int(GeomType.HFIELD.value):
```
Kenny-Vilella: Just wonder, this is to avoid the calculation of the prism verts in the primitive narrowphase?

vreutskyy: yes. the support function may be called many-many times for a single GJK-EPA test, so we want to minimize the calculations we do there

Kenny-Vilella: It should not be `hftri_index == -1` instead?

vreutskyy: no. all -1 were changed to something >-1 in hfield_midphase. actually there's no need to check hftri_index here. just a leftover from the old version. removing it

Kenny-Vilella: Yes you are right

mujoco_warp/_src/types.py
```diff
@@ -598,6 +599,10 @@ class Model:
     nmeshface: number of faces for all meshes                ()
     nlsp: number of step sizes for parallel linsearch        ()
     npair: number of predefined geom pairs                   ()
+    nlsp: number of step sizes for parallel linsearch        ()
+    nhfield: number of heightfields                          ()
+    nhfielddata: size of elevation data                      ()
+    nlsp: number of step sizes for parallel linsearch        ()
```
Kenny-Vilella: You added twice nlsp, while it is already present above.

mujoco_warp/_src/types.py
```diff
@@ -1149,6 +1166,7 @@ class Data:
     sap_segment_index: broadphase context                       (nworld+1,)
     dyn_geom_aabb: dynamic geometry axis-aligned bounding boxes (nworld, ngeom, 2)
     collision_pair: collision pairs from broadphase             (nconmax,)
+    collision_index: collision index for hfield pairs           (nconmax,)
```
Kenny-Vilella: collision_index -> collision_hftri_index

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,409 @@
...
+  one for each potentially colliding triangle.
+
+  Args:
+      geom_type: Array of geometry types
+      geom_dataid: Array of geometry data IDs
+      geom_rbound: Array of geometry bounding radii
+      geom_margin: Array of geometry margins
+      hfield_nrow: Array of heightfield rows
+      hfield_ncol: Array of heightfield columns
+      hfield_size: Array of heightfield sizes
```
Kenny-Vilella, line 270: nconmax_in is missing here

Kenny-Vilella (commented):
Just a few formatting comments, otherwise everything looks good

With default parameter the collision looks a bit weird but it's because of the contact generation.
Increasing the number of gjk iteration fix the contact generation and everything works well.

Kenny-Vilella (approved):

mujoco_warp/_src/io_test.py
erikfrey: Yes, you can delete it.

mujoco_warp/_src/types.py
erikfrey: Still looks like nlsp is duplicated in the current PR.

mujoco_warp/_src/collision_driver.py
```diff
@@ -440,6 +458,9 @@ def collision(m: Model, d: Data):
   else:
     sap_broadphase(m, d)
 
+  # Process heightfield collisions
+  hfield_midphase(m, d)
```
erikfrey: This launches a kernel with dim = nconmax irrespective of whether there's any hfields.  Should we guard against running this function if nhfield == 0?

I could certainly be wrong, but I would expect this to slow down humanoid.xml benchmark by a little bit, for example.  Can you try running it before/after this PR?

mujoco_warp/_src/collision_hfield.py
```diff
@@ -0,0 +1,409 @@
...
+    other_geom,
+    worldid,
+  )
+
+  # Get hfield dimensions for triangle index calculation
+  dataid = geom_dataid[hfield_geom]
+  ncol = hfield_ncol[dataid]
+
+  # Loop through grid cells and add pairs for all triangles
+  for j in range(min_j, max_j + 1):
```
erikfrey, line 333: There's some pathological scenarios where `max_j - min_j` is large or `max_i - min_i` is large.

MuJoCo has a compile-time constant, `mjMAXCONPAIR` set to 50:

https://github.com/google-deepmind/mujoco/blob/main/include/mujoco/mjmodel.h#L30

Basically we assume that no geom pair can ever produce more than 50 contacts.  That might be useful to enforce here.

The challenge is that if we enforce `MAXCONPAIR` in the midphase, we might exclude the best contacts with the most penetration in favor of some other contact candidates that have less penetration.

What do you think?  I don't think there is a perfect answer.

erikfrey (commented):
This is nicely written code!  I'm not going to review it line by line as @Kenny-Vilella and you already tag teamed on cleaning it up.  Really appreciate the cleanliness and docstring comments, awesome.

Let's get this merged after we do two things:

1) Run benchmarks to ensure we didn't slow down the collision driver for non-hfield scenes (probably OK but let's double check)

2) Decide on a strategy for capping contacts

Once those two are figured out I think we are good to go.

mujoco_warp/_src/collision_hfield.py
vreutskyy: As MuJoCo docs already warn users about this, I don't think we need to do something special here.

"The number of possible contacts between a height field and a geom is limited to 50 ([mjMAXCONPAIR](https://mujoco.readthedocs.io/en/stable/APIreference/APIglobals.html#glnumeric)); any contacts beyond that are discarded. To avoid penetration due to discarded contacts, the spatial features of the height field should be large compared to the geoms it collides with."

~~I could limit the number of triangles I check to say 9 (3x3 - pessimistic) or 16 (4x4 - optimistic). Just to avoid wasting time calculating contacts that will be discarded anyway.~~

(edit) No, it's a bit more actually. 3x3 is 18 triangles, not 9. And with up to 4 contact points per triangle it's 72 points already. MuJoCo (C version) uses this limit on 50 contact points per pair because it allocates a temporary buffer to pass it to a collision function. MJX doesn't seem to use this constant. Though I could still limit the number of triangle I process if it feels safer.

vreutskyy:
As GJK/EPA doesn't work atm (issue #258), I disabled my UT. Will enable it back what it's fixed.

vreutskyy:
fixed it

vreutskyy:
Ran `testspeed.py --mjcf=humanoid.xml` a few times on `main` and my branch. Seem close

```
feature/71-height-fields:

 Total simulation time: 11.77 s
 Total steps per second: 696,229
 Total realtime factor: 3,481.14 x
 Total time per step: 1436.31 ns

 Total simulation time: 11.67 s
 Total steps per second: 702,199
 Total realtime factor: 3,510.99 x
 Total time per step: 1424.10 ns

 Total simulation time: 11.76 s
 Total steps per second: 696,697
 Total realtime factor: 3,483.49 x
 Total time per step: 1435.34 ns

main:

 Total simulation time: 11.92 s
 Total steps per second: 687,193
 Total realtime factor: 3,435.97 x
 Total time per step: 1455.19 ns

 Total simulation time: 11.73 s
 Total steps per second: 698,333
 Total realtime factor: 3,491.67 x
 Total time per step: 1431.98 ns

 Total simulation time: 11.66 s
 Total steps per second: 702,297
 Total realtime factor: 3,511.49 x
 Total time per step: 1423.90 ns
```

mujoco_warp/test_data/hfield/hfield.xml
btaba: FWIW "test_data" is typically reserved for files used in UTs or benchmarks (at least for our code in MJX land)

mujoco_warp/_src/collision_hfield.py
btaba: Hello, I was asked to comment a bit here: yes the entire grid overlap region needs to be checked for potential contacts! In MJX we go a step further, and cull the contacts to a max of 4 (maximizing manifold area) for each geom<>hfield collision pair. MJ has a max of 50 as mentioned. In this PR, I see no maximum (as also mentioned). 

I don't think it makes sense to limit the number of triangles to check (a collision could equally be in the center of the grid or the edge). I do think it makes sense to cull contacts for a geom<>hfield pair after the midphase step. Although I appreciate that this may lead to a redesign of the collision driver, and maybe this is better left as a follow-up (using Apollo<>hfield env as a benchmark to hill-climb on).

erikfrey: OK, I trust you both have thought through the implications here.  Let's leave a TODO and get this merged.

erikfrey (approved):
Approve pending adding a TODO for capping max contacts.  Let's get this merged!

adenzler-nvidia:
closes #258 as well.

PR #220, opened 2025-05-07
Kernel API Refactor

### LINKED ISSUE #50: Remove @wp.struct from Model, Data

erikfrey:
* Moves most kernel functions to the module level
* Tile kernels are wrapped in a function that takes a tile size or tile object and returns a kernel
* Change Model and Data from wp.struct to dataclasses.dataclass
* Give kernel functions explicit in/out params. This is helpful for:
  * Moving towards making MJWarp differentiable
  * Easier integration into JAX/Newton that want control of providing separate fine-grained params.

This change should generally address the kernel parameter size issue we're hitting when users try MJWarp with older CUDA installs (4kb for older CUDA versions, 32kb for CUDA 12.8)

The downside of this change is that the kernel signatures are much more verbose, and explicitly writing out the kernel parameters is slow and could be a vector for bugs. As such, this PR also introduces a kernel analyzer (that will run in github CI and is also a vscode plugin) that checks for many basic bugs.

Although tests pass, there's a significant chance this PR introduces bugs.  Here's what we did to test it:

* Ensure all unit tests pass
* Loaded up multiple benchmark models in `mjwarp-viewer` and ensured they still look correct.

Given the size of this PR, if it introduces any bugs, let's roll forward.

This PR closes #50


mujoco_warp/_src/solver_test.py
```diff
@@ -32,6 +32,9 @@
 # due to float precision
 _TOLERANCE = 5e-3
 
+wp.config.verify_cuda = True
+wp.config.debug = True
+
```
adenzler-nvidia: randomly came across this, is this expected to stay there?

adenzler-nvidia (commented):
This is very hard to review, found one thing randomly but otherwise I think we just have to rely on the tests and us iteratively fixing issues that will come up later.

Thanks for all the hard work, it looks like a painful refactor.

mujoco_warp/_src/solver_test.py
erikfrey: ah yeah some left over debugging, good catch thanks

mujoco_warp/_src/math.py
```diff
@@ -234,19 +228,15 @@ def closest_segment_point(a: wp.vec3, b: wp.vec3, pt: wp.vec3) -> wp.vec3:
 
 
 @wp.func
-def closest_segment_point_and_dist(
-  a: wp.vec3, b: wp.vec3, pt: wp.vec3
-) -> Tuple[wp.vec3, wp.float32]:
+def closest_segment_point_and_dist(a: wp.vec3, b: wp.vec3, pt: wp.vec3) -> Tuple[wp.vec3, wp.float32]:
```
thowell: can the return type be `Tuple[wp.vec3, float]`?

erikfrey: yes, and i'll add a TODO to kernel analyzer to check return types

PR #222, opened 2025-05-07
Add minimal Franka test data example

StafaH:
Adding a minimal example (franka robot and floor) to test data.


erikfrey (changes requested):
@StafaH I would leave the primitive colliders as they were (basically undo your last commit) - those work OK and have already been validated in sim2real contexts.

Was there something you had in mind by explicitly removing them?

mujoco_warp/test_data/franka_emika_panda/scene.xml
```diff
@@ -0,0 +1,32 @@
+<mujoco model="panda scene">
+  <include file="panda.xml"/>
+
+  <statistic center="0.3 0 0.4" extent="1"/>
+
+  <option timestep="0.005" iterations="5" ls_iterations="8" integrator="implicitfast">
```
erikfrey: for MJWarp, we should be able to remove `iterations` and `ls_iterations`

mujoco_warp/test_data/franka_emika_panda/scene.xml
```diff
@@ -0,0 +1,32 @@
...
+    <flag eulerdamp="disable"/>
+  </option>
+
+  <custom>
+    <numeric data="12" name="max_contact_points"/>
```
erikfrey: for MJWarp, we should be able to remove this whole `<custom>` section

mujoco_warp/test_data/franka_emika_panda/scene.xml
StafaH: Done!

mujoco_warp/test_data/franka_emika_panda/scene.xml
StafaH: Done!

StafaH:
Reverted the primitive collision change. Agree with the sim2real comment to align with previous work. Originally I was just thinking this test data/test case should more more MJWarp related, since the original motivation for the primitive colliders was that mesh collision was too expensive in MJX, but your rationale makes more sense as well.

erikfrey (approved):
Awesome, thanks @StafaH !

PR #223, opened 2025-05-07
add utilities for tendon wrapping

thowell:
adds utilities:

- `is_intersect`
- `wrap_circle`
- `wrap_inside`
- `wrap`

for tendon wrapping

mujoco reference: https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_util_misc.c


mujoco_warp/_src/util_misc.py
```diff
@@ -0,0 +1,386 @@
...
+
+import warp as wp
+
+from . import math
+from .types import MJ_MINVAL
+from .types import WrapType
+
+
+@wp.func
+def is_intersect(p1: wp.vec2, p2: wp.vec2, p3: wp.vec2, p4: wp.vec2) -> bool:
```
erikfrey, line 28: We haven't been great about this but let's add some docstrings.  It's on my list to add more as well.  So for example, this could read:

```py
def is_intersect(p1: wp.vec2, p2: wp.vec2, p3: wp.vec2, p4: wp.vec2) -> bool:
  """Check for intersection of two 2D line segments."""
```

Don't need to be super long, and feel free to crib the comment strings from MJC `engine_util_misc.c`

mujoco_warp/_src/math.py
```diff
@@ -217,6 +217,14 @@ def make_frame(a: wp.vec3):
   # fmt: on
 
 
+@wp.func
+def normalize_with_norm(x: wp.vec2):
```
erikfrey: can you just define a single `normalize_with_norm` but it takes `Any`?

It looks like wp.func allows any: https://github.com/NVIDIA/warp/blob/73d5a063505e64b35d90894f8f452a00fa82c897/warp/math.py#L57

erikfrey (approved):
Looks good, just two nits.

mujoco_warp/_src/util_misc.py
thowell: done

mujoco_warp/_src/math.py
thowell: done

PR #227, opened 2025-05-08
update solver: store cost in prev_cost before zeroing cost

thowell:


btaba (approved):
Wow tricky!! Nice find

PR #231, opened 2025-05-09
Model Batching V4

adenzler-nvidia:
Previous discussions in #109 , another attempt in #195 

@btaba could you help test-driving this from the JAX side once that's ready?


mujoco_warp/_src/types.py
```diff
@@ -843,27 +843,27 @@ class Model:
...
-  dof_invweight0: wp.array(dtype=float)
-  dof_frictionloss: wp.array(dtype=float)
-  dof_solimp: wp.array(dtype=vec5)
-  dof_solref: wp.array(dtype=wp.vec2)
+  dof_armature: wp.array2d(dtype=float)
+  dof_damping: wp.array2d(dtype=float)
+  dof_invweight0: wp.array2d(dtype=float)
+  dof_frictionloss: wp.array2d(dtype=float)
+  dof_solimp: wp.array2d(dtype=vec5)
+  dof_solref: wp.array2d(dtype=wp.vec2)
```
btaba, line 887: This looks like a good selection! Essentially I'm looking at type annotations with jax.Array in [mjx/_src/types.py](https://github.com/google-deepmind/mujoco/blob/d77126dc58860f60710ed2a78a80f4f81ec1a786/mjx/mujoco/mjx/_src/types.py#L679), maybe that's what you're also doing. In any case, I think it's easy enough to modify this as we need.

btaba (commented):
Hi @adenzler-nvidia I'm not quite ready with the JAX shim to test, but I expect maybe by EOW. Nevertheless, I think this approach will work a lot better with JAX than prior versions, I can't foresee any big issues.

mujoco_warp/_src/types.py
```diff
@@ -630,27 +630,27 @@ class Model:
     jnt_bodyid: id of joint's body                           (njnt,)
     jnt_limited: does joint have limits                      (njnt,)
     jnt_actfrclimited: does joint have actuator force limits (njnt,)
-    jnt_solref: constraint solver reference: limit           (njnt, mjNREF)
-    jnt_solimp: constraint solver impedance: limit           (njnt, mjNIMP)
-    jnt_pos: local anchor position                           (njnt, 3)
+    jnt_solref: constraint solver reference: limit           (nworld, njnt, mjNREF)
+    jnt_solimp: constraint solver impedance: limit           (nworld, njnt, mjNIMP)
+    jnt_pos: local anchor position                           (nworld,njnt, 3)
```
thowell: nit: `(nworld, njnt, 3)`

mujoco_warp/_src/types.py
```diff
@@ -630,27 +630,27 @@ class Model:
...
     jnt_axis: local joint axis                               (njnt, 3)
-    jnt_stiffness: stiffness coefficient                     (njnt,)
-    jnt_range: joint limits                                  (njnt, 2)
-    jnt_actfrcrange: range of total actuator force           (njnt, 2)
-    jnt_margin: min distance for limit detection             (njnt,)
+    jnt_stiffness: stiffness coefficient                     (nworld, njnt)
+    jnt_range: joint limits                                  (nworld, njnt, 2)
+    jnt_actfrcrange: range of total actuator force           (nworldnjnt, 2)
```
thowell: nit: `(nworld, njnt, 2)`

mujoco_warp/_src/types.py
```diff
@@ -630,27 +630,27 @@ class Model:
...
-    dof_invweight0: diag. inverse inertia in qpos0           (nv,)
-    dof_frictionloss: dof friction loss                      (nv,)
-    dof_solimp: constraint solver impedance: frictionloss    (nv, NIMP)
-    dof_solref: constraint solver reference: frictionloss    (nv, NREF)
+    dof_armature: dof armature inertia/mass                  (nworld, nv)
+    dof_damping: damping coefficient                         (nworld, nv)
+    dof_invweight0: diag. inverse inertia in qpos0           (nworld, nv)
+    dof_frictionloss: dof friction loss                      (nworld, nv)
+    dof_solimp: constraint solver impedance: frictionloss    (nworld, nv, NIMP)
+    dof_solref: constraint solver reference: frictionloss    (nworld,nv, NREF)
```
thowell: nit: `(nworld, nv, NREF)`

thowell:
@adenzler-nvidia thanks for making the updates to enable model batching for these fields! 

similar to @btaba i think this pr has a good selection of batched `Model` fields and additional fields can easily be updated with a batch dimension if needed.

added a few minor comments for docstring formatting.

thanks!

erikfrey (approved):
OK, looks great and conforms exactly to our plan and discussions.  Let's go!

mujoco_warp/_src/types.py
adenzler-nvidia: Thanks - took another look at MJX and added/removed some. It's pretty much the same list now, but agreed we can add/remove as needed.

Issue #88, opened 2025-04-01
`JacobianType.SPARSE`


erikfrey:
This is a pretty hard feature and may need to be broken down into subtasks.

We may need to study how to do efficient batched sparse matmul in Warp.

We might want to start off with a feasibility study to make sure Sparse is fast enough on GPU to be worth the extra complexity of supporting it.

erikfrey:
@thowell I'm going to ask you to break this down into a series of sub-issues that we think are more approachable?

thowell:
@erikfrey updated the issue description with specific fields that we can represent as sparse matrices.

thowell:
# MuJoCo Warp Sparse Jacobian Roadmap

Add sparse representations for `efc_J`, `ten_J`, `actuator_moment`, and `flexedge_J` to MuJoCo Warp. 

These fields often have sparse rows, for example, a limit constraint row only has 1 non-zero element 
corresponding to the limited DOF. Utilizing sparse representations can significantly 
**reduce memory requirements** and **improve performance by skipping unnecessary operations** like 
memory reads/writes or multiplication by zero.

## MuJoCo sparse representation

MuJoCo Warp adopts MuJoCo's sparse representation, a Compressed Sparse Row (CSR)-like format to store sparse matrices. For a given field `Z`, the sparse representation consists of:

| Field | Description |
|-------|-------------|
| `Z` | Non-zero values stored contiguously |
| `Z_rownnz` | Number of non-zeros per row |
| `Z_rowadr` | Starting index in `Z` for each row |
| `Z_colind` | Column index for each non-zero value |

This format enables efficient row-wise access and avoids storing zero entries, reducing memory and enabling sparse arithmetic operations. Note that `Z` and `Z_colind` have the same number of elements (total non-zeros), while `Z_rownnz` and `Z_rowadr` have the same number of elements (number of rows).


## Constraints

The constraint Jacobian `efc_J` is currently always represented as a dense `njmax` x `nv` matrix. This field can require significant memory and its dense construction iterates over all DOFs for each row for many constraints.

### efc_J

The sparse representation introduces new fields:

| Field | Description |
|-------|-------------|
| `types.Constraint.J` | Non-zero Jacobian values stored contiguously |
| `types.Constraint.J_rownnz` | Number of non-zeros per constraint row |
| `types.Constraint.J_rowadr` | Starting index in `types.Constraint.J` for each constraint |
[...]

PR #234, opened 2025-05-09
Make JacobianType.auto switch to sparse in put_model

adenzler-nvidia:
This is no solution for #88 - but I think we should at least switch to sparse automatically if the user has set the option to auto.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -64,7 +64,7 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
 
   # TODO(team): remove after solver._update_gradient for Newton solver utilizes tile operations for islands
   nv_max = 60
-  if mjm.nv > nv_max and (not mjm.opt.jacobian == mujoco.mjtJacobian.mjJAC_SPARSE):
+  if mjm.nv > nv_max and (not mjm.opt.jacobian >= int(mujoco.mjtJacobian.mjJAC_SPARSE)):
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/io.py
```diff
@@ -64,7 +64,7 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
 
   # TODO(team): remove after solver._update_gradient for Newton solver utilizes tile operations for islands
   nv_max = 60
-  if mjm.nv > nv_max and (not mjm.opt.jacobian == mujoco.mjtJacobian.mjJAC_SPARSE):
+  if mjm.nv > nv_max and (not mjm.opt.jacobian >= int(mujoco.mjtJacobian.mjJAC_SPARSE)):
```
thowell: it looks like `Model.is_sparse` is set utilizing `mujoco.mj_isSparse` https://github.com/google-deepmind/mujoco_warp/blob/e1604a1dec2169d913d3abf2e83e17bf15f883a9/mujoco_warp/_src/io.py#L287
so the `mjJAC_AUTO` case should be handled by the `is_sparse` field.

can we make the check here `if mjm.nv > nv_max and mjm.opt.jacobian == mujoco.mjtJacobian.mjJAC_DENSE`?

adenzler-nvidia: yeah we can do that, it's just some boolean algebra. I agree it's clearer. Thanks.

thowell (approved):

PR #235, opened 2025-05-09
Add 2d Flex and cloth example to test models

quagla:
Current limitations:
- One flex per scene
- Dimension 2
- Collision with vertices only
- No bending stiffness (need to port bending plugin to the MuJoCo engine first)


mujoco_warp/_src/passive.py
```diff
@@ -161,6 +161,92 @@ def _qfrc_passive_gravcomp(
...
+  f = 0  # TODO(quaglino): this should become a function of t
+
+  # TODO(quaglino): support dim != 2
+  dim = flex_dim[f]
+  nedge = 3
+  nvert = 3
+  kD = flex_damping[f] / timestep
+
+  edges = wp.mat(1, 2, 2, 0, 0, 1, shape=(3, 2), dtype=int)
+  gradient = wp.mat(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, shape=(3, 6))
```
erikfrey: nit: I think these inits let you do a shortcut and will fill if you pass a single scalar, so:

```py
gradient = wp.mat(0.0, shape=(3,6))
```

mujoco_warp/_src/passive.py
```diff
@@ -161,6 +161,92 @@ def _qfrc_passive_gravcomp(
...
+    for ed2 in range(nedge):
+      for i in range(2):
+        for x in range(3):
+          force[edges[ed2, i], x] -= elongation[ed1] * gradient[ed2, 3 * i + x] * metric[ed1, ed2]
+
+  for v in range(nvert):
+    vert = flex_elem[(dim + 1) * t + v]
+    bodyid = flex_vertbodyid[flex_vertadr[f] + vert]
+    for x in range(3):
+      qfrc_spring_out[worldid, body_dofadr[bodyid] + x] += force[v, x]
```
erikfrey: I'm not familiar with the flex model - can two flex elements reference the same dof?

If so, this must become an atomic add (`wp.atomic_add`).  If not, we're OK.

mujoco_warp/_src/passive.py
```diff
@@ -161,6 +161,92 @@ def _qfrc_passive_gravcomp(
...
+  # Data in:
+  flexvert_xpos_in: wp.array2d(dtype=wp.vec3),
+  flexedge_length_in: wp.array2d(dtype=float),
+  flexedge_velocity_in: wp.array2d(dtype=float),
+  # In:
+  timestep: float,
+  # Data out:
+  qfrc_spring_out: wp.array2d(dtype=float),
+):
+  worldid, t = wp.tid()
```
erikfrey: nit: worldid, elementid

erikfrey (approved):
Super exciting!  Thanks Alessio - just a few nits.

mujoco_warp/_src/passive.py
```diff
@@ -185,6 +271,34 @@ def passive(m: Model, d: Data):
     inputs=[m.dof_damping, d.qvel, d.qfrc_spring],
     outputs=[d.qfrc_damper, d.qfrc_passive],
   )
+  wp.launch(
```
erikfrey: Kernel launches have a small overhead cost.  let's do something like:

```py
if m.nflex:
  wp.launch(_flex_elasticity, ...
```

Also it looks like `_damper_passive` already writes to `qfrc_passive`.  It's good to reduce kernel launches and avoid writing back out to global memory where possible.  Do you think you could move the `_flex_elasticity` launch after `_spring_passive` but before `_damper_passive`?

That way, you can delete `_combine_qfrc` and instead `_damper_passive` will write the sum to `qfrc_passive`

erikfrey (approved):

mujoco_warp/_src/passive.py
erikfrey: Actually ignore the first part of this suggestion.  I just remembered that if any of the launch dims are zero, Warp is smart enough to skip enqueuing a kernel launch, so because `n.flexelem` == 0 we are OK, you don't need the if.

I would still do the second part of my recommendation, which is remove the `_combine_qfrc`

mujoco_warp/_src/passive.py
quagla: Done.

mujoco_warp/_src/passive.py
quagla: Done.

mujoco_warp/_src/passive.py
quagla: I didn't realize the order mattered, thanks! Done.

mujoco_warp/_src/passive.py
quagla: It can, dofs are at vertices and the loop is over triangles. Done, thanks.

mujoco_warp/_src/passive.py
```diff
@@ -286,6 +286,80 @@ def _qfrc_passive(
...
+  flex_elemedge: wp.array(dtype=int),
+  flexedge_length0: wp.array(dtype=float),
+  flex_stiffness: wp.array(dtype=float),
+  flex_damping: wp.array(dtype=float),
+  # Data in:
+  flexvert_xpos_in: wp.array2d(dtype=wp.vec3),
+  flexedge_length_in: wp.array2d(dtype=float),
+  flexedge_velocity_in: wp.array2d(dtype=float),
+  # In:
+  timestep: float,
```
thowell: can we update `timestep` -> `opt_timestep` and move this input under `#Model:` in order to match the convention in the repository?

for example
https://github.com/google-deepmind/mujoco_warp/blob/12256ba37842738d8684d44c611cb659cda28930/mujoco_warp/_src/forward.py#L57

thowell:
@quagla thanks for the contribution!

mujoco_warp/_src/passive.py
quagla: Done.

PR #237, opened 2025-05-09
add tendon spring damper

thowell:
mujoco reference: https://github.com/google-deepmind/mujoco/blob/baf84265b8627e6f868bc92ea6422e4e78dacb9c/src/engine/engine_passive.c#L336


mujoco_warp/_src/passive.py
```diff
@@ -72,7 +77,16 @@ def _spring_passive(
     qfrc_spring_out[worldid, dofid + 3] = -stiffness * dif[0]
     qfrc_spring_out[worldid, dofid + 4] = -stiffness * dif[1]
     qfrc_spring_out[worldid, dofid + 5] = -stiffness * dif[2]
+
+    # damper
+    qfrc_damper_out[worldid, dofid + 0] = -damping * qvel_in[worldid, dofid + 0]
```
erikfrey, line 86: Maybe at some point we might take a pass and try some for loops, I would bet there's no perf impact.  So:

```py
for i in range(6):
  qfrc_damper_out[worldid, dofid + i] = -damping * qvel_in[worldid, dofid + i]
```

Anyway no action for now, just something to think about.

erikfrey (approved):
Looks great.

PR #238, opened 2025-05-09
add energy_pos and energy_vel

thowell:
mujoco reference: https://github.com/google-deepmind/mujoco/blob/baf84265b8627e6f868bc92ea6422e4e78dacb9c/src/engine/engine_sensor.c#L918


mujoco_warp/_src/sensor.py
```diff
@@ -1254,3 +1256,227 @@ def sensor_acc(m: Model, d: Data):
...
+  elif length < lower:
+    displacement = lower - length
+  else:
+    displacement = 0.0
+
+  energy = wp.vec2(0.5 * stiffness * displacement * displacement, 0.0)
+  wp.atomic_add(energy_out, worldid, energy)
+
+
+def energy_pos(m: Model, d: Data):
```
erikfrey, line 1599: `energy_pos` is run inside `forward` as long as `mjENABLED(mjENBL_ENERGY)` - should we do the same in mjwarp?

mujoco_warp/_src/sensor.py
```diff
@@ -1254,3 +1256,227 @@ def sensor_acc(m: Model, d: Data):
...
+    # Out:
+    energy_out: wp.array(dtype=wp.vec2),
+  ):
+    worldid = wp.tid()
+
+    qvel_tile = wp.tile_load(qvel_in[worldid], shape=wp.static(nv))
+    Mqvel_tile = wp.tile_load(Mqvel[worldid], shape=wp.static(nv))
+
+    # qvel * (M @ qvel)
+    qvelMqvel_tile = wp.tile_map(wp.mul, qvel_tile, Mqvel_tile)
```
erikfrey, line 1655: nice tile code!

erikfrey (approved):
Looks great, just one q.

mujoco_warp/_src/sensor.py
thowell: added todos to `forward`

PR #239, opened 2025-05-10
spatial tendon: sphere geom wrap, cylinder geom wrap, and pulley

thowell:
adds

- sphere geom wrapping: internal and external
- cylinder geom wrapping: internal and external
- pulleys

to spatial tendons

mujoco reference: https://github.com/google-deepmind/mujoco/blob/924ee3070ae8addb16a3a00e037113601be2111d/src/engine/engine_core_smooth.c#L651

depends on #223


mujoco_warp/_src/smooth.py
```diff
@@ -1941,69 +1942,232 @@ def _spatial_site_tendon(
...
+  vec, length = math.normalize_with_norm(dif)
+  wp.atomic_add(ten_length_out[worldid], ten_adr, length * pulley_scale)
+
+  if length < MJ_MINVAL:
+    vec = wp.vec3(1.0, 0.0, 0.0)
+
+  body0 = site_bodyid[id0]
+  body1 = site_bodyid[id1]
+  if body0 != body1:
+    for i in range(nv):
```
erikfrey, line 2050: I could be wrong and the perf impact is not significant, but any time I see a `for i in range(nv)`, I would be inclined to add a TODO to fix it later

erikfrey (approved):
Looks good, just one nit

mujoco_warp/_src/smooth.py
thowell: added todos

PR #224, opened 2025-05-07
add muscle actuator utilities

thowell:
add utilities:

- `muscle_gain_length`
- `muscle_gain`
- `muscle_bias`
- `_sigmoid`
- `muscle_dynamics_timescale`
- `muscle_dynamics`

for muscle actuators

mujoco implementation reference: https://github.com/google-deepmind/mujoco/blob/6cfea719850f7dbc91b9d6fb36deafb3c5d04eaf/src/engine/engine_util_misc.c#L500
mujoco test reference: https://github.com/google-deepmind/mujoco/blob/6cfea719850f7dbc91b9d6fb36deafb3c5d04eaf/test/engine/engine_util_misc_test.cc#L144

PR #240, opened 2025-05-11
add muscle actuator

thowell:
add muscle actuator

mujoco reference: https://github.com/google-deepmind/mujoco/blob/924ee3070ae8addb16a3a00e037113601be2111d/src/engine/engine_forward.c#L277
mjx reference: https://github.com/google-deepmind/mujoco/blob/924ee3070ae8addb16a3a00e037113601be2111d/mjx/mujoco/mjx/_src/forward.py#L100

depends on #224


erikfrey (approved):
Looks great.

PR #241, opened 2025-05-11
inverse dynamics

thowell:
add inverse dynamics

mujoco reference: https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_inverse.c
mjx reference: https://github.com/google-deepmind/mujoco/blob/main/mjx/mujoco/mjx/_src/inverse.py


mujoco_warp/_src/solver.py
```diff
@@ -2452,13 +2452,7 @@ def solve_done(
...
 
 
-@event_scope
-def solve(m: types.Model, d: types.Data):
-  """Finds forces that satisfy constraints."""
-
-  # warmstart
-  wp.copy(d.qacc, d.qacc_warmstart)
-
+def solve_context(m: types.Model, d: types.Data, grad: bool = True):
```
erikfrey: should this be called `create_context`?

erikfrey (approved):
Looks great, just one nit

mujoco_warp/_src/solver.py
thowell: changed `solve_context` to `create_context`

PR #242, opened 2025-05-12
WIP: Add conditional graph for solver iteration

Kenny-Vilella:
This is still WIP.

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -2440,6 +2440,7 @@ def solve_done(
   efc_done_in: wp.array(dtype=bool),
   # Data out:
   efc_done_out: wp.array(dtype=bool),
+  condition_iteration_out: wp.array(dtype=int),
 ):
   # TODO(team): static m?
   worldid = wp.tid()
@@ -2449,47 +2450,18 @@ def solve_done(
 
   improvement = _rescale(nv, stat_meaninertia, efc_prev_cost_in[worldid] - efc_cost_in[worldid])
   gradient = _rescale(nv, stat_meaninertia, wp.math.sqrt(efc_grad_dot_in[worldid]))
-  efc_done_out[worldid] = (improvement < opt_tolerance) or (gradient < opt_tolerance)
+  done = (improvement < opt_tolerance) or (gradient < opt_tolerance)
+  if done:
+    efc_done_out[worldid] = True
+    wp.atomic_add(condition_iteration_out, 0, -1)
 
 
 @event_scope
-def solve(m: types.Model, d: types.Data):
-  """Finds forces that satisfy constraints."""
-
-  # warmstart
-  wp.copy(d.qacc, d.qacc_warmstart)
-
-  # initialize some efc arrays
-  wp.launch(
-    solve_init_efc,
-    dim=(d.nworld),
-    outputs=[d.efc.search_dot, d.efc.cost, d.efc.solver_niter, d.efc.done],
-  )
-
-  # jaref = d.efc_J @ d.qacc - d.efc_aref
-  d.efc.Jaref.zero_()
-  wp.launch(
-    solve_init_jaref,
-    dim=(d.njmax, m.nv),
-    inputs=[m.nv, d.nefc, d.qacc, d.efc.worldid, d.efc.J, d.efc.aref],
-    outputs=[d.efc.Jaref],
-  )
-
-  # Ma = qM @ qacc
-  support.mul_m(m, d, d.efc.Ma, d.qacc, d.efc.done)
-
-  _update_constraint(m, d)
-  _update_gradient(m, d)
-
-  # search = -Mgrad
-  wp.launch(
-    solve_init_search,
-    dim=(d.nworld, m.nv),
-    inputs=[d.efc.Mgrad],
-    outputs=[d.efc.search, d.efc.search_dot],
-  )
-
-  for i in range(m.opt.iterations):
+def _solver_iteration(
+  condition_iteration: wp.array(dtype=wp.int32),
+  m: types.Model,
+  d: types.Data,
+):
     _linesearch(m, d)
 
     if m.opt.solver == types.SolverType.CG:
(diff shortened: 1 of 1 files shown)


Kenny-Vilella:
@adenzler-nvidia  This seems to be OK and provides perf improvements for all the scenes tested.
Not so happy with how I deal with the max number of iterations, but not sure how to do it better.

mujoco_warp/_src/solver.py
```diff
@@ -2489,65 +2565,17 @@ def solve(m: types.Model, d: types.Data):
...
-        d.efc.grad_dot,
-        d.efc.cost,
-        d.efc.prev_cost,
-        d.efc.done,
-      ],
-      outputs=[d.efc.done],
+  if m.opt.iterations != 0:
+    # note: we only launch the iteration kernel if everything is not done
+    condition_iteration = wp.array([d.nworld], dtype=wp.int32)
+    number_iterations = wp.full(shape=d.nworld, value=m.opt.iterations, dtype=wp.int32)
```
adenzler-nvidia: This inline allocation likely breaks jax-compatibility. Could you move this to io.py?

mujoco_warp/_src/solver.py
```diff
@@ -2440,16 +2440,92 @@ def solve_done(
...
   if efc_done_in[worldid]:
     return
 
   improvement = _rescale(nv, stat_meaninertia, efc_prev_cost_in[worldid] - efc_cost_in[worldid])
   gradient = _rescale(nv, stat_meaninertia, wp.math.sqrt(efc_grad_dot_in[worldid]))
-  efc_done_out[worldid] = (improvement < opt_tolerance) or (gradient < opt_tolerance)
+  done = (improvement < opt_tolerance) or (gradient < opt_tolerance)
+  if done or number_iterations_out[worldid] == 0:
+    efc_done_out[worldid] = True
+    wp.atomic_add(condition_iteration_out, 0, -1)
```
adenzler-nvidia: maybe some comments that put this into prose could help here. But maybe that's just because I'm not used to using these conditionals yet.

Kenny-Vilella: Added some comments

mujoco_warp/_src/solver.py
Kenny-Vilella: Humm it looks like that conditional in general breaks JAX compatibility.
Tried to add a dummy capture_if on trunk and is what I get when running jax_test:

> INTERNAL: CUDA error: Failed to create a child graph node and add it to a CUDA graph: CUDA_ERROR_NOT_SUPPORTED: operation not supported

Is adding a check and disabling conditionals for jax a possibility?

Kenny-Vilella: Jax has come kind of conditional with lax.cond.
So it may be an issue with my machine.

thowell: @Kenny-Vilella i am also getting an error when running `jax_test` with this pr

Kenny-Vilella: @thowell  Thanks for checking !
So we either need to giving up on conditional (and their potentially large speedup) or make special case for JAX.

adenzler-nvidia:
ok, short update here: we are providing a repro to the JAX team for them to check. It's still a bit unclear whether it's a jax issue or an issue with the conditional implementation in warp.

In the meantime, we will hide this functionality behind a flag such that we can make the JAX workflow work. In parallel, we should explore whether there is a workaround of moving the sim step out of the JAX parts and just calling things in sequence.

mujoco_warp/_src/solver.py
Kenny-Vilella: @thowell  The test pass on my side If I commented out this line in jax_test.py

`os.environ["XLA_FLAGS"] = "--xla_gpu_graph_min_graph_size=1"`

Is it the same on your side?

erikfrey: @Kenny-Vilella I'm pretty sure that flag is required in order to maintain good performance in JAX as it ensures cuda graph capture across both JAX and Warp.  If you remove the flag, throughput goes down by a quite a bit.

erikfrey: @Kenny-Vilella we have a field you can use in `MJData` for this, it's called `solver_niter`:

https://github.com/google-deepmind/mujoco/blob/main/include/mujoco/mjdata.h#L185

You can probably ignore the `mjNISLAND` dimension size and assume relevant data should go into / come from `solver_niter[0]` during `get_data` and `put_data`

Note the semantics are reversed - in MuJoCo, `solver_niter` counts up, not down.  The nice thing about populating these solver statistics it they show up in the visualizer like so:

image

erikfrey: Ah yes, we're already using it here:

https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/solver.py#L1370

Can you reuse that array?

Kenny-Vilella: Will try to reuse this array, but it's quite critical that the number goes to 0.
But maybe they are some workaround , will try a bit.

For the flag is more about trying to pinpoint the origin of the issue.
I sent to the Jax team some code that do work and some that do not work, let's see if they can find the issue.

In the meantime, I will try to add a guard such that the conditional nodes are only used when Jax is not used.

erikfrey: Sounds good - I think it's OK to reverse the direction as long as we document it, and "undo" the reverse in get_data/put_data.  Let me know if you need help with this.

thowell: updates for `solver_niter` in #296

Kenny-Vilella: This is now done.

Kenny-Vilella:
@adenzler-nvidia @erikfrey  It looks like everything is corrected now.
I added a flag to turn on/off conditional in order to let the code being jax compatible.
It will have been better to detect automatically whether we are using JAX or not, but I did not find any good way of doing that.

mujoco_warp/_src/types.py
```diff
@@ -427,6 +427,7 @@ class Option:
     wind: wind (for lift, drag, and viscosity)
     density: density of medium
     viscosity: viscosity of medium
+    graph_conditional: flag to use cuda graph conditional, should be False when JAX is used
```
adenzler-nvidia, line 430: what about driver support, is this the same as what we state in the readme?

adenzler-nvidia:
looks good for me but let's give @erikfrey and @thowell time to object.

mujoco_warp/_src/types.py
Kenny-Vilella: I don't see any mention of driver support in the readme.
For the record, I guess 12.4 should be enough with the feature we are using in this PR, but in the future we will probably use features introduced in 12.8

mujoco_warp/_src/solver.py
```diff
@@ -2470,19 +2474,93 @@ def solve_done(
   if efc_done_in[worldid]:
     return
 
-  solver_niter_out[worldid] += 1
+  solver_niter_out[worldid] -= 1
```
erikfrey: if we are going to maintain `nsolving` and `solver_niter` (which is fine by me) let's keep the original semantics, which is that it starts at zero and goes up.

mujoco_warp/_src/solver.py
```diff
@@ -2520,65 +2598,26 @@ def solve(m: types.Model, d: types.Data):
...
+    # note: we only launch the iteration kernel if everything is not done
+    d.nsolving.fill_(d.nworld)
+    wp.capture_while(
+      d.nsolving,
+      while_body=_solver_iteration,
+      m=m,
+      d=d,
     )
+  else:
+    # This branch is mostly for when JAX is used as it is currently not compatible
```
erikfrey, line 2613: Yep this is OK for now but hopefully we can delete this branch soon!

erikfrey (approved):
Looks great, one nit and then let's merge

One thought for @thowell: after this merge, can we reduce the number of places in the code where we read `efc_done` and short circuit?

erikfrey:
Thanks @Kenny-Vilella - this is an awesome new feature.

mujoco_warp/_src/solver.py
Kenny-Vilella: You are correct, it is now changed. Thanks for finding this !

For the record, we have to maintain `nsolving `and `solver_niter `because they are measuring different things.
`nsolving `=> unsolved world
`solver_niter `=> solver iterations

PR #243, opened 2025-05-12
Fix njmax overflow in JTDAJ

adenzler-nvidia:
With the optimized launch pattern there, we don't have the guarantee that we cannot overflow njmax. So we need to check that we are not exceeding njmax there.


thowell:
@adenzler-nvidia we should probably update constraints.py to check `nefc <= njmax` before adding a constraint?

for example, here https://github.com/google-deepmind/mujoco_warp/blob/406d46b9396a968551f7eaabb7ba2a7de235650b/mujoco_warp/_src/constraint.py#L143 we should probably have something like
```
if efcid >= njmax:
  return
```

adenzler-nvidia:
That is a good point, we should have overflow checks in all places where we add to the efc structures, similar to the check in contact gen.

I think the reading is covered in all cases except for the one this PR fixes, because we only launch njmax threads. This here just broke with the optimization.

Should I append to this PR or shall we have a separate one?

thowell:
@adenzler-nvidia #245 should prevent potential overflow from `make_constraint`. do we still need this pr's proposed check in `solver.update_gradient_JTDAJ`?

adenzler-nvidia:
yes it's still needed as far as I can see, because we round up to the next multiple of sm_count * 6 * 256 for the launch, which might be larger than njmax. So we might still read beyond njmax.

thowell:
@adenzler-nvidia with #245 `constraint.make_constraint` should set `nefc` to be `0 <= nefc <= njmax`. 

in the case where the number of threads launched for the `solver.update_gradient_JTDAJ` kernel is greater than `njmax`, is the check `efcid <= nefc` in the kernel sufficient since `nefc <= njmax`?

is there a scenario we should consider where `nefc > njmax`?

adenzler-nvidia:
Because of the atomic add we should always assume that nefc can be larger than njmax. We keep on adding, even if we don't write out the data after that. It also helps us figure out how big njmax actually needs to be, using --measure_alloc option in testspeed.

thowell:
@adenzler-nvidia ok makes sense, thanks for pointing out that it is possible `nefc > njmax` due to the atomic adds!

should we clamp `nefc` at the end of `constraint.make_constraint` so that `0 <= nefc <= njmax`? 

should we also consider the possibility of `ncon > nconmax` due to atomic adds?

thanks!

adenzler-nvidia:
I think there is value in reading out the high watermark of nefc - then you can tune your njmax parameter. Same thing with nconmax.

We should do another pass on the code looking specifically at nconmax, but I remember we had a PR + discussion a while ago and figured out that all of those kernels are launched with nconmax threads, so we would never process any of the overflow.

thowell (approved):

PR #244, opened 2025-05-12
Add block cholesky implementation

nvtw:
Adds a block cholesky implementation. For testing purposes, it's always active at the moment but for system sizes below a certain threshold (probably around 32), the old implementation should be faster.

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -2346,7 +2346,7 @@ def _update_gradient(m: types.Model, d: types.Data):
       )
 
     # TODO(team): Define good threshold for blocked vs non-blocked cholesky
-    if m.nv < 1:
+    if m.nv < 32:
       wp.launch_tiled(
         update_gradient_cholesky(m.nv),
         dim=(d.nworld,),


mujoco_warp/_src/solver.py
```diff
@@ -2294,14 +2330,22 @@ def _update_gradient(m: types.Model, d: types.Data):
...
 
-    # TODO(team): support cholesky on larger nv, using tiles here limits the max matrix size
-    wp.launch_tiled(
-      update_gradient_cholesky(m.nv),
-      dim=(d.nworld,),
-      inputs=[d.efc.grad, d.efc.h, d.efc.done],
-      outputs=[d.efc.Mgrad],
-      block_dim=32,
-    )
+    if m.nv < 1:
```
nvtw: Here we should choose a reasonalbe threshold value before merging, e. g. 32

nvtw:
This PR seems to require a new warp beta package @shi-eric

nvtw:
This PR needs a bugfix for the lower and upper triangular solve when the warp cpu backend is used (the cuda backend works fine). Warp MR is open

nvtw:
NaNs show up in some tests with block cholesky but I was not successful in finding the root cause so far. Isolated tests with nans as input in the upper triangular matrix of h were not an issue since block cholesky does not care about them.

mujoco_warp/_src/block_cholesky.py
```diff
@@ -0,0 +1,164 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from functools import lru_cache
+
+import warp as wp
+
+
+@lru_cache(maxsize=None)
```
erikfrey, line 21: OOC why do you need the lru_cache?

I thought Warp cached at the kernel level, but this suggests it caches also wp.func, is that it?

mujoco_warp/_src/solver.py
erikfrey: Do we need both?  e.g. is it the case that if `active_matrix_size` <= `block_size`, the perf of `update_gradient_cholesky_blocked` ~= `update_gradient_cholesky`?

erikfrey (approved):
@nvtw ok finally got to this in my queue - eyeballing the code I don't see anything immediately wrong that might cause NaNs.  Please let us know how we can help!  In the meantime, two questions for my own curiosity.

mujoco_warp/_src/block_cholesky.py
nvtw: On windows I get way faster kernel load times with @lru_cache. I suspect @lru_cache avoids hard drive access while warp caching might search for files but I cannot tell for sure.

mujoco_warp/_src/solver.py
```diff
@@ -2316,14 +2345,30 @@ def _update_gradient(m: types.Model, d: types.Data):
...
 
-    # TODO(team): support cholesky on larger nv, using tiles here limits the max matrix size
-    wp.launch_tiled(
-      update_gradient_cholesky(m.nv),
-      dim=(d.nworld,),
-      inputs=[d.efc.grad, d.efc.h, d.efc.done],
-      outputs=[d.efc.Mgrad],
-      block_dim=32,
-    )
+    # TODO(team): Define good threshold for blocked vs non-blocked cholesky
```
adenzler-nvidia, line 2348: let's capture this TODO in a github issue.

adenzler-nvidia:
let's get this in and iterate on the actual threshold value once we have the benchmarks in place. Thanks!

mujoco_warp/_src/solver.py
erikfrey: Added as #320

PR #245, opened 2025-05-12
update make_constraint kernels to return if efcid >= njmax

thowell:


mujoco_warp/_src/constraint.py
```diff
@@ -142,6 +143,9 @@ def _efc_equality_connect(
   necid = wp.atomic_add(ne_connect_out, 0, 3)
   efcid = nefc_in[0] + necid
 
+  if efcid >= njmax_in:
+    return
```
adenzler-nvidia, line 147: This does not check whether we have space for all 3 constraints, right? Because we might be just below njmax but +3 overflows it?

mujoco_warp/_src/constraint.py
```diff
@@ -531,6 +551,10 @@ def _efc_equality_weld(
 
   newid = wp.atomic_add(ne_weld_out, 0, 6)
   efcid = nefc_in[0] + ne_connect_in[0] + newid
+
+  if efcid >= njmax_in:
+    return
```
adenzler-nvidia, line 556: same point here, this needs a check for each of the 6 constraints.

thowell: done

mujoco_warp/_src/constraint.py
thowell: done

adenzler-nvidia (approved):
LGTM, thanks!

PR #252, opened 2025-05-14
Set random qpos in tests, fix underlying bugs.

btaba:


mujoco_warp/_src/test_util.py
```diff
@@ -100,6 +101,10 @@ def fixture(
   mjd = mujoco.MjData(mjm)
   if keyframe > -1:
     mujoco.mj_resetDataKeyframe(mjm, mjd, keyframe)
+  elif qpos0:
+    mjd.qpos[:] = mjm.qpos0
+  else:
+    mjd.qpos[:] = np.random.random(mjm.nq)
```
thowell, line 108: can we apply the joint limits here?

thowell, line 108: can we normalize quaternions here so they are units quaternions?

btaba: I'm explicitly not normalizing quaternions or applying joint limits here! This uncovers a bunch of issues in underlying code where quats are not normalized where they should be

btaba: see other comment

thowell: ok, makes sense for testing!

thowell (approved):

thowell:
@btaba thanks!

PR #253, opened 2025-05-14
Implement tile sort in sap broad phase

nvtw:
Implement tile sort in sap broad phase

### DIFF
--- a/mujoco_warp/_src/collision_driver.py
+++ b/mujoco_warp/_src/collision_driver.py
@@ -257,6 +257,28 @@ def _sap_broadphase(
     worldgeomid += nsweep_in
 
 
+def create_segmented_sort_kernel(tile_size: int):
+  @wp.kernel
+  def segmented_sort_kernel(
+    sap_projection_lower: wp.array2d(dtype=float),
+    sap_sort_index: wp.array2d(dtype=int),
+  ):
+    worldid = wp.tid()
+
+    # Load input into shared memory
+    keys = wp.tile_load(sap_projection_lower[worldid], shape=tile_size, storage="shared")
+    values = wp.tile_load(sap_sort_index[worldid], shape=tile_size, storage="shared")
+
+    # Perform in-place sorting
+    wp.tile_sort(keys, values)
+
+    # Store sorted shared memory into output arrays
+    wp.tile_store(sap_projection_lower[worldid], keys)
+    wp.tile_store(sap_sort_index[worldid], values)
+
+  return segmented_sort_kernel
+
+
 def sap_broadphase(m: Model, d: Data):
   """Broadphase collision detection via sweep-and-prune."""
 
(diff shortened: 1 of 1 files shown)


thowell:
@nvtw can we test `sap_broadphase` with both sort options?

nvtw:
I guess manually selecting the sort algorithm would need some global config parameter. Would it be ok to introduce one?

thowell:
@nvtw since the switch between sort options depends on `nworld = d.nworld * m.ngeom` it should be possible to set `nworld` in order to specify a specific sort option for testing.

nvtw:
@thowell I had a bug, the borad phase type should only be decided on m.ngeom, it's fixed now. But that means that I would need to create a test with 1000 bodies to triger the segmented_sort. Does that make sense? The threshold value of when switching to tile_sort could be a model parameter, then the UT could set it to a low value. Maybe that's the better approach than UTs with 1000 bodies in a scene. 1000 was chosen since benchmarks showed that tile_sort is faster than segmented sort up to around 2000 key value pairs.

erikfrey:
I think it's reasonable for choice of broadphase to be specified in `Model.opt`, and we move the heuristic from `collision_driver.collision` to `io.put_model`

mujoco_warp/_src/types.py
```diff
@@ -453,6 +453,7 @@ class Option:
   wind: wp.vec3
   density: float
   viscosity: float
+  broad_phase_tile_sort_threshold: int
```
erikfrey: Hmm, sorry I should have been more explicit.  Following the pattern we use elsewhere, I was imagining something like:

```py
class BroadphaseType(enum.IntEnum):
  """Type of broadphase algorithm.

  Attributes:
     NXN: foo
     SAP_TILE: bar
     SAP_SEGMENTED: baz
  """
  NXN = 0
  SAP_TILE = 1
  SAP_SEGMENTED = 2
 ```

then down in in Option:

```py
broadphase: int
```

Can you try something like that?

Don't forget to document this in the docstring for Option!

nvtw: Thanks for the clarification. I implemented the suggestions, hope everything is as expected.

mujoco_warp/_src/io.py
```diff
@@ -360,6 +360,7 @@ def create_nmodel_batched_array(mjm_array, dtype):
       epa_iterations=12,
       epa_exact_neg_distance=wp.bool(False),
       depth_extension=0.1,
+      broadphase=types.BroadphaseType.NXN,
```
erikfrey: you could put the logic in here to decide which broadphase based on ngeom, did you want to do that?

if you want to keep it a fancy one-liner you can do:

```py
broadphase=np.argmax(np.array([100, 1000, np.inf]) > ngeom)
```

okay maybe that's too fancy      :-)          sometimes i can't avoid a good code golf.

you could also just do if/else

erikfrey (approved):
Looks very nice, just one last question

mujoco_warp/_src/io.py
nvtw: I wen't for if/else but the code golf is actually also elegant :)

PR #255, opened 2025-05-14
update io.put_data to check qM and qLD before calling np.linalg.cholesky

thowell:


powertj (approved):
lgtm, thanks!

adenzler-nvidia (approved):
Thanks!

PR #257, opened 2025-05-15
dtype fixes for numpy 1.26.0

adenzler-nvidia:
This makes MjWarp work with numpy 1.26.0 (which is not that old, from sept 2023..) I think these changes are generally ok, so I would like to merge them if possible.

Otherwise we need to be explicit about only supporting numpy 2.0+


adenzler-nvidia:
@erikfrey any opposition here? Would like to merge this if possible to unblock people.

erikfrey (approved):
No concerns, looks good.  It might be nice to have a test to guard against breaking compatibility again but we can save that for later.

adenzler-nvidia:
thanks!

PR #265, opened 2025-05-17
add constraint type

thowell:
adds constraint type (mjtConstraint) https://mujoco.readthedocs.io/en/latest/APIreference/APItypes.html#mjtconstraint

### DIFF
--- a/mujoco_warp/_src/constraint.py
+++ b/mujoco_warp/_src/constraint.py
@@ -18,6 +18,7 @@ import warp as wp
 from . import math
 from . import support
 from . import types
+from .types import ConstraintType
 from .types import vec5
 from .types import vec11
 from .warp_util import event_scope
@@ -40,8 +41,10 @@ def _update_efc_row(
   margin: float,
   Jqvel: float,
   frictionloss: float,
+  type: int,
   id: int,
   # Data out:
+  efc_type_out: wp.array(dtype=int),
   efc_id_out: wp.array(dtype=int),
   efc_pos_out: wp.array(dtype=float),
   efc_margin_out: wp.array(dtype=float),
@@ -88,6 +91,7 @@ def _update_efc_row(
   efc_pos_out[efcid] = pos_aref + margin
   efc_margin_out[efcid] = margin
   efc_frictionloss_out[efcid] = frictionloss
+  efc_type_out[efcid] = type
   efc_id_out[efcid] = id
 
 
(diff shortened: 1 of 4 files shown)


erikfrey:
Hmmmm... :-)

Why though?  I've wondered about this in C MuJoCo too.  No control flow depends on this field.  It's not reported anywhere (to my knowledge).  I suppose we might use it in printf debugging.

Is it worth the memory/ops?

cc @yuvaltassa who might have a useful POV on this one too.

thowell:
here is an example in mujoco that utilizes `mjCNSTR_LIMIT_JOINT` in the `mjSENS_JOINTLIMITPOS` case https://github.com/google-deepmind/mujoco/blob/62a9d5b98d80ad6cc8b7756e9284194d05b108ad/src/engine/engine_sensor.c#L291

yuvaltassa:
MJC has lots of `if (d->efc_type[i] != mjCNSTR_CONTACT_ELLIPTIC)` for the special handling of elliptic, you don't need that?

yuvaltassa:
Oh I see, elliptic is a whole different code path, this makes sense. I guess the C code is forward looking so that one day we can have a mix of elliptic and pyramidal contacts. That said, @thowell's example with sensors is a valid one. You don't know in advance where the limit constraint lives since they pop in and out depending on whether you're past a limit or not, so you kind of have to scan and look for it. I mean in MJX I guess nothing pops in and out and you know the index in advance but that's not true in mjWrap, is it?

erikfrey:
OK, works for me!

mujoco_warp/_src/constraint.py
```diff
@@ -221,7 +226,9 @@ def _efc_equality_connect(
       0.0,
       Jqvel[i],
       0.0,
+      int(ConstraintType.EQUALITY.value),
```
erikfrey: nit: I think you can drop the `.value`

(comment applies here and elsewhere)

erikfrey (approved):
thanks taylor

mujoco_warp/_src/constraint.py
thowell: it seems like `int(ConstraintType.EQUALITY)` does not work, but `ConstraintType.EQUALITY.value` works. should i update relevant cases in `constraint.py` to the latter format?

erikfrey: Sure that works.

PR #266, opened 2025-05-17
add limitpos sensors

thowell:
adds jointlimitpos and tendonlimitpos sensors

mujoco references:
https://github.com/google-deepmind/mujoco/blob/178fb49c2b2ff48f59653515ab09b9cafca31b7a/src/engine/engine_sensor.c#L288
https://github.com/google-deepmind/mujoco/blob/178fb49c2b2ff48f59653515ab09b9cafca31b7a/src/engine/engine_sensor.c#L298


mujoco_warp/_src/sensor.py
```diff
@@ -495,53 +535,79 @@ def _sensor_pos(
...
-      d.subtree_com,
-      d.actuator_length,
-      d.ten_length,
-    ],
-    outputs=[d.sensordata],
-  )
+  if m.sensor_pos_adr.size > 0:
+    wp.launch(
+      _sensor_pos,
+      dim=(d.nworld, m.sensor_pos_adr.size),
```
erikfrey: I think because dim here evaluates to `(d.nworld, 0)` Warp is smart enough to skip the kernel launch, which means you don't need the `if m.sensor_pos_adr.size > 0` guard above

Comment applies here and below

erikfrey (approved):
Some kernel analyzer and other issues to fix, but code generally looks OK barring one nit

mujoco_warp/_src/sensor.py
```diff
@@ -182,6 +183,45 @@ def _ball_quat(jnt_qposadr: wp.array(dtype=int), qpos_in: wp.array2d(dtype=float
...
+  nl = nl_in[0]
+
+  # skip if not limit
+  if efcid < ne + nf or efcid >= ne + nf + nl:
+    return
+
+  sensorid = sensor_limitpos_adr[limitposid]
+  if efc_id_in[efcid] == sensor_objid[sensorid]:
+    efc_type = efc_type_in[efcid]
+    if efc_type == int(ConstraintType.LIMIT_JOINT.value) or efc_type == int(ConstraintType.LIMIT_TENDON.value):
```
erikfrey, line 244: Actually there's one thing I'm not following here:

What if we during this step we do not emit a constraint row - does that mean we don't write a sensor value?  Shouldn't we be writing a sensor value no matter what?  Presumably zero or something if the limit is not reached?

mujoco_warp/_src/sensor.py
thowell: guards are removed

mujoco_warp/_src/sensor.py
thowell: added `_limit_pos_zero` kernel that initializes limitpos sensor values to zero

PR #267, opened 2025-05-17
add limit force sensors

thowell:
add jointlimitfrc and tendonlimitfrc sensors

mujoco references:
https://github.com/google-deepmind/mujoco/blob/178fb49c2b2ff48f59653515ab09b9cafca31b7a/src/engine/engine_sensor.c#L838
https://github.com/google-deepmind/mujoco/blob/178fb49c2b2ff48f59653515ab09b9cafca31b7a/src/engine/engine_sensor.c#L848

depends on #265 and #266


mujoco_warp/_src/sensor.py
```diff
@@ -1455,6 +1492,29 @@ def sensor_acc(m: Model, d: Data):
       outputs=[d.sensordata],
     )
 
+  if m.sensor_limitfrc_adr.size > 0:
```
erikfrey: same comment here, I believe you can skip this check

mujoco_warp/_src/sensor.py
```diff
@@ -1112,6 +1112,43 @@ def _joint_actuator_force(
...
+  # skip if not limit
+  if efcid < ne + nf or efcid >= ne + nf + nl:
+    return
+
+  sensorid = sensor_limitfrc_adr[limitfrcid]
+  if efc_id_in[efcid] == sensor_objid[sensorid]:
+    efc_type = efc_type_in[efcid]
+    if efc_type == int(ConstraintType.LIMIT_JOINT.value) or efc_type == int(ConstraintType.LIMIT_TENDON.value):
+      worldid = efc_worldid_in[efcid]
+      _write_scalar(sensor_datatype, sensor_adr, sensor_cutoff, sensorid, efc_force_in[efcid], sensordata_out[worldid])
```
erikfrey, line 1331: don't forget to zero out!

erikfrey (approved):
Looks great, two nits

mujoco_warp/_src/sensor.py
thowell: removed guards from `sensor_acc`

mujoco_warp/_src/sensor.py
thowell: added a kernel `_limit_frc_zero` that initializes limit force sensors to zero

PR #268, opened 2025-05-17
add tendon actuator force sensor

thowell:
add tendonactfrc sensor

mujoco reference: https://github.com/google-deepmind/mujoco/blob/178fb49c2b2ff48f59653515ab09b9cafca31b7a/src/engine/engine_sensor.c#L828

### DIFF
--- a/mujoco_warp/_src/sensor.py
+++ b/mujoco_warp/_src/sensor.py
@@ -30,6 +30,7 @@ from .types import JointType
 from .types import Model
 from .types import ObjType
 from .types import SensorType
+from .types import TrnType
 from .warp_util import event_scope
 from .warp_util import kernel as nested_kernel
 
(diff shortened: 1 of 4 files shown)


mujoco_warp/_src/sensor.py
```diff
@@ -1455,6 +1511,50 @@ def sensor_acc(m: Model, d: Data):
       outputs=[d.sensordata],
     )
 
+  if m.sensor_tendonactfrc_adr.size > 0:
```
erikfrey: guard shouldn't be needed right?

erikfrey (approved):
Looks great!  one nit

mujoco_warp/_src/sensor.py
thowell: guards are removed from `sensor_acc`

PR #270, opened 2025-05-17
add limit velocity sensors

thowell:
add jointlimvel and tendonlimvel sensors

mujoco references:
https://github.com/google-deepmind/mujoco/blob/178fb49c2b2ff48f59653515ab09b9cafca31b7a/src/engine/engine_sensor.c#L589
https://github.com/google-deepmind/mujoco/blob/178fb49c2b2ff48f59653515ab09b9cafca31b7a/src/engine/engine_sensor.c#L599

depends on #265 and #269


mujoco_warp/_src/sensor.py
```diff
@@ -608,6 +608,43 @@ def _ball_ang_vel(jnt_dofadr: wp.array(dtype=int), qvel_in: wp.array2d(dtype=flo
...
+  # skip if not limit
+  if efcid < ne + nf or efcid >= ne + nf + nl:
+    return
+
+  sensorid = sensor_limitvel_adr[limitvelid]
+  if efc_id_in[efcid] == sensor_objid[sensorid]:
+    efc_type = efc_type_in[efcid]
+    if efc_type == int(ConstraintType.LIMIT_JOINT.value) or efc_type == int(ConstraintType.LIMIT_TENDON.value):
+      worldid = efc_worldid_in[efcid]
+      _write_scalar(sensor_datatype, sensor_adr, sensor_cutoff, sensorid, efc_vel_in[efcid], sensordata_out[worldid])
```
erikfrey, line 833: zero out

mujoco_warp/_src/sensor.py
```diff
@@ -978,50 +1015,74 @@ def _sensor_vel(
...
+        d.ten_velocity,
+        d.actuator_velocity,
+        d.cvel,
+        d.subtree_linvel,
+        d.subtree_angmom,
+      ],
+      outputs=[d.sensordata],
+    )
+
+  if m.sensor_limitvel_adr.size > 0:
```
erikfrey: can skip guard

erikfrey (approved):
@thowell I kind of wonder if we should fuse these sensor kernels together that are all scanning over efc which is quite a large array, maybe we can look into that at some point using the benchmarks?  No action for now, just thinking out loud.

mujoco_warp/_src/sensor.py
thowell: added a kernel `_limit_vel_zero` to initialize limit velocity sensors to zero

mujoco_warp/_src/sensor.py
thowell: removed guards from `sensor_vel`

PR #274, opened 2025-05-19
Collision api building blocks

nvtw:
This MR is intended for discussing how to make the collision building blocks more independent of mujoco specific code and potentially move some code into the newton repository such that other solvers can use it without the need of a mujoco_warp dependency. I'm open to all suggestions, not sure if the proposed splitting is the right way to go.


mujoco_warp/_src/collision_primitive_newton.py
```diff
@@ -0,0 +1,1756 @@
...
+      contact6 = contact
+    elif i == 6:
+      contact7 = contact
+    elif i == 7:
+      contact8 = contact
+
+  return contact1, contact2, contact3, contact4, contact5, contact6, contact7, contact8, min(8, n)
+
+
+# newton/geometry/types.py (or within geometry/__init__.py)
```
nvtw: Some definitions from here and below already exists in Newton and would be included using imports instead of redefining it.

mujoco_warp/_src/convex_collision_newton.py
```diff
@@ -0,0 +1,886 @@
...
+from .support import all_same
+from .support import any_different
+
+wp.clear_kernel_cache()
+
+
+# newton/geometry/types.py (or within geometry/__init__.py)
+
+# Shape geometry types
+GEO_SPHERE = wp.constant(0)
```
nvtw: Some definitions int this file (e. g. all the GEO_x defs) already exists in Newton and would be included using imports instead of redefining it.

mujoco_warp/_src/collision_primitive_newton.py
```diff
@@ -0,0 +1,1756 @@
+import warp as wp
+
+
+from .math import closest_segment_point
```
nvtw: The functions imported there would need to be present in the newton repo too - probably a copy would be fine since they're quite small methods

mujoco_warp/_src/convex_collision_newton.py
```diff
@@ -0,0 +1,886 @@
+import warp as wp
+
+
+wp.config.enable_backward = False
+
+
+from .collision_primitive_newton import Geom
```
nvtw: The functions (and also the Geom type) imported there would need to be present in the newton repo too - probably a copy would be fine since they're quite small methods/structs

nvtw:
@erikfrey @adenzler-nvidia I need some advice for this MR: The separation of general collision code and mujoco specific code is done and matches what we discussed in the in-person-meeting last week. Shall I attempt to merge it as it is? Currently the general purpose collision code that would move out of mujoco in a second, later MR is in the file with the _geometry postfix.

adenzler-nvidia:
maybe let's wait for @kbayes GJK code to be merged as that will change some stuff again. Or are we not considering that for the moment?

nvtw:
Ok, I'll try to separate general purpose broad phase code from mujoco specific code in the meantime.

adenzler-nvidia:
@nvtw - is there still any relevance to this PR or is it all replaces by your newer ones?

nvtw:
I would like to merge this MR. But I was advised to wait until local memory warp arrays are available since that will change the API. Shall I close the MR in the meantime?

adenzler-nvidia:
we can leave it open then. I just though maybe it's replaced by your newer stuff.

nvtw:
This MR is replaced by https://github.com/google-deepmind/mujoco_warp/pull/526

PR #276, opened 2025-05-19
Flex bending passive force

quagla:


mujoco_warp/viewer.py
```diff
@@ -49,10 +49,6 @@ def key_callback(key: int) -> None:
 
 def _load_model():
```
erikfrey, line 50: do we anticipate we will need this again in the future?

if not, shall we revert to just calling `MjModel.from_xml_path`?

mujoco_warp/_src/passive.py
```diff
@@ -422,6 +422,47 @@ def _flex_elasticity(
...
+  )
+
+  if v[3] == -1:
+    return
+
+  force = wp.mat(0.0, shape=(nvert, 3))
+  for i in range(nvert):
+    for j in range(nvert):
+      for x in range(3):
+        force[i, x] -= flex_bending[16 * edgeid + 4 * i + j] * flexvert_xpos_in[worldid, v[j]][x]
```
erikfrey: it looks like you are doing some manual index computation here - could you instead give `flex_bending` some extra dims so it's faster/easier to index?

erikfrey (approved):
looks great, just two nits

mujoco_warp/_src/passive.py
quagla: Done.

mujoco_warp/viewer.py
quagla: I think it's nice to have a place where one can make edits. For doing the passive forces, it was quicker to prototype this, instead of first making a change to mujoco, submitting the CL, and waiting for the nightly build. Performance wise is the same, since from_xml_path creates a spec and then compiles it.

PR #285, opened 2025-05-21
WIP SDF collisions

jprozorova:
This MR is a starting point for SDF collisions.  In Mujoco SDF collisions are performed when at least 1 geometry is of SDF type and there is a plugin which implements distance and gradient method.  Mujoco already contains a couple of sdf plugins and offers a possibility to add more. So far I ported the general sdf collision logic and some distance and gradient functions of convex and non-convex shapes e.g.  from nut and bolt plugins.

But, thats not flexible enough and does not generalize to other non convex shapes. How do we proceed from here ? Do we intend to extend the plugin mechanism to warp?

### DIFF
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -20,6 +20,7 @@ from absl.testing import absltest
 from absl.testing import parameterized
 
 import mujoco_warp as mjwarp
+from .types import SDFType
 
 from . import test_util
 
@@ -27,6 +28,49 @@ from . import test_util
 class CollisionTest(parameterized.TestCase):
   """Tests the collision contact functions."""
 
+  _SDF_SDF = {
+      "_NUT_NUT": """
+<mujoco>
+  <extension>
+    <plugin plugin="mujoco.sdf.nut">
+      <instance name="nut1">
+        <config key="radius" value="0.27"/>
+      </instance>
+      <instance name="nut2">
+        <config key="radius" value="0.27"/>
+      </instance>
+    </plugin>
+  </extension>
+  <compiler autolimits="true"/>
+  <option sdf_iterations="10"/> <!-- Increased for better collision accuracy -->
+  <asset>
+    <mesh name="nut_mesh1">
+      <plugin instance="nut1"/>
+    </mesh>
+    <mesh name="nut_mesh2">
+      <plugin instance="nut2"/>
+    </mesh>
+  </asset>
+  <worldbody>
+    <!-- First nut (floating) -->
+    <body pos="0 0 0.5">
+      <joint type="free"/>
+      <geom type="sdf" name="nut1" mesh="nut_mesh1" rgba="0.83 0.68 0.4 1">
+        <plugin instance="nut1"/>
+      </geom>
+    </body>
+    <!-- Second nut (positioned to intersect) -->
+    <body pos="0 0 0.4">
+      <geom type="sdf" name="nut2" mesh="nut_mesh2" rgba="0.9 0.4 0.2 1">
+        <plugin instance="nut2"/>
+      </geom>
+    </body>
+    <light name="left" pos="-1 0 2" cutoff="80"/>
+    <light name="right" pos="1 0 2" cutoff="80"/>
+  </worldbody>
+</mujoco>
+"""
+  }
   _FIXTURES = {
     "box_plane": """
         <mujoco>
@@ -287,6 +331,36 @@ class CollisionTest(parameterized.TestCase):
   #        </worldbody>
   #      </mujoco>
   #    """,
+
+  @parameterized.parameters(_SDF_SDF.keys())
+  def test_sdf_collision(self, fixture):
+    """Tests collisions with different geometries."""
+    mjm, mjd, m, d = test_util.fixture(xml=self._SDF_SDF[fixture])
+
+    # Exempt GJK collisions from exact contact count check
+    # because GJK generates more contacts
+    allow_different_contact_count = True
+    mujoco.mj_collision(mjm, mjd)
+    mjwarp.collision(m, d)
+    for i in range(1):
+      actual_dist = mjd.contact.dist[i]
+      actual_pos = mjd.contact.pos[i]
+      actual_frame = mjd.contact.frame[i][0:3]
+      result = False
+      for j in range(d.ncon.numpy()[0]):
+        test_dist = d.contact.dist.numpy()[j]
+        test_pos = d.contact.pos.numpy()[j, :]
+        test_frame = d.contact.frame.numpy()[j].flatten()[0:3]
+        check_dist = np.allclose(actual_dist, test_dist, rtol=5e-2, atol=1.0e-2)
+        check_pos = np.allclose(actual_pos, test_pos, rtol=5e-2, atol=1.0e-2)
+        check_frame = np.allclose(actual_frame, test_frame, rtol=5e-2, atol=1.0e-2)
+        if check_dist and check_pos and check_frame:
+          result = True
+          break
+      np.testing.assert_equal(result, True, f"Contact {i} not found in Gjk results")
+
+    if not allow_different_contact_count:
+      self.assertEqual(d.ncon.numpy()[0], mjd.ncon)
 
   @parameterized.parameters(_FIXTURES.keys())
   def test_collision(self, fixture):
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -29,7 +29,7 @@ class CollisionTest(parameterized.TestCase):
   """Tests the collision contact functions."""
 
   _SDF_SDF = {
-      "_NUT_NUT": """
+    "_NUT_NUT": """
 <mujoco>
   <extension>
     <plugin plugin="mujoco.sdf.nut">
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -38,8 +38,33 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     if unsupported.any():
       raise NotImplementedError(f"{field_str} {field[unsupported]} not supported.")
 
-  #if mjm.nplugin > 0:
-  #  raise NotImplementedError("Plugins are unsupported.")
+  if mjm.nplugin > 0:
+    sdf_types = []
+    attr = []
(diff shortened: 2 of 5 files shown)


quagla:
Hi @jprozorova, I am the author of SDFs in MuJoCo, excited to see it ported to mjWarp!

I had a look at the code, I think the plugins will not be extended soon to Warp. Instead of using SDFType to choose between built-in functions, could we move them to be user callbacks (saved somewhere in mjModel)?

For general nonconvex geometries, it would be great to use NanoVDB..

mujoco_warp/_src/types.py
```diff
@@ -219,10 +219,16 @@ class GeomType(enum.IntEnum):
   CYLINDER = mujoco.mjtGeom.mjGEOM_CYLINDER
   BOX = mujoco.mjtGeom.mjGEOM_BOX
   MESH = mujoco.mjtGeom.mjGEOM_MESH
+  SDF = mujoco.mjtGeom.mjGEOM_SDF
   # unsupported: HFIELD,
   # NGEOMTYPES, ARROW*, LINE, SKIN, LABEL, NONE
 
 
+class SDFType(enum.IntEnum):
```
quagla: I think SDFType can be removed from the engine to the test as it's something only the specific user function supplied by the user knows about.

mujoco_warp/_src/io.py
```diff
@@ -370,6 +397,8 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     dof_tri_row=wp.array(dof_tri_row, dtype=int),
     dof_tri_col=wp.array(dof_tri_col, dtype=int),
     geom_type=wp.array(mjm.geom_type, dtype=int),
+    geom_sdf_plugin_type=wp.array(sdf_types, dtype=int),
```
quagla: Do we need geom_sdf_plugin_type? I think the only information that now sdf_narrowphase_kernel needs is whether it is a user supplied sdf or now.

mujoco_warp/_src/io.py
```diff
@@ -370,6 +397,8 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
...
+    geom_sdf_plugin_attr=wp.array(attr, dtype=wp.vec3f),
```
quagla: I think we should try to match mujoco's names, so it would be only plugin_attr as all plugins read from the same array.

quagla: @erikfrey do you have an opinion on whether this is important?

mujoco_warp/_src/collision_driver_test.py
```diff
@@ -288,6 +390,58 @@ class CollisionTest(parameterized.TestCase):
...
+      @wp.func
+      def user_sdf_grad(p: wp.vec3, attr: wp.vec3, sdf_type: int) -> wp.vec3:
+            if sdf_type == int(SDFType.NUT.value):
+              return nut_sdf_grad(p, attr)
+            elif sdf_type == int(SDFType.BOLT.value):
+              return bolt_sdf_grad(p, attr)
+            return wp.vec3()
+
+
+      collision_sdf.user_sdf = user_sdf
```
quagla: It there a way to make this more general, e.g. passing a dictionary of functions in order to make the model more composable? I guess not because it would imply a callback. @nimrod-gileadi for visibility

mujoco_warp/_src/io.py
jvonmuralt: we need to somehow pass information to the user_sdf() about the type (bolt/nut/etc) or am I missing something? Is there maybe a better place to do it?

mujoco_warp/_src/collision_driver_test.py
jvonmuralt: no, I think thats not possible either. In warp kernels/ functions called from the kernels we cannot dynamically select functions from dictionaries

mujoco_warp/_src/io.py
quagla: I guess in my mind I was thinking that each sdf would have a different function but you are right with the limitation that the user can only pass one custom sdf, we have to pass the type.

mujoco_warp/_src/collision_driver_test.py
quagla: Since - for a given run - the SDF attributes are not allowed to change dynamically, what if we ask the user to declare them as wp.static? From https://nvidia.github.io/warp/codegen.html#example-function-pointers it looks like we could then store in `Model` the index corresponding to each geom of the functions in a user-provided dict (or list) and call the appropriate one from the `sdf_eval` kernel by just knowing which geom we are dealing with.

jvonmuralt: For example, we define an array of len = number of geometries where each entry contains a number - key from the dictionary (which sdf should be used for this particular geom). While this is fixed, the  geom index will be worked out based on tid in the kernel and therefore it is not static anymore.  So we can modify this example like:

op_handlers = {
    0: do_add,
    1: do_sub,
    2 : do_mul,
}
types = [0, 1]
@wp.kernel
def operate(input: wp.array(dtype=inputs.dtype, ndim=2), output: wp.array(dtype=wp.float32)):
        tid = wp.tid()
        a, b = input[tid, 0], input[tid, 1]
        output[tid] =  wp.static(op_handlers[types[tid]])(a, b)

and this would fail  with ;Error evaluating static expression: name 'tid' is not defined. Make sure all variables used in the static expression are constant.

so you need to know the key at compile time to be able to use wp.static, and we only can know it at runtime based on the geom this thread is working with

or do you have something else in mind?

quagla: What about https://nvidia.github.io/warp/codegen.html#dynamic-kernel-creation

jvonmuralt: I do not see how to use it. It seems to be "dynamic" at  kernel creation, so it can generate a new kernel specialized with values/functions.  but since our single kernel launch processes different pairs of different types, again we cannot fix that at kernel creation time. 

I only see  that https://nvidia.github.io/warp/codegen.html#advanced-example-branching-elimination-with-static-loop-unrolling can be used to improve performance to eliminate branching, but from code design it is still the same if-else stuff.  I tried something similar and when I checked cuda code very many functions got generated because sdf is called in a loop and in different contexts, so one would need to evaluate carefully which approach is actually better here

mujoco_warp/_src/types.py
```diff
@@ -1081,6 +1082,8 @@ class Model:
   dof_tri_row: wp.array(dtype=int)  # warp only
   dof_tri_col: wp.array(dtype=int)  # warp only
   geom_type: wp.array(dtype=int)
+  geom_sdf_plugin_type: wp.array(dtype=int)
```
quagla: Can  we rename it to geom_plugin?

mujoco_warp/_src/types.py
```diff
@@ -1081,6 +1082,8 @@ class Model:
...
+  plugin_attr: wp.array(dtype=wp.vec3f)
```
quagla: Please move it after sensors (like in mjModel.h).

mujoco_warp/_src/types.py
quagla: Oh I see you want to store plugin[geom_plugin]. Maybe geom_plugin_global_id then.

mujoco_warp/_src/collision_driver_test.py
```diff
@@ -354,6 +455,54 @@ class CollisionTest(parameterized.TestCase):
   #        </worldbody>
   #      </mujoco>
   #    """,
+  class SDFType(enum.IntEnum):
+    BOLT = 2
```
quagla: It's a bit of a hack but if you want to be safer about global ids you could load the xml into a temporary mjModel and read the ids when you set the class up.

mujoco_warp/_src/io.py
jvonmuralt: With the current implementation both geom_plugin_global_id and geom_plugin_attr are not for all plugins but for geometry plugins only (thats why I called the second geom_plugin_attr instead of plugin_attr).
Alternatively we can have an array for plugins like in mujoco:  **plugin** with plugin_ids and **plugin_attr** and an array for geometries geom_plugin_index (?) which maps each geometry to the plugin index. 
Let me know if I should change it to that

quagla: Yes please it would be better to follow mujoco and optimize later if needed (and if so change mujoco accordingly).

quagla (approved):

quagla:
Amazing, thanks for the quick update! Have you tested it? I have forked your branch and copied `nutbolt.xml` (and `scene.xml`) to `test_data` and tried `python mujoco_warp/viewer.py --mjcf mujoco_warp/test_data/collision_sdf/nutbolt.xml` but I see no collisions happening

quagla:
From my point of view, the PR is ready for merging @erikfrey

erikfrey:
Ok thanks!  I'll have time to review it tomorrow, but I won't do a deep review as I appreciate the attention to detail you both have already put in.

mujoco_warp/__init__.py
```diff
@@ -71,3 +72,4 @@
 from ._src.types import Option as Option
 from ._src.types import Statistic as Statistic
 from ._src.types import TrnType as TrnType
+from .test_data.collision_sdf.utils import register_sdf_plugins as register_sdf_plugins
```
erikfrey, line 75: I wouldn't expect us to add this to the public API - can we just have code import from `mujoco_warp.test_data.collision_sdf.utils` ?

mujoco_warp/viewer.py
```diff
@@ -84,6 +85,7 @@ def _main(argv: Sequence[str]) -> None:
     print("Engine: MuJoCo C")
   else:  # mjwarp
     print("Engine: MuJoCo Warp")
+    register_sdf_plugins(mjwarp.collision_sdf)
```
erikfrey, line 88: what if instead we did like:

```py
# check if the file has any mujoco.sdf test plugins
spec = mujoco.MjSpec.from_file(_MODEL_PATH.value)
if any(p.plugin_name.startswith("mujoco.sdf") for p in spec.plugins):
  from mujoco_warp.test_data.collision_sdf.utils import register_sdf_plugins as register_sdf_plugins
  register_sdf_plugins(mjwarp.collision_sdf)
```

erikfrey:
@jprozorova we are going to submit this and then address my comments separately.  Thanks again for this awesome contribution!

PR #294, opened 2025-05-23
Optimization: fuse linesearch_zero_jv and linesarch_init_jv

adenzler-nvidia:
For humanoid, but I'm seeing good stuff for many other things as well. This is a tiny piece of an effort to work on all of these tiny kernels, so sometimes it might not even show up in the overall runtime. But looking at low-level traces, there is a big difference.

Time spent in the 2 kernels before/after for 8k humanoids: 11ms -> 1.5ms.

main:

```

 Total JIT time: 1.18 s
 Total simulation time: 0.16 s
 Total steps per second: 2,555,648
 Total realtime factor: 12,778.24 x
 Total time per step: 391.29 ns

```
this:

```

 Total JIT time: 1.12 s
 Total simulation time: 0.15 s
 Total steps per second: 2,773,893
 Total realtime factor: 13,869.46 x
 Total time per step: 360.50 ns

```

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -1282,7 +1282,10 @@ def _linesearch(m: types.Model, d: types.Data):
 
   # jv = efc_J @ search
   # TODO(team): is there a better way of doing batched matmuls with dynamic array sizes?
-  dofs_per_thread = 50
+  if m.nv > 50:
+    dofs_per_thread = 20
+  else:
+    dofs_per_thread = 50
 
   threads_per_efc = ceil(m.nv / dofs_per_thread)
   # we need to clear the jv array if we're doing atomic adds.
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -1268,16 +1283,28 @@ def _linesearch(m: types.Model, d: types.Data):
...
-    outputs=[d.efc.jv],
-  )
+
+  # if we are only using 1 thread, it makes sense to do more dofs as we can also skip the
+  # init kernel. If we use more than 1 thread, dofs_per_thread is lower for better load balancing.
+
+  if m.nv > 50:
+    dofs_per_thread = 20
+  else:
+    dofs_per_thread = 50
```
adenzler-nvidia, line 1293: I made a best-effort attempt at figuring out this threshold using some local experiments. We can still tune it as we get more and better benchmarks.

erikfrey:
Wow, very cool!  Deferring to @thowell for full review.

PR #295, opened 2025-05-23
add body transmission 

thowell:
add body transmission

mujoco reference: https://github.com/google-deepmind/mujoco/blob/e389872c7db149f9efb1f1501a963f6ed6f12421/src/engine/engine_core_smooth.c#L1348


erikfrey (commented):
@thowell there seems to be some changes to add `contact_exclude` in `collision_driver` but it's not clear if that's related to this PR?  Just want to make sure I'm reviewing the right thing

thowell:
@erikfrey `contact_exclude` is added in this pr because the body transmission utilizes this information. mujoco reference https://github.com/google-deepmind/mujoco/blob/072c872deb3d8d41e85857674f3ec294a79e6f78/src/engine/engine_core_smooth.c#L1395 for the body transmission use case

instead of adding the `contact_exclude` field the status https://github.com/google-deepmind/mujoco/blob/072c872deb3d8d41e85857674f3ec294a79e6f78/src/engine/engine_collision_driver.c#L1400 could be computed when needed

erikfrey:
OK, I like the idea of computing it when needed because the logic to compute it seems quite simple anyway, and the field is only read in one place, at least in this PR.  Let's try that!

thowell:
@erikfrey `Contact.exclude` is removed and this information is now computed in `_transmission_body_moment`

mujoco_warp/_src/smooth.py
```diff
@@ -2038,10 +2048,141 @@ def _transmission(
...
+
+  # irrelevant contact, continue
+  if b1 != bodyid and b2 != bodyid:
+    return
+
+  contact_exclude = int(contact_dist_in[conid] >= contact_includemargin_in[conid])
+
+  # mark contact normals in efc_force
+  if contact_exclude == 0:
+    if dofid == 0:
```
erikfrey: do we need this `if dofid == 0:` in both branches?

seems like we could move it up out of the if statements?

mujoco_warp/_src/smooth.py
```diff
@@ -2038,10 +2048,141 @@ def _transmission(
...
+  # Data out:
+  actuator_moment_out: wp.array3d(dtype=float),
+):
+  worldid, trnbodyid, dofid = wp.tid()
+
+  ncon = actuator_trntype_body_ncon_in[worldid, trnbodyid]
+
+  if ncon > 0:
+    actid = actuator_trntype_body_adr[trnbodyid]
+    actuator_moment_out[worldid, actid, dofid] /= -float(ncon)
```
erikfrey, line 2180: I wish there was a way to fuse this into the previous kernel but I can't think of any good way right now!  No action required.

mujoco_warp/_src/smooth.py

PR #299, opened 2025-05-25
add contact override

thowell:
contact override for parameters: `margin`, `solref`, `solimp`, and `friction` with `EnableBit.OVERRIDE`


erikfrey:
Hmm this is a pretty esoteric MuJoCo feature and it reaches pretty deep into every single contact function.

Is this used in practice?  @thowell can you check the benchmarks and make sure it doesn't slow down perf?

erikfrey:
@thowell given that in the documentation, this parameter is justified as a debug / experimentation parameter, maybe we can just leave that to C MuJoCo to handle, and avoid it in MJWarp, WDYT?

PR #300, opened 2025-05-25
configurable `block_dim` settings for `wp.launch_tiled`

thowell:
`wp.launch_tiled` settings `block_dim`, organized in `BlockDim`, are configurable fields of `Model`

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -22,6 +22,7 @@ MJ_MINVAL = mujoco.mjMINVAL
 MJ_MAXVAL = mujoco.mjMAXVAL
 MJ_MINIMP = mujoco.mjMINIMP  # minimum constraint impedance
 MJ_MAXIMP = mujoco.mjMAXIMP  # maximum constraint impedance
+MJ_MAXCONPAIR = mujoco.mjMAXCONPAIR
 MJ_NREF = mujoco.mjNREF
 MJ_NIMP = mujoco.mjNIMP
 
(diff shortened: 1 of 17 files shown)


erikfrey:
@adenzler-nvidia - in your experience, how sensitive is performance to block_dim?

I haven't done an exhaustive study but the few times I've futzed with it, I've found that setting it to 32 and then forgetting about it worked pretty well.

But I don't know if that will hold true for different scenes or different cards.  Generally speaking, adding config is an invitation to tune or tweak - so all else even, less configuration is better.  Are we sure this warrants an invitation for twiddling?

adenzler-nvidia:
I think the fact that we set everything to 32 is just because we mainly use it for cholesky. The differences can be huge. Usually the perf profile is pretty similar for the same generation but it can change between generations. Or sometimes also for different driver/compiler versions.

Maybe we can test-drive this a bit by adding an autotuning script that does an exhaustive search, and see if there are any wild differences coming out of it?

mujoco_warp/_src/types.py
```diff
@@ -26,6 +26,30 @@
 MJ_NIMP = mujoco.mjNIMP
 
 
+class BlockDim:
```
erikfrey, line 31: @thowell Can you add a docstring here and make it clear that this is an experimental option that we may remove?

thowell: added a docstring

thowell:
@adenzler-nvidia added a todo regarding configurable `block_dim` settings

PR #301, opened 2025-05-25
update smooth.py

thowell:
update smooth.py

- `array2df` -> `wp.array2d(dtype=float)`
- `array3df` -> `wp.array3d(dtype=float)`


adenzler-nvidia (approved):
LGTM thanks for cleaning up!

PR #302, opened 2025-05-26
Optimization: launch fewer threads and loop over nv in init_qfrc_constraint

adenzler-nvidia:
Seems to have a good effect on many assets, especially if you look at low-level traces. Makes sense if you think about it - we only need to load the force once.

Total time spent in this kernel for some assets (main -> this):
humanoid: 4.4ms -> 1.9ms
n_humanoids: 333ms -> 29.8ms
internal asset with nv==70: 94ms -> 14.58ms
Ant: 2.7s -> 294ms
cartpole: 693ms -> 286ms

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -1757,6 +1757,37 @@ def update_constraint_init_qfrc_constraint(
     efc_J_in[efcid, dofid] * efc_force_in[efcid],
   )
 
+@wp.kernel
+def update_constraint_init_qfrc_constraintv2(
+  # Model in:
+  nv: int,
+  # Data in:
+  nefc_in: wp.array(dtype=int),
+  efc_worldid_in: wp.array(dtype=int),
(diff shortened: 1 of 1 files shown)


erikfrey:
This is interesting, I've been leaving @thowell review comments to undo `for i in range(nv)` in kernels.

I think there's something important here to learn - @adenzler-nvidia is there some rule of thumb on when to parallelize?  Naively I would have expected that serially unrolling a for loop in a single thread would be slower than launching more threads.

thowell:
@adenzler-nvidia should we prioritize creating a suite of benchmarks in order to more efficiently optimize performance and confirm that improvements generalize to various environments?

adenzler-nvidia:
@thowell I think we're quite good once we have the kitchen in - then what's missing to make tuning these things a bit more efficient is some automation/scripting such that we don't need to run things manually all the time.

That being said, for these tiny kernels the only real indication you get is from low-level profiling tools, and then at some point it's going to compound.

PR #307, opened 2025-05-27
updates for multiple contacts

thowell:
- `_multiple_contacts` is updated with additional inputs:
  - `ncontact` with default `4`
  - `npolygon` with default `8`
  - `perturbation_angle` with default `1e-3` (radians) (https://github.com/google-deepmind/mujoco/blob/568620dd2f0f661dae2baafe3c5dee6d1d9f88ae/src/engine/engine_collision_convex.c#L964)
- `geom_type` check determines settings for `_multiple_contacts`
  - mujoco reference: https://github.com/google-deepmind/mujoco/blob/6b919162d1ecc32ee36d2cd48c99b86d867d4830/src/engine/engine_collision_convex.c#L958

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -27,7 +27,6 @@ from .support import all_same
 from .support import any_different
 from .types import MJ_MINVAL
 from .types import Data
-from .types import EnableBit
 from .types import GeomType
 from .types import Model
 from .types import vec5
@@ -756,8 +755,6 @@ def _gjk_epa_pipeline(
     collision_pairid_in: wp.array(dtype=int),
     collision_worldid_in: wp.array(dtype=int),
     ncollision_in: wp.array(dtype=int),
-    # In:
-    enable_multiccd: bool,
     # Data out:
     ncon_out: wp.array(dtype=int),
     contact_dist_out: wp.array(dtype=float),
@@ -859,8 +856,7 @@ def _gjk_epa_pipeline(
     # TODO(btaba): split get_multiple_contacts into a separate kernel.
 
     if (
-      enable_multiccd
-      and geom_type[g1] != int(GeomType.ELLIPSOID.value)
+      geom_type[g1] != int(GeomType.ELLIPSOID.value)
       and geom_type[g1] != int(GeomType.SPHERE.value)
       and geom_type[g2] != int(GeomType.ELLIPSOID.value)
       and geom_type[g2] != int(GeomType.SPHERE.value)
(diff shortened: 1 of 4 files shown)


mujoco_warp/_src/collision_convex.py
```diff
@@ -848,8 +857,18 @@ def gjk_epa_sparse(
       return
 
     # TODO(btaba): split get_multiple_contacts into a separate kernel.
-    # TODO(team): multiccd enablebit
-    count, points = _multiple_contacts(geom1, geom2, depth, normal)
+
+    if (
+      enable_multiccd
```
btaba: Sorry for the comment, but I'd almost prefer this flag not be introduced here. When is there a case when you don't want multi contacts enabled? These flags seem more like legacy to me

kbayes:
I can't imagine needing a MULTICCD flag in mjWarp, as many models are unstable without multicontact. This was an experimental flag in C MuJoCo as the original multiccd would need to run the narrowphase multiple times to gather the contact points, hence a huge hit in performance.

thowell:
@btaba @kbayes `EnableBit.MULTICCD` has been removed

btaba:
> @btaba @kbayes `EnableBit.MULTICCD` has been removed

yay!

mujoco_warp/_src/collision_convex.py
```diff
@@ -848,8 +854,17 @@ def gjk_epa_sparse(
...
-    # TODO(team): multiccd enablebit
-    count, points = _multiple_contacts(geom1, geom2, depth, normal)
+
+    if (
+      geom_type[g1] != int(GeomType.ELLIPSOID.value)
+      and geom_type[g1] != int(GeomType.SPHERE.value)
+      and geom_type[g2] != int(GeomType.ELLIPSOID.value)
+      and geom_type[g2] != int(GeomType.SPHERE.value)
+    ):
+      count, points = _multiple_contacts(geom1, geom2, depth, normal)
```
btaba: just from a readability perspective, I would not set defaults on `_multiple_contacts` and I'd be explicit about the args at the callsite here

btaba: Also I'm a bit confused, capsule<>capsule generates 4 points?

btaba: Ok also for readability, I'd maybe do:

```
if geom_type[g1] in (ellipsoid, sphere) and geom_type[g2] in (ellipsoid, sphere):
  # do stuff
else:
  # stuff only involving a mesh contact?
```

thowell: removed `_multiple_contacts` defaults

thowell: i don't think the operation `value in (value1, value2)` is supported in `@wp.kernel`/`@wp.func`

following the readability comment, the implementation is updated to have the `_multiple_contacts` call with 4 contacts switched to the `else` case

btaba: LGTM!

PR #308, opened 2025-05-27
height field maxconpair check

thowell:
limits the number of contacts with a height field to `mjMAXCONPAIR`

depends on #307

mujoco reference: https://github.com/google-deepmind/mujoco/blob/568620dd2f0f661dae2baafe3c5dee6d1d9f88ae/src/engine/engine_collision_convex.c#L1327

### DIFF
--- a/mujoco_warp/_src/collision_hfield.py
+++ b/mujoco_warp/_src/collision_hfield.py
@@ -19,6 +19,7 @@ from typing import Any
 
 import warp as wp
 
+from .types import MJ_MAXCONPAIR
 from .types import Data
 from .types import GeomType
 from .types import Model
@@ -329,7 +330,7 @@ def _hfield_midphase(
   ncol = hfield_ncol[dataid]
 
   # Loop through grid cells and add pairs for all triangles
-  # TODO(vreutskyy): propose a way to limit the number of contacts between a hfield and the other geom
+  count = int(0)
   for j in range(min_j, max_j + 1):
     for i in range(min_i, max_i + 1):
       # Each grid cell contains two triangles
@@ -351,6 +352,10 @@ def _hfield_midphase(
         collision_hftri_index_out[new_pairid] = base_idx + t
         collision_pairid_out[new_pairid] = pair_id
         collision_worldid_out[new_pairid] = worldid
+
+        count += 1
+        if count >= MJ_MAXCONPAIR:
+          return
 
 
 def hfield_midphase(m: Model, d: Data):
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -22,6 +22,7 @@ from absl.testing import parameterized
 import mujoco_warp as mjwarp
 
 from . import test_util
+from . import types
 
 
 class CollisionTest(parameterized.TestCase):
@@ -664,6 +665,31 @@ class CollisionTest(parameterized.TestCase):
     mjwarp.collision(m, d)
 
     self.assertEqual(d.ncon.numpy()[0], mjd.ncon)
+
+  def test_hfield_maxconpair(self):
+    _XML = f"""
+    <mujoco>
+      <asset>
+        <hfield name="hfield" nrow="10" ncol="10" size="1e-6 1e-6 1 1"/>
+      </asset>
+      <worldbody>
+        <body>
+          <joint type="slide" axis="0 0 1"/>
+          <geom type="sphere" size=".1"/>
+        </body>
+        <geom type="hfield" hfield="hfield"/>
+      </worldbody>
+      <keyframe>
+        <key qpos=".0999"/>
+      </keyframe>
+    </mujoco>
+    """
+
+    _, _, m, d = test_util.fixture(xml=_XML, keyframe=0)
+
+    mjwarp.collision(m, d)
+
+    np.testing.assert_equal(d.ncon.numpy()[0], types.MJ_MAXCONPAIR)
 
   # TODO(team): test contact parameter mixing
 


mujoco_warp/_src/collision_hfield.py
```diff
@@ -352,6 +353,10 @@ def _hfield_midphase(
         collision_pairid_out[new_pairid] = pair_id
         collision_worldid_out[new_pairid] = worldid
 
+        count += 1
+        if count >= MJ_MAXCONPAIR:
```
btaba: Hello! I'm confused by this PR. Correct me if I'm wrong but this seems to be culling the number of triangles to check for collisions, not the actual collisions themselves.

kbayes: What's the difference between this line and the check on lines 348 - 349? The docstrings says - "nconmax: Maximum number of contacts"

vreutskyy: There can be confusion in the terminology, since "contact" sometimes refers to a "contact pair"—a pair of colliding geometries—or a "contact point," which is the actual point with a position, normal, and depth. Both are called "contact," but they mean different things.

For a convex-convex contact pair, all contact points will have the same normal and lie in the same plane (the separating axis and plane), so we can use only four contact points forming the largest quadrilateral. However, for an hfield-convex contact pair, the situation is different because the hfield can be concave, and we may need many more contact points for stability.

The goal of this task is to limit the number of contact points generated for the hfield-convex contact pair to 50, not the number of triangles we test. (I wouldn't add this TODO comment if it were so trivial.) My thoughts are that we could allocate an extra buffer to store contact point counters for every contact pair, and then, in gjk_narrowphase, keep track of the number of contact points we create so we can stop when we reach 50—or something like that.

thowell: @kbayes the pr is updated so there is no longer the `count >= MJ_MAXCONPAIR` check in `collision_hfield._hfield_midphase`. 

the `MJ_MAXCONPAIR` check was limiting the number of collision pairs for a geom pair that included a height field. the check `new_pairid >= ncon_max_in` limits the total number of collision pairs for all geom pair types for all worlds.

thowell: @btaba @kbayes @vreutskyy the pr is updated

`ncon_hfield` is added to `Data` and stores counts for number of contacts added for a geom pair that includes a height field. the number of contacts for a geom pair with a height field is limited in `collision_convex.gjk_epa_sparse`.

mujoco_warp/_src/types.py
```diff
@@ -1153,6 +1154,7 @@ class Data:
     njmax: maximum number of constraints                        ()
     solver_niter: number of solver iterations                   (nworld,)
     ncon: number of detected contacts                           ()
+    ncon_hfield: number of contacts per geom pair with hfield   (ncon,)
```
btaba: In io.py you have this as (nconmax,) not (ncon,). But are you sure the dim is nconmax? You index by the upper_tri_index.

btaba (changes requested):

mujoco_warp/_src/collision_hfield.py
btaba: Nice thanks for fixing that bit! Resolving this thread

mujoco_warp/_src/collision_convex.py
```diff
@@ -853,6 +857,13 @@ def gjk_epa_sparse(
 
     frame = make_frame(normal)
     for i in range(count):
+      # limit maximum number of contacts with height field
+      if geomtype1 == int(GeomType.HFIELD.value) or geomtype2 == int(GeomType.HFIELD.value):
```
btaba: Should this be done in a separate kernel, that way it can be more easily extended? Say we want to cull points a different way; 50 is too high for batch sim and picking a random 4 isn't great either – I picked 4 on the manifold in MJX but TBD if folks want to do that here or not. Either way I think this PR should make things easier to extend in the collision driver (keep in mind this GJK codepath will get replaced soon anyways).

mujoco_warp/_src/types.py
thowell: updates to the pr should address this issue. thanks!

mujoco_warp/_src/collision_convex.py
thowell: moved this functionality to `_max_contacts_height_field`

mujoco_warp/_src/io.py
```diff
@@ -19,10 +19,21 @@
...
+from . import math
 from . import types
 
 
+def _hfield_geom_pair(mjm: mujoco.MjModel) -> Tuple[int, np.array]:
+  geom1, geom2 = np.triu_indices(mjm.ngeom, k=1)
+  geom_type_hf = mujoco.mjtGeom.mjGEOM_HFIELD
+  has_hfield = (mjm.geom_type[geom1] == geom_type_hf) | (mjm.geom_type[geom2] == geom_type_hf)
+  nhfieldgeompair = np.sum(has_hfield)
+  geompair2hfgeompair = -1 * np.ones(mjm.ngeom * (mjm.ngeom - 1) // 2, dtype=int)
```
btaba, line 31: It's a bit unfortunate that `mjm.ngeom * (mjm.ngeom - 1) // 2` needs to get allocated for this functionality, but I can't really think of a more clever way (besides a sparse array)

mujoco_warp/_src/collision_convex.py
```diff
@@ -187,6 +189,30 @@ def _expand_polytope(count: int, prev_count: int, dists: vecc3, tris: mat2c3, p:
...
+  g2: int,
+  worldid: int,
+  # Data out:
+  ncon_hfield_out: wp.array2d(dtype=int),
+):
+  hfield = int(GeomType.HFIELD.value)
+  if geom_type[g1] == hfield or (geom_type[g2] == hfield):
+    geompairid = upper_tri_index(ngeom, g1, g2)
+    hfgeompairid = geompair2hfgeompair[geompairid]
+    hfgeompairid = wp.atomic_add(ncon_hfield_out[worldid], hfgeompairid, 1)
```
btaba: nit: `hfncon` or something ?

btaba (approved):

mujoco_warp/_src/collision_convex.py

PR #314, opened 2025-05-28
update for `qLD_updates`

thowell:
see the mujoco reference lines:
- https://github.com/google-deepmind/mujoco/blob/287f46b220513dde292f47dc4eb7dd4647610123/src/engine/engine_core_smooth.c#L1807
- https://github.com/google-deepmind/mujoco/blob/287f46b220513dde292f47dc4eb7dd4647610123/src/engine/engine_core_smooth.c#L1858

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -109,6 +109,8 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   qLD_updates, dof_depth = {}, np.zeros(mjm.nv, dtype=int) - 1
 
   for k in range(mjm.nv):
+    if mjd.M_rownnz[k] == 1:
+      continue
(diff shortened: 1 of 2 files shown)


kbayes (approved):

mujoco_warp/_src/io.py
```diff
@@ -109,6 +109,8 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   qLD_updates, dof_depth = {}, np.zeros(mjm.nv, dtype=int) - 1
 
   for k in range(mjm.nv):
+    if mjd.M_rownnz[k] == 1:
```
erikfrey, line 113: nit: let's add the same comment here "skip diagonal rows" as this was unobvious enough that I forgot to do it      :-)

thowell: done

PR #322, opened 2025-05-30
Improve documentation and clean up formatting of smooth.py

erikfrey:
Also remove some needless if-guards - Warp is smart enough to skip kernel launches for size-0 dims.

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 # ==============================================================================
 
-import mujoco
 import warp as wp
 
 from . import math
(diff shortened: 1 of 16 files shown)


mujoco_warp/_src/smooth.py
```diff
@@ -663,52 +633,46 @@ def _light_fn(
...
-        d.light_xpos,
-        d.subtree_com,
-      ],
-      outputs=[d.light_xpos, d.light_xdir],
-    )
+  """
+  Computes camera and light positions and orientations.
+
+  Updates the global positions and orientations for all cameras and lights in the model, including special handling for
+  tracking and target modes.
```
yuvaltassa: odd wrapping

yuvaltassa: oh, are we going with 120?

erikfrey: Oh whoops, no we are not!  Good catch.  We have max doc length set to 92.

Our ruff was misconfigured and was not catching this - fixed now and also fixed the errors resulting from ruff catching more issues.

PR #327, opened 2025-05-31
add slidercrank transmission

thowell:
mujoco reference: https://github.com/google-deepmind/mujoco/blob/fe81373ffe7e648a5cecef16c8cb72dac38fcfa6/src/engine/engine_core_smooth.c#L1103


erikfrey (approved):
looks great

PR #330, opened 2025-06-01
update viewer

thowell:
on reset, options are updated and if any options have changed `step` is recompiled

### DIFF
--- a/mujoco_warp/viewer.py
+++ b/mujoco_warp/viewer.py
@@ -52,6 +52,16 @@ def _load_model():
   return spec.compile()
 
 
+def _compile_step(m, d):
+  mjwarp.step(m, d)
+  # double warmup to work around issues with compilation during graph capture:
+  mjwarp.step(m, d)
+  # capture the whole step function as a CUDA graph
+  with wp.ScopedCapture() as capture:
+    mjwarp.step(m, d)
+  return capture.graph
+
+
 def _main(argv: Sequence[str]) -> None:
   """Launches MuJoCo passive viewer fed by MJWarp."""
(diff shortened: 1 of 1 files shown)


mujoco_warp/viewer.py
```diff
@@ -100,14 +104,37 @@ def _main(argv: Sequence[str]) -> None:
...
         wp.copy(d.xfrc_applied, wp.array([mjd.xfrc_applied.astype(np.float32)]))
         wp.copy(d.qpos, wp.array([mjd.qpos.astype(np.float32)]))
         wp.copy(d.qvel, wp.array([mjd.qvel.astype(np.float32)]))
         wp.copy(d.time, wp.array([mjd.time], dtype=wp.float32))
 
+        # on reset, update options and potentially recompile step
+        if mjd.time == 0.0:
+          recompile = False
+
+          for field in m.opt.__dataclass_fields__:
```
erikfrey: i wonder if there's something more general we can do here, something like:

```py
import pickle

...

model_hash = pickle.dumps(mjm)

...

with viewer:
  while True:
    ...
    if model_hash != pickle.dumps(mjm):
      model_hash = pickle.dumps(mjm)
      m = mjwarp.put_model(mjm)
      graph = _compile_step(m, d)
```

erikfrey (approved):
Looks great, just one suggestion

mujoco_warp/viewer.py
thowell: great suggestion, this concept is now utilized. thanks!

no need now to reset for the option changes to take effect.

PR #331, opened 2025-06-01
address TODOs in util_misc.py

thowell:


mujoco_warp/_src/util_misc.py
```diff
@@ -194,13 +200,7 @@ def wrap_inside(end: wp.vec4, radius: float):
   Returns:
     0.0 if wrap else -1.0, pair of 2D wrap points
   """
-  # defaults
-  # TODO(team): update kernel analyzer to allow defaults
```
erikfrey, line 198: we should still update the kernel analyzer to allow defaults, that seems reasonable

(you can leave the TODO)

erikfrey (approved):
looks great, just one nit

mujoco_warp/_src/util_misc.py
thowell: done

PR #337, opened 2025-06-02
Skip solver if there are no constraints

adenzler-nvidia:
For inverted pendulum without any joint limits:

main

```

 Total JIT time: 1.10 s
 Total simulation time: 0.49 s
 Total steps per second: 16,787,727
 Total realtime factor: 335,754.54 x
 Total time per step: 59.57 ns

```

this:

```

 Total JIT time: 0.15 s
 Total simulation time: 0.15 s
 Total steps per second: 53,858,953
 Total realtime factor: 1,077,179.06 x
 Total time per step: 18.57 ns
```

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -293,10 +293,6 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
 
     nxn_pairid[pairid] = i
 
-  # skip contact generation if there are no potential contact pairs
-  if np.all(nxn_pairid == -2):
-    mjm.opt.disableflags |= types.DisableBit.CONTACT.value
-
   def create_nmodel_batched_array(mjm_array, dtype):
     array = wp.array(mjm_array, dtype=dtype)
     array.ndim += 1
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/io.py
```diff
@@ -293,6 +293,10 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
 
     nxn_pairid[pairid] = i
 
+  # skip contact generation if there are no potential contact pairs
+  if np.all(nxn_pairid == -2):
```
erikfrey: hmm, let's skip this one if you don't mind, for two reasons:

1) we have an upcoming PR where we will filter `nxn_pairid` to not include any -2 values.  once this happens then your optimization will naturally occur because we'll call the kernel launch with a zero in the dims which means no kernel launch

2) i'd rather we don't write opt values as those are user settings, and overwriting them may suprise user code that expects them to remain frozen

erikfrey (approved):
LGTM for one change, but request for the other

mujoco_warp/_src/io.py
adenzler-nvidia: I agree it's really a hack, removed.

PR #347, opened 2025-06-06
update passive

thowell:
- fixes a bug where `qfrc_fluid` was not initialized to zero before calling `support.apply_ft`
- removes unused `import Option`
- `test_passive` is updated

### DIFF
--- a/mujoco_warp/_src/passive.py
+++ b/mujoco_warp/_src/passive.py
@@ -46,8 +46,8 @@ def _spring_damper_dof_passive(
   stiffness = jnt_stiffness[worldid, jntid]
   damping = dof_damping[worldid, dofid]
 
-  has_stiffness = stiffness == 0.0
-  has_damping = damping == 0.0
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/passive.py
```diff
@@ -466,16 +465,15 @@ def _flex_bending(
 @event_scope
 def passive(m: Model, d: Data):
   """Adds all passive forces."""
+  d.qfrc_spring.zero_()
```
erikfrey: In #355 we learned that zeroing arrays does show up in benchmarks and is not insignificant!

Can you think of a way to structure this code so that we don't need to zero out all the arrays at every step?

erikfrey (changes requested):

mujoco_warp/_src/passive.py
thowell: restructured the code to avoid zeroing arrays (except `qfrc_gravcomp`) at each step

PR #349, opened 2025-06-09
Add missing dependency, packaging, to pyproject.toml

nimrod-gileadi:
In smooth.py, `packaging.version` is imported, but it's not declared as a package dependency.


btaba:
The dep was removed in 754470a

PR #351, opened 2025-06-10
Vector types are now array.

btaba:


mujoco_warp/_src/passive.py
```diff
@@ -238,9 +241,9 @@ def _box_fluid(
   lvel_torque = rotT @ torque
   lvel_force = rotT @ force
 
-  if opt_wind[0] or opt_wind[1] or opt_wind[2]:
+  if opt_has_wind:
```
erikfrey: unlike the JAX code, we've been able to avoid this `has_*` stuff so far in the warp code.  why do we need it here?

erikfrey (approved):
Looks good, just one question.

mujoco_warp/_src/passive.py
btaba: if wind is an array, we can't use it for static compile-time stuff, lmk wyt

btaba: vector types can be used as static args, but they are just not compatible with JAX

erikfrey: Hmm I must be missing something obvious, you have this:

```py
wind = opt_wind[worldid]
```

so then why can't we do:

```py
if wind[0] or wind[1] or wind[2]:
  ...
```

?

btaba: Changed this line, kept the host-side code similar. Vector types look like they are eval'd at runtime not compile time inside kernels, but they are also available on the host for compilation logic, slightly confusing.

PR #352, opened 2025-06-11
Optimization for geom_local_to_global

Kenny-Vilella:
Currently, the function geom_local_to_global is slowing down the apptronik apollo benchmark because we have a lot of geom composing the ground.
We recalculate at each timestep their position while it is unnecessary.
This change aim to skip unnecessary calculation.

Benchmark on a L40:
`python mujoco_warp/testspeed.py --function=step --mjcf=/home/horde/Documents/optimization/mujoco_warp/benchmark/apptronik_apollo/scene_terrain.xml --batch_size=8192 --nstep=1000`

Trunk:
```
 Total JIT time: 1.40 s
 Total simulation time: 2.39 s
 Total steps per second: 342,123
 Total realtime factor: 1,710.61 x
 Total time per step: 2922.93 ns
```

This branch:
```
 Total JIT time: 1.56 s
 Total simulation time: 20.52 s
 Total steps per second: 399,152
 Total realtime factor: 1,995.76 x
 Total time per step: 2505.31 ns
```

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -158,14 +158,15 @@ def _geom_local_to_global(
   # Data in:
   xpos_in: wp.array2d(dtype=wp.vec3),
   xquat_in: wp.array2d(dtype=wp.quat),
+  geom_skip_in: wp.array(dtype=bool),
   # Data out:
-  geom_sameframe_out: wp.array(dtype=int),
+  geom_skip_out: wp.array(dtype=bool),
   geom_xpos_out: wp.array2d(dtype=wp.vec3),
   geom_xmat_out: wp.array2d(dtype=wp.mat33),
 ):
   worldid, geomid = wp.tid()
   bodyid = geom_bodyid[geomid]
-  if geom_sameframe_out[geomid] == 0:
+  if not geom_skip_in[geomid]:
     # Calculate only if necessary
     xpos = xpos_in[worldid, bodyid]
     xquat = xquat_in[worldid, bodyid]
@@ -174,7 +175,7 @@ def _geom_local_to_global(
 
     if bodyid == 0:
       # static geom pose are calculated only once
-      geom_sameframe_out[geomid] = 1
+      geom_skip_out[geomid] = True
 
 
 @wp.kernel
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/types.py
```diff
@@ -1441,6 +1442,7 @@ class Data:
   ximat: wp.array2d(dtype=wp.mat33)
   xanchor: wp.array2d(dtype=wp.vec3)
   xaxis: wp.array2d(dtype=wp.vec3)
+  geom_sameframe: wp.array(dtype=int)  # warp only
```
erikfrey: why not bool?

mujoco_warp/_src/types.py
```diff
@@ -1315,6 +1315,7 @@ class Data:
     ximat: Cartesian orientation of body inertia                (nworld, nbody, 3, 3)
     xanchor: Cartesian position of joint anchor                 (nworld, njnt, 3)
     xaxis: Cartesian joint axis                                 (nworld, njnt, 3)
+    geom_sameframe: Indicate whether to compute geom pose       (ngeom,)
```
erikfrey: I don't think the semantics are the same as C MuJoCo so I'm not sure we should call it `geom_sameframe` - it's just that the geom body is the worldbody - it might be in a different frame than the world body, right?  And we would still want to apply this optimization.

Also we want the user to understand that we will not write to `geom_xpos` in some scenarios.  This might be surprising to them, e.g. imagine if they store temp data in `geom_xpos` and then expect it to be fixed after running `step`.

So my suggestion would be call this `geom_skip` and then clearly document its behavior, e.g.:

```py
geom_skip: skip calculating `geom_xpos` and `geom_xmat` during step, reuse previous value
```

WDYT?

mujoco_warp/_src/smooth.py
```diff
@@ -159,15 +159,22 @@ def _geom_local_to_global(
   xpos_in: wp.array2d(dtype=wp.vec3),
   xquat_in: wp.array2d(dtype=wp.quat),
   # Data out:
+  geom_sameframe_out: wp.array(dtype=int),
```
erikfrey: We discussed with Miles that in the future we may want to turn on differentiability, and to do so we may need to explicitly track in/out variables.  `geom_sameframe` is an in param and an out param, so we should repeat it, once under `# Data in:` and once under `# Data out:`

erikfrey (approved):
Darn I wish I had thought of this myself, nice job.  Just some nits

mujoco_warp/_src/types.py
Kenny-Vilella: Good question.

I thought maybe we wanted to add a flag to skip either the pos or quat calculation like in mujoco C.
But I don't know the usefulness of this, so wanted to have your opinion.

erikfrey: I can't imagine any reason we'd want to distinguish, I think just a bool is OK.

Kenny-Vilella: Changed it

mujoco_warp/_src/types.py
Kenny-Vilella: Done

mujoco_warp/_src/smooth.py
Kenny-Vilella: Thanks for noticing. I was a bit too careless, sorry for that.

PR #353, opened 2025-06-11
Update .gitignore to ignore generated benchmark data for kitchen

adenzler-nvidia:
Seems like the paths were still pointing at an older version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,12 +7,14 @@ MUJOCO_LOG.TXT
 ## Cloned repositories
 ## -------------------
 
-mujoco_warp/test_data/mujoco_menagerie
+benchmark/mujoco_menagerie/
+
 
 ## Created directories
 ## --------------------
 
-mujoco_warp/test_data/kitchen/combined_assets/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -7,12 +7,14 @@ MUJOCO_LOG.TXT
...
-mujoco_warp/test_data/mujoco_menagerie
+benchmark/mujoco_menagerie/
+
 
 ## Created directories
 ## --------------------
 
-mujoco_warp/test_data/kitchen/combined_assets/
+benchmark/kitchen/combined_assets/
+benchmark/kitchen/kitchen_robot.xml
```
Kenny-Vilella: Not sure about this path, by default the file is saved where the script is launched.
We could also change the script to put in there by default.

adenzler-nvidia: let's wildcard ignore that file for the entire repo - we don't ever want to have it submitted unless someone really intentionally does it.

erikfrey (approved):

PR #355, opened 2025-06-11
Avoid zeroing the constraint jacobian upfront

adenzler-nvidia:
We do end up writing into all of the relevant memory right now afterwards anyway. For the kitchen benchmark, we clear 10 GB of VRAM which takes around 8ms on my RTX 6000 Pro Blackwell.

This change does require us to be a bit more careful about how we're initializing all the constraints, but the win is that we can do things in one go, writing the relevant data and clearing some of the nearby data that is assumed to be zero. Another upside is that we only need to clear nefc number of entries, compared to njmax.

This started showing up in the kitchen benchmark with @Kenny-Vilella 's linesearch optimizations. Still need to combine these 2 - I don't see any regressions on trunk so I think this is worthwhile without a big number right now.

### DIFF
--- a/mujoco_warp/_src/constraint.py
+++ b/mujoco_warp/_src/constraint.py
@@ -245,6 +245,7 @@ def _efc_equality_connect(
 @wp.kernel
 def _efc_equality_joint(
   # Model:
+  nv: int,
   opt_timestep: float,
   qpos0: wp.array2d(dtype=float),
   jnt_qposadr: wp.array(dtype=int),
@@ -291,6 +292,9 @@ def _efc_equality_joint(
     return
 
   efc_worldid_out[efcid] = worldid
+
+  for i in range(nv):
+    efc_J_out[efcid, i] = 0.0
 
   jntid_1 = eq_obj1id[i_eq]
   jntid_2 = eq_obj2id[i_eq]
@@ -460,6 +464,7 @@ def _efc_equality_tendon(
 @wp.kernel
 def _efc_friction_dof(
   # Model:
+  nv: int,
   opt_timestep: float,
   dof_invweight0: wp.array2d(dtype=float),
   dof_frictionloss: wp.array2d(dtype=float),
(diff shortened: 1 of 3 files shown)


Kenny-Vilella:
I think you also need it in equality_connect and equality_weld, no?

adenzler-nvidia:
both equality_connect and equality_connect have a for loop over all dofs - I think these are already covered.

mujoco_warp/_src/constraint.py
```diff
@@ -292,6 +293,9 @@ def _efc_equality_joint(
 
   efc_worldid_out[efcid] = worldid
 
+  for i in range(nv):
```
erikfrey, line 296: Not that I mind passing in `nv` but generally I wonder why we don't do this:

```py
for i in range(efc_J_out.shape[1]):
  efc_J_out[efcid, i] = 0.0
```

do you think patterns like this are good?

erikfrey (approved):
Nice, makes sense.

Do you think there's some `nv` for which these for loops starting hurting more than helping?  Or is it for such an astronomically high `nv` that it's not worth even considering?

mujoco_warp/_src/constraint.py
adenzler-nvidia: I can see an argument for both cases - we are already passing the shape so that would be available.

The ideal thing would be to have nv be a wp.static, but I think that's a bit overkill for this kernel. I personally lean towards nv, as it is a bit more self-documenting. But I think most important is that we are somewhat consistent about it.

adenzler-nvidia:
For sure there is a point where the amount of loop iterations start hurting. I think we are not seeing the complete story yet, we might come back to this at some point. Most likely there is a threshold somewhere, we would have to figure out whether we care about both sides of the threshold.

At some point I would really like to add some autotuning framework/helpers that helps with all those thresholds and let's you tune these parameters to your model.

PR #357, opened 2025-06-11
Optimization: fuse linesearch_init_quad_gauss

adenzler-nvidia:
removes the atomic add, no more kernel to clear the memory. 1 kernel launch less, from 7us + 2us to 2us per solver iteration. Tiny change, but one more of these small things that are adding up.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -15,6 +15,7 @@
 
 import warp as wp
 
+
 def pytest_addoption(parser):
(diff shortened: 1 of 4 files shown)


adenzler-nvidia:
converted to draft - some other stuff got dragged into it. will revise.

erikfrey (approved):
Nice!

PR #358, opened 2025-06-11
Initial EPA implementation

kbayes:


mujoco_warp/_src/collision_gjk.py
```diff
@@ -40,6 +40,61 @@ class GJKResult:
...
+  # TODO(kbayes): look into if a linear map actually improve performance
+  map: wp.array(dtype=int)
+  nmap: int
+
+  # edges that make up the horizon when adding new vertices to polytope
+  nedges: int
+  edges: wp.array(dtype=int)
+
+
+# kernel_analyzer: off
```
erikfrey: please use the kernel analyzer!  if it's doing something you disagree with, let me know

mujoco_warp/_src/collision_gjk.py
```diff
@@ -450,3 +505,582 @@ def gjk(
...
+    pt.map[i] = i
+    pt.face_index[i] = i
+    pt.nmap = 4
+  pt.status = 0
+  pt.nfaces = 4
+  return pt
+
+
+@wp.func
+def _epa(tolerance2: float, epa_iterations: int, pt: Polytope, geom1: Geom, geom2: Geom, geomtype1: int, geomtype2: int):
```
erikfrey, line 948: you indicated gjk is a public wp.func, should we do the same here?  e.g. rename `_epa` to `epa`

that seems reasonable that users might want to graph the func and use it themselves.

if so can you add a docstring?

mujoco_warp/_src/collision_gjk.py
```diff
@@ -40,6 +40,61 @@ class GJKResult:
...
+  pt.face_v[idx] = r
+
+  pt.face_dist2[idx] = wp.dot(r, r)
+  pt.face_index[idx] = -1
+  return pt.face_dist2[idx]
+
+
+@wp.func
+def _epa_support(pt: Polytope, idx: int, geom1: Geom, geom2: Geom, geom1_type: int, geom2_type: int, dir: wp.vec3):
+  _, s1 = gjk_support_geom(geom1, geom1_type, dir)
```
erikfrey, line 87: What will be the relationship between `collision_gjk.py` and `collision_convex.py`?  Should we just pull the support function definition into here?

Don't need to do it in this PR, just curious what your plan is.

mujoco_warp/_src/collision_gjk.py
```diff
@@ -450,3 +505,582 @@ def gjk(
...
+  x2 = wp.vec3()
+  x2[0] = v1[0] * l1 + v2[0] * l2 + v3[0] * l3
+  x2[1] = v1[1] * l1 + v2[1] * l2 + v3[1] * l3
+  x2[2] = v1[2] * l1 + v2[2] * l2 + v3[2] * l3
+
+  return x1, x2
+
+
+@wp.func
+def _polytope2(
```
erikfrey, line 705: Could you add docstrings for the polytope functions?  They're pretty complicated!

mujoco_warp/_src/collision_gjk.py
```diff
@@ -450,3 +505,582 @@ def gjk(
...
+  face_dist2: wp.array(dtype=float),
+  face_index: wp.array(dtype=int),
+  map: wp.array(dtype=int),
+  edges: wp.array(dtype=int),
+):
+  """General convex collision detection via GJK/EPA."""
+  result = gjk(tolerance, gjk_iterations, geom1, geom2, x_1, x_2, geomtype1, geomtype2)
+
+  # recover contact data
+  if result.dist <= tolerance and result.dim > 1:
```
erikfrey: nit: might be a bit nicer to short circuit return here:

```py
if result.dist > tolerance or result.dim <= 1:
  return result
```

mujoco_warp/_src/collision_gjk.py
```diff
@@ -450,3 +505,582 @@ def gjk(
...
+  simplex2 = mat43()
+  simplex2[0] = pt.verts2[v1]
+  simplex2[1] = pt.verts2[v2]
+  simplex2[2] = pt.verts2[v3]
+
+  return simplex, simplex1, simplex2
+
+
+@wp.func
+def _rotmat(axis: wp.vec3):
```
erikfrey, line 607: this could probably go in `math.py` right below `quat_to_mat` - they're pretty similar

erikfrey (approved):
Looks great!  Just some nits.

mujoco_warp/_src/collision_gjk.py
kbayes: This has some hardwired constants that only apply to polytope2 so that's why it's there.

mujoco_warp/_src/collision_gjk.py
kbayes: Done.

mujoco_warp/_src/collision_gjk.py
kbayes: This is just a temporary file. If it's more accurate and performant than the current GJK/EPA we can replace it with this one. Otherwise, I'll have to merge them together somehow.

mujoco_warp/_src/collision_gjk.py
kbayes: GJK should now be private. EPA is way too complicated to initialize as a public API.

mujoco_warp/_src/collision_gjk.py
kbayes: There's no kernels in this code, but some reason it was tripping up on the structs, so I disabled it for now.

mujoco_warp/_src/collision_gjk.py
kbayes: Done.

PR #361, opened 2025-06-12
Optimization parallel linesearch (Grouping constraints per world)

Kenny-Vilella:
This introduces a new set of optimization:
- Constraints are grouped per world
- All parallel linesearch kernels are merged into one

Below are the results for our 4 main benchmarks on a L40.

Branch        |  linesearch | Scene  | Real time factor |
------------ | ------------ | ------- | ------------------ | 
this branch  |  parallel   | aloha pot    |  539.13 x   => about 10% speedup
main            |  parallel   | aloha pot    | 490.57 x
this branch  |  iterative  | aloha pot    |  299.72 x    => slightly faster
main            |  iterative  | aloha pot    |  288.44 x
------------ | ------------ | ------- | ------------------ | 
this branch  |  parallel   | apollo        |  551.37 x    => slightly faster
main            |  parallel   | apollo        |  545.11 x
this branch  |  iterative  | apollo        |  537.27 x    => slightly faster
main            |  iterative  | apollo        |  534.22 x
------------ | ------------ | ------- | ------------------ | 
this branch  |  parallel   | kitchen       |  75.91 x   => about 3 times faster
main            |  parallel   | kitchen       |  25.19 x
this branch  |  iterative  | kitchen       |  25.28 x   => about the same speed
main            |  iterative  | kitchen       |  25.59 x
------------ | ------------ | ------- | ------------------ | 
this branch  |  parallel   | humanoid   |  11,439.53 x   => about 10% slowdown
main            |  parallel   | humanoid   |  12,841.93 x
this branch  |  iterative  | humanoid   |  7,655.97 x     => about 20% slowdown
main            |  iterative  | humanoid   |  9,197.56 x

Very simple scene like the humanoid are slower but scenes that are stressing heavily the solver like the kitchen benchmark are much faster.
For the humanoid, I think that further optimizations could reduce the slowdown.
But maybe better to not put too much stuff in one PR.

Also I didn't spend too much time optimizing the iterative linesearch, but it can be made faster (conditional, fusing pyramidal kernels, etc.)


erikfrey:
Hmm, changing the constraint grouping to group by world is a pretty deep design / architectural change.  This is what our JAX code does today, and it has some major downsides.

The biggest downside is that we have many rows of zeros interspersed in the constraint arrays and we must allocate a worst case constraint array size that is larger than the current heterogenous setup.

I don't think you will see this in the benchmarks as they are designed today because there isn't much diversity of constraints amongst the n worlds.  You will see this during RL training where you are doing exploration, and world 0 has 2 constraints but world 1 has 200 constraints.  In such a setting, you're potentially wasting a lot of memory by adding world as a batch dimension.

You mention two speedups: grouping by world and fusing parallel linesearch kernels.  Can you do the latter without the former?

What specifically do you think is leading to such a big speedup in the kitchen benchmark?  Is there a way to achieve the speedup with the current heterogenous constraint array setup?

Kenny-Vilella:
First of all, the speedup is particularly clear in the kitchen benchmark because it is the only one that stress the linesearch with a lot of constraints (about 3000 per world). But in all the benchmarks we make basically linesearch disappear from the nsight trace. So it's really optimize linesearch across the board.

The main reason for the speedup isthat we get rid of the `linesearch_parallel_quad_total_candidate` kernel.
Not 100% sure why this kernel is so slow because I can't use nsight compute in my current temporary setup, but I guess the atomic add is not great.

I would see if I can get good performance by fusing kernels without grouping constraints per world.

Kenny-Vilella:
I looked at it again, and I do not think that we can optimize the parallel linesearch kernels without grouping constraints per world.

erikfrey:
Got it, thanks Kenny.

Really appreciate the big speedup on the kitchen benchmark!  I think we have another way to attack this, though: after 1.0 release, (or maybe before, depending on the interest) I would like us to implement **sleeping**.  @yuvaltassa already has a plan for this in C MuJoCo, and looking over his notes I think the approach to sleeping will be GPU friendly.

Once we implement sleeping, the number of constraints in kitchen should drop from 3000 to 30, and the kitchen benchmark will be :fire: :fire: :fire: 

Let's discuss this more next week.  My general hesitation on this grouping is we've been discussing improving our support for heterogeneous worlds and I think this moves us away from that, not towards it.

Kenny-Vilella:
@erikfrey Great to hear !

What is the expected timeline for the sleeping feature?
Just want to get a rough idea because I have the feeling that sleeping will improve performance in addition to this change rather than replace it.
But of course we need testing to be sure of that.

For supporting heterogeneous envs, I don't think that it will have a significant impact as long as all the envs are pretty similar.
And if the envs are very different, why batching them in the first place?

Would you be favorable to this change if the perf improvement is significant enough? And if yes what perf improvement would you like to see to be convinced?

Kenny-Vilella:
I think that there is still a bug somewhere.
In the aloha pot benchmark, I can occasionally see some envs explode.
It happens quite often for the parallel linesearch, while it happens rarely for the iterative linesearch.

erikfrey:
@Kenny-Vilella we have observed that parallel linesearch does not yet find the best alpha compared to iterative, so this may be an existing bug.  You can check if you're seeing the same instability on HEAD.

Kenny-Vilella:
@erikfrey  Yes it is something that I have considered. Unfortunately, there are other issues on the main branch that makes testing the benchmark difficult.
I spent last week trying to fix issues on the main branch, will keep doing until I could distinguish more easily error coming from the main branch vs bug in my implementation.

erikfrey:
OK!  This is a very cool improvement- please let us know if we can help with blockers you are facing.

Kenny-Vilella:
Finally found the issue with the tile_sort, so I started to make more tests with this branch vs trunk.

For the aloha pot, I see explosions on both trunk and this branch for any combination of parallel/iterative linesearch and sparse/dense Jacobian.

But the explosion rate is much higher on this branch for all cases.

Also the n_humanoid explodes on this branch for parallel lineserach, while it runs smoothly on trunk.

So there are three different issues:
- There is at least one issue with this branch that makes things more unstable
- The aloha pot is always unstable
- Parallel lineserach is always more unstable than the iterative one

For some numbers with he aloha pot benchmark:
On trunk, ~1000/8192 world explodes for parallel linesearch ||  ~20/8192 world explodes for iterative linesearch
On this branch, ~7000/8192 world explodes for parallel linesearch ||  ~2000/8192 world explodes for iterative linesearch

erikfrey:
@Kenny-Vilella OK, it might be as simple as a model tuning issue.  Let me see if I can repro on my side and I'll make some suggestions.

Kenny-Vilella:
@erikfrey  After fixing the bug with ccd_kernel_builder, I can see that the aloha_pot benchmark is fully stable with the iterative linesearch, but it is still fairly unstable with the parallel linesearch.
70% explosion rate vs less than 0.1% on the main branch.

So yes still has a bug in this branch, but at least now I can start to actually do some bug tracking.

Kenny-Vilella:
Found one bug due to a missing synchronization.
Now we are down to 1% explosion rate, still higher than main but it becomes more reasonable.

The fix will probably impact perf, so I need to do some check, but it should be fairly modest.

Kenny-Vilella:
Following the last fix, we now get the same explosion rate for the aloha pot benchmark on the main branch and this branch (2%).
So it looks like most of the bug got fix.

Performance wise, my optimized branch is twice faster for the humanoid, apollo and kitchen benchmark.
No perf improvement for the aloha_pot benchmark, but I suspect that it is caused by the explosion.
Let's wait to have a stable main branch to reevaluate the aloha_pot benchmark.

Kenny-Vilella:
@erikfrey @adenzler-nvidia  If we want to proceed with this, I think it is finally ready for review.

Below I rerun the benchmark with latest trunk on my RTX 4080 and compared with this [optimized branch](https://github.com/Kenny-Vilella/mujoco_warp/tree/dev/kvilella/small_optimizations_solver_kernels):

Branch        |  linesearch | Scene  | Real time factor |  Speedup |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
optimized branch  |  parallel   | aloha pot    |  540.78 x   |  No measurable difference
main            |  parallel   | aloha pot    |  535.84 x  | |
optimized branch  |  iterative  | aloha pot    |  1,141.15 x   |  No measurable difference
main            |  iterative  | aloha pot    |  1,152.73 x   | |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
optimized branch  |  parallel   | apollo        |  9,217.54 x  |  2.1x
main            |  parallel   | apollo        |  4,359.18 x  | |
optimized branch  |  iterative  | apollo        |  6,333.80 x  |  1.6x
main            |  iterative  | apollo        |  3,935.24 x  | |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
optimized branch  |  parallel   | kitchen       |  15.06 x  |  10.0x 
main            |  parallel   | kitchen       |  1.45 x  | |
optimized branch  |  iterative  | kitchen       |  9.94 x  |   7.0x 
main            |  iterative  | kitchen       |  1.48 x  | |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
optimized branch  |  parallel   | humanoid   |  9,277.90 x  |  2.7x 
main            |  parallel   | humanoid   |  3,348.09 x  | |
optimized branch  |  iterative  | humanoid   |  6,722.26 x  |  2.0x 
main            |  iterative  | humanoid   |  3,399.70 x  | |

Number for the aloha pot benchmark is not changing because 50% of the time is spent on the broadphase (next kernel is at 4%).
So optimizing the solver has currently a very minor effect on the total number.
For the parallel linesearch, a few envs are exploding which may partially explain the bad number compared to the iterative linesearch.

EDIT: Adding commands to generate the result:
For [optimized branch](https://github.com/Kenny-Vilella/mujoco_warp/tree/dev/kvilella/small_optimizations_solver_kernels):
```
python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=52 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/apptronik_apollo/scene_flat.xml --batch_size=8192 --nconmax=150000 --njmax=43 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=kitchen_robot.xml --batch_size=2048  --nconmax=1600000 --njmax=2990 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/aloha_pot/scene.xml --keyframe=0 --batch_size=8192 --nconmax=500000 --njmax=68 --is_sparse=True --ls_parallel=False

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/aloha_pot/scene.xml --keyframe=0 --batch_size=8192 --nconmax=500000 --njmax=936 --is_sparse=True --ls_parallel=True
```

Note that for the aloha pot benchmark values are different for parallel linesearch, this is because we generate some NaNs.

For trunk:
```
python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=410000 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/apptronik_apollo/scene_flat.xml --batch_size=8192 --njmax=380000 --nconmax=150000 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=/home/kvilella/Documents/Codes/mujoco_warp/optim/mujoco_warp/kitchen_robot.xml --batch_size=2048 --nconmax=1600000 --njmax=6123520 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/aloha_pot/scene.xml --keyframe=0 --batch_size=8192 --nconmax=500000 --njmax=250000 --is_sparse=True --ls_parallel=True
```

mujoco_warp/_src/io.py
```diff
@@ -822,7 +822,7 @@ def make_data(mjm: mujoco.MjModel, nworld: int = 1, nconmax: int = -1, njmax: in
     nconmax = nworld * 20
   if njmax == -1:
     # TODO(team): heuristic for njmax
-    njmax = nworld * 20 * 6
+    njmax = 20 * 6
```
adenzler-nvidia, line 825: question to me is whether we need some more hand-holding for users here. The behavior/meaning of njmax is not a bit different between mjwarp and MJX, right?

mujoco_warp/_src/constraint.py
```diff
@@ -1862,14 +1815,6 @@ def make_constraint(m: types.Model, d: types.Data):
...
         )
 
-      if limit_ball or limit_slide_hinge or limit_tendon:
-        wp.launch(
-          _update_nefc,
-          dim=(1,),
-          inputs=[d.nl],
-          outputs=[d.nefc],
-        )
-
```
adenzler-nvidia, line 1872: Do we still need a kernel that aggregates the total constraint count? Otherwise the optimization/fast path of not having to run the solver at all with the conditional breaks.
I'm not sure myself, maybe it's ok now to just run 1 iteration and do the early termination. WDYT?

I can totally see this not being an issue with parallel linesearch, but with iterative it's probably a bit of a cost?

mujoco_warp/_src/io.py
Kenny-Vilella: Do we have njmax in mjx?
I think it is a concept only in mujoco_warp.

There are this github issue #411 to try improving this.
I agree that in general the default is not great.
But now that we have some print help a lot to determine a good value.

mujoco_warp/_src/constraint.py
Kenny-Vilella: At all depends on how often we hit this fast path.
Intuitively, I thought that it was an edge case so it was not worthy to add a kernel launch for that.

But maybe it deserves to be tested instead of relying on intuition.
I will make some tests and report here.

thowell:
@Kenny-Vilella please post the command that generated each testspeed result. thanks!

Kenny-Vilella:
@thowell  Added the commands in the post !

mujoco_warp/_src/constraint.py
Kenny-Vilella: I run our 4 P0 benchmarks with 1 and 8192 envs and never hit the fast path.
Don't think it's worth the extra kernel launch, what do you think?

adenzler-nvidia: Thanks - I think the fast path would only be hit for pure forward dynamics without any additional constraints on the joints. I only know cartpole is like this, but that's not a realistic scenario. We can always reconsider if it starts to be an issue.

Could you remove the capture_if that skips the solver then at the same time? Or is it already gone?

mujoco_warp/_src/io.py
adenzler-nvidia: Thanks - let's not have this block this PR.

mujoco_warp/_src/constraint.py
Kenny-Vilella: Already [gone](https://github.com/google-deepmind/mujoco_warp/pull/361/files#diff-3df65962f4c1247d15f9b9fbe9d41a8934e2eff21ed90ce931b1e8b5e96d10a5L2655) !

mujoco_warp/_src/io.py
```diff
@@ -1338,13 +1363,12 @@ def padtile(x, length, dtype=None):
       mid=wp.empty(shape=(nworld,), dtype=wp.vec3),
       mid_alpha=wp.empty(shape=(nworld,), dtype=float),
       cost_candidate=wp.empty(shape=(nworld, mjm.opt.ls_iterations), dtype=float),
-      quad_total_candidate=wp.empty(shape=(nworld, mjm.opt.ls_iterations), dtype=wp.vec3f),
-      # TODO(team): skip allocation if not elliptic
+      # TODO(team): skip allocation if not ellpitic
```
thowell: 'ellpitic' -> 'elliptic'

mujoco_warp/_src/forward.py
```diff
@@ -219,10 +219,10 @@ def _next_time(
   worldid = wp.tid()
   time_out[worldid] = time_in[worldid] + opt_timestep[worldid]
 
-  if worldid == 0:
-    if nefc_in[0] > njmax_in:
-      wp.printf("nefc overflow - please increase njmax to %u\n", nefc_in[0])
+  if nefc_in[worldid] > njmax_in:
```
thowell: can we create a local variable `nefc = nefc_in[worldid]` so there is only one read to global memory instead of two (one here and one on the next line)?

mujoco_warp/_src/forward.py
```diff
@@ -219,10 +219,10 @@ def _next_time(
...
+    wp.printf("nefc overflow - please increase njmax to %u\n", nefc_in[worldid])
 
+  if worldid == 0:
     if ncollision_in[0] > nconmax_in:
```
thowell: same thing here, can we create a local variable `ncollision = ncollision_in[0]` so there is only one read to global memory?

mujoco_warp/_src/io.py
```diff
@@ -1165,7 +1163,35 @@ def put_data(
...
+  efc_D_fill = np.zeros((nworld, njmax))
+  efc_vel_fill = np.zeros((nworld, njmax))
+  efc_pos_fill = np.zeros((nworld, njmax))
+  efc_aref_fill = np.zeros((nworld, njmax))
+  efc_frictionloss_fill = np.zeros((nworld, njmax))
+  efc_force_fill = np.zeros((nworld, njmax))
+  efc_margin_fill = np.zeros((nworld, njmax))
+
+  nefc = mjd.nefc
+  for i in range(nworld):
```
thowell: instead of a loop, can we do something like `efc_type_fill[:, :nefc] = np.tile(mjd.efc_type, (nworld, 1))` for each field?

mujoco_warp/_src/io.py
```diff
@@ -1288,18 +1314,17 @@ def padtile(x, length, dtype=None):
...
-      J=padtile(efc_J, njmax),
-      pos=padtile(mjd.efc_pos, njmax),
-      margin=padtile(mjd.efc_margin, njmax),
-      D=padtile(mjd.efc_D, njmax),
-      vel=padtile(mjd.efc_vel, njmax),
-      aref=padtile(mjd.efc_aref, njmax),
-      frictionloss=padtile(mjd.efc_frictionloss, njmax),
-      force=padtile(mjd.efc_force, njmax),
-      Jaref=wp.empty(shape=(njmax,), dtype=float),
+      type=wp.array(efc_type_fill, dtype=int, ndim=2),
```
thowell: `wp.array(..., ..., ndim=2)` -> `wp.array2d(...)`
`wp.array(..., ..., ndim=3)` -> `wp.array3d(...)`

mujoco_warp/_src/solver.py
```diff
@@ -2035,44 +1931,38 @@ def update_gradient_JTDAJ(
...
-    if efcid >= min(nefc, njmax_in):
-      return
+  dofi = dof_tri_row[elementid]
+  dofj = dof_tri_col[elementid]
 
-    worldid = efc_worldid_in[efcid]
-    if efc_done_in[worldid]:
-      continue
+  for efcid in range(nefc):
+    # if efcid >= min(nefc, njmax_in):
```
thowell: can we remove the commented out lines?

mujoco_warp/_src/solver.py
```diff
@@ -2035,44 +1931,38 @@ def update_gradient_JTDAJ(
...
+  efc_active_in: wp.array2d(dtype=bool),
   efc_done_in: wp.array(dtype=bool),
   # In:
-  nblocks_perblock: int,
-  dim_x: int,
   # Data out:
   efc_h_out: wp.array3d(dtype=float),
 ):
-  efcid_temp, elementid = wp.tid()
+  # TODO(team): static m?
```
thowell: can we remove this todo?

mujoco_warp/_src/solver.py
```diff
@@ -2305,47 +2195,40 @@ def _update_gradient(m: types.Model, d: types.Data):
     # of SMs on the GPU. We can now query the SM count:
     # https://github.com/NVIDIA/warp/commit/f3814e7e5459e5fd13032cf0fddb3daddd510f30
 
-    # make dim_x and nblocks_perblock static arguments for _JTDAJ to allow loop unrolling
+    # make dim_block and nblocks_perblock static for _JTDAJ to allow loop unrolling
```
thowell: can we move the code that computes `dim_block` and `nblocks_perblock` into the elliptic cone case below?

can we update the comment `_JTDAJ` -> `update_gradient_JTCJ`?

thowell:
@Kenny-Vilella great results!! thank you for this awesome contribution!

after addressing the comments this pr should be ready to merge

mujoco_warp/_src/io.py
Kenny-Vilella: Done

mujoco_warp/_src/io.py
Kenny-Vilella: Done

mujoco_warp/_src/forward.py
Kenny-Vilella: Done

mujoco_warp/_src/forward.py
Kenny-Vilella: Done

mujoco_warp/_src/solver.py
Kenny-Vilella: Not sure what was the purpose of these lines.
But of course always happy to remove commented stuff.

mujoco_warp/_src/solver.py
Kenny-Vilella: I guess at that point nobody remember what it initially means?
Removed it.

mujoco_warp/_src/solver.py
Kenny-Vilella: Done

mujoco_warp/_src/io.py
Kenny-Vilella: It is indeed much smarter.
Changed it.

Kenny-Vilella (commented):
@thowell  Thanks for the thorough reviews and the good suggestions !

I will still do some double check tomorrow to compare results with and without my change to make sure everything is alright.

thowell (approved):

thowell:
great work @Kenny-Vilella!

PR #367, opened 2025-06-13
Add blocked cholesky safeguards

nvtw:
Fixes a couple of issues to make the blocked cholesky solver compatible with the latest warp

### DIFF
--- a/mujoco_warp/_src/block_cholesky.py
+++ b/mujoco_warp/_src/block_cholesky.py
@@ -147,8 +147,8 @@ def create_blocked_cholesky_solve_func(block_size: int):
         num_tile_elements = block_size * block_size
         num_iterations = (num_tile_elements + num_threads_per_block - 1) // num_threads_per_block
 
-        for i in range(num_iterations):
-          linear_index = tid_block + i * num_threads_per_block
+        for ii in range(num_iterations):
+          linear_index = tid_block + ii * num_threads_per_block
           linear_index = linear_index % num_tile_elements
           row = linear_index // block_size
           col = linear_index % block_size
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/block_cholesky.py
```diff
@@ -128,37 +125,94 @@ def blocked_cholesky_solve_func(
...
+        num_tile_elements = block_size
+        num_iterations = (num_tile_elements + num_threads_per_block - 1) // num_threads_per_block
+
+        for ii in range(num_iterations):
+          linear_index = tid_block + ii * num_threads_per_block
+          linear_index = linear_index % num_tile_elements
+          value = rhs_tile[linear_index, 0]
+          if i + linear_index >= active_matrix_size:
+            value = float(0)
+          rhs_tile[linear_index, 0] = value
```
adenzler-nvidia, line 170: could you fuse the two blocks here? Then we only need to loop once.

mujoco_warp/_src/block_cholesky.py
```diff
@@ -128,37 +125,94 @@ def blocked_cholesky_solve_func(
...
```
adenzler-nvidia, line 213: I think these should be functions, this is a lot of non-intuitive code. It's the same as above, right?

mujoco_warp/_src/block_cholesky.py
```diff
@@ -128,37 +125,94 @@ def blocked_cholesky_solve_func(
...
+
+        for ii in range(num_iterations):
+          linear_index = tid_block + ii * num_threads_per_block
+          linear_index = linear_index % num_tile_elements
+          row = linear_index // block_size
+          col = linear_index % block_size
+          value = L_tile[row, col]
+          if i + row >= active_matrix_size or i + col >= active_matrix_size:
+            value = wp.where(row == col, float(1), float(0))
+          L_tile[row, col] = value
```
adenzler-nvidia, line 158: Why is the zero-padding needed exactly, isn't the tile zero-initialized anyway? Especially if it's going beyond the end of the source tile?

mujoco_warp/_src/block_cholesky.py
nvtw: I can try to move it into functions. Warp did not support tiles as function arguments originally, but now it does. Also the non-intuitive code comes partially from the fact that I need to work around __syncthread calls under the hood that freeze the full function if I don't do some strange work-arounds.

mujoco_warp/_src/block_cholesky.py
nvtw: Rhs tile gets modified after load (see rhs_tile -= L_T_x_tile). And I need to be 100% sure that even after the modification there are just zeros in the part outside of the matrix bounds. Otherwise the results are wrong now that we must have non-zero-L-rows outside of the matrix bounds.

mujoco_warp/_src/block_cholesky.py
nvtw: I can try but the loops don't have the same amount of iterations.

adenzler-nvidia: you're right forget about this then.

mujoco_warp/_src/block_cholesky.py
adenzler-nvidia: this may be another stupid idea that does not work, but would it be cleaner to just generate different functions for the remainder parts that are not block-size at the end and call them manually after the block-size parts? We should know the dimensions upfront?

Do you understand what I mean?

PR #375, opened 2025-06-18
Report ncon/ncollision/nefc overflows

adenzler-nvidia:
I think we can pay the price in that kernel at the end. Makes a lot of sense to let users know what's going on.


btaba (approved):
LGTM as a good step in the right direction for reporting overflow!

PR #377, opened 2025-06-18
Work around crashes with graph_conditional == True

adenzler-nvidia:
We are hitting random crashes with these kernels that are dynamically created. I have not been able to figure out why it's not working, maybe it's because it's a throwaway python object.

Either way, this fixes the issue and allows us to actually enable the feature.

### DIFF
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -1102,10 +1102,7 @@ def forward(m: Model, d: Data):
   fwd_acceleration(m, d)
   sensor.sensor_acc(m, d)
 
-  if d.njmax == 0:
-    wp.copy(d.qacc, d.qacc_smooth)
-  else:
-    solver.solve(m, d)
+  solver.solve(m, d)
 
 
 @event_scope
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -2637,12 +2637,21 @@ def create_context(m: types.Model, d: types.Data, grad: bool = True):
     _update_gradient(m, d)
 
 
+def _copy_acc(m: types.Model, d: types.Data):
+  wp.copy(d.qacc, d.qacc_smooth)
+  wp.copy(d.qacc_warmstart, d.qacc_smooth)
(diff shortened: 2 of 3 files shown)


btaba:
I checked just OOC, and jax still fails with graph conditional unfortunately

mujoco_warp/_src/derivative.py
```diff
@@ -190,14 +191,18 @@ def deriv_smooth_vel(m: Model, d: Data, flg_forward: bool = True):
     # TODO(team): sparse version
 
     for tile_nu, tile_nv in zip(m.actuator_moment_tiles_nu, m.actuator_moment_tiles_nv):
-      wp.launch_tiled(
-        _tile_qderiv_actuator_passive(
+
+      if DERIV_SMOOTH_VEL_KERNELS.get((tile_nu.size, tile_nv.size)) is None:
+        DERIV_SMOOTH_VEL_KERNELS[(tile_nu.size, tile_nv.size)] = _tile_qderiv_actuator_passive(
```
btaba: would this be overkill?

```
# in warp_util.py
_KERNEL_CACHE = {}

def cache_kernel(func):
  # caching because of some graph_conditional thing?
  @functools.wraps(func)
  def wrapper(*args):
    key = tuple(a.size if hasattr(a, 'size') else hash(a) for a in args)
    if key not in _KERNEL_CACHE:
       _KERNEL_CACHE[key] = func(*args)
    return _KERNEL_CACHE[key]
  return wrapper
```

then

```
@warp_util.cache_kernel
def _tile_qderic_actuator_passive(...):
   ...
```


OTOH, could there be potential bugs by no caching by `actuation_enabled` etc. ?

adenzler-nvidia:
yeah unfortunately this does not fix the crashes with JAX in the loop. Will check what the status is there..

mujoco_warp/_src/derivative.py
adenzler-nvidia: we can generalize it, yes. Makes sense.

Let's just try to cache everything, I think other things would break if we allow users to change some if these options but we can stay on the safe side for this change for sure.

PR #379, opened 2025-06-18
update collision_gjk.py

thowell:


kbayes:
Thanks for being an extra pair of eyes on this! I’ll be online on Monday to review.

mujoco_warp/_src/collision_gjk.py
```diff
@@ -855,22 +855,22 @@ def _polytope3(
     return pt
 
   # create hexahedron for EPA
-  if _attach_face(pt, 0, 4, 0, 1) < MJ_MINVAL2:
```
kbayes: Revert these changes and the lines below.

kbayes (approved):
Thank you!

mujoco_warp/_src/collision_gjk.py

PR #383, opened 2025-06-19
convex collision pair

thowell:
only create and launch kernels for convex collision pairs that are included in `Model`

```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=450000 --njmax=450000
```
this pr:
```

 Total JIT time: 0.31 s
 Total simulation time: 30.10 s
 Total steps per second: 272,121
 Total realtime factor: 1,360.61 x
 Total time per step: 3674.83 ns
```

main:
```

 Total JIT time: 0.98 s
 Total simulation time: 31.85 s
 Total steps per second: 257,208
 Total realtime factor: 1,286.04 x
 Total time per step: 3887.90 ns
```

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -965,10 +965,10 @@ _collision_kernels = {}
 
 def gjk_narrowphase(m: Model, d: Data):
   if len(_collision_kernels) == 0:
-    for i in range(m.convex_collision_pair.size):
+    for i in range(m.convex_collision_pair.shape[0]):
       convex_collision_pair = m.convex_collision_pair[i]
-      t1 = convex_collision_pair[0]
-      t2 = convex_collision_pair[1]
+      t1 = int(convex_collision_pair[0])
+      t2 = int(convex_collision_pair[1])
       _collision_kernels[(t1, t2)] = _gjk_epa_pipeline(
         t1,
         t2,
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -965,9 +965,10 @@ _collision_kernels = {}
 
 def gjk_narrowphase(m: Model, d: Data):
   if len(_collision_kernels) == 0:
-    for types in _CONVEX_COLLISION_FUNC:
-      t1 = types[0]
-      t2 = types[1]
+    for i in range(m.convex_collision_pair.size):
+      convex_collision_pair = m.convex_collision_pair[i]
+      t1 = convex_collision_pair[0]
+      t2 = convex_collision_pair[1]
       _collision_kernels[(t1, t2)] = _gjk_epa_pipeline(
         t1,
         t2,
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -21,6 +21,7 @@ import warp as wp
 
 from . import math
 from . import types
+from .collision_convex import _CONVEX_COLLISION_FUNC
 
 
 def _hfield_geom_pair(mjm: mujoco.MjModel) -> Tuple[int, np.array]:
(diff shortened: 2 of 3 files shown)


mujoco_warp/_src/collision_convex.py
```diff
@@ -965,9 +965,10 @@ def gjk_epa_sparse(
...
 def gjk_narrowphase(m: Model, d: Data):
   if len(_collision_kernels) == 0:
-    for types in _CONVEX_COLLISION_FUNC:
-      t1 = types[0]
-      t2 = types[1]
+    for i in range(m.convex_collision_pair.shape[0]):
+      convex_collision_pair = m.convex_collision_pair[i]
+      t1 = int(convex_collision_pair[0])
+      t2 = int(convex_collision_pair[1])
       _collision_kernels[(t1, t2)] = _gjk_epa_pipeline(
```
erikfrey: With this approach do we still need a module level cache?  Why not just do:

```py
for t1, t2 in m.convex_collision_pair:
  kernel = _gjk_epa_pipeline(...
  wp.launch(kernel, ...
```

mujoco_warp/_src/types.py
```diff
@@ -1260,6 +1263,7 @@ class Model:
   mat_rgba: wp.array2d(dtype=wp.vec4)
   geompair2hfgeompair: wp.array(dtype=int)  # warp only
   block_dim: BlockDim  # warp only
+  convex_collision_pair: np.array  # warp only
```
erikfrey: do we need numpy for this?  can we keep it a tuple?

mujoco_warp/_src/io.py
```diff
@@ -296,6 +297,9 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
 
     nxn_pairid[pairid] = i
 
+  convex_collision_pair = _CONVEX_COLLISION_FUNC & set(zip(mjm.geom_type[geom1], mjm.geom_type[geom2]))
```
erikfrey: is there a way to just store a tuple of the distinct geom types here so we don't have to import `collision_convex`?  that way perhaps we can reuse this same mechanism for collision_sdf and collision_primitive?

erikfrey (changes requested):
I like it, but can we generalize this so it also applies to the primitive and sdf collision functions?

mujoco_warp/_src/collision_convex.py

mujoco_warp/_src/types.py

mujoco_warp/_src/io.py
thowell: the import from `collision_convex` is removed by moving the geom type pairs information to io.py

thowell:
@erikfrey should we follow up in separate PRs for primitives and sdfs that create fields in `Model` similar to `convex_collision_pair`, like `primitive_collision_pair` and `sdf_collision_pair`, and then loop over these tuples in `primitive_narrowphase` and `sdf_narrowphase`?

thowell:
@erikfrey the PR is updated to utilize the `Model` field `collision_geom_pair`, a `tuple` containing all geom type pairs (including primitive and sdf)

mujoco_warp/_src/types.py
```diff
@@ -1260,6 +1261,7 @@ class Model:
   mat_rgba: wp.array2d(dtype=wp.vec4)
   geompair2hfgeompair: wp.array(dtype=int)  # warp only
   block_dim: BlockDim  # warp only
+  collision_geom_pair: tuple  # warp only
```
erikfrey: I think you can do `tuple[tuple[int, int], ...]` or some such?

Also maybe we should make it clearer these are geom types?  Maybe name it `geom_type_pair`?

erikfrey (approved):
OK great, just one nit.

Also can you redo the benchmark one last time to make sure our simplification (no longer caching the functions explicitly at the module level) didn't hurt anything?

mujoco_warp/_src/types.py

thowell:
The PR description is updated with current testspeed results and the current testspeed results are similar to the previous testspeed results.

PR #384, opened 2025-06-19
update convex collision

thowell:


erikfrey:
Is this still relevant?

PR #395, opened 2025-06-24
Make import of register_sdf_plugin conditional

quagla:


mujoco_warp/testspeed.py
```diff
@@ -99,7 +97,13 @@ def _main(argv: Sequence[str]):
   if path.suffix == ".mjb":
     mjm = mujoco.MjModel.from_binary_path(path.as_posix())
   else:
-    mjm = mujoco.MjModel.from_xml_path(path.as_posix())
+    spec = mujoco.MjSpec.from_file(path.as_posix())
```
erikfrey: you convinced me that `_load_model` is a nice function     :-)

do you want to do it here too?

erikfrey (approved):
Just one suggestion, otherwise LGTM

mujoco_warp/testspeed.py
quagla: Done.

erikfrey (approved):

PR #409, opened 2025-06-26
Adjust constant in GJK

kbayes:
1e-15 was too small and now 1e-10 is too large. Looks like sweet spot is near 1e-12.


mujoco_warp/_src/collision_gjk.py
```diff
@@ -533,7 +533,7 @@ def _gjk(
   for k in range(gjk_iterations):
     xnorm = wp.dot(x_k, x_k)
     # TODO(kbayes): determine new constant here
```
erikfrey, line 535: should this TODO be deleted?

kbayes: Not yet. I'd like to have some unit tests in place before removing the TODO.

PR #412, opened 2025-06-27
Default graph_conditional to True.

erikfrey:


btaba:
is the idea to copybara replace until NV folks can fix this?

erikfrey:
@btaba Yes let's overwrite `graph_conditional=False` in MJX for now.

btaba (changes requested):
I can't find an existing test that runs all of forward or step in a systematic way (which partly explains the trouble I'm having adding such tests in MJX).

An e2e test should be added and parametrized on graph_conditional=[True, False] to make sure we don't unintentionally break the MJX integration while NV folks work on JAX compat.

And AFAIU this [test](https://github.com/google-deepmind/mujoco_warp/blob/6a17c38abb45c5f9492c929a8be2991809980ef3/mujoco_warp/_src/jax_test.py#L38) is excluded from github CI ?

adenzler-nvidia:
test_graph_capture does a step for humanoid/pendula/collision.xml/constraints.xml. But we should parametrize that one for conditional true/false indeed, and maybe run more than 1 step.

erikfrey:
OK @adenzler-nvidia @btaba updated the test to check `graph_conditional` on and off, and runs multiple steps.

@adenzler-nvidia if you want to absorb this into your PR go for it.

adenzler-nvidia:
combined this with #485 - closing this one.

PR #416, opened 2025-06-27
Improve geom_type_pair and fix missing collision type pairs bug.

kbayes:


mujoco_warp/_src/types.py
```diff
@@ -1274,7 +1274,7 @@ class Model:
   actuator_trntype_body_adr: wp.array(dtype=int)  # warp only
   geompair2hfgeompair: wp.array(dtype=int)  # warp only
   block_dim: BlockDim  # warp only
-  geom_type_pair: tuple[int, ...]  # warp only
+  geom_type_pair: wp.array2d(dtype=int)
```
btaba: the annotation says this is a device array but in io.py you're not making it a device array. Then in collision_convex you're using the value on the host. Can you keep the current annotation?

kbayes: My bad. Forgot to add it back in. Thanks.

btaba: I mean `tuple[int, ...]`

kbayes: Ah, it should be a np array. Thanks again.

btaba: It should not be a np.ndarray, please convert to tuple[int, ...]

Hopefully [this](https://github.com/google-deepmind/mujoco_warp/blob/13bcf2e17815fde7f67fa8380bdfbd9af7dd8f77/mujoco_warp/_src/io_test.py#L158) local test was failing, I'll have to fix it if not

mujoco_warp/_src/io.py
```diff
@@ -336,8 +336,18 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
...
+  for g1 in geom1:
+    geom1_types[int(mjm.geom_type[g1])] = True
+  geom2_types = np.zeros(len(types.GeomType), dtype=bool)
+  for g2 in geom2:
+    geom2_types[int(mjm.geom_type[g2])] = True
+
+  for type1 in range(len(types.GeomType)):
+    for type2 in range(type1, len(types.GeomType)):
+      if (geom1_types[type1] and geom2_types[type2]) or (geom1_types[type2] and geom2_types[type1]):
+        geom_type_pair.append((type1, type2))
```
btaba: unfortunately the data here doesn't match the annotation (you are doing `tuple[tuple[int, int], ...]`). I'll add a separate test to catch things like this. Can you really not use `tuple[int, ...]` ?

kbayes: Done.

mujoco_warp/_src/io.py
```diff
@@ -336,8 +336,19 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
...
+    geom1_types[int(mjm.geom_type[g1])] = True
+  geom2_types = np.zeros(len(types.GeomType), dtype=bool)
+  for g2 in geom2:
+    geom2_types[int(mjm.geom_type[g2])] = True
+
+  for type1 in range(len(types.GeomType)):
+    for type2 in range(type1, len(types.GeomType)):
+      if (geom1_types[type1] and geom2_types[type2]) or (geom1_types[type2] and geom2_types[type1]):
+        geom_type_pair.append(type1)
+        geom_type_pair.append(type2)
```
btaba, line 352: This can be shortened a bit, but up to you. Like this does the same thing (I think)

```
  type_pairs = np.stack([mjm.geom_type[geom1], mjm.geom_type[geom2]], axis=1)
  type_pairs.sort()
  geom_type_pair = set(tuple(tp) for tp in type_pairs)
  geom_type_pair = tuple(int(t) for tp in geom_type_pair for t in tp)
```

PR #421, opened 2025-06-27
add __all__ to __init__.py

thowell:
this addition to `__init__.py` is for documentation


mujoco_warp/__init__.py
```diff
@@ -72,3 +72,63 @@
 from ._src.types import Option as Option
 from ._src.types import Statistic as Statistic
 from ._src.types import TrnType as TrnType
+
+__all__ = [
```
erikfrey, line 76: These look the same as above - I guarantee that in the future, we'll add an import above and forget to add it to `__all__` - out of curiosity, why does the documentation need this?

erikfrey (commented):
General question to start.

PR #426, opened 2025-06-30
Add kernel caching to GJK narrowphase

adenzler-nvidia:
This avoids crashes in CUDA conditional code and helps a bit with compile times.


erikfrey (commented):
Thanks @adenzler-nvidia - did you see this during graph capture?  We should add some tests that explicitly test graph capture.

erikfrey (approved):

adenzler-nvidia:
yes, it was crashing on me, but I had a local change to enable conditional nodes. I think once we default that to true we should be set. Maybe in the mean time we could just have another version of the graph capture test with conditionals on?

PR #427, opened 2025-06-30
Avoid running SDF collision detection without SDFs in model

adenzler-nvidia:
Adds some static computation to avoid another launch with nconmax size.


kbayes (approved):
Nice!

PR #459, opened 2025-07-04
Combine qM factorization and solve if entire forward function is called

adenzler-nvidia:
Makes the solve almost free for dense codepath because we can keep the matrix in registers.

### DIFF
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -525,7 +525,7 @@ def implicit(m: Model, d: Data):
 
 
 @event_scope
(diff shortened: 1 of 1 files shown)


thowell:
does it make sense to change the logic from `deferred_factorization` to `factorize` or alternatively rename `deferred_factorization` -> `factorize_skip`/ `skip_factorization`?

adenzler-nvidia:
changing to `factorize` which makes the individual APIs a bit more self-documenting.

thowell (approved):

PR #464, opened 2025-07-04
Don't run collision detection if there are no potential pairs.

adenzler-nvidia:
This is as simple as it gets.

We could also do that by prefiltering the nxn_geom_pairs and then checking the shape - I went down that avenue but it wasn't trivial to combine it with the way filtering is done in the SAP broadphase that also relies on the pairid array. So leaving this for the future. @thowell I think you wanted to do this as well at some point, right?

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -356,8 +356,7 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   )
 
   # Disable collisions if there are no potentially colliding pairs
-  valid_mask = nxn_pairid > -2
-  if not valid_mask.any():
+  if np.sum(geom_type_pair_count) == 0:
     mjm.opt.disableflags |= types.DisableBit.CONTACT.value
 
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/io.py
```diff
@@ -355,6 +355,11 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     minlength=len(types.GeomType) * (len(types.GeomType) + 1) // 2,
   )
 
+  # Disable collisions if there are no potentially colliding pairs
+  valid_mask = nxn_pairid > -2
+  if not valid_mask.any():
```
kbayes: Use geom_type_pair_count from above. This counts total number of potentially colliding pairs for each geom type pair instead of using the large nxn_pairid array. i.e.

if np.sum(geom_type_pair_count) == 0:
      mjm.opt.disableflags |= types.DisableBit.CONTACT.value

adenzler-nvidia: makes sense, thanks for the hint. Updated the code.

kbayes (approved):

PR #470, opened 2025-07-07
remove legacy gjk epa

thowell:


kbayes:
I think you should hold off until when there's parity with multicontact.

Issue #326, opened 2025-05-31
Pre-filter `nxn_geom_pair` and `nxn_pair_id`


erikfrey:
For a scene like `benchmark/apptronik_apollo/scene_terrain.xml`, most geom pairs could never collide because they belong to the world body.

We already ignore invalid pairs by setting `nxn_pairid` = -2, but we should go a step further and shrink `nxn_pairid` and `nxn_geom_pair` - this may require adding a third id array.

For the apptronik scene above, before filtering we have 13,989,405 candidate pairs.  After filtering, we have 95,021.

On a 4090, collision detection using SAP and the unfiltered candidate pairs takes 1500ns, and using nxn and the filtered candidate pairs takes 500ns.

It's possible that such pre-filtering will harm performance for very small simple scenes, so it may be better as an optional behavior that is statically switched on using wp.static

PR #471, opened 2025-07-07
nxn_broadphase filtered geom pairs

thowell:
`nxn_broadphase` is updated to only launch threads and consider geom pairs that are not filtered (ie, `nxn_pairid > -2`). this update can have a significant impact on performance for environments with many filtered geom pairs like benchmark/apptronik_apollo/scene_terrain.xml

`nxn_broadphase` (filtered)
```
mjwarp-testspeed --function=nxn_broadphase --mjcf=benchmark/apptronik_apollo/scene_terrain.xml --batch_size=8192 --event_trace=True --nconmax=1000000 --njmax=1000000
```
```

 Total JIT time: 0.00 s
 Total simulation time: 18.94 s
 Total steps per second: 432,462
 Total realtime factor: 2,162.31 x
 Total time per step: 2312.34 ns

Event trace:

nxn_broadphase: 2309.58
```

`sap_broadphase`

```
mjwarp-testspeed --function=sap_broadphase --mjcf=benchmark/apptronik_apollo/scene_terrain.xml --batch_size=8192 --event_trace=True --nconmax=1000000 --njmax=1000000
```
```

 Total JIT time: 0.01 s
 Total simulation time: 46.91 s
 Total steps per second: 174,629
 Total realtime factor: 873.15 x
 Total time per step: 5726.42 ns

Event trace:

sap_broadphase: 5721.0
```

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -381,6 +381,7 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   rangefinder_sensor_adr = np.full(mjm.nsensor, -1)
   rangefinder_sensor_adr[sensor_rangefinder_adr] = np.arange(len(sensor_rangefinder_adr))
 
+  # TODO(team): improve heuristic for selecting broadphase routine
   if mjm.ngeom > 1000:
     broadphase = types.BroadphaseType.SAP_SEGMENTED
   elif mjm.ngeom > 100:


erikfrey:
@thowell Curious why you're comparing to `sap_broadphase` as a baseline?  How does it compare to unfiltered `nxn_broadphase` as a baseline?  Also do you think `sap_broadphase` can also benefit from the filtered candidates?  I see in `_sap_broadphase` kernel where it checks `nxn_paird[idx] < -1`.

thowell:
@erikfrey

- comparison between nxn (filtered) and sap in order to follow comment in #326
- nxn (main) was taking a while to run...

not sure how much sap can benefit from filtered geom pairs unless sap filters out geoms that are filtered with all other geoms.

erikfrey:
@thowell got it, but then does that suggest we need a more sophisticated criteria here in order to trigger `nxn`:

https://github.com/google-deepmind/mujoco_warp/blob/c89ba300f21ec26a9af416eca253beb722906038/mujoco_warp/_src/io.py#L380-L385

thowell:
@erikfrey we should probably perform an empirical study with the benchmark scenes, each evaluated at a number of keyframes, to determine good default values for automatically switching between broadphase routines

thowell:
@erikfrey added a todo for improving the heuristic that selects the broadphase routine

kbayes (approved):

PR #474, opened 2025-07-08
Add control noise to benchmark

Kenny-Vilella:
Following the example in Mujoco C, we now introduce control noise in the benchmark to have our benchmarks being closer to an RL loop.


mujoco_warp/_src/test_util.py
```diff
@@ -157,6 +157,50 @@ def _sum(stack1, stack2):
...
+  worldid, nuid = wp.tid()
+
+  center = 0.0
+  radius = 1.0
+  ctrlrange = actuator_ctrlrange[0, nuid]
+  if actuator_ctrllimited[nuid]:
+    center = (ctrlrange[1] + ctrlrange[0]) / 2.0
+    radius = (ctrlrange[1] - ctrlrange[0]) / 2.0
+  radius *= ctrlnoise
+  noise = 2.0 * halton(step * worldid, nuid + 2) - 1.0
```
Kenny-Vilella: Here I added step * worldid to get some difference between world.
I am not familiar with the Halton sequence, so maybe good to double check that it is the proper way to do this.

Kenny-Vilella:
It doesn't seem to work as expected.
The control seems to be set properly, but it does not impact the simulation.
Looking into it.

EDIT: The control noise works as expected.
The issue is that we always hit the actuator force range limit so at the end the noise has no impact on the results.
Still looking at what modification would be best.

Kenny-Vilella:
The issue seems to be because I am running benchmark with the is_sparse option, and this option is currently broken.
I will try tomorrow wiith the dense version and check that this works as expected.

Kenny-Vilella:
It is indeed working with the dense version.
So this feature is working as expected.

thowell:
there is an implementation `halton` in collision_sdf.py https://github.com/google-deepmind/mujoco_warp/blob/4dd4de8e7d747e0cdf2e3cde55c74df451889f47/mujoco_warp/_src/collision_sdf.py#L272

can we move `halton` to util_misc.py and have the function be imported by collision_sdf and test_util?

mujoco_warp/_src/test_util.py
```diff
@@ -157,6 +157,50 @@ def _sum(stack1, stack2):
...
+  # Model:
+  actuator_ctrllimited: wp.array(dtype=bool),
+  actuator_ctrlrange: wp.array2d(dtype=wp.vec2),
+  # In:
+  step: int,
+  ctrlnoise: float,
+  # Data out:
+  ctrl_out: wp.array2d(dtype=float),
+):
+  worldid, nuid = wp.tid()
```
thowell: `nuid` -> `actid`?

mujoco_warp/_src/test_util.py
```diff
@@ -157,6 +157,50 @@ def _sum(stack1, stack2):
...
+    r = n0 - n1 * base
+    hn += f * float(r)
+    f /= b
+    n0 = n1
+
+  return hn
+
+
+@wp.kernel
+def get_ctrl_noise(
```
thowell: `get_ctrl_noise` -> `ctrl_noise` ?

mujoco_warp/_src/test_util.py
thowell: maybe `(step + 1) * (worldid + 1)`? 
- `(worldid + 1)` to generate different first inputs to `halton` at each step for `worldid=0`
- `(step + 1)` to generate different first inputs to `halton` for `step=0`

mujoco_warp/_src/test_util.py
```diff
@@ -157,6 +157,50 @@ def _sum(stack1, stack2):
   return ret
 
 
+@wp.func
+def halton(
+  # In:
```
erikfrey: nit: we added a rule to kernel analyzer that, if you keep the function signature as one line, you don't need to follow the comment rules.  So I think this could also be:

```py
@wp.func
def halton(index: int, base: int) -> float:
  ...
```

mujoco_warp/_src/test_util.py
```diff
@@ -204,9 +248,21 @@ def benchmark(
...
+        get_ctrl_noise,
+        dim=(d.nworld, m.nu),
+        inputs=[
+          m.actuator_ctrllimited, m.actuator_ctrlrange, i, 0.01
+        ],
+        outputs=[d.ctrl])  # fmt: skip
+
+      run_beg = time.perf_counter()
       wp.capture_launch(graph)
+      wp.synchronize()
```
erikfrey, line 244: I don't know enough about warp internals, are we sure we're still measuring the same thing:

at HEAD:

```py
for i in range(nstep):
  ...
  wp.capture_launch(graph)
  ...
wp.synchronize()
```

your change:

```py
for i in range(nstep):
  ...
  wp.capture_launch(graph)
  wp.synchronize()
  ...
```

the former can enqueue multiple graph launches without a synchronize in between them.

Maybe as a sanity check can you confirm that benchmark for one or two scenes don't change significantly as a result of this PR?

mujoco_warp/_src/test_util.py
Kenny-Vilella: Good idea.
Changed it.

mujoco_warp/_src/test_util.py
Kenny-Vilella: I moved the function to `util_misc` following Taylor's suggestion and changed the formatting.

mujoco_warp/_src/test_util.py
Kenny-Vilella: Done

mujoco_warp/_src/test_util.py
Kenny-Vilella: Done

mujoco_warp/_src/test_util.py
Kenny-Vilella: Yeah sorry I should have highlighted this change more clearly.

Below are the numbers:

benchmark                    |  this branch  |  main              |
-------------------------- | --------------- | --------------- |
humanoid                      |  5,078.99 x   |  5,003.54 x   |  
apollo on a flat terrain |  398.77 x      |   391.90 x      |  
aloha pot                       |  72.66 x      |   70.49 x       |  

For the sake of completeness, these are the commands for the benchmark:

`python mujoco_warp/testspeed.py --function=step --is_sparse=False --mjcf=benchmark/humanoid/humanoid.xml --ls_parallel=True --batch_size=8192 --nconmax=450000 --njmax=450000`

`python mujoco_warp/testspeed.py --function=step --is_sparse=False --mjcf=/home/horde/Documents/optimization/test/mujoco_warp/benchmark/apptronik_apollo/scene_flat.xml --ls_parallel=True --batch_size=8192 --nconmax=150000 --njmax=417792`

`python mujoco_warp/testspeed.py --function=step --is_sparse=False --mjcf=/home/horde/Documents/linesearch_optimization/trunk/mujoco_warp/benchmark/aloha_pot/scene.xml --ls_parallel=True --batch_size=8192 --nconmax=6700000 --njmax=7036928 --keyframe=0`

erikfrey: OK!  Thanks Kenny.

PR #484, opened 2025-07-09
Add flag to disable collision detection

nvtw:
Add flag to optionally disable collision detection. This allows to inject contacts from other sources, e. g. newton.


adenzler-nvidia:
is it not possible to just edit the disablebit for collisions? DisableBit.CONTACT?

nvtw:
DisableBit.CONTACT still zeros ncon_out and therefore discards the contacts injected from Newton, see function below. I could move the flag check up but it might have side effects. Inputs from mujoco experts are welcome.
```@event_scope
def collision(m: Model, d: Data):
  """Collision detection."""

  # zero collision-related arrays
  wp.launch(
    _zero_collision_arrays,
    dim=d.nconmax,
    inputs=[
      d.nworld,
      d.ncon_hfield.shape[1],
      d.ncon,
      d.ncon_hfield.reshape(-1),
      d.collision_hftri_index,
      d.ncollision,
    ],
  )

  if d.nconmax == 0:
    return

  dsbl_flgs = m.opt.disableflags
  if (dsbl_flgs & DisableBit.CONSTRAINT) | (dsbl_flgs & DisableBit.CONTACT):
    return

  if m.opt.broadphase == int(BroadphaseType.NXN):
    nxn_broadphase(m, d)
  else:
    sap_broadphase(m, d)

  narrowphase(m, d)
```

adenzler-nvidia:
I see your point. What does your pipeline look like, you will generate contacts before the call to mjwarp.step()? How are you getting the correct geom positions in this case? FK does run as part of the step as well.

would it make sense to give you a callback or something along these lines?

nvtw:
A callback does not really help since the contacts get already passed into newton's step function. I simply convert them to mujoco. So far it works quite good but I need to test more examples. The following code shows the step function in solver_mujoco.py from newton. The signature of step is fixed, so I don't know if a callback to run contact detection again (=inefficient) makes sense.
```    @override
    def step(self, state_in: State, state_out: State, control: Control, contacts: Contacts, dt: float):
        if self.use_mujoco:
            self.apply_mjc_control(self.model, state_in, control, self.mj_data)
            if self.update_data_interval > 0 and self._step % self.update_data_interval == 0:
                # XXX updating the mujoco state at every step may introduce numerical instability
                self.update_mjc_data(self.mj_data, self.model, state_in)
            self.mj_model.opt.timestep = dt
            self.mujoco.mj_step(self.mj_model, self.mj_data)
            self.update_newton_state(self.model, state_out, self.mj_data)
        else:
            self.apply_mjc_control(self.model, state_in, control, self.mjw_data)
            if self.update_data_interval > 0 and self._step % self.update_data_interval == 0:
                self.update_mjc_data(self.mjw_data, self.model, state_in)
            self.mjw_model.opt.timestep.fill_(dt)
            with wp.ScopedDevice(self.model.device):
                use_mujoco_contacts = False
                if not use_mujoco_contacts:
                    self.convert_contacts_to_mjwarp(self.model, state_in, contacts)
                    self.mujoco_warp.step(self.mjw_model, self.mjw_data, False)
                else:
                    self.mujoco_warp.step(self.mjw_model, self.mjw_data, True)
                print("step done")
            self.update_newton_state(self.model, state_out, self.mjw_data)
        self._step += 1
        return state_out
```

adenzler-nvidia:
Does not make any sense to run contact gen twice indeed. So you are relying on the newton FK to get shape positions? I still don't really get how that would give you updated geom positions. Plus there is a lot of duplication that could be removed if you are relying on an external forward kinematics step anyway.

Back to the main topic, I guess in this setup your proposal makes sense. Is there a world where we would generate contacts for part of the geom in MjWarp, and other geoms have externally generated contacts? Maybe it would make sense to design the API around this.

erikfrey:
The MuJoCo Way™ for public API functions to take only Model and Data, and control flow is dictated by Model properties.  So I think you have two options here:

1) Don't call step. Instead call the sub-functions of step (skipping collision detection), all of which are in the public API.
2) Propose a property on Model that gives you the behavior you want.

I think either of those we could work with.  Long term, if you think you'll be customizing how you use mjwarp a lot, I think 1) is the better bet.

nvtw:
@erikfrey I added a new flag to the options in the model. I tried option 1 but it involves too much code I would need to copy over to newton.

mujoco_warp/_src/forward.py
```diff
@@ -536,7 +536,8 @@ def fwd_position(m: Model, d: Data, factorize: bool = True):
   smooth.tendon_armature(m, d)
   if factorize:
     smooth.factor_m(m, d)
-  collision_driver.collision(m, d)
+  if run_collision_detection:
```
erikfrey: Here you can do:

```py
if m.opt.run_collision_detection:
  ...
```

Then you can remove the `run_collision_detection` param that you added.

Similarly we will remove `factorize` param in a later cleanup.

mujoco_warp/_src/forward.py
```diff
@@ -1091,11 +1092,14 @@ def _zero_energy(
 
 
 @event_scope
-def forward(m: Model, d: Data):
+def forward(
```
erikfrey: nit: can revert this change, function signature just on one line

mujoco_warp/_src/types.py
```diff
@@ -535,6 +535,8 @@ class Option:
     graph_conditional: flag to use cuda graph conditional, should be False when JAX is used
     sdf_initpoints: number of starting points for gradient descent
     sdf_iterations: max number of iterations for gradient descent
+    run_collision_detection: flag to run collision detection (if the collision detection is
```
erikfrey: nit: maybe make it clearer to the reader why this exists, something like:

```
    run_collision_detection: if False, skips collision detection and allows user-populated
                             contacts during the physics step  (as opposed to `DisableBit.CONTACT`
                             which explicitly zeros out the contacts at each step)
```

Something like that.

erikfrey (approved):
LGTM, just a few nits

mujoco_warp/_src/forward.py
nvtw: Done

mujoco_warp/_src/types.py
nvtw: Done

mujoco_warp/_src/forward.py
nvtw: Done

PR #485, opened 2025-07-09
Gracefully handle missing conditional node support for older CUDA drivers

adenzler-nvidia:
Adds a helper function that checks the versions, and disables conditional nodes for older drivers.

### DIFF
--- a/mujoco_warp/_src/collision_driver.py
+++ b/mujoco_warp/_src/collision_driver.py
@@ -28,7 +28,7 @@ from .types import Data
 from .types import DisableBit
 from .types import GeomType
 from .types import Model
-from .warp_util import event_scope
+from .warp_util import conditional_graph_enabled_and_supported, event_scope
 
 wp.set_module_options({"enable_backward": False})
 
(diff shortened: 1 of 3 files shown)


adenzler-nvidia:
@btaba if there is a good way to infer whether this is part of JAX workflow, would be cool to add that check into the function as well.

Potentially we could even get rid of the flag, but I think it's still useful to have a switch for debugging. Maybe it can be a disablebit in the future.

erikfrey:
@adenzler-nvidia this makes sense, thanks.  I think we want these kinds of capability checks in `io.py` rather than in the runtime, mostly for transparency to the user.

Basically combine this with #412 and make the default behavior more intelligent to also call your `assert_conditional_graph_support` - WDYT?

adenzler-nvidia:
I don't really mind - I mostly put it into the runtime because I didn't want to mess with how we're setting the options in io but I can put it there. should I combine this with #412? Do we have a good way of identifying whether this is using JAX or are we hardcoding hardcoding it in the MJX integration right now?

erikfrey:
We will hardcode it in the MJX.  OK give me a few minutes to add the tests @btaba is asking for in the other PR and then we can smoosh it all together.

erikfrey:
OK @adenzler-nvidia let me know if you want to pull #412 into this one, or I can pull yours, either way is fine.

adenzler-nvidia:
ok, updated with changes from #412.

PR #486, opened 2025-07-09
broadphase filter

### LINKED ISSUE #405: Investigate AABB vs Sphere for Broadphase
For a batch size of 4k, `testspeed` reports that `aloha_pot` at keyframe 0 contains ~3.2 million collision candidates after broadphase, but only ~10,000 collisions after narrowphase.

It's possible that the sphere bounds collision is generating too many candidates and that AABB checking would produce significantly fewer.  `aloha_pot` is a good testcase as it contains many oblong shapes that are better fit by AABB, for example these grippers do not collide with AABB but would with spheres:

Image

thowell:
add AABB and OBB implementations for broadphase filtering

closes #405

adds class `BroadphaseFilter`:
plane: 1
sphere: 2
aabb: 4
obb: 8

for combining filter types

the `broadphase_filter` value is the sum of each applied filter's value (eg: plane + sphere: 1 + 2 = 3)

**benchmark/aloha_pot/scene.xml**

baseline:
- broadphase: nxn
- filtering: plane, sphere
- nconmax: 3,500,000
- nworld: 4096

note: `nconmax` = 3,500,000 to avoid overflow
```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=4096 --event_trace=True --nconmax=3500000 --njmax=3500000 --broadphase=0 --broadphase_filter=3 --iterations=5 --ls_iterations=10
```
```
Summary for 4096 parallel rollouts

 Total JIT time: 0.57 s
 Total simulation time: 98.49 s
 Total steps per second: 41,588
 Total realtime factor: 83.18 x
 Total time per step: 24045.44 ns

Event trace:

      collision: 22078.21
        nxn_broadphase: 429.31
        convex_narrowphase: 21523.50
        primitive_narrowphase: 115.45

```

filtering: plane -> aabb:
- broadphase: nxn
- nconmax: 3,500,000
- nworld: 4096

```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=4096 --event_trace=True --nconmax=3500000 --njmax=3500000 --broadphase=0 --broadphase_filter=5 --iterations=5 --ls_iterations=10 --graph_conditional=False --ctrl_noise=False
```
```
Summary for 4096 parallel rollouts

 Total JIT time: 0.60 s
 Total simulation time: 13.88 s
 Total steps per second: 295,174
 Total realtime factor: 590.35 x
 Total time per step: 3387.83 ns

Event trace:

      collision: 1389.72
        nxn_broadphase: 662.02
        convex_narrowphase: 658.14
        primitive_narrowphase: 59.67
```

filtering: plane -> obb:
- broadphase: nxn
- nconmax: 3,500,000
- nworld: 4096

```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=4096 --event_trace=True --nconmax=3500000 --njmax=3500000 --broadphase=0 --broadphase_filter=9 --iterations=5 --ls_iterations=10 --graph_conditional=False --ctrl_noise=False
```
```
Summary for 4096 parallel rollouts

 Total JIT time: 0.55 s
 Total simulation time: 12.48 s
 Total steps per second: 328,140
 Total realtime factor: 656.28 x
 Total time per step: 3047.48 ns

Event trace:

      collision: 1059.04
        nxn_broadphase: 451.73
        convex_narrowphase: 548.46
        primitive_narrowphase: 48.94
```
---

reduced nconmax/njmax
---

filtering: plane -> aabb:
- broadphase: nxn
- nconmax: 200,000
- nworld: 8196

```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=8196 --event_trace=True --nconmax=200000 --njmax=200000 --broadphase=0 --broadphase_filter=5 --iterations=5 --ls_iterations=10 --graph_conditional=False --ctrl_noise=False
```
```
Summary for 8196 parallel rollouts

 Total JIT time: 0.61 s
 Total simulation time: 19.17 s
 Total steps per second: 427,616
 Total realtime factor: 855.23 x
 Total time per step: 2338.55 ns

Event trace:

      collision: 1088.82
        nxn_broadphase: 686.48
        convex_narrowphase: 387.39
        primitive_narrowphase: 13.73
```

filtering: plane -> obb:
- broadphase: nxn
- nconmax: 200,000
- nworld: 8196

```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=8196 --event_trace=True --nconmax=200000 --njmax=200000 --broadphase=0 --broadphase_filter=9 --iterations=5 --ls_iterations=10 --graph_conditional=False --ctrl_noise=False
```
```
Summary for 8196 parallel rollouts

 Total JIT time: 0.56 s
 Total simulation time: 16.53 s
 Total steps per second: 495,800
 Total realtime factor: 991.60 x
 Total time per step: 2016.94 ns

Event trace:

      collision: 776.83
        nxn_broadphase: 464.16
        convex_narrowphase: 300.47
        primitive_narrowphase: 10.98
```

filtering: plane, sphere->aabb:
- broadphase: nxn
- nconmax: 200,000
- nworld: 8196

```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=8196 --event_trace=True --nconmax=200000 --njmax=200000 --broadphase=0 --broadphase_filter=7 --iterations=5 --ls_iterations=10 --graph_conditional=False --ctrl_noise=False
```
```
Summary for 8196 parallel rollouts

 Total JIT time: 0.62 s
 Total simulation time: 16.80 s
 Total steps per second: 487,856
 Total realtime factor: 975.71 x
 Total time per step: 2049.78 ns

Event trace:

      collision: 815.04
        nxn_broadphase: 418.26
        convex_narrowphase: 379.90
        primitive_narrowphase: 15.66
```

filtering: plane, sphere->obb:
- broadphase: nxn
- nconmax: 200,000
- nworld: 8196

```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=8196 --event_trace=True --nconmax=200000 --njmax=200000 --broadphase=0 --broadphase_filter=11 --iterations=5 --ls_iterations=10 --graph_conditional=False --ctrl_noise=False
```
```
Summary for 8196 parallel rollouts

 Total JIT time: 0.56 s
 Total simulation time: 15.97 s
 Total steps per second: 513,163
 Total realtime factor: 1,026.33 x
 Total time per step: 1948.70 ns

Event trace:

      collision: 713.62
        nxn_broadphase: 427.55
        convex_narrowphase: 267.42
        primitive_narrowphase: 17.43
```

---
notes:
- a new filter 'plane' has been added that i think should always be part of the broadphase filter
- benchmark results have been updated following updates to the implementation
- was seeing nondeterminisitic results (?) with nworld=4096, nconmax=75,000, njmax=75,000. the updated benchmarks with the reduced nconmax/njmax values now have nworld=8196 and nconmax=200,000, njmax=200,000
- based on these results, the filter with best sps is **plane, sphere->obb** so the default broadphase_filter option is set to: (1 + 2 + 8 = 11)


erikfrey:
@thowell fantastic!  How does aabb on its own without the sphere filter look perf wise?

I think we can just switch to some other default besides sphere filter in this PR - my guess would be aabb on its own, but curious what you think.

mujoco_warp/_src/collision_driver.py
```diff
@@ -98,6 +98,184 @@ def _sphere_filter(
...
+  # 2: obb
+  # 3: sphere -> aabb
+  # 4: sphere -> obb
+  # 5: sphere -> aabb -> obb
+
+  sphere = True
+  aabb = True
+  obb = True
+
+  if opt_broadphase_filter == 0 or opt_broadphase_filter == 3 or opt_broadphase_filter == 4 or opt_broadphase_filter == 5:
```
kbayes: Can these values be in an enum?

0: sphere, 
1: aabb, 
2: obb,
3: sphere -> aabb,
4: sphere -> obb,
5: sphere -> aabb -> obb

kbayes (approved):

thowell:
@erikfrey

seems like aabb and obb do not work as well on their own compared to running sphere filtering first. running aabb/obb on their own leads to ncollision overflow warnings for nconmax=75000 (this value does not have overflow warnings with sphere -> aabb/obb).

not sure if this result is expected or highlighting issues with the aabb/obb implementations

@quagla any thoughts?

erikfrey:
That's definitely not expected and sounds like a bug, aabb should be strictly smaller bounding volume than sphere no?  Maybe just tiny box corners could poke outside the sphere boundary in the worst case.

thowell:
@erikfrey @kbayes the implementation is updated and the pr description has updated results. i think the issue was that the sphere filter had special cases for plane geoms and if the sphere filter was not run then plane geoms were not processed correctly

mujoco_warp/_src/collision_driver.py
```diff
@@ -60,8 +61,162 @@ def _zero_collision_arrays(
...
+    return True
+
+  center1 = xmat1 @ center1 + xpos1
+  size1 = xmat1 @ size1
+
+  center2 = xmat2 @ center2 + xpos2
+  size2 = xmat2 @ size2
+
+  margin = wp.max(margin1, margin2)
+  infinite = wp.vec2i(int(inf1[0] or inf1[1] or inf1[2]), int(inf2[0] or inf2[1] or inf2[2]))
```
erikfrey: why not move this up so the two infinite checks are together and both early return?

mujoco_warp/_src/collision_driver.py
```diff
@@ -72,30 +227,42 @@ def _sphere_filter(
...
-  pos1 = geom_xpos_in[worldid, geom1]
-  pos2 = geom_xpos_in[worldid, geom2]
-  size1 = geom_rbound[worldid, geom1]
-  size2 = geom_rbound[worldid, geom2]
+  xpos1 = geom_xpos_in[worldid, geom1]
+  xpos2 = geom_xpos_in[worldid, geom2]
+  xmat1 = geom_xmat_in[worldid, geom1]
+  xmat2 = geom_xmat_in[worldid, geom2]
+
+  if rbound1 == 0.0 or rbound2 == 0.0:
```
erikfrey, line 284: you're already checking these conditions inside `_plane_filter` and `_sphere_filter`, no need to check them twice?

this could probably just be the four conditions:

```py

if opt_broadphase_fitler & int(BroadphaseFilter.PLANE.value):
  ...
if opt_broadphase_fitler & int(BroadphaseFilter.SPHERE.value):
  ...
if opt_broadphase_fitler & int(BroadphaseFilter.AABB.value):
  ...
if opt_broadphase_fitler & int(BroadphaseFilter.OBB.value):
  ...
```

mujoco_warp/_src/types.py
```diff
@@ -62,6 +62,22 @@ class BlockDim:
   mul_m_dense: int = 32
 
 
+class BroadphaseFilter(enum.IntFlag):
+  """Broadphase filter.
```
erikfrey: nit, I think we can be a bit more descriptive:

```py
class BroadphaseFilter(enum.IntFlag):
  """Bitmask specifying which collision functions to run during broadphase.
  ...
  """
```

erikfrey (approved):
Such a great PR, thanks @thowell - just some nits

mujoco_warp/_src/collision_driver.py
thowell: done

mujoco_warp/_src/types.py

mujoco_warp/_src/collision_driver.py
```diff
@@ -60,8 +61,162 @@ def _zero_collision_arrays(
...
+  margin1: float,
+  margin2: float,
+  xpos1: wp.vec3,
+  xpos2: wp.vec3,
+  xmat1: wp.mat33,
+  xmat2: wp.mat33,
+) -> bool:
+  """Axis aligned boxes collision.
+
+  references: mju_raySlab: see Ericson, Real-time Collision Detection section 5.3.3.
```
kbayes: I looked this reference up. It's the ray AABB intersection check. How is this relevant?

erikfrey (changes requested):
@thowell I just realized these changes have no accompanied tests.  We should explicitly test these new filters, ideally with an XML that explicitly rules in or out one filter vs. the other based on what we expect from the XML, so we can validate they are doing what we want.

Probably moving forward we have to be stricter about test coverage, so let's do this before merging.

mujoco_warp/_src/collision_driver.py
thowell: updated to remove the infinite checks from the aabb and obb filters. instead, these filters are not called if one of the geoms is a plane (ie, geom_rbound=0)

mujoco_warp/_src/collision_driver.py
thowell: checking for plane geoms is removed from the individual filters. `_broadphase_filter` checks if either geom is a plane and there are two code paths: if one of the geoms is a plane: 1) plane filter, else: 2) sphere filter, aabb filter, obb filter

mujoco_warp/_src/collision_driver.py
thowell: mujoco's mju_raySlab is one of the functions i referenced as i developed this pr and the reference is from the comment associated with the mju_raySlab implementation

kbayes: I mean this is a standard AABB test. I'd just leave the reference out.

thowell:
the aabb implementation was incorrect and has been updated. the pr description is updated with new testspeed results.

thowell:
the pr also add testspeed flags for ctrl_noise and graph_conditional

thowell:
updated the aabb implementation to align bounding boxes with the world frame. the pr description has updated testspeed results for filters that include aabb.

thowell:
@erikfrey tests are added for broadphase filters

mujoco_warp/_src/collision_driver.py
thowell: updated the reference

mujoco_warp/_src/collision_driver.py
```diff
@@ -60,8 +61,198 @@ def _zero_collision_arrays(
...
+  min_y2 = MJ_MAXVAL
+  min_z2 = MJ_MAXVAL
+
+  sign = wp.vec2(-1.0, 1.0)
+
+  for i in range(2):
+    for j in range(2):
+      for k in range(2):
+        corner1 = wp.vec3(sign[i] * size1[0], sign[j] * size1[1], sign[k] * size1[2])
+        pos1 = center1 + xmat1 @ corner1
```
erikfrey: nit: would it be an eensy weensy teeny tiny bit more performant if we remove the `center1 +` and `center2 + ` from the for loop and instead move them down to the overlap checks below?

mujoco_warp/_src/test_util.py
```diff
@@ -231,13 +232,14 @@ def benchmark(
...
     time_vec = np.zeros(nstep)
     for i in range(nstep):
-      wp.launch(
-        ctrl_noise,
-        dim=(d.nworld, m.nu),
-        inputs=[
-          m.actuator_ctrllimited, m.actuator_ctrlrange, i, 0.01
-        ],
-        outputs=[d.ctrl])  # fmt: skip
+      if ctrl_noise:
```
erikfrey: I would be in favor of moving the `ctrl_noise` and `graph_conditional` changes to  a separate PR.

If you want them in this PR, let's change the `ctrl_noise` param to be a float instead of a bool that defaults to 0.01 (here and in testspeed).  if `ctrl_noise == 0.0` then you can skip the kernel launch.

mujoco_warp/testspeed.py
```diff
@@ -151,6 +155,14 @@ def _main(argv: Sequence[str]):
         m.opt.integrator = k
 
   m.opt.ls_parallel = _LS_PARALLEL.value
+  if _BROADPHASE.value is not None:
+    m.opt.broadphase = _BROADPHASE.value
+  if _BROADPHASE_FILTER.value is not None:
+    m.opt.broadphase_filter = _BROADPHASE_FILTER.value
+
```
erikfrey, line 161: nit: either delete this newline or add newlines above the other conditionals.

erikfrey (approved):
Awesome outcome, just a few nits.

mujoco_warp/_src/collision_driver.py
thowell: done

mujoco_warp/_src/test_util.py
thowell: removed ctrl_noise and graph_conditional changes

mujoco_warp/testspeed.py
thowell: removed graph_conditional so there is no longer a newline between the conditionals. there is a newline between the conditionals and the call to `put_data` below.

mujoco_warp/_src/test_util.py
thowell: missed removing ctrl_noise flag from testspeed.py before merging this pr, fixed with #513

PR #488, opened 2025-07-10
Add distance cutoff to GJK

kbayes:
Performance wise this gives roughly 2x speed up for convex_narrowphase in the Aloha pot scene.


thowell:
@kbayes should we add an option `gjk_cutoff`?

kbayes:
@thowell as in the model? No in that case, it gets set when using the distance sensors.

thowell (approved):

PR #491, opened 2025-07-10
Use convex hull for plane <-> mesh collisions

### LINKED ISSUE #489: Convex <-> Plane collisions not using convex hull
The code is still iterating on the vertices of the visual mesh, which can be very expensive:

for i in range(convex.vertnum):
    support = wp.dot(plane_pos - convex.vert[convex.vertadr + i], n)

This should also walk the convex hull graph for efficiency.

adenzler-nvidia:
closes #489 

I blindly followed the GJK code and am not a geometry expert, so I need a thorough review on this.

Also added a test that covers the >10 vertices case, as well as adjusted one test that didn't even have a collision.

### DIFF
--- a/mujoco_warp/_src/collision_primitive.py
+++ b/mujoco_warp/_src/collision_primitive.py
@@ -2677,97 +2677,97 @@ def primitive_narrowphase_builder(m: Model):
     type1 = geom_type[g1]
     type2 = geom_type[g2]
 
+    worldid = collision_worldid_in[tid]
+
+    _, margin, gap, condim, friction, solref, solreffriction, solimp = contact_params(
+      geom_condim,
+      geom_priority,
+      geom_solmix,
+      geom_solref,
+      geom_solimp,
+      geom_friction,
+      geom_margin,
+      geom_gap,
+      pair_dim,
+      pair_solref,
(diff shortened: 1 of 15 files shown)


mujoco_warp/_src/collision_primitive.py
```diff
@@ -838,68 +838,197 @@ def plane_convex(
...
+    edge_localid = convex.graphadr + 2 + 2 * numvert
+
+    # Find support points
+    max_support = wp.float32(-_HUGE_VAL)
+
+    # hillclimb until no change
+    prev = int(-1)
+    imax = int(0)
+
+    while True:
```
kbayes, line 914: Is possible (probably not) to put all these hill climbs into a helper function?

adenzler-nvidia: I did try it but there was not a lot of code saved, it still needs 3 functions because of the 3 different ways we calculate the distances to the existing contacts.

gattis:
I think this pull is causing some mesh-plane collisions to not work properly as I reported here: [612](https://github.com/google-deepmind/mujoco_warp/issues/612).  In the example provided, the meshes being passed into mujoco_warp are already convex.

PR #493, opened 2025-07-10
Document collision functions.

erikfrey:
Also deleted `collision_box.py` - seems unused?


mujoco_warp/_src/collision_driver.py
```diff
@@ -495,16 +520,22 @@ def _narrowphase(m, d):
...
-  """Collision detection."""
+  """Runs the full collision detection pipeline.
+
+  This function orchestrates the broadphase and narrowphase collision detection stages. It
+  first identifies potential collision pairs using a broadphase algorithm (either N-squared
+  or Sweep-and-Prune, based on `m.opt.broadphase`). Then, for each potential pair, it
+  performs narrowphase collision detection to compute detailed contact information like
+  distance, position, and frame.
+
+  The results are used to populate the `d.contact` array, and the total number of contacts
```
thowell, line 533: we might want to note that `ncon` can be greater than the total number of contacts

mujoco_warp/_src/collision_primitive.py
```diff
@@ -2671,10 +2671,25 @@ def _primitive_narrowphase(
...
 def primitive_narrowphase(m: Model, d: Data):
+  """Runs collision detection on primitive geom pairs discovered during broadphase.
+
+  This function processes collision pairs involving primitive shapes that were
+  identified during the broadphase stage. It computes detailed contact information
+  such as distance, position, and frame, and populates the `d.contact` array.
+
+  The primitive geom types handled are PLANE, SPHERE, CAPSULE, CYLINDER, BOX.
+
+  It also handles collisions between planes and meshes, treating the mesh as a
```
thowell: convex hull?

thowell:
@erikfrey i think we can also remove `BlockDim.box_box`

mujoco_warp/_src/collision_convex.py
```diff
@@ -1053,6 +1053,20 @@ def gjk_epa_sparse(
 
 @event_scope
 def convex_narrowphase(m: Model, d: Data):
+  """Runs narrowphase collision detection for convex geom pairs.
+
+  This function handles collision detection for pairs of convex geometries that were
+  identified during the broadphase. It uses the Gilbert-Johnson-Keerthi (GJK) algorithm to
+  determine the distance between shapes and the Expanding Polytope Algorithm (EPA) to find
```
kbayes, line 1060: Why should the implementation details be in the docstring?

erikfrey: Cause documentation is for humans!  And your GJK code is so lovely, we should definitely think of it as pedagogical, e.g. people will read it to understand physics.  Your fault for making the code so pretty         :-)

PR #498, opened 2025-07-13
add plane ellipsoid primitive collision

thowell:


kbayes:
Does plane<->convex not support this?

thowell:
@kbayes the current plane<>convex implementation doesn't work for plane<>ellipsoid. 

the mujoco reference mjc_PlaneConvex has a special case for ellipsoids https://github.com/google-deepmind/mujoco/blob/11e695b4c2af9886ac4415300cca250493d01669/src/engine/engine_collision_convex.c#L1100

PR #499, opened 2025-07-13
update readme: differentiability

thowell:
note in the readme that differentiability is not currently available


adenzler-nvidia (approved):
Thanks!

PR #502, opened 2025-07-14
io.put_model check for flex collisions

thowell:
- add check in io.put_model for `flex_conaffinity` and `flex_contype`
- update flex/cloth.xml and flex/floppy.xml to have `conaffinity=0` and `contype=0`


quagla (approved):

adenzler-nvidia (approved):
LGTM, thanks!

Issue #458, opened 2025-07-04
Crash in aloha pot env with SAP_TILE broadphase


Kenny-Vilella:
The aloha pot environment exhibits crash (segmentation fault).

For instance, this crash on my computer:
```python mujoco_warp/testspeed.py --function=step --is_sparse=True --mjcf=benchmark/aloha_pot/scene.xml --nstep=10000 --ls_parallel=True --batch_size=1024 --nconmax=3000000 --njmax=400000```

No crash happen when the broadphase is changed to `NXN` or `SAP_SEGMENTED`.
My preliminary investigation has shown that the crash occurs in `_sap_range` when accessing `sap_projection_lower_in`.
Compute sanitizer indicates an illegal read.
It looks like the array get corrupted when calling `tile_sort` in `_segmented_sort`.

The seg fault is still present without graph capture.
I noticed that it stopped to crash when I added some lines of code to save the array and print some results, but it still crashing when I added synchronization or and sleeping.

Kenny-Vilella:
I do not see the crash when running on CPU.

Kenny-Vilella:
The crash appears after some NaN are generated for some environments.
I am still tracking what is happening exactly, The issue is either in the broadphase or narrowphase.
So far, I did not get good results by only fixing the tile_sort, so it may not be due exclusively to the tile_sort function.

Kenny-Vilella:
It looks like there are two sources of crash.
1) There are uninitialized values in the tile_sort functions, when sending NaN values we may take back sorting indices that are out of bound causing crash. This will be solved at the warp level.
2) It looks like that even if the tile_sort output all 0 for sorting indices, it will crash down the line because the collision_pairid would not necessarily be correct. I'm not 100% sure what is happening here, but it looks like that we independently of how we handle NaN values in the tile_sort function, it will always causes problem down the line in mujoco_warp.

So my suggestion is to just not send NaN values to the tile_sort function. It seems to be the easiest and best solution.

Kenny-Vilella:
Changed has been made on the warp side in this [commit](https://github.com/NVIDIA/warp/commit/7368bf7d630c58ca9999f5284076cd5c0e9f2dc3).

yuvaltassa:
@Kenny-Vilella for my own education, what is the difference between `0` and `static_cast<V>(0)`?

Kenny-Vilella:
@yuvaltassa  It's too avoid an implicit conversion by the compiler.
[...]

PR #510, opened 2025-07-16
Fix crash in aloha_pot benchmark

Kenny-Vilella:
Issue was described in #458.

In summary, instability of the benchmark was generating NaN values.
It looks like the tile_sort does not handle the NaN values correctly and output sorting indices that are larger than ngeom, which causes out of bound memory issue and a subsequent crash

I am still figuring out what is happening on the tile_sort and whether there are bugs there.
This PR can be view as a quickfix, if I can fix it on the tile_sort, then I would revert this PR.

### DIFF
--- a/mujoco_warp/_src/collision_driver.py
+++ b/mujoco_warp/_src/collision_driver.py
@@ -179,9 +179,13 @@ def _sap_project(
   radius = rbound + geom_margin[worldid, geomid]
   center = wp.dot(direction_in, xpos)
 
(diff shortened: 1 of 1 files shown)


Kenny-Vilella:
As mentioned in the github issue, I will make some change to the tile_sort in warp, but we still need this.
In any case, it is probably much safer to not provide NaN value to the sort function.

erikfrey (approved):

PR #516, opened 2025-07-17
Set graph conditional to false with event trace in testspeed

kbayes:
Also,

* refactor options a little bit to use Enums (makes code a little bit more robust)
* use enum for broadphase options

### DIFF
--- a/mujoco_warp/testspeed.py
+++ b/mujoco_warp/testspeed.py
@@ -15,6 +15,7 @@
 
 """Run benchmarks on various devices."""
 
+import enum
 import inspect
 from typing import Sequence
 
@@ -27,6 +28,28 @@ from etils import epath
 
 import mujoco_warp as mjwarp
 
+
+class OutputOptions(enum.IntEnum):
+  TEXT = 0
+  TSV = 1
+
+
+class BroadphaseOptions(enum.IntEnum):
+  NXN = 0
+  SAP_TILE = 1
+  SAP_SEGMENTED = 2
+
+
+class SolverOptions(enum.IntEnum):
+  CG = mujoco.mjtSolver.mjSOL_CG
+  NEWTON = mujoco.mjtSolver.mjSOL_NEWTON
+
+
+class ConeOptions(enum.IntEnum):
+  PYRAMIDAL = mujoco.mjtCone.mjCONE_PYRAMIDAL
+  ELLIPTIC = mujoco.mjtCone.mjCONE_ELLIPTIC
+
+
 _FUNCTION = flags.DEFINE_enum(
   "function",
   "step",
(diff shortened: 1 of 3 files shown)


mujoco_warp/testspeed.py
```diff
@@ -27,6 +28,28 @@
 
 import mujoco_warp as mjwarp
 
+
+class OutputOptions(enum.IntEnum):
+  TEXT = 0
+  TSV = 1
+
+
+class BroadphaseOptions(enum.IntEnum):
```
erikfrey: oh, cool!  I didn't know you could do this with absl flags.

But then, why not use the proper enums defined in types.py?

so use `mjwarp.BroadphaseType` and `mjwarp.SolverType` etca

mujoco_warp/testspeed.py
```diff
@@ -142,6 +160,8 @@ def _main(argv: Sequence[str]):
 
   with wp.ScopedDevice(_DEVICE.value):
     m = mjwarp.put_model(mjm)
+    if _EVENT_TRACE.value:
+      m.opt.graph_conditional = False  # graph conditional doesn't work with event trace
```
erikfrey: nice, I like this.  thanks for thinking of it.

any time we manually override an option, let's print to stdout so the user knows we've done this.

mujoco_warp/viewer.py
```diff
@@ -29,18 +30,35 @@
 
 import mujoco_warp as mjwarp
 
+
+class BroadphaseOptions(enum.IntEnum):
```
erikfrey: same note here - we can use the actual types yeah?

erikfrey (approved):
thanks Kyle - just a few nits

mujoco_warp/testspeed.py
btaba: can we hold off on enums in types.py?

kbayes: @btaba How come? Just a clarification, the enums are only used for absl to prettify the CLI, only ints should be passed into the model.

btaba: I can see they are used in the CLI and that's awesome. I wrote "...in **types.py**"

kbayes: @btaba sorry, for further clarification, I'm referring to @erikfrey comment. Some of the enums appear in types.py already, and I can just use them directly. I wasn't sure if you were saying not to use those enums or something else. I don't know where you're coming from.

btaba: Ok sorry clarifying: "enums" materialized as "int" is all I want. I read Erik's comment wrong! FWIU he's suggesting to use "BroadphaseType" from "types.py" rather than defining a new one here (as you mentioned also).

mujoco_warp/viewer.py
kbayes: Done.

mujoco_warp/testspeed.py
kbayes: Done.

mujoco_warp/testspeed.py
kbayes: Done.

PR #518, opened 2025-07-18
Potential fix for overflow of face data in EPA

adenzler-nvidia:
`mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=4096 --nconmax=4000000 --njmax=4000000 --iterations=5 --ls_iterations=10`

is overflowing the face array leading to asserts in debug and overwriting memory of the adjacent threads. This is a hack-fix that avoids the overflow, but I'm deferring to @kbayes to decide what the real fix is going to be.


adenzler-nvidia:
closing in favor of #520

PR #519, opened 2025-07-18
box-box: avoid illegal memory access

adenzler-nvidia:
warp does not short-circuit if statements, so a combined statement will always evaluate all the conditions. This means that the second part can access invalid memory there.


camevor:
Nice spot, thanks!
looks good to me

thowell (approved):

PR #520, opened 2025-07-18
Increase EPA face memory and gracefully return if EPA runs out of memory

kbayes:


mujoco_warp/_src/io.py
```diff
@@ -1026,11 +1026,11 @@ def make_data(mjm: mujoco.MjModel, nworld: int = 1, nconmax: int = -1, njmax: in
...
-    epa_face=wp.zeros(shape=(nconmax, 6 + 3 * MJ_CCD_ITERATIONS), dtype=wp.vec3i),
-    epa_pr=wp.zeros(shape=(nconmax, 6 + 3 * MJ_CCD_ITERATIONS), dtype=wp.vec3),
-    epa_norm2=wp.zeros(shape=(nconmax, 6 + 3 * MJ_CCD_ITERATIONS), dtype=float),
-    epa_index=wp.zeros(shape=(nconmax, 6 + 3 * MJ_CCD_ITERATIONS), dtype=int),
-    epa_map=wp.zeros(shape=(nconmax, 6 + 3 * MJ_CCD_ITERATIONS), dtype=int),
+    epa_face=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=wp.vec3i),
+    epa_pr=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=wp.vec3),
+    epa_norm2=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=float),
+    epa_index=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=int),
+    epa_map=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=int),
```
adenzler-nvidia, line 1033: what are you basing these numbers on? Would it make sense to implement some debugging switch that would give the user some information if this overflows?

kbayes: @adenzler-nvidia  In C MuJoCo there's warnings that get printed to the user (for only the sake of reporting the issue). Can I do something similar with just using wp.printf or do we have special util for reporting warnings?

adenzler-nvidia: I would advise against using prints directly from the kernels because of the impact on register usage (and perf) unless it's a switch we can compile out in release mode. But we added prints to the next_time kernel to report nconmax/njmax overflows and I think it was a good idea, that's a very small kernel and we can afford to pay the price for the registers there.

So maybe just write a flag here, check it in next_time, and report there? We can always remove it again or downgrade to a debug-only thing if it starts showing up on a profile.

Issue #95, opened 2025-04-01
Benchmarks that automatically publish.


erikfrey:
Could use something like [airspeed velocity](https://asv.readthedocs.io/en/latest/) to regularly run benchmarks and publish them.  Could be great to catch perf regressions.

Could couple this with a destination site that justs shows benchmarks.

bkolligs:
Would it be in scope of these benchmarks to compare performance against the existing MJX implementation?

PR #522, opened 2025-07-18
Initial implementation of airspeed velocity benchmarks.

erikfrey:
Introduces an initial benchmark suite that any user can run on their local machine using `asv` - the commandline tool for [airspeed velocity](https://github.com/airspeed-velocity/asv).

With this PR, we can run benchmarks on most of the major MJWarp API calls that comprise step:

* `collision`
* `fwd_position`
* `fwd_velocity`
* `fwd_actuation`
* `fwd_acceleration`
* `solve`
* `step`

We can do so for branches, for commit ranges... lots of flexibility on how we use this.

This PR is half the solution for #95 - later we will figure out how to automatically run these benchmarks and publish them somewhere public.

Also, we aren't benchmarking the kitchen scene yet.  We'll look into that in the future.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,0 +1,56 @@
+# MuJoCo Warp Benchmark Suite
+
+MJWarp uses [airspeed velocity](https://github.com/airspeed-velocity/asv) for benchmarks.
+
+Make sure you install MJWarp in develop so you can run the `asv` command:
+
+```
+pip install -e .[dev,cuda]
+```
+
+Run benchmarks like so at the top level of the `mujoco_warp` checkout:
+
+```
+asv run
+```
+
+You should see output that looks like this:
+
+```
+Couldn't load asv.plugins._mamba_helpers because
+No module named 'libmambapy'
+· Creating environments
+· Discovering benchmarks
+·· Uninstalling from virtualenv-py3.12
+·· Installing c6b97641 <asv> into virtualenv-py3.12.
+· Running 7 total benchmarks (1 commits * 1 environments * 7 benchmarks)
+[ 0.00%] · For mujoco-warp commit c6b97641 <asv>:
+[ 0.00%] ·· Benchmarking virtualenv-py3.12
+[14.29%] ··· benchmark.AlohaPot.time_function
+[14.29%] ··· ================== ============
+                  function
+             ------------------ ------------
+                 collision        365±10ns
+                fwd_position      443±10ns
+                fwd_velocity     31.2±0.2ns
+               fwd_actuation     13.3±0.2ns
+              fwd_acceleration   8.58±0.2ns
+                   solve         612±200ns
+                    step          1.90±1μs
+             ================== ============
+...
+```
+
+Benchmarks are slow to run - if you would like to quickly verify benchmarks are working, you can run with `-q`:
+
+```
+asv run -q
+```
+
+You can also benchmark your own branch:
+
+```
+asv run $MYBRANCH
+```
+
+See the [airspeed velocity documentation](https://asv.readthedocs.io/en/latest/index.html) for more information.
--- a/benchmark/__init__.py
+++ b/benchmark/__init__.py
@@ -1,0 +1,14 @@
+# Copyright 2025 The Newton Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -1,0 +1,63 @@
+import mujoco_warp
+
+
+class AlohaPot(mujoco_warp.BenchmarkSuite):
+  """Aloha robot with a pasta pot on the workbench."""
+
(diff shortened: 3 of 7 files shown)


benchmark/benchmark.py
```diff
@@ -0,0 +1,68 @@
...
+  """Aloha robot with a pasta pot on the workbench."""
+
+  path = "aloha_pot/scene.xml"
+  batch_size = 8192
+  nconmax = 200_000
+  njmax = 200_000
+
+
+class ApptronikApolloFlat(mujoco_warp.BenchmarkSuite):
+  """Apptronik Apollo locomoting on an infinite plain."""
```
thowell: `plain` -> `plane`

benchmark/benchmark.py
```diff
@@ -0,0 +1,68 @@
...
+  """Apptronik Apollo locomoting on Isaac-style pyramids made of thousands of boxes."""
+
+  path = "apptronik_apollo/scene_terrain.xml"
+  batch_size = 8192
+  nconmax = 400_000
+  njmax = 600_000
+
+
+class FrankaEmikaPanda(mujoco_warp.BenchmarkSuite):
+  """Franka Emika Panda on an infinite plain."""
```
thowell: `plain` -> `plane`

benchmark/benchmark.py
```diff
@@ -0,0 +1,68 @@
...
+
+  path = "franka_emika_panda/scene.xml"
+  batch_size = 32768
+  nconmax = 10_000
+  njmax = 150_000
+
+
+class Humanoid(mujoco_warp.BenchmarkSuite):
+  """MuJoCo humanoid on an infinite plain."""
```
thowell: `plain` -> `plane`

benchmark/__init__.py
thowell, line 1: should we add the apache license header?
https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/__init__.py

thowell:
wasn't able to run `asv run` successfully using the instructions in the benchmark readme

```
       class AlohaPot(mujoco_warp.BenchmarkSuite):
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
   AttributeError: module 'mujoco_warp' has no attribute 'BenchmarkSuite'

·· Failed to build the project and import the benchmark suite.
```

it looks like `BenchmarkSuite` is part of the `mujoco_warp` module

```
Python 3.12.9 (main, Mar 12 2025, 00:49:25) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mujoco_warp
>>> mujoco_warp.BenchmarkSuite
<class 'mujoco_warp._src.test_util.BenchmarkSuite'>
```

erikfrey:
@thowell since the code isn't checked in yet, you need to run asv on the branch that contains the class it's complaining it can't reference, so probably something like `asv run asv^!`

The `^!` is git notation so it only tests the most recent commit in the branch.

Once this code is checked in, you will be able to run `asv run` to default to main.

thowell:
@erikfrey that worked, thanks!

erikfrey:
@thowell PTAL

mujoco_warp/_src/test_util.py
```diff
@@ -265,3 +270,106 @@ def benchmark(
...
+  def teardown(self, fn):
+    if fn != "step":
+      # advance time to profile a range of states not just the initial keyframe state
+      wp.capture_launch(self._step_graph)
+
+    self._step += 1
+    self._prev_fn = fn
+
+  def time_function(self, fn):
+    wp.capture_launch(self._fn_graph)
```
adenzler-nvidia, line 374: (sorry if this is obvious from the asv docs) So just to confirm - are running time_function 1000 times here, and rely on the teardown function to advance the state such that we aren't measuring the same thing over and over, right?

It's a bit non-obvious to me but indeed a nice way to make the benchmarking for individual functions work.

No action needed.

Kenny-Vilella: One small thing to add.
Because some forward functions take the same properties as input and output, then the second time the function is called (when time is measured) is not strictly equivalent to the first time it is called (to setup everything).
But I agree that it is probably close enough to not cause issue.

benchmark/README.md
```diff
@@ -0,0 +1,56 @@
...
+                fwd_velocity     31.2±0.2ns
+               fwd_actuation     13.3±0.2ns
+              fwd_acceleration   8.58±0.2ns
+                   solve         612±200ns
+                    step          1.90±1μs
+             ================== ============
+...
+```
+
+Benchmarks are slow to run - if you would like to quickly verify benchmarks are working, you can run with `-q`:
```
Kenny-Vilella, line 44: One small detail is maybe we can add a line about the `--bench` option which allow to filter the benchmark.

benchmark/benchmark.py
```diff
@@ -0,0 +1,83 @@
...
+class ThreeHumanoids(mujoco_warp.BenchmarkSuite):
+  """Three MuJoCo humanoids on an infinite plane.
+
+  Ideally, simulation time scales linearly with number of humanoids.
+  """
+
+  repeat = 100
+  path = "humanoid/n_humanoid.xml"
+  batch_size = 1024
+  nconmax = 50_000
```
Kenny-Vilella, line 82: Om my machine running testspeed with this value leads to overflow.
Maybe 75k is better?

thowell:
@erikfrey might be useful to update the top level readme under 'Benchmarking' to note this airspeed velocity functionality

erikfrey:
Thanks @Kenny-Vilella - after sleeping on this for a few nights I'm not satisfied that we have two different ways of benchmarking in the codebase.  I'm going to try to improve the underlying mechanism used by `testspeed` so we can have both approaches use the same approach, and ensure it's accurate and reliable.

erikfrey:
Closing this in favor of #545

PR #524, opened 2025-07-21
add support.sign following mju_sign convention

thowell:
mujoco reference: https://github.com/google-deepmind/mujoco/blob/0eb2f9f11881b937c4c093b1061e062323059731/src/engine/engine_util_misc.c#L1049

warp.sign reference: https://nvidia.github.io/warp/modules/functions.html#warp.sign


thowell:
discussing with @kbayes, we should update the mujoco implementation (eg, https://github.com/google-deepmind/mujoco/blob/0eb2f9f11881b937c4c093b1061e062323059731/src/engine/engine_collision_convex.c#L307) to utilize sign(x): 1 if x >=0 else -1

PR #526, opened 2025-07-22
Separate reusable collision detection code from Mujoco specific code

nvtw:
First step towards making narrow phase collision code reusable in other solvers. The MR does not change the results the code produces, it mainly moves and separates code.


nvtw:
A very recent warp version that includes this commit (https://github.com/NVIDIA/warp/commit/3889aec8f8a909491b60d616aec59d3bbccc23b9) is required to make the code in this MR run.

erikfrey:
@nvtw starting to look at this - can you let us know whether this has any impact on benchmarks?

nvtw:
The performance was a bit worse but I changed the API such that the code now does exactly the same thing as before (I tested a variant that caches the contacts and writes them at the end but that seems to use a lot of registers). The benchmarks also show that the perf is identical to before the change with the current version. The API got slightly more complicated, that's why I first went for the caching approach. Now there is a callback function for writing contacts - this is also the most flexible approach.

mujoco_warp/_src/collision_primitive.py
```diff
@@ -44,6 +41,37 @@ class mat83f(wp.types.matrix(shape=(8, 3), dtype=wp.float32)):
   pass
 
 
+@wp.struct
+class GeomCore:
```
kbayes: GeomCore is really vague. 

Could we try something like PrimitiveGeom with type, pos, rot, and size? And have one for MeshGeom?

nvtw: I'll try to find a better solution. I could also just keep the Geom struct but that one gets really bloated. The problem is that all methods postfixed with _wrapper must have the exact same argument list due to how these methods get called. It's not possible that one takes a Geom and one a MeshGeom. Maybe I'll just go back to the Geom struct and remove GeomCore.

nvtw: I went back to using the Geom struct. The goal of this MR is to make the contact generation reusable (I hope to move the content in the file collision_primitive_core.py into Newton in a later step). Introducing GeomCore was a unnecessary drive by fix to avoid struct bloat. It's better to address that in a separate MR.

nvtw:
@erikfrey @kbayes would be great if we could merge this MR until the end of the week since this newton MR depends on it: https://github.com/newton-physics/newton/pull/482

erikfrey:
@nvtw whew this is tricky, I wonder if we can change this a bit to simplify it.  Right now it looks like you are doing this:

```py
def get_plane_capsule(contact_writer: Any):
  @wp.func
  def plane_sphere(
    # In:
    plane_normal: wp.vec3,
    plane_pos: wp.vec3,
    sphere_center: wp.vec3,
    sphere_radius: float,
    contact_writer_args: Any,
  ) -> int:
  ```

Instead do:

```py
@wp.func
def plane_sphere(
  # In:
  plane_normal: wp.vec3,
  plane_pos: wp.vec3,
  sphere_center: wp.vec3,
  sphere_radius: float,
  # Out:
  dist_out: wp.array(dtype=float)
  pos_out: wp.array(dtype=wp.vec3)
  frame_out: wp.array(dtype=wp.mat33)
) -> int:
```

So basically the function only writes the **core** data, then the user outside in the calling kernel can write extra information (condim etc)

nvtw:
I'll see what I can do. I had a similar API as you just proposed but it was noticeably slower. The exact API you propose is tricky since I still need to report the indices of the contacts that got written (it's all parallel, the contacts are not necessarily located next to each other) and I would prefer not to force the user to have an array of full frames because technically storing the 3rd coordinate axis is redundant as it can always be computed with the cross product from the first two frame axes.

nvtw:
@erikfrey  I decided that I really would like to keep the approach with the contact writer callback for newton because of its maximal flexibility. Shall I write wrappers around it for mujoco with the API you suggested (I sill need an additional indexer and max contact argument as well as a way to return the indices of the generated contacts)? There are already wrappers of some kind in collision_primitive.py in this MR since they are a technical requirement for how the different contact gen functions are called. We would end up with two layers of wrappers. I can do it but honestly it feels a bit like overkill?

nvtw:
I could add the following code to bridge the callback API and the API proposed by Erik. But it feels/looks like boilerplate code to me, I would prefer to go directly for the callback API as it is currently implemented in this MR.
```python
@wp.struct
class WriterArgs:
  max_num_contacts: int
  contact_indexer: wp.array(dtype=int)
  dist_out: wp.array(dtype=float)
  pos_out: wp.array(dtype=wp.vec3)
  frame_out: wp.array(dtype=wp.mat33)
  indices_out: wp.array(dtype=int)
  margin: float
  local_indexer: wp.array(dtype=int)


@wp.func
def _contact_writer(contact: ContactPoint, args: WriterArgs):
  active = (contact.dist - args.margin) < 0
  if active:
    cid = wp.atomic_add(args.contact_indexer, 0, 1)
    if cid < args.max_num_contacts:
      args.dist_out[cid] = contact.dist
      args.pos_out[cid] = contact.pos
      args.frame_out[cid] = contact.frame
      args.indices_out[args.local_indexer[0]] = cid
      args.local_indexer[0] = args.local_indexer[0] + 1


@wp.func
def plane_sphere(
  # In:
  plane_normal: wp.vec3,
  plane_pos: wp.vec3,
  sphere_center: wp.vec3,
  sphere_radius: float,
  margin: float,
  max_num_contacts: int,
  # Out:
  contact_indexer: wp.array(dtype=int),
  dist_out: wp.array(dtype=float),
  pos_out: wp.array(dtype=wp.vec3),
  frame_out: wp.array(dtype=wp.mat33),
  indices_out: wp.array(dtype=int),  # Must offer enough space for the maximal number of contacts this func can generate
) -> int:
  local_indexer = wp.zeros(shape=(1,), dtype=int)

  args = WriterArgs()
  args.max_num_contacts = max_num_contacts
  args.contact_indexer = contact_indexer
  args.dist_out = dist_out
  args.pos_out = pos_out
  args.frame_out = frame_out
  args.indices_out = indices_out
  args.margin = margin
  args.local_indexer = local_indexer
  return wp.static(get_plane_sphere(_contact_writer))(plane_normal, plane_pos, sphere_center, sphere_radius, args)
```

mujoco_warp/_src/io.py
```diff
@@ -1591,7 +1594,14 @@ def get_data_into(
 
   result.contact.dist[:] = d.contact.dist.numpy()[:ncon]
   result.contact.pos[:] = d.contact.pos.numpy()[:ncon]
-  result.contact.frame[:] = d.contact.frame.numpy()[:ncon].reshape((-1, 9))
+  frame = np.zeros((ncon, 9))
```
thowell: maybe something like
```
results.contact.frame = np.hstack([normal, tangent, tangent2])
```

mujoco_warp/_src/io.py
```diff
@@ -971,7 +971,9 @@ def make_data(mjm: mujoco.MjModel, nworld: int = 1, nconmax: int = -1, njmax: in
     contact=types.Contact(
       dist=wp.zeros((nconmax,), dtype=float),
       pos=wp.zeros((nconmax,), dtype=wp.vec3f),
-      frame=wp.zeros((nconmax,), dtype=wp.mat33f),
+      # frame=wp.zeros((nconmax,), dtype=wp.mat33f),
```
thowell: we can remove this commented out line

mujoco_warp/_src/math.py
thowell, line 1: maybe `frame`, `tangent`, `normal`?

we should add docstrings to these functions

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2144 @@
...
+  elif face_idx == 5:
+    rotmore[0, 0] = -1.0
+    rotmore[1, 1] = +1.0
+    rotmore[2, 2] = -1.0
+
+  return rotmore
+
+
+@wp.func
+def make_tangent(a: wp.vec3):
```
thowell: we can add return type `wp.vec3`

mujoco_warp/_src/io.py
nvtw: Done

mujoco_warp/_src/io.py
nvtw: Done

mujoco_warp/_src/math.py
nvtw: I added some missing docstrings, hope I took the methods you are referring to.

mujoco_warp/_src/collision_primitive_core.py
nvtw: Done

mujoco_warp/_src/collision_driver_test.py
```diff
@@ -492,7 +492,9 @@ def test_sdf_collision(self, fixture):
       result = False
       test_dist = d.contact.dist.numpy()[i]
       test_pos = d.contact.pos.numpy()[i, :]
-      test_frame = d.contact.frame.numpy()[i].flatten()[0:3]
+      # Reconstruct frame from normal and tangent
+      test_normal = d.contact.normal.numpy()[i, :]
+      test_frame = test_normal[0:3]
       check_dist = np.allclose(actual_dist, test_dist, rtol=5e-2, atol=1.0e-1)
       check_frame = np.allclose(actual_frame, test_frame, rtol=5e-2, atol=1.0e-1)
```
thowell, line 498: this is testing normals? should we update the comment above?

mujoco_warp/_src/collision_primitive.py
```diff
@@ -144,14 +141,53 @@ def _geom(
   return geom
 
 
+@wp.func
+def _write_contact_ext(
```
thowell: maybe rename to something that indicates that it is writing additional parameters including solver parameters?

mujoco_warp/_src/collision_primitive.py
```diff
@@ -16,10 +16,7 @@
 import warp as wp
 
 from .collision_hfield import hfield_triangle_prism
-from .math import closest_segment_point
-from .math import closest_segment_to_segment_points
-from .math import make_frame
-from .math import normalize_with_norm
+from .collision_primitive_core import *
```
thowell: might be helpful to `import .collision_primitive_core as core` and then call methods/types using `core.`

mujoco_warp/_src/collision_primitive.py
thowell: this function and the one below look similar?

we should include docstrings for this function and the one below to better distinguish between the utility of each function

mujoco_warp/_src/collision_primitive.py
```diff
@@ -194,1325 +232,349 @@ def write_contact(
...
 @wp.func
-def _plane_sphere(plane_normal: wp.vec3, plane_pos: wp.vec3, sphere_pos: wp.vec3, sphere_radius: float):
-  dist = wp.dot(sphere_pos - plane_pos, plane_normal) - sphere_radius
-  pos = sphere_pos - plane_normal * (sphere_radius + 0.5 * dist)
-  return dist, pos
-
-
-@wp.func
-def plane_sphere(
+def plane_sphere_wrapper(
```
thowell: this is not an api function, so maybe `_plane_sphere_wrapper`? (similar comment for all the wrappers in this file)

thowell: is this function missing a return type?

mujoco_warp/_src/collision_primitive.py
```diff
@@ -2829,16 +959,31 @@ def _primitive_narrowphase(
...
 
+    start = -1
+    end = -1
     for i in range(wp.static(len(_primitive_collisions_func))):
       collision_type1 = wp.static(_primitive_collisions_types[i][0])
       collision_type2 = wp.static(_primitive_collisions_types[i][1])
 
       if collision_type1 == type1 and collision_type2 == type2:
-        wp.static(_primitive_collisions_func[i])(
+        start, end = wp.static(_primitive_collisions_func[i])(
```
thowell, line 985: what are the performance implications for this change that splits writing contacts into two steps? please provide testspeed results for a few of the benchmark scenes comparing the current approach with this pr.

mujoco_warp/_src/collision_primitive.py
```diff
@@ -2829,16 +959,31 @@ def _primitive_narrowphase(
...
-          worldid,
+          margin,
+          ncon_out,
+          contact_dist_out,
+          contact_pos_out,
+          contact_normal_out,
+          contact_tangent_out,
+        )
+
+    # Write contact metadata for all contacts in the range
```
thowell: metadata -> solver parameters?

mujoco_warp/_src/collision_primitive.py
```diff
@@ -2829,16 +959,31 @@ def _primitive_narrowphase(
...
+    if start >= 0:
+      for i in range(start, end):
+        _write_contact_ext(
```
thowell: maybe rename to indicate that this function writes solver parameters?

mujoco_warp/_src/collision_primitive.py
```diff
@@ -144,14 +141,53 @@ def _geom(
...
+    contact_includemargin_out[cid] = includemargin
+    contact_dim_out[cid] = condim_in
+    contact_friction_out[cid] = friction_in
+    contact_solref_out[cid] = solref_in
+    contact_solreffriction_out[cid] = solreffriction_in
+    contact_solimp_out[cid] = solimp_in
+
+
 @wp.func
 def write_contact(
```
thowell, line 193: maybe rename this function to indicate that it is writing information about collisions?

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2151 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import warp as wp
+
+from .math import closest_segment_point
```
thowell, line 20: should these functions be part of core instead of math?

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2151 @@
...
+  """Represents a contact point with position, normal, tangent, and distance."""
+
+  pos: wp.vec3
+  normal: wp.vec3
+  tangent: wp.vec3
+  dist: float
+
+
+@wp.func
+def pack_contact(pos: wp.vec3, normal: wp.vec3, tangent: wp.vec3, dist: float) -> ContactPoint:
```
thowell: `contact`?

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2151 @@
...
+
+@wp.func
+def pack_contact_auto_tangent(pos: wp.vec3, normal: wp.vec3, dist: float) -> ContactPoint:
+  """Creates a ContactPoint with automatically generated tangent from normal."""
+  tangent = make_tangent(normal)
+  return ContactPoint(pos=pos, normal=normal, tangent=tangent, dist=dist)
+
+
+@wp.func
+def extract_frame(c: ContactPoint) -> wp.mat33:
```
thowell: `frame`?

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2151 @@
...
+  contact: ContactPoint,
+  max_contacts: int,
+  # Out:
+  dist_out: wp.array(dtype=float),
+  pos_out: wp.array(dtype=wp.vec3),
+  normal_out: wp.array(dtype=wp.vec3),
+  tangent_out: wp.array(dtype=wp.vec3),
+):
+  """Writes contact data to output arrays if index is within bounds."""
+  if index < max_contacts:
```
thowell, line 124: maybe check the `index < max_contacts` outside of this function

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2151 @@
...
+
+
+@wp.func
+def pack_contact(pos: wp.vec3, normal: wp.vec3, tangent: wp.vec3, dist: float) -> ContactPoint:
+  """Creates a ContactPoint from position, normal, tangent, and distance."""
+  return ContactPoint(pos=pos, normal=normal, tangent=tangent, dist=dist)
+
+
+@wp.func
+def pack_contact_auto_tangent(pos: wp.vec3, normal: wp.vec3, dist: float) -> ContactPoint:
```
thowell: maybe `contact_auto_tangent`?

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2151 @@
...
+  contact_counter_out: wp.array(dtype=int),
+  dist_out: wp.array(dtype=float),
+  pos_out: wp.array(dtype=wp.vec3),
+  normal_out: wp.array(dtype=wp.vec3),
+  tangent_out: wp.array(dtype=wp.vec3),
+):
+  """Calculates contacts between a plane and a convex object.
+
+  Args:
+    plane_normal: Normal vector of the plane.
```
thowell: we should update the docstrings to include the type in parentheses `plane_normal (wp.vec3): ...`

mujoco_warp/_src/collision_primitive_core.py
nvtw: I would like to keep the check inside the write function for maximal safety. Because if the caller needs to do the check, it might get forgotten.

mujoco_warp/_src/collision_primitive.py
nvtw: Done

mujoco_warp/_src/collision_primitive.py
nvtw: Done

nvtw: Added the return type. Other wrapper functions missed them too.

mujoco_warp/_src/collision_primitive.py
nvtw: Done

mujoco_warp/_src/collision_primitive_core.py
nvtw: I'd prefer to keep these functions in math for now (they were already there before my changes). But when the core part gets moved into newton, I'll indeed need to include a copy of the function to have it available in newton.

mujoco_warp/_src/collision_primitive_core.py
nvtw: Done

mujoco_warp/_src/collision_primitive_core.py
nvtw: Done

mujoco_warp/_src/collision_primitive_core.py
nvtw: Done

mujoco_warp/_src/collision_primitive_core.py
nvtw: Done

mujoco_warp/_src/collision_primitive.py
nvtw: I would like to keep the name of this function since it was already here before and is not affected by my changes. The smaller contact writer functions I added are prefixed with _ to indicate that they should not be used outside of the file they're declared in.

mujoco_warp/_src/collision_primitive.py
nvtw: Done

mujoco_warp/_src/collision_primitive.py
nvtw: Done

nvtw: Done, it's now called _write_contact_solver_params

mujoco_warp/_src/collision_driver_test.py
nvtw: I removed the comment since it is misleading and not really needed

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -0,0 +1,2158 @@
...
+
+
+@wp.func
+def contact(pos: wp.vec3, normal: wp.vec3, tangent: wp.vec3, dist: float) -> ContactPoint:
+  """Creates a ContactPoint from position, normal, tangent, and distance."""
+  return ContactPoint(pos=pos, normal=normal, tangent=tangent, dist=dist)
+
+
+@wp.func
+def contact_auto_tangent(pos: wp.vec3, normal: wp.vec3, dist: float) -> ContactPoint:
```
thowell: not sure that we need this function. i think it is ok to only have `contact` and then construct a tangent vector using `make_tangent` to then provide to `contact`

thowell:
@nvtw

- i think we should should move the math functions imported in collision_primitive_core.py to the file itself so that this file is self contained. we probably need to think of this file as containing everything that will be imported into collision_primitive.py

- one of the outcomes we would like for collision_primitive_core.py is for the utilities provided to be modular and eventually work across engines. it seems like there should be a more general and flexible version of the api. i think the collision functions (eg, plane_sphere, etc) could return fixed size arrays, maybe something like:

```
# core
mat23 = wp.types.matrix(shape=(2, 3), dtype=float)

@wp.func
def plane_capsule(
  plane_normal: wp.vec3,
  plane_pos: wp.vec3,
  cap_center: wp.vec3,
  cap_axis: wp.vec3,
  cap_radius: float,
  cap_half_length: float,
) -> Tuple[wp.vec2, mat23, mat23, mat23]:
  dist = wp.vec2()
  pos = mat23()
  normal = mat23()
  tangent = mat23()

  ...

  return dist, pos, normal, tangent

# engine specific code that utilizes core functions
import core

@wp.func
def plane_capsule_wrapper(
  ...
):
  dist, pos, normal, tangent = core.plane_capsule(...)

  for i in range(2):
    if dist[i] - margin[i] < 0.0:
      conid = wp.atomic_add(ncon, 0, 1)

      if conid < nconmax:
        core.write_contact(...)
```

nvtw:
@thowell 
I strongly prefer not to use types like mat23 to store 2 vec3. Also box box can create up to 8 contacts. This uses so many registers that the code becomes significantly slower. In one of the many old commits in this MR, there was exactly an API like this that I discarded due to performance issues.
For functions that can only produce 1 contact, I could return just the contact with no speed impact. But then the API gets fragmented since functions returning multiple contacts would have a different signature.

mujoco_warp/_src/collision_primitive_core.py
nvtw: Done

mujoco_warp/_src/collision_primitive.py
nvtw: **THIS PR:**

(env) C:\git7\mujoco_warp>python mujoco_warp/testspeed.py --function=step benchmark/apptronik_apollo/scene_flat.xml --nworld=8192 --nstep=1000 --njmax=51 --nconmax=150000 -o opt.is_sparse=True -o opt.ls_parallel=True --event_trace=False
Loading model from: benchmark\apptronik_apollo\scene_flat.xml...
  nbody: 37 nv: 25 ngeom: 19 nu: 19 is_sparse: True
  solver: NEWTON cone: PYRAMIDAL iterations: 10 ls_iterations: 20 ls_parallel: True
  integrator: EULER graph_conditional: True
Data
  nworld: 8192 nconmax: 150000 njmax: 51

Total JIT time: 0.54 s
Total simulation time: 15.02 s
Total steps per second: 545,287
Total realtime factor: 2,726.43 x
Total time per step: 1833.90 ns

**MAIN**

(env) C:\git7\mujoco_warp>python mujoco_warp/testspeed.py --function=step benchmark/apptronik_apollo/scene_flat.xml --nworld=8192 --nstep=1000 --njmax=51 --nconmax=150000 -o opt.is_sparse=True -o opt.ls_parallel=True --event_trace=False
Loading model from: benchmark\apptronik_apollo\scene_flat.xml...
  nbody: 37 nv: 25 ngeom: 19 nu: 19 is_sparse: True
  solver: NEWTON cone: PYRAMIDAL iterations: 10 ls_iterations: 20 ls_parallel: True
  integrator: EULER graph_conditional: True
Data
  nworld: 8192 nconmax: 150000 njmax: 51

Total JIT time: 0.53 s
Total simulation time: 15.15 s
Total steps per second: 540,784
Total realtime factor: 2,703.92 x
Total time per step: 1849.17 ns

(env) C:\git7\mujoco_warp>

nvtw: Perf seems to have not changed. I sometimes also get runs where this PR is a tiny bit slower than main but I assume these are measurement fluctuations.

nvtw: **THIS PR**

(env) C:\git7\mujoco_warp>python mujoco_warp/testspeed.py --function=step benchmark/humanoid/humanoid.xml --nworld=8192 --nstep=1000 --njmax=52 --nconmax=200000 -o opt.is_sparse=True -o opt.ls_parallel=True --event_trace=False
Loading model from: benchmark\humanoid\humanoid.xml...
  nbody: 17 nv: 27 ngeom: 20 nu: 21 is_sparse: True
  solver: NEWTON cone: PYRAMIDAL iterations: 10 ls_iterations: 20 ls_parallel: True
  integrator: EULER graph_conditional: True
Data
  nworld: 8192 nconmax: 200000 njmax: 52

Total JIT time: 1.06 s
Total simulation time: 11.87 s
Total steps per second: 690,220
Total realtime factor: 3,451.10 x
Total time per step: 1448.81 ns

**MAIN**

(env) C:\git7\mujoco_warp>python mujoco_warp/testspeed.py --function=step benchmark/humanoid/humanoid.xml --nworld=8192 --nstep=1000 --njmax=52 --nconmax=200000 -o opt.is_sparse=True -o opt.ls_parallel=True --event_trace=False
Loading model from: benchmark\humanoid\humanoid.xml...
  nbody: 17 nv: 27 ngeom: 20 nu: 21 is_sparse: True
  solver: NEWTON cone: PYRAMIDAL iterations: 10 ls_iterations: 20 ls_parallel: True
  integrator: EULER graph_conditional: True
Data
  nworld: 8192 nconmax: 200000 njmax: 52

Total JIT time: 0.85 s
Total simulation time: 11.94 s
Total steps per second: 686,205
Total realtime factor: 3,431.03 x
Total time per step: 1457.29 ns

(env) C:\git7\mujoco_warp>

nvtw: **THIS PR**

(env) C:\git7\mujoco_warp>python mujoco_warp/testspeed.py --function=step benchmark/aloha_pot/scene.xml --keyframe=0 --nworld=8192 --nstep=1000 --njmax=100 --nconmax=600000 -o opt.is_sparse=True -o opt.ls_parallel=True --event_trace=False
Loading model from: benchmark\aloha_pot\scene.xml...
  nbody: 26 nv: 23 ngeom: 204 nu: 14 is_sparse: True
  broadphase: SAP_TILE broadphase_filter: PLANE|SPHERE|OBB
  solver: NEWTON cone: ELLIPTIC iterations: 100 ls_iterations: 50 ls_parallel: True
  integrator: EULER graph_conditional: True
Data
  nworld: 8192 nconmax: 600000 njmax: 100

Rolling out 1000 steps at dt = 0.002...

Total JIT time: 65.31 s
Total simulation time: 43.50 s
Total steps per second: 188,320
Total realtime factor: 376.64 x
Total time per step: 5310.10 ns

**MAIN**

(env) C:\git7\mujoco_warp>python mujoco_warp/testspeed.py --function=step benchmark/aloha_pot/scene.xml --keyframe=0 --nworld=8192 --nstep=1000 --njmax=100 --nconmax=600000 -o opt.is_sparse=True -o opt.ls_parallel=True --event_trace=False
Loading model from: benchmark\aloha_pot\scene.xml...
  nbody: 26 nv: 23 ngeom: 204 nu: 14 is_sparse: True
  broadphase: SAP_TILE broadphase_filter: PLANE|SPHERE|OBB
  solver: NEWTON cone: ELLIPTIC iterations: 100 ls_iterations: 50 ls_parallel: True
  integrator: EULER graph_conditional: True
Data
  nworld: 8192 nconmax: 600000 njmax: 100

Rolling out 1000 steps at dt = 0.002...

Total JIT time: 60.10 s
Total simulation time: 44.11 s
Total steps per second: 185,710
Total realtime factor: 371.42 x
Total time per step: 5384.73 ns

(env) C:\git7\mujoco_warp>

thowell:
@nvtw @erikfrey @adenzler-nvidia if we are happy with the api i think this pr is in a good state

will leave it up to @erikfrey and/or @adenzler-nvidia to approve the api

thanks @nvtw!

adenzler-nvidia:
No objections from my side. I hear the API has been discussed in detail, so I expect @nvtw to have implemented the necessary changes. Let's give people one more day to object, otherwise I suggest to merge to avoid dragging this out even longer.

kbayes:
Thanks @adenzler-nvidia and @nvtw. I'll take another pass later today.

kbayes:
> @thowell I strongly prefer not to use types like mat23 to store 2 vec3. Also box box can create up to 8 contacts. This uses so many registers that the code becomes significantly slower. In one of the many old commits in this MR, there was exactly an API like this that I discarded due to performance issues. For functions that can only produce 1 contact, I could return just the contact with no speed impact. But then the API gets fragmented since functions returning multiple contacts would have a different signature.

@nvtw We can get away with only 4 contacts for box-box collisions. I just added the pruning logic for multicontact. Maybe we can export that and prune box-box contacts. Better yet, we can just use GJK/EPA in place of box-box if this is not a performance hit. I'm starting to see better behavior with GJK/EPA computing box-box collisions.

kbayes:
@nvtw have you considered handling positive distances given a cutoff like we discussed in Zurich?

nvtw:
I will close this MR as there is no consensus and it became to messy. The MR will be replaced by this other MR: https://github.com/google-deepmind/mujoco_warp/pull/646

PR #531, opened 2025-07-23
Adding UT to compare parallel and iterative linesearch

Kenny-Vilella:
Added a simple UT that compare the resulting alpha from iterative and parallel linesearch.
It also check that increasing nlsp improve convergence with iterative linesearch.

### DIFF
--- a/mujoco_warp/_src/solver_test.py
+++ b/mujoco_warp/_src/solver_test.py
@@ -68,6 +68,58 @@ class SolverTest(parameterized.TestCase):
 
       _assert_eq(mjwarp_cost, mj_cost, name="cost")
 
+  @parameterized.parameters(ConeType.PYRAMIDAL, ConeType.ELLIPTIC)
+  def test_equivalent_linesearch(self, cone):
+    """Test that iterative and parallel linesearch leads to equivalent results."""
+    _, _, m, d = test_util.fixture(
+      "humanoid/humanoid.xml",
+      cone=cone,
+      ls_parallel=False,
+      iterations=50,
+      ls_iterations=50,
+    )
+
+    # One step to obtain more non-zeros results
+    mjwarp.step(m, d)
+
+    # Preparing for linesearch
+    m.opt.iterations = 0
+    mjwarp.fwd_velocity(m, d)
+    mjwarp.fwd_acceleration(m, d, factorize=True)
+    solver.solve(m, d)
+
+    # Storing some initial values
+    d_efc_Ma = d.efc.Ma.numpy()
+    d_efc_Jaref = d.efc.Jaref.numpy()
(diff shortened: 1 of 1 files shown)


Kenny-Vilella:
@thowell  Do you think that this kind of test has some value?

mujoco_warp/_src/solver_test.py
```diff
@@ -68,6 +68,58 @@ def cost(qacc):
 
       _assert_eq(mjwarp_cost, mj_cost, name="cost")
 
+  @parameterized.parameters(ConeType.PYRAMIDAL, ConeType.ELLIPTIC)
+  def test_equivalent_linesearch(self, cone):
```
thowell: `test_parallel_linesearch`?

mujoco_warp/_src/solver_test.py
```diff
@@ -68,6 +68,58 @@ def cost(qacc):
...
+    """Test that iterative and parallel linesearch leads to equivalent results."""
+    _, _, m, d = test_util.fixture(
+      "humanoid/humanoid.xml",
```
thowell, line 80: can we change this to `constraints.xml` and test keyframes: 0, 1, 2

mujoco_warp/_src/solver_test.py
Kenny-Vilella: Yes I agree that the name is not great, it is difficult to find an informative name.
I follow your suggestion.

mujoco_warp/_src/solver_test.py
Kenny-Vilella: At first, I was doing this, but I am not sure that the test is really meaningful as `alpha` is always equal to 1.0.
With the humanoid I get alpha ~ 0.52 for pyramidal and 0.16 for elliptic.

Also, one of my assert was wrong, after fixing it the code UT fail for elliptic and parallel on my workstation.
But it seems to pass on CI. I am a bit confused here.
If I go back to an old commit one month back, it also fails, so it is not due to recent change.

EDIT: it works on my machine when running on cpu.

thowell: yeah it makes sense to keep `humanoid.xml` if the steps size for `constraints.xml` are 1.0 for its keyframes

Kenny-Vilella:
The UT is not working on GPU.
So this PR should not be merged as long as we did not figure out the reason for the fail.

Kenny-Vilella:
Solved the discrepancy between CPU and GPU by adding some `copy()`.
Now the UT if failing on CI as it should be.

Kenny-Vilella:
I deactivated the failing case until the elliptic case has been fixed.
Maybe we can merge this first as it still has value for the pyramidal case.

PR #532, opened 2025-07-23
Default to c mujoco ccd_iterations and expose override flag in viewer and testspeed

kbayes:


mujoco_warp/_src/io.py
```diff
@@ -1071,6 +1071,7 @@ def put_data(
   nworld: Optional[int] = None,
   nconmax: Optional[int] = None,
   njmax: Optional[int] = None,
+  ccd_iterations: Optional[int] = None,
```
erikfrey: shouldn't this be a model opt?

erikfrey (changes requested):
@thowell @kbayes `put_data` should be kept as slim as possible, and wherever possible we should rely on `types.Option` for algorithmic parameters

mujoco_warp/_src/io.py
kbayes: @erikfrey It is already, it's awkward because put_data doesn't take model as a parameter. Can we just pass model as the sole parameter and have nworld, nconmax, etc live in model? They're constants anyway.

erikfrey: Ah yeah agreed this is an annoyance.  Ideally we would have 1:1 between Warp Model and MjModel but I understand that's not always practical.

Given you said you're still investigating an issue where C MuJoCo converges faster, should we hold off on this change until that investigation is complete?

kbayes: Done.

kbayes:
@erikfrey this is ready now

erikfrey (changes requested):
@kbayes OK nice - #551 completely changes the commandline flags in a way that will support ccd_iterations, so if you remove the changes to `testspeed` and `viewer`, it'll make merging this PR and that one much easier.  Otherwise looks great.

erikfrey:
@kbayes what's the status of this PR?

kbayes:
@erikfrey I haven't circle back to this yet. There were some bugs / performance issues that popped up when increasing iterations from 12 -> 50. I fixed all of what I could find recently.

kbayes:
Closing in favor of #715.

PR #536, opened 2025-07-23
add option for maxconpair

thowell:
- add field `maxconpair` to `Option` that limits the maximum number of contacts per geom pair
- default is changed from `mujoco.mjMAXCONPAIR` (50) to 4


btaba:
Hi @thowell I don't think this will result in correct behavior. We want to be judicious about which 4 points to pick. IMO it would make more sense to have an option to either [1] do 50 points (although I'm not sure who really wants this...TBD), [2] do 4 with a fixed selection criteria.

Option [3] could be: pick N with a selection criteria that supports N, but I'm not really sure that's necessary right now.

Lmk wyt, I'm in favor of just doing [2]

PR #537, opened 2025-07-24
Work around asserts for warp arrays with stride 0

adenzler-nvidia:
I suggest to just declare a max count for the number of worlds.


mujoco_warp/_src/io.py
```diff
@@ -816,13 +820,26 @@ def make_data(mjm: mujoco.MjModel, nworld: int = 1, nconmax: int = -1, njmax: in
   Returns:
     Data: The data object containing the current state and output arrays (device).
   """
+
+  nworld = nworld or 1
```
btaba: why not just use 1 as the default kwarg? I must be missing something but not sure what

mujoco_warp/_src/io.py
```diff
@@ -366,10 +369,11 @@ def create_nmodel_batched_array(mjm_array, dtype, expand_dim=True):
     array._is_batched = True
     if not expand_dim:
       array.strides = (0,) + array.strides[1:]
+      array.shape = (MAX_WORLDS,) + array.shape[1:]
       return array
     array.strides = (0,) + array.strides
     array.ndim += 1
-    array.shape = (1,) + array.shape
+    array.shape = (MAX_WORLDS,) + array.shape
```
btaba, line 376: Ok, just checked and this works for me. I just need to do a 

```
    if hasattr(field, '_is_batched'):
      field.shape = (1,) + field.shape[1:]
```

on my end, so the numpy shapes aren't borked before I convert them to jax arrays

btaba (approved):
Thanks @adenzler-nvidia !!

mujoco_warp/_src/io.py
adenzler-nvidia: would it be an option for you to have an optional argument use_jax that just leaves the shapes as-is? Then you don't need to do this weird revert.

mujoco_warp/_src/io.py
adenzler-nvidia: I copied the pattern from put_data and didn't think too much about it. I can revert. We should clean up the epilogue of the two functions a bit to unify some of the error handling and the API.

btaba: Ah ok, either way is fine. If it were up to me, I'd save the extra line of code and get rid of the Optional

mujoco_warp/_src/io.py
btaba: I think this approach is fine. Maybe update the test `test_put_model_nworld_array` in `io.py` to check that the leading dim is MAX_WORLDS, this would help catch a breaking API change.

mujoco_warp/_src/io.py
adenzler-nvidia: fixed

mujoco_warp/_src/io.py
adenzler-nvidia: updated the tests.

adenzler-nvidia:
@btaba could you have a look at jax_test? I think it's a bit awkward but I'm not sure if there's a better option. We do have to fix the shapes afterwards again to avoid the asserts.. but at least we can then provide the correct values.

I still don't really like the approach but this is probably as good as it gets.

mujoco_warp/_src/jax_test.py
```diff
@@ -89,9 +89,14 @@ def step(carry, _):
       graph_compatible=True,
     )
 
+    # reset qpos0 to get the right numpy shape again
+    m.qpos0.shape = (1, ) + m.qpos0.shape[1:]
```
btaba: optional: you could create a new warp array here so you don't need L98

btaba:
> @btaba could you have a look at jax_test? I think it's a bit awkward but I'm not sure if there's a better option. We do have to fix the shapes afterwards again to avoid the asserts.. but at least we can then provide the correct values.
> 
> I still don't really like the approach but this is probably as good as it gets.

I agree, I think this is as good as it gets. I do think it's weird that you can modify the shape like that, I wonder if that will be supported in warp in the long run.

Issue #528, opened 2025-07-22
Hfield ncon balloons and should be culled further


btaba:
Currently prim<>hfield collisions are culled to mjMAXCONPAIR (50). In MJX, we culled to [4 contacts](https://github.com/google-deepmind/mujoco/blob/4f43a7d08e4f1289f4d76a4c48b8f42aca5bb38b/mjx/mujoco/mjx/_src/collision_convex.py#L1196), with no loss of generality (e.g. results in MuJoCo Playground). The 50 contacts quickly balloons memory over a large batch dim, and thus the contacts for prim<>hfield should be culled (ideally to a max of 4 contacts).

thowell:
should we make `MJ_MAXCONPAIR`
https://github.com/google-deepmind/mujoco_warp/blob/94d6928ebc51412cc4b1e58bd0815a023c3d0cc0/mujoco_warp/_src/types.py#L25
an option eg, `Option.maxconpair` and default the option value to 4?

btaba:
This hasn't been resolved?

btaba:
The tl;dr about why y'all might want to fix this before release is:

* max-contacts for a single geom <> hfield is set to 50
* these max numbers eventually get multiplied by num_envs O(10^3)
* this large constraint footprint for batch envs should be mitigated at some point
* we have some logic [here](https://github.com/google-deepmind/mujoco/blob/4b18009f47559a23e2c90b662769e0118b77c61e/mjx/mujoco/mjx/_src/collision_convex.py#L1202C18-L1207)  that fixes this, that could be replicated here

PR #539, opened 2025-07-24
add contact sensor

thowell:
initial implementation for contact sensor

mujoco documentation: https://mujoco.readthedocs.io/en/latest/XMLreference.html#sensor-contact
mujoco reference: https://github.com/google-deepmind/mujoco/blob/5eab59d02650eb1be97655de01b79469554a2802/src/engine/engine_sensor.c#L959

current features:
- matching: `geom1= geom2=`
- reduction: `mindist`
- data: all

### DIFF
--- a/mujoco_warp/_src/sensor.py
+++ b/mujoco_warp/_src/sensor.py
@@ -1565,6 +1565,7 @@ def _sensor_acc(
   sensor_acc_adr: wp.array(dtype=int),
   sensor_contact_adr: wp.array(dtype=int),
   # Data in:
+  njmax_in: int,
   ncon_in: wp.array(dtype=int),
   xpos_in: wp.array2d(dtype=wp.vec3),
   xipos_in: wp.array2d(dtype=wp.vec3),
@@ -1665,6 +1666,7 @@ def _sensor_acc(
       if force or torque:
         contact_forcetorque = support.contact_force_fn(
           opt_cone,
+          njmax_in,
           ncon_in,
           contact_frame_in,
           contact_friction_in,
@@ -2044,6 +2046,7 @@ def sensor_acc(m: Model, d: Data):
       m.sensor_cutoff,
       m.sensor_acc_adr,
       m.sensor_contact_adr,
+      d.njmax,
       d.ncon,
       d.xpos,
       d.xipos,


mujoco_warp/_src/sensor_test.py
```diff
@@ -405,6 +407,62 @@ def test_energy(self, xml):
...
+      </sensor>
+    </mujoco>
+    """
+
+    _, mjd, m, d = test_util.fixture(xml=_MJCF, keyframe=0)
+
+    d.sensordata.zero_()
+    mjwarp.forward(m, d)
+
+    _assert_eq(d.sensordata.numpy()[0], mjd.sensordata, "sensordata")
```
btaba: nice test as always! is it worth asserting that some of the sensordata isn't zero (so it's actually checking stuff)?

mujoco_warp/_src/io.py
```diff
@@ -115,6 +115,20 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
...
 
+  # contact sensor
+  is_contact_sensor = mjm.sensor_type == types.SensorType.CONTACT
+  if is_contact_sensor.any():
+    # matching
+    if (
+      (mjm.sensor_objtype[is_contact_sensor] != types.ObjType.GEOM)
+      | (mjm.sensor_reftype[is_contact_sensor] != types.ObjType.GEOM)
+    ).any():
+      raise NotImplementedError("Contact sensor matching not implemented.")
```
btaba: better to say something like "only geom-geom contact sensor mathcing is implemented"

mujoco_warp/_src/io.py
```diff
@@ -115,6 +115,20 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
...
+
+    # reduction
+    if (mjm.sensor_intprm[is_contact_sensor, 1] != 1).any():
+      raise NotImplementedError(f"Contact sensor reduction not implemented.")
```
btaba: same here, just note the one that's supported, it's easier for the user

mujoco_warp/_src/sensor.py
```diff
@@ -1588,7 +1601,139 @@ def _sensor_acc(
...
+
+    dataspec = sensor_intprm[sensorid, 0]
+    dim = sensor_dim[sensorid]
+    objtype = sensor_objtype[sensorid]
+    reftype = sensor_reftype[sensorid]
+    refid = sensor_refid[sensorid]
+    reduce = sensor_intprm[sensorid, 1]
+
+    # TODO(thowell): matching: none-none, geom1/geom2, site, body, subtree
+    assert (objtype == int(ObjType.GEOM.value)) and (reftype == int(ObjType.GEOM.value))
```
btaba: do these asserts have zero overhead? why is this needed here even though there's a check in io.py?

mujoco_warp/_src/sensor.py
```diff
@@ -1588,7 +1601,139 @@ def _sensor_acc(
...
+          tangent = True
+          size += 3
+
+    num = dim // size  # number of slots
+
+    adr = sensor_adr[sensorid]
+
+    # find match
+    slot = int(0)
+    for i in range(ncon):
```
btaba: I'm a bit confused, how is this not really slow? If I'm reading this correctly, every thread needs to loop through the entire ncon for all worlds (can be O(10^5) or more).

We might need a step before that sorts the contacts by worldid. If we get the start index of each world's contacts as well (after the sort), it could save significant work here for every `_sensor_acc` thread.

I think that would be an improvement, but I'm sure there's an even better design. We may also want to add the contact sensors to the benchmark envs, just so this kind of stuff gets flagged.

mujoco_warp/_src/sensor.py
```diff
@@ -1780,6 +1966,35 @@ def sensor_acc(m: Model, d: Data):
...
+        d.efc.force,
+      ],
+      outputs=[d.sensor_contact_id.reshape(-1), d.sensor_contact_dist.reshape(-1), d.sensor_contact_force],
+    )
+
+    # sort contacts by mindist
+    # TODO(thowell): maxforce reduction
+
+    # TODO(thowell): tile sort option
+    wp.utils.segmented_sort_pairs(
```
btaba: If `num="1"` is it really worth doing this sort? I was hoping we would avoid the overhead when `num="1"`. 

In the medium/long-run, we might need a few kernels to do some pre-work, grouped by contact sensor type

mujoco_warp/_src/sensor.py
btaba: (I would suggest starting to measure perf early in the PR, with one or two of the benchmark envs and a good keyframe)

mujoco_warp/_src/sensor_test.py
thowell: done

mujoco_warp/_src/io.py
thowell: updated

mujoco_warp/_src/io.py
thowell: updated

mujoco_warp/_src/sensor.py
thowell: updated:
- pre-sort contact ids: first sort by world, second sort per-world by reduction criteria
- per-world contact iterations (instead of iterating over contacts from all worlds)

thowell:
benchmarking:

humanoid (no sensors)
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=64 --event_trace=True
```

```

 Total JIT time: 0.25 s
 Total simulation time: 18.88 s
 Total steps per second: 433,963
 Total realtime factor: 2,169.82 x
 Total time per step: 2304.34 ns

Event trace:

step: 2299.61
    sensor_acc: 0.18
```

add sensors to humanoid
```
  <sensor>
    <contact geom1="foot1_right" geom2="floor" num="1" reduce="mindist" data="found"/>
    <contact geom1="foot2_right" geom2="floor" num="1" reduce="mindist" data="found"/>
    <contact geom1="foot1_left" geom2="floor" num="1" reduce="mindist" data="found"/>
    <contact geom1="foot2_left" geom2="floor" num="1" reduce="mindist" data="found"/>
  </sensor>
 ```

humanoid (4 contact sensors - previous commit (d86adf2 merged with main) loop over all contacts)
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=64 --event_trace=True
```

```

 Total JIT time: 0.30 s
 Total simulation time: 43.94 s
 Total steps per second: 186,449
 Total realtime factor: 932.24 x
 Total time per step: 5363.41 ns

Event trace:

step: 5358.44
    sensor_acc: 3137.80
  euler: 3.55
```

humanoid (4 contact sensors - current commit: sort contact ids, per-world loop over contacts)
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=64 --event_trace=True
```

```

 Total JIT time: 0.28 s
 Total simulation time: 25.59 s
 Total steps per second: 320,087
 Total realtime factor: 1,600.43 x
 Total time per step: 3124.15 ns

Event trace:

step: 3119.29
    sensor_acc: 858.22
```

**Comment by erikfrey** [comment]
@thowell we know the most common use case will probably be 1 contact per sensor right.  In that setting we shouldn't really need any kind of segmented sort.  I wonder if we should just do something that is O(num) like so:

```
for i in range(0, max(sensor_contact_num)):
  wp.launch(write_to_sensor_pos(i))
```

so linear in kernel launches per sensor contact num, first kernel finds the minimum dist that fits in pos 0, second kernel launch finds the remaining minimum dists that fit in pos1, etc.  You might need to use something like [wp.atomic_cas](https://nvidia.github.io/warp/modules/functions.html#warp.atomic_cas).

**Comment by thowell** [comment]
@btaba @erikfrey the implementation is updated, here are the latest testspeed results

this pr: humanoid + four contact sensors
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=64 --event_trace=True
```
```

 Total JIT time: 0.31 s
 Total simulation time: 19.16 s
 Total steps per second: 427,516
 Total realtime factor: 2,137.58 x
 Total time per step: 2339.09 ns

Event trace:

step: 2334.09
    sensor_acc: 37.41
```

main: humanoid no contact sensors
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=64 --event_trace=True
```
```

 Total JIT time: 0.28 s
 Total simulation time: 18.83 s
 Total steps per second: 435,137
 Total realtime factor: 2,175.68 x
 Total time per step: 2298.13 ns

Event trace:

step: 2293.06
    sensor_acc: 0.18
```

File: mujoco_warp/_src/sensor.py (hunk 6 above)
Comment (thowell) [reply mujoco_warp/_src/sensor.py]: the updated implementation has a case that skips sorting if the sensor has one match or less

File: mujoco_warp/_src/sensor.py (hunk 4 above)
Comment (thowell) [reply mujoco_warp/_src/sensor.py]: the asserts have been removed

**Comment by thowell** [comment]
the docstring for `Data` exceeds the doc line limit
```
mujoco_warp/_src/types.py:1510:93: W505 Doc line too long (98 > 92)
mujoco_warp/_src/types.py:1511:93: W505 Doc line too long (98 > 92)
mujoco_warp/_src/types.py:1512:93: W505 Doc line too long (98 > 92)
``` 

should we update the setting/s in pyproject.toml to accommodate more characters per line?

File: mujoco_warp/_src/sensor.py (hunk 5 above)
Comment (thowell) [reply mujoco_warp/_src/sensor.py]: updated:
- there is now a kernel that performs matching `_contact_match` and a kernel that performs sorting `_contact_sort`, followed by a case for `SensorType.CONTACT` in the `_sensor_acc` kernel that utilizes the matched and sorted contact ids to compute values for `sensordata`

File: mujoco_warp/_src/sensor.py [hunk 7]
```diff
@@ -1726,6 +1858,82 @@ def _sensor_touch(
...
+  # contact information
+  geom = contact_geom_in[contactid]
+
+  # geom-geom match
+  geom0geom1 = objid == geom[0] and refid == geom[1]
+  geom1geom0 = objid == geom[1] and refid == geom[0]
+  if geom0geom1 or geom1geom0:
+    worldid = contact_worldid_in[contactid]
+
+    contactmatchid = wp.atomic_add(sensor_contact_nmatch_out[worldid], contactsensorid, 1)
```
btaba, line 1899: nice!

mujoco_warp/_src/sensor.py
```diff
@@ -1726,6 +1858,82 @@ def _sensor_touch(
...
+):
+  worldid, contactsensorid = wp.tid()
+
+  nmatch = sensor_contact_nmatch_in[worldid, contactsensorid]
+
+  # skip sort
+  if nmatch <= 1:
+    return
+
+  criteria_tile = wp.tile_load(sensor_contact_criteria_in[worldid, contactsensorid], shape=MJ_MAXCONPAIR)
```
btaba, line 1933: what's the perf for nmatch>"1"? just curious how big the perf drop is (because we'll hit this code path quickly for manip). In the contact sensor design we discussed potentially not supporting num>"1" depending on perf, which would nix the sort in favor of a top-1, which may be faster than a sort over tile size 50.

Loading a tile of size 50 already seems like something we should optimize away after https://github.com/google-deepmind/mujoco_warp/issues/528

mujoco_warp/_src/sensor.py
btaba: LGTM thank you Taylor!

btaba:
Does the humanoid example cover the nmatch>1 case? I just want to make sure there isn't a hidden slow-down once we get to manip use-cases

thowell:
humanoid has plane<>capsule so there should potentially be two matches per plane<>capsule pair

testspeed defaults to keyframe 0:

image

some frames in to the simulation:

image

even more frames in to the simulation:

image

btaba:
> humanoid has plane<>capsule so there should potentially be two matches per plane<>capsule pair
> 
> testspeed defaults to keyframe 0:
>  ...

Ok interesting! So the tile_sort with tile size 50 has zero overhead? I find it a bit hard to believe but your screenshots do indicate `nmatch>1`. I would do a printf to be doubly sure, but either way I believe you that there are no more gremlins in the PR!

btaba (approved):

mujoco_warp/_src/sensor.py
btaba: oh sorry, this comment was stale and didn't get posted until I hit "submit review"

thowell:
@btaba

for humanoid the maximum number of contacts for any of the geom pairs is two, so to test the overhead from the tile sort using `MJ_MAXCONPAIR`:

```
  criteria_tile = wp.tile_load(sensor_contact_criteria_in[worldid, contactsensorid], shape=2)
  matchid_tile = wp.tile_load(sensor_contact_matchid_in[worldid, contactsensorid], shape=2)
```

```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=64 --event_trace=True
```

tile sort: 2
```

 Total JIT time: 5.51 s
 Total simulation time: 19.09 s
 Total steps per second: 429,101
 Total realtime factor: 2,145.50 x
 Total time per step: 2330.46 ns

Event trace:

step: 2325.34

    sensor_acc: 25.62
```

---

baseline

tile sort: MJ_MAXCONPAIR = 50 (**this pr**)
```

 Total JIT time: 5.42 s
 Total simulation time: 19.19 s
 Total steps per second: 426,999
 Total realtime factor: 2,134.99 x
 Total time per step: 2341.93 ns

Event trace:

step: 2337.10

    sensor_acc: 36.82
```

Issue #538, opened 2025-07-24
Commit `ca08923` causes large performance regression in sampling-based MPC application


alberthli:
The recently-merged commit (`ca08923`) that groups constraints per world for parallel line search causes large performance regressions in my downstream sampling-based MPC application.

Here is a video of the application in the previous commit (`e041fe0`).

https://github.com/user-attachments/assets/377ce632-137f-4071-96fc-160ee4a807b8

Now, here is a video when using the new commit.

https://github.com/user-attachments/assets/6064fe90-eeee-4d6e-9244-d03e2a5a0d19

These videos were taken back to back, with the only difference being the commit hash of `mujoco_warp`. Besides the visuals, the main thing you should be looking at is the plan time in the GUI.

We can see that the new version can be up to 10x slower in this application. I also noticed that the memory consumption is significantly higher. On the new commit, if I try to instantiate an environment with 10000 rollouts, I get the error
```
  File "/home/albert/judo/judo/rollout/mjw.py", line 64, in setup_mjwarp_backend
    self.mwd = mjw.put_data(
               ~~~~~~~~~~~~^
        self.mjm,
        ^^^^^^^^^
    ...<3 lines>...
        njmax=40000,
        ^^^^^^^^^^^^
    )  # TODO: expose these options as parameters in a rollout backend
    ^
  File "/home/albert/judo/.pixi/envs/dev/lib/python3.13/site-packages/mujoco_warp/_src/io.py", line 1341, in put_data
    jv=wp.empty(shape=(nworld, njmax), dtype=float),
       ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/albert/judo/.pixi/envs/dev/lib/python3.13/site-packages/warp/context.py", line 5334, in empty
    return warp.array(shape=shape, dtype=dtype, device=device, requires_grad=requires_grad, pinned=pinned, **kwargs)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/albert/judo/.pixi/envs/dev/lib/python3.13/site-packages/warp/types.py", line 1974, in __init__
    self._init_new(dtype, shape, strides, device, pinned)
[...]

PR #540, opened 2025-07-25
Small optimizations to solver kernels

Kenny-Vilella:
Second round of optimization. Number in the previous PR was given based on this branch.
I rerun all the benchmarks below to show the effect of this change only.
Mainly improves simple scene that are solver heavy.

Change:
- `update_constraint_init_qfrc_constraint` and `update_constraint_gauss_cost` have been modified to launch less threads, but each one having more work to do. It also eliminate the atomic.
- `update_gradient_JTDAJ` have also been optimized.

Branch        |  linesearch | Scene  | Real time factor |  Speedup |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
this branch  |  parallel   | aloha pot    |  514.76 x   |  No measurable difference
main            |  parallel   | aloha pot    |  459.35 x  | |
this branch  |  iterative  | aloha pot    |  1,341.47 x   |  No measurable difference
main            |  iterative  | aloha pot    |  1,316.76 x  | |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
this branch  |  parallel   | apollo        |  11,532.10 x  |  15% speedup
main            |  parallel   | apollo        |  10,015.33 x  | |
this branch  |  iterative  | apollo        |  7,327.56 x  |  17% speedup
main            |  iterative  | apollo        |  6,259.50 x  | |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
this branch  |  parallel   | kitchen       |  222.09 x  |  No measurable difference
main            |  parallel   | kitchen       |  223.47 x  | |
this branch  |  iterative  | kitchen       |  163.45 x  |   No measurable difference
main            |  iterative  | kitchen       |  162.70 x  | |
------------------- | ------------ | ------- | ---------------------- |  ------------------------------- | 
this branch  |  parallel   | humanoid   |  12,665.90 x  |  28% speedup
main            |  parallel   | humanoid   |  9,927.09 x  | |
this branch  |  iterative  | humanoid   |  8,864.60 x   |  22% speedup
main            |  iterative  | humanoid   |  7,285.12 x  | |

Command used to run the benchmarks:
```
python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/humanoid/humanoid.xml --batch_size=8192 --nconmax=200000 --njmax=52 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/apptronik_apollo/scene_flat.xml --batch_size=8192 --nconmax=150000 --njmax=45 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=kitchen_robot.xml --batch_size=2048  --nconmax=1600000 --njmax=2990 --is_sparse=True --ls_parallel=True

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/aloha_pot/scene.xml --keyframe=0 --batch_size=8192 --nconmax=500000 --njmax=80 --is_sparse=True --ls_parallel=False

python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/aloha_pot/scene.xml --keyframe=0 --batch_size=8192 --nconmax=2000000 --njmax=2000 --is_sparse=True --ls_parallel=True
```

Note that for the aloha pot benchmark values are different for parallel linesearch, this is because we generate some NaNs.


mujoco_warp/_src/solver.py
```diff
@@ -1947,20 +1941,26 @@ def update_gradient_JTDAJ(
 
   nefc = nefc_in[worldid]
 
-  dofi = dof_tri_row[elementid]
-  dofj = dof_tri_col[elementid]
+  dofi = (int(sqrt(float(1 + 8 * elementid))) - 1) // 2
+  dofj = elementid - (dofi * (dofi + 1)) // 2
```
Kenny-Vilella, line 1966: It will be great to have someone double check this.

mujoco_warp/_src/solver.py
```diff
@@ -1661,25 +1660,24 @@ def update_constraint_init_qfrc_constraint(
...
 
-  force = efc_force_in[worldid, efcid]
-  for i in range(nv):
-    wp.atomic_add(
-      qfrc_constraint_out[worldid],
-      i,
-      efc_J_in[worldid, efcid, i] * force,
-    )
+  sum_qfrc = float(0.0)
+  for efcid in range(nefc_in[worldid]):
```
adenzler-nvidia, line 1669: I think this pattern might actually be one of the reasons we are seeing some perf regressions like the one in #538 . Would make sense to check the changes in this PR with the benchmark there.

Maybe in the future we need to rethink this approach and need something smarter. But unless we have a huge regression, this can be a future thing.

mujoco_warp/_src/solver.py
```diff
@@ -1690,18 +1688,18 @@ def update_constraint_gauss_cost(
...
-  )
-  wp.atomic_add(efc_gauss_out, worldid, gauss_cost)
-  wp.atomic_add(efc_cost_out, worldid, gauss_cost)
+  gauss_cost = float(0.0)
+  for dofid in range(nv):
+    gauss_cost += (efc_Ma_in[worldid, dofid] - qfrc_smooth_in[worldid, dofid]) * (
+      qacc_in[worldid, dofid] - qacc_smooth_in[worldid, dofid]
+    )
+  efc_gauss_out[worldid] += 0.5 * gauss_cost
+  efc_cost_out[worldid] += 0.5 * gauss_cost
```
adenzler-nvidia: for the loop over nv, there is already a pattern I introduced in other kernels that will split to more threads once nv is large, at the cost of doing some atomic. Might be worth investigating here.

mujoco_warp/_src/solver.py
```diff
@@ -1947,20 +1941,26 @@ def update_gradient_JTDAJ(
...
+    active = efc_active_in[worldid, jj]
+    efc_Ji = efc_J_in[worldid, jj, dofi]
+    efc_Jj = efc_J_in[worldid, jj, dofj]
 
-    # TODO(team): sparse efc_J
-    value = efc_J_in[worldid, efcid, dofi] * efc_J_in[worldid, efcid, dofj] * efc_D
-    if value != 0.0:
-      wp.atomic_add(efc_h_out[worldid, dofi], dofj, value)
+  sum_h += efc_Ji * efc_Jj * efc_D * float(active)
+  efc_h_out[worldid, dofi, dofj] += sum_h
```
adenzler-nvidia, line 1984: is there a chance we could also fit the initialization of efc_h into the same kernel as well?

adenzler-nvidia:
let's double-check these changes against the benchmark in #538.

mujoco_warp/_src/solver.py
Kenny-Vilella: Yeah we should definitely be smarter.
It should not be difficult to create a small script to test this type of small kernels and find threshold where different approach is better.
It looks quite fun actually.

adenzler-nvidia: have a look at `def linesearch_jv_fused(nv: int, dofs_per_thread: int):` - I think we could potentially autotune the threshold there as well. I made quite some experiments and once you go to 2 threads, it makes sense to even things out a bit.

mujoco_warp/_src/solver.py
Kenny-Vilella: Yeap it seems to be slightly faster.
Changed it !

mujoco_warp/_src/solver.py
Kenny-Vilella: That is a good question.
After some preliminary investigations, it may be possible but not easy in the sparse case.
It really needs careful consideration and testing.
I added a todo and will investigate this after the PR is merged.

mujoco_warp/_src/solver.py
Kenny-Vilella: There is an added complexity that nefc is not constant across the world and can't be accessed easily from host side.
We could reason with njmax instead, but then there may be a decoupling between the chosen path and the actual optimized path for that conditions.

Keep investigating because there is probably a way do implement something similar.

adenzler-nvidia: Ah I missed this, yeah that makes it more complicated. Let's try to figure this out later then, I'm sure there are things we can do about it.

mujoco_warp/_src/solver.py
adenzler-nvidia: I'm not sure if the launch dim is correct now. Do I understand this correctly - this just gets from a flat index to a 2d index for the lowerl-left of the matrix including the diagonal, right? In that case the formula is correct.

If m.dof_tri_row.size contains the number of elements of the lower-left including diagonal, then that's correct. The code does suggest it, but the naming is non-intuitive.

mujoco_warp/_src/solver.py
Kenny-Vilella: I tried for the update_gradient_JTDAJ kernel and it did not went as expected.
My feeling is that we are too memory bound so spawning more thread is not helpful.
But not quite sure.

mujoco_warp/_src/solver.py
Kenny-Vilella: My understanding is that it is indeed the lowerl-left of the matrix including the diagonal.
It is true that I could change the launch dimension to something more obvious.
Will do it.

adenzler-nvidia (approved):
Looks good now, thanks!

PR #542, opened 2025-07-25
Fix testspeed and viewer cone flag bug.

erikfrey:
Testspeed and viewer were overriding cone type by default.

`aloha_pot` scene uses elliptic cone, and this override has been causing confusion in perf testing and debugging.


Kenny-Vilella:
Good catch !
With your fix, I can see a lot of NaNs generated for both parallel and iterative linesearch on the `aloha_pot` benchmark.
Probably, partly a tuning issue.

Kenny-Vilella (approved):

erikfrey:
Yep created #546 for NaN

PR #543, opened 2025-07-25
Collection of fixes for njmax overflows

adenzler-nvidia:
Now that debug runs properly and we have per-world njmax, there were 2 issues coming up where we overflow njmax. This should fix them


Kenny-Vilella (approved):
Looks good to me.

PR #545, opened 2025-07-25
Second attempt of airspeed velocity benchmarks.

erikfrey:
I decided I didn't like #522 for two reasons:

1) it's slow
2) it introduces a second way to benchmark when we already have `testspeed`

So I decided to redo it.  This new approach uses the same core code as `testspeed` by fixing the limitation in `testspeed` that `event_trace` param could not work with `graph_conditional` - this PR fixes that, and removes the recent change that forced `graph_conditional=False` when using event tracing.

After fixing this, I'm finding that using Warp events are pretty reliable.  I see only ~0.5% stddev from run to run, and the numbers all make sense to me.  Also, these tests run much faster - the entire benchmark suite runs in <2 mins on my machine, where the previous approach was probably closer to 10 minutes or more.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -8,7 +8,7 @@ Make sure you install MJWarp in develop so you can run the `asv` command:
 pip install -e .[dev,cuda]
 ```
 
-Run benchmarks like so at the top level of the `mujoco_warp` checkout:
+To execute benchmarks, from the `mujoco_warp` directory run:
 
 ```
 asv run
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,0 +1,60 @@
+# MuJoCo Warp Benchmark Suite
+
+MJWarp uses [airspeed velocity](https://github.com/airspeed-velocity/asv) for benchmarks.
+
+Make sure you install MJWarp in develop so you can run the `asv` command:
+
+```
+pip install -e .[dev,cuda]
+```
+
+Run benchmarks like so at the top level of the `mujoco_warp` checkout:
+
+```
+asv run
+```
+
+You should see output that looks like this:
+
+```
+Couldn't load asv.plugins._mamba_helpers because
+No module named 'libmambapy'
+· Creating environments
+· Discovering benchmarks
+· Running 6 total benchmarks (1 commits * 1 environments * 6 benchmarks)
+[ 0.00%] · For mujoco-warp commit 603429ca <asv-2>:
+[ 0.00%] ·· Benchmarking virtualenv-py3.12
+[16.67%] ··· Setting up benchmark:85                                                                                                                                   ok
+[16.67%] ··· benchmark.ApptronikApolloFlat.track_metric                                                                                                                ok
+[16.67%] ··· =============================================== =====================
+                                 function                                         
+             ----------------------------------------------- ---------------------
+                               jit_duration                   0.21160659193992615 
+                            solver_niter_mean                  3.263658447265625  
+                             solver_niter_p95                         5.0         
+                         device_memory_allocated                   887095296      
+                                   step                        758.2121635787189  
+                               step.forward                    753.6720001371577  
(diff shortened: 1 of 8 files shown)


mujoco_warp/_src/warp_util.py
```diff
@@ -97,6 +98,11 @@ def wrapper(*args, **kwargs):
     global _STACK
     if _STACK is None:
       return fn(*args, **kwargs)
+
+    for frame_info in inspect.stack():
+      if frame_info.function in ("capture_while", "capture_if"):
```
erikfrey, line 103: @adenzler-nvidia @Kenny-Vilella note here that I'm checking the call stack and not emitting `wp.Event` if I'm inside a `capture_while` or `capture_if`.  This approach is a bit brittle as if you guys ever change the API function names, this will break.

Let me know if there's a better way of asking "Am I inside a conditional graph"

benchmark/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+MJWarp uses [airspeed velocity](https://github.com/airspeed-velocity/asv) for benchmarks.
+
+Make sure you install MJWarp in develop so you can run the `asv` command:
+
+```
+pip install -e .[dev,cuda]
+```
+
+Run benchmarks like so at the top level of the `mujoco_warp` checkout:
```
Kenny-Vilella: Is this sentence grammatically correct?

benchmark/benchmark.py
```diff
@@ -0,0 +1,96 @@
...
+  njmax = 5
+
+
+class Humanoid(mujoco_warp.BenchmarkSuite):
+  """MuJoCo humanoid on an infinite plane."""
+
+  path = "humanoid/humanoid.xml"
+  params = mujoco_warp.BenchmarkSuite.params + ("step.euler",)
+  batch_size = 8192
+  nconmax = 150_000
```
Kenny-Vilella: On my computer, this is not enough and leads to explosion.
200k seems better.

benchmark/benchmark.py
```diff
@@ -0,0 +1,96 @@
...
+  njmax = 64
+
+
+class ThreeHumanoids(mujoco_warp.BenchmarkSuite):
+  """Three MuJoCo humanoids on an infinite plane.
+  Ideally, simulation time scales linearly with number of humanoids.
```
Kenny-Vilella, line 80: If the purpose of this is to look at the scaling, which is a great idea, then we should have some consistency between `Humanoid` and `ThreeHumanoids` benchmarks, that is, same batch size and njmax/nconmax x 3.

Kenny-Vilella (approved):
Everything looks good, just a few minor comments.
Now the benchmark time are consistent between asv and testspeed and also with and without event trace.
It's great !

One comment not directly related to this PR.
The three humanoids scene is surprisingly slow compared to the humanoid one.
I already noted some weirdness in the past.
I created an issue #550 to keep it in mind.

mujoco_warp/_src/warp_util.py
adenzler-nvidia: I think this approach is fine, I don't see those names changing anytime soon.

benchmark/benchmark.py
erikfrey: yeah agreed, it's just too slow with the really big batch size.  gotta fix the batch cholesky then we can raise this number.  added a TODO.

benchmark/benchmark.py
erikfrey: ok!

benchmark/README.md
erikfrey: Yes!  Although it's a bit of a [buffalo](https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo) sentence - cleaned it up.

thowell (approved):

kurdas111-commits:
Эх

PR #554, opened 2025-07-28
per-world contact

thowell:
inspired by the performance improvements from changing constraints from all-worlds to per-world #361, this pr investigates changing contacts from all-worlds to per-world.

tldr
performance seems to be about the same, but there might be other benefits from changing contacts to per-world.

**humanoid**

main
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=196608 --njmax=64
```
```

 Total JIT time: 0.25 s
 Total simulation time: 16.96 s
 Total steps per second: 483,045
 Total realtime factor: 2,415.22 x
 Total time per step: 2070.20 ns
```

this pr
```
mjwarp-testspeed --function=step --mjcf=test_data/humanoid/humanoid.xml --batch_size=8192 --nconmax=24 --njmax=64
```
```

 Total JIT time: 0.88 s
 Total simulation time: 16.88 s
 Total steps per second: 485,250
 Total realtime factor: 2,426.25 x
 Total time per step: 2060.79 ns
```

**aloha + pot**

main
```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=8192 --nconmax=131136 --njmax=96 --cone=pyramidal --keyframe=0
```
```

 Total JIT time: 0.65 s
 Total simulation time: 33.28 s
 Total steps per second: 246,178
 Total realtime factor: 492.36 x
 Total time per step: 4062.09 ns
```

this pr
```
mjwarp-testspeed --function=step --mjcf=benchmark/aloha_pot/scene.xml --batch_size=8192 --nconmax=16 --njmax=96 --cone=pyramidal --keyframe=0
```
```

 Total JIT time: 0.58 s
 Total simulation time: 33.28 s
 Total steps per second: 246,161
 Total realtime factor: 492.32 x
 Total time per step: 4062.38 ns
```

additional benefits from changing contacts to per-world:

- `njmax` and `nconmax` would share per-world semantics
- i think `njmax` and `nconmax` could be moved from `Data` to `Model` and utilize the existing fields in `mjModel`

there might be performance improvements for elliptic friction cones. once open issues for elliptic friction cones are resolved we can revisit potential performance improvements from this pr.

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -178,17 +178,16 @@ def ccd_kernel_builder(
     # Data out:
     ncon_out: wp.array(dtype=int),
     ncon_hfield_out: wp.array2d(dtype=int),
-    contact_dist_out: wp.array(dtype=float),
-    contact_pos_out: wp.array(dtype=wp.vec3),
-    contact_frame_out: wp.array(dtype=wp.mat33),
-    contact_includemargin_out: wp.array(dtype=float),
-    contact_friction_out: wp.array(dtype=vec5),
-    contact_solref_out: wp.array(dtype=wp.vec2),
-    contact_solreffriction_out: wp.array(dtype=wp.vec2),
-    contact_solimp_out: wp.array(dtype=vec5),
-    contact_dim_out: wp.array(dtype=int),
-    contact_geom_out: wp.array(dtype=wp.vec2i),
-    contact_worldid_out: wp.array(dtype=int),
+    contact_dist_out: wp.array2d(dtype=float),
+    contact_pos_out: wp.array2d(dtype=wp.vec3),
+    contact_frame_out: wp.array2d(dtype=wp.mat33),
+    contact_includemargin_out: wp.array2d(dtype=float),
+    contact_friction_out: wp.array2d(dtype=vec5),
+    contact_solref_out: wp.array2d(dtype=wp.vec2),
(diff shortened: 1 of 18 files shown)


adenzler-nvidia:
Thanks Taylor - interesting. As you already pointed out, I would expect this to enable some performance improvements for elliptic friction cones. In the contact generation there is likely little benefit from this separation. There might be some potential for data layout improvements in the constraint prep as well, but I think the potential is limited there.

I agree it makes sense to first get the functionality right for the elliptic cones though.

erikfrey:
OK interesting!  As mentioned in #361 I'm actually not sure we want per-world grouping as it challenges the design around heterogenous batches of simulation, and that we might go back at some point if we can figure out how to do it without sacrificing performance.

So let's keep this in our back pocket if we need it but I wouldn't be in a hurry to merge this without evidence of big speedups.

Separately we can adjust our API functions, e.g. we can make `make_data` and `put_data` accept per-world nconmax if it's more intuitive for the two params to be more aligned that way.

Kenny-Vilella:
That is very interesting.
One thing that I looked at in the past is further improving the iterative linesearch, but it was blocked due to the contact in the elliptic cone case.
This change may enable to combine the kernels, and we could even reevaluate conditional.
If parallel linesearch is finally not accurate enough, this could be a solid plan B to get better perf.

PR #555, opened 2025-07-29
Sdf collisions general case

jvonmuralt:
The PR adds sdf processing using wp.Volume:
1. wp.Volume is constructed from mujoco octree 
2. sdf and grad_sdf methods are implemented similarly to the logic in MuJoCo and integrated in the existing sdf kernel 
3. Tests for the equivalence of octree and wp.Volume as well as for the correct format and size of data are added
4. Cow-Torus example is ported including implementation of torus sdf plugin


mujoco_warp/test_data/collision_sdf/torus.obj
quagla, line 1: I think you shouldn't need this file? Mujoco will compute the mesh using marching cubes for you.

mujoco_warp/viewer.py
quagla, line 1: I think testspeed needs updating too.

mujoco_warp/_src/collision_sdf.py
```diff
@@ -109,68 +94,130 @@ def grad_ellipsoid(p: wp.vec3, size: wp.vec3) -> wp.vec3:
...
     return sphere(p, attr)
   elif type == int(GeomType.ELLIPSOID.value):
     return ellipsoid(p, attr)
   elif type == int(GeomType.SDF.value):
-    return user_sdf(p, attr, sdf_type)
+    if sdf_type == -1:
+      return sample_volume_sdf_world(p, volume_id)
+    else:
+     return user_sdf(p, attr, sdf_type)
   wp.printf("ERROR: SDF type not implemented\n")
```
quagla, line 282: If I now try to run the nutbolt.xml example I am hitting this error.

mujoco_warp/viewer.py
jvonmuralt: done

mujoco_warp/_src/collision_sdf.py
jvonmuralt: Yes, the plane implementation was missing and it complained because bolt is located on the plane, I added it now

mujoco_warp/test_data/collision_sdf/cow.xml
quagla, line 1: Can you move this scene to the `benchmark` folder?

mujoco_warp/_src/types.py
```diff
@@ -905,6 +905,9 @@ class Model:
     mesh_polymapadr: first polygon address per vertex        (nmeshvert,)
     mesh_polymapnum: number of polygons per vertex           (nmeshvert,)
     mesh_polymap: vertex to polygon map                      (nmeshpolymap,)
+    volume_ids: Warp volume IDs for SDF sampling             (nvolume,)
+    volumes: Warp volume objects containing mesh SDF data    (nvolume,)
```
quagla: I would remove "mesh" from the comment and remain agnostic about where the SDF comes from (to avoid confusion with actual mujoco meshes, which are always convex for collisions).

mujoco_warp/_src/sensor.py
```diff
@@ -1950,7 +1952,8 @@ def _sensor_tactile(
 
   # compute distance
   plugin_id = geom_plugin_index[geom]
-  depth = wp.min(sdf(int(GeomType.SDF.value), lpos, plugin_attr[plugin_id], plugin[plugin_id]), 0.0)
+  volume_data = make_volume_data(wp.uint64(0), wp.vec3(0.0), wp.vec3(0.0))
```
quagla: Can you please explain this? Will the sensor return the correct value when colliding with a Volume SDF? Will this slow down the sensor when colliding with a primitive SDF?

mujoco_warp/_src/io_test.py
```diff
@@ -69,6 +69,9 @@ def _check_type_matches_annotation(test_obj, obj: Any, prefix: str = ""):
 
   for field in dataclasses.fields(obj):
     field_name = field.name
+
+    if field_name == "volumes":
```
quagla: Maybe we should have a small test? Like asserting the size is correct?

mujoco_warp/_src/io.py
```diff
@@ -843,10 +846,85 @@ def create_nmodel_batched_array(mjm_array, dtype, expand_dim=True):
...
+                sdf_val = sample_octree_sdf(pos, oct_child, oct_aabb, oct_coeff)
+              else:
+                clamped_pos = np.clip(pos, original_mins, original_maxs)
+                sdf_val = sample_octree_sdf(clamped_pos, oct_child, oct_aabb, oct_coeff)
+              sdf_values[x, y, z] = sdf_val
+
+        background_value = 1.0
+        device = wp.get_device()
+        if device.is_cuda:
+          volume = wp.Volume.load_from_numpy(sdf_values, mins, voxel_size, background_value, device=device)
```
quagla: Out of curiosity: is there also the option to create an octree or is it always a dense voxel grid? I thought Volume could load from NanoVDB.

mujoco_warp/_src/collision_driver_test.py
```diff
@@ -507,6 +527,93 @@ def test_sdf_collision(self, fixture):
       result = check_dist
       np.testing.assert_equal(result, True, f"Contact {i} not found in Gjk results")
 
+  @pytest.mark.skipif(not wp.get_device().is_cuda, reason="SDF volumes require CUDA device")
+  def test_sdf_volumes(self):
```
quagla, line 529: Nice!

mujoco_warp/_src/collision_sdf.py
```diff
@@ -167,7 +176,66 @@ def user_sdf_grad(p: wp.vec3, attr: wp.vec3, sdf_type: int) -> wp.vec3:
 
 
 @wp.func
-def sdf(type: int, p: wp.vec3, attr: wp.vec3, sdf_type: int) -> float:
+def sample_volume_sdf_world(xyz: wp.vec3, volume_data: VolumeData) -> float:
```
quagla: I am missing something, why is it called world?

mujoco_warp/_src/io.py
jvonmuralt: From my understanding the backend of Volume is NanoVDB, so it is not storing the dense data even if we create it from numpy. I think it will run this on init https://github.com/NVIDIA/warp/blob/bf8c5f82ea337d056b3e84c2394f829a41669ca7/warp/utils.py#L1088

mujoco_warp/_src/collision_sdf.py
jvonmuralt: Yes, thats a bit misleading. I named it similarly to the wp volume functionality e.g. volume_world_to_index but since we are in the body frame, I will rename

mujoco_warp/_src/io.py
jvonmuralt: Here on slides 37-38 and onwards there are methods to load and create it https://peterchencyc.com/assets/pdf/3664475.3664543.pdf
Let me know if we should do it differently

quagla: Sounds good, we can always optimize later.

mujoco_warp/_src/collision_sdf.py
```diff
@@ -44,24 +44,36 @@ class AABB:
...
 
+@wp.struct
+class VolumeData:
+  volume_id: wp.uint64
+  center: wp.vec3
+  half_size: wp.vec3
+
+
+@wp.func
+def make_volume_data(volume_id: wp.uint64, center: wp.vec3, half_size: wp.vec3) -> VolumeData:
```
quagla: Can we maybe call it `get_volume_data`? I got confused by the name earlier and I thought you were allocating something at every step.

mujoco_warp/_src/sensor.py
```diff
@@ -1948,9 +1952,20 @@ def _sensor_tactile(
   tmp = xpos - geom_xpos_in[worldid, geom]
   lpos = wp.transpose(geom_xmat_in[worldid, geom]) @ tmp
 
-  # compute distance
   plugin_id = geom_plugin_index[geom]
-  depth = wp.min(sdf(int(GeomType.SDF.value), lpos, plugin_attr[plugin_id], plugin[plugin_id]), 0.0)
+  if plugin_id != -1:
```
quagla: I am confused by the logic, can we call `collision_sdf.sdf` directly instead of replicating the logic?

jvonmuralt: Thats a good point, I refactored a bit, made get_sdf_params function which can be called for a geom and it returns all the needed params for sdf

mujoco_warp/_src/io_test.py
jvonmuralt: added

quagla (approved):
To test that the tactile sensor works, you can load the model in `test_data/tactile.xml` and press `c` on your keyboard (or turn the contacts on in the rendering panel). If you also press `t` for making objects transparent you will see the contact area is colored.

mujoco_warp/_src/collision_driver_test.py
```diff
@@ -16,6 +16,7 @@
 
 import mujoco
 import numpy as np
+import pytest
```
thowell: we can remove this import (pytest is an optional dependency) and instead use `self.skipTest` (i think this is from the dependency absl-py) below. example: https://github.com/google-deepmind/mujoco_warp/blob/3f1197edd97a96a21384fcb418aa566306a22654/mujoco_warp/_src/jax_test.py#L38

mujoco_warp/_src/io.py
```diff
@@ -847,6 +851,83 @@ def create_nmodel_batched_array(mjm_array, dtype, expand_dim=True):
   return m
 
 
+def mujoco_octree_to_warp_volume(mjm):
```
thowell: since this is not part of the public api, maybe `_mujoco_octree_to_warp_volume`?

thowell: please add an output type

thowell: we can add the input type `mujoco.MjModel`

mujoco_warp/_src/io.py
```diff
@@ -1651,3 +1732,53 @@ def get_data_into(
 
   # sensors
   result.sensordata[:] = d.sensordata.numpy()
+
+
+def sample_octree_sdf(point, oct_child, oct_aabb, oct_coeff):
```
thowell: please add types for inputs and output

mujoco_warp/_src/io.py
```diff
@@ -1651,3 +1732,53 @@ def get_data_into(
...
+  eps = 1e-6
```
thowell: should this be a kwarg with a default value?

mujoco_warp/_src/io_test.py
```diff
@@ -21,6 +21,7 @@
 
 import mujoco
 import numpy as np
+import pytest
```
thowell: since pytest is an optional dependency, we should probably remove this import and below instead of `@pytest.mark.skipif` use `skipTest` from absl-py

mujoco_warp/test_data/collision_sdf/asset/spot.obj
thowell, line 1: please include this readme in the asset directory https://github.com/google-deepmind/mujoco/blob/main/model/plugin/sdf/asset/README.md

mujoco_warp/test_data/collision_sdf/asset/spot.png
thowell, line 1: see comment on spot.obj

mujoco_warp/test_data/collision_sdf/cow.xml
```diff
@@ -0,0 +1,53 @@
...
+      <geom type="sdf" name="cow1" mesh="spot" material="matspot"/>
+    </body>
+    <body pos="0.05 .25 2.2">
+      <freejoint/>
+      <geom type="sdf" name="cow2" mesh="spot" material="matspot"/>
+    </body>
+    <light name="left" pos="0 0 1"/>
+    <light name="right" pos="1 0 1"/>
+  </worldbody>
+</mujoco>
```
thowell: nit: please include a new line at the end of the file

mujoco_warp/test_data/collision_sdf/torus.py
```diff
@@ -0,0 +1,32 @@
+import math
+
+import warp as wp
+
+
+@wp.func
+def torus(p: wp.vec3, attr: wp.vec3):
```
thowell: please include a return type

mujoco_warp/test_data/collision_sdf/torus.py
```diff
@@ -0,0 +1,32 @@
...
+
+@wp.func
+def torus_sdf_grad(p: wp.vec3, attr: wp.vec3) -> wp.vec3:
+  grad = wp.vec3()
+  major_radius = attr[0]
+  minor_val = attr[1]
+
+  len_xy = math.sqrt(p[0] * p[0] + p[1] * p[1])
+  q = len_xy - major_radius
+  grad_q_x = p[0] / len_xy
```
thowell, line 25: is division by zero possible here (and/or the line below)?

mujoco_warp/test_data/collision_sdf/torus.py
```diff
@@ -0,0 +1,32 @@
...
+  grad_q_y = p[1] / len_xy
+  len_qz = math.sqrt(q * q + p[2] * p[2])
+
+  grad[0] = q * grad_q_x / wp.max(len_qz, minor_val)
```
thowell: is division by zero possible here when computing `grad`?

thowell: precompute the denominator?

thowell:
@jvonmuralt left some comments, thank you for this contribution!

mujoco_warp/test_data/collision_sdf/torus.py
jvonmuralt: yes, added clamping by a small value.

thowell (approved):

PR #557, opened 2025-07-29
update `Model` docstring

thowell:
update `Model` docstring to remove documentation for fields that are not part of the dataclass


adenzler-nvidia (approved):
Thanks for cleaning up!

PR #558, opened 2025-07-29
contact parameter mixing

thowell:
- add tests for contact parameter mixing
- update `collision_primitive.contact_params`
  - priority for `friction`


mujoco_warp/_src/collision_primitive.py
```diff
@@ -1554,26 +1554,32 @@ def contact_params(
...
+      mix = 0.0
+      condim = condim2
+      max_geom_friction = geom_friction[worldid, g2]
+    else:
+      mix = solmix1 / (solmix1 + solmix2)
+      mix = wp.where((solmix1 < MJ_MINVAL) and (solmix2 < MJ_MINVAL), 0.5, mix)
+      mix = wp.where((solmix1 < MJ_MINVAL) and (solmix2 >= MJ_MINVAL), 0.0, mix)
+      mix = wp.where((solmix1 >= MJ_MINVAL) and (solmix2 < MJ_MINVAL), 1.0, mix)
+      condim = wp.max(condim1, condim2)
+      max_geom_friction = wp.max(geom_friction[worldid, g1], geom_friction[worldid, g2])
```
adenzler-nvidia, line 1582: I'm wondering why you didn't go for the initial approach with wp.where here for friction as well, is this just for readability or did you see performance changes?

thowell: not sure about the reason for this approach--should we change this to utilize `wp.where`?

adenzler-nvidia: I think it's ok for now, likely same codegen and it's more readable like this.

adenzler-nvidia (approved):
Thanks, let's get this merged!

PR #566, opened 2025-07-31
Optimization of solve_beta for the CG solver

Kenny-Vilella:
Combined the three kernels for `solve_beta` and removed the corresponding properties.

Perf improvements is difficult to measure, but on Nsight system, I can see on the main branch that the three kernels took around 2% of the total time (10us per solver iteration), after the change it's only 0.7% (4us per solver iteration).

Command used:
`python mujoco_warp/testspeed.py --function=step --mjcf=benchmark/apptronik_apollo/scene_flat.xml --batch_size=8192 --nconmax=150000 --njmax=53 --is_sparse=True --ls_parallel=False --solver=cg`

I saw similar results for the humanoid.


adenzler-nvidia (approved):
Thanks - looks good. Might be a candidate to switch to multiple threads for high dof counts as well at some point in the future.

PR #570, opened 2025-08-05
Adding UT for the update_gradient function with the CG solver

Kenny-Vilella:
- Added a UT for testing the update_gradient function by comparing with results from Mujoco C.

### DIFF
--- a/mujoco_warp/_src/solver_test.py
+++ b/mujoco_warp/_src/solver_test.py
@@ -26,6 +26,7 @@ import mujoco_warp as mjwarp
 from . import solver
 from . import test_util
 from .types import ConeType
+from .types import DisableBit
 from .types import SolverType
 
 # tolerance for difference between MuJoCo and MJWarp solver calculations - mostly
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/solver_test.py
```diff
@@ -68,6 +69,37 @@ def cost(qacc):
...
+    (ConeType.ELLIPTIC, False),
+    (ConeType.PYRAMIDAL, True),
+    (ConeType.ELLIPTIC, True),
+  )
+  def test_update_gradient_CG(self, cone, sparse):
+    """Test _update_gradient function is correct for the CG solver."""
+    mjm, mjd, m, d = test_util.fixture(
+      "humanoid/humanoid.xml",
+      cone=cone,
+      solver=SolverType.CG,
```
Kenny-Vilella, line 82: As we do not have access to the corresponding mujoco C internal functions, we can only check for the CG solver.
For the record, I created a UT in mujoco C and could verify that results for the Newton solver is also correct.

We could eventually hardcode the results for the Newton solver and compare to this reference, but not sure we want to add this kind of hardcoded values.

thowell:
is it possible to modify the test to call `mjwarp.solve` instead of `mjwarp.forward`?

mujoco_warp/_src/solver_test.py
```diff
@@ -26,6 +26,7 @@
 from . import solver
 from . import test_util
 from .types import ConeType
+from .types import DisableBit
```
thowell: is this import utilized somewhere?

Kenny-Vilella: Nope, just a leftover.
Removed !

PR #571, opened 2025-08-05
update solver

### LINKED ISSUE #546: Physics instability with elliptic cone friction and aloha pot scene
Pyramidal friction works great, but I see this when I use elliptic cone:

Image

thowell:
updates for the engine solver

resolves #546

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -1048,6 +1048,8 @@ def linesearch_prepare_quad(
     for j in range(1, dim):
       # complete vector quadratic (for bottom zone)
       efcidj = contact_efc_address_in[conid, j]
+      if efcidj < 0:
+        return
       jvj = efc_jv_in[worldid, efcidj]
       jarefj = efc_Jaref_in[worldid, efcidj]
       dj = efc_D_in[worldid, efcidj]
@@ -1357,12 +1359,17 @@ def update_constraint_efc(
     mu = friction[0] / wp.sqrt(opt_impratio[worldid])
 
     efcid0 = contact_efc_address_in[conid, 0]
+    if efcid0 < 0:
+      return
+
     N = efc_Jaref_in[worldid, efcid0] * mu
 
     ufrictionj = float(0.0)
     TT = float(0.0)
     for j in range(1, dim):
       efcidj = contact_efc_address_in[conid, j]
+      if efcidj < 0:
+        return
       frictionj = friction[j - 1]
       uj = efc_Jaref_in[worldid, efcidj] * frictionj
       TT += uj * uj
--- a/mujoco_warp/_src/constraint.py
+++ b/mujoco_warp/_src/constraint.py
@@ -1162,6 +1162,7 @@ def _efc_contact_pyramidal(
     efcid = wp.atomic_add(nefc_out, worldid, 1)
 
     if efcid >= njmax_in:
+      contact_efc_address_out[conid, dimid] = -1
       return
 
     timestep = opt_timestep[worldid]
@@ -1325,6 +1326,7 @@ def _efc_contact_elliptic(
     efcid = wp.atomic_add(nefc_out, worldid, 1)
 
     if efcid >= njmax_in:
+      contact_efc_address_out[conid, dimid] = -1
       return
 
     timestep = opt_timestep[worldid]


Kenny-Vilella:
For the moment only tested the change and not looked at the code in details.

I still see some explosions with the `aloha_pot` benchmark.
This is the command used:
```
python mujoco_warp/testspeed.py --function=step benchmark/aloha_pot/scene.xml --keyframe=0 --nworld=8192 --nstep=1000 --njmax=1000 --nconmax=500000 -o opt.is_sparse=True -o opt.ls_parallel=False/True
```
Results: 5/8192 explosions for iterative linesearch, 2540/8192 explosions for parallel linesearch.

We can also make the scene explodes on the viewer:
[mujoco_elliptic_bug.webm](https://github.com/user-attachments/assets/96e190bd-4534-433d-b2dc-70d46fa4b5b9)

My first impression is that the current change does improve the situation.
But the issue mentioned with `efc_address` in  #546  is still not fully solved.

EDIT: The humanoid benchmark also explodes with parallel linesearch, while it does not explode with iterative linesearch but perf significantly decrease (6,736.23x vs 3,598.41x)

mujoco_warp/_src/solver.py
```diff
@@ -1967,23 +1658,28 @@ def update_gradient_JTDAJ(
 
   sum_h = float(0.0)
   efc_D = efc_D_in[worldid, 0]
-  active = efc_active_in[worldid, 0]
   efc_Ji = efc_J_in[worldid, 0, dofi]
   efc_Jj = efc_J_in[worldid, 0, dofj]
+  efc_state = efc_state_in[worldid, 0]
   for efcid in range(nefc - 1):
+    if efc_D == 0.0 or efc_state != int(types.ConstraintState.QUADRATIC.value):
```
Kenny-Vilella: This should be changed.
With this implementation, as soon as we hit these conditions all the following iterations would be skipped.

I think what we want is to guard:
`sum_h += efc_Ji * efc_Jj * efc_D`
with:
`if efc_state == int(types.ConstraintState.QUADRATIC.value)`

And on line 1677, it should be `==` and not `!=` as only quadratic constraints are active?

mujoco_warp/_src/solver.py
```diff
@@ -2028,37 +1724,41 @@ def update_gradient_JTCJ(
...
-    tt = efc_uu_in[conid]
-    if tt <= 0.0:
-      t = 0.0
-    else:
-      t = wp.sqrt(tt)
-
-    middle_zone = (t > 0) and (n < (mu * t)) and ((mu * n + t) > 0.0)
-
-    if not middle_zone:
+    efcid0 = contact_efc_address_in[conid, 0]
```
Kenny-Vilella, line 1744: I am not sure about this one.
Are we sure at that point that this is actually an active contact?

mujoco_warp/_src/solver.py
```diff
@@ -39,68 +41,142 @@ def _in_bracket(x: wp.vec3, y: wp.vec3) -> bool:
...
+  ne = ne_in[worldid]
+  nf = nf_in[worldid]
+
+  # equality
+  if efcid < ne:
+    wp.atomic_add(out, worldid, _eval_pt(efc_quad_in[worldid, efcid], alpha))
+  # friction
+  elif efcid < ne + nf:
+    # search point, friction loss, bound (rf)
+    start = efc_Jaref_in[worldid, efcid]
```
Kenny-Vilella, line 91: Nitpick: We could eventually put this into a function to share between iterative and parallel linesearch.
But you could choose what you think is better.

mujoco_warp/_src/solver.py
```diff
@@ -1351,288 +1279,119 @@ def update_constraint_init_cost(
...
-  efc_active_out: wp.array2d(dtype=bool),
+  efc_state_out: wp.array2d(dtype=int),
 ):
   worldid, efcid = wp.tid()
 
   if efcid >= nefc_in[worldid]:
     return
 
-  if efc_done_in[worldid]:
-    return
```
Kenny-Vilella: Just wonder why you remove this?

mujoco_warp/_src/solver.py
```diff
@@ -1351,288 +1279,119 @@ def update_constraint_init_cost(
...
+    elif Jaref >= rf:
+      efc_force_out[worldid, efcid] = -f
+      efc_state_out[worldid, efcid] = int(types.ConstraintState.LINEARPOS.value)
+      wp.atomic_add(efc_cost_out, worldid, -f * (0.5 * rf - Jaref))
     else:
+      efc_force_out[worldid, efcid] = -efc_D * Jaref
+      efc_state_out[worldid, efcid] = int(types.ConstraintState.QUADRATIC.value)
+      wp.atomic_add(efc_cost_out, worldid, 0.5 * efc_D * Jaref * Jaref)
+  elif efc_type_in[worldid, efcid] != int(types.ConstraintType.CONTACT_ELLIPTIC.value):
+    # limit, contact
```
Kenny-Vilella: Nitpick: Maybe mentioned that it is pyramidal contact

mujoco_warp/_src/solver.py
```diff
@@ -1351,288 +1279,119 @@ def update_constraint_init_cost(
...
       efc_force_out[worldid, efcid] = 0.0
-      efc_active_out[worldid, efcid] = False
-      return
+      efc_state_out[worldid, efcid] = int(types.ConstraintState.SATISFIED.value)
+    else:
+      efc_force_out[worldid, efcid] = -efc_D * Jaref
+      efc_state_out[worldid, efcid] = int(types.ConstraintState.QUADRATIC.value)
+      wp.atomic_add(efc_cost_out, worldid, 0.5 * efc_D * Jaref * Jaref)
   else:
-    # contact
```
Kenny-Vilella, line 1564: Nitpick: and here elliptic contact for consistency

mujoco_warp/_src/solver.py
```diff
@@ -1351,288 +1279,119 @@ def update_constraint_init_cost(
...
-  condim = contact_dim_in[conid]
-
-  if condim == 1 or dimid >= condim:
-    return
+    if conid >= ncon_in[0]:
+      return
 
-  friction = contact_friction_in[conid]
-  efcid = contact_efc_address_in[conid, dimid]
+    dim = contact_dim_in[conid]
```
Kenny-Vilella, line 1357: Should we return if `condim =1` ?

mujoco_warp/_src/solver.py
```diff
@@ -1351,288 +1279,119 @@ def update_constraint_init_cost(
...
-    mu2 = mu * mu
-    dm = efc_D_in[worldid, efcid0] / wp.max(mu2 * float(1.0 + mu2), types.MJ_MINVAL)
+      force = -dm * nmt * mu
 
-    nmt = n - mu * t
+      if efcid == efcid0:
+        efc_force_out[worldid, efcid] = force
+        wp.atomic_add(efc_cost_out, worldid, 0.5 * dm * nmt * nmt)
+      else:
+        efc_force_out[worldid, efcid] = -math.safe_div(force, T) * ufrictionj
```
Kenny-Vilella, line 1404: In the previous version, there was a `+=` instead of a `=`.
It was a bug right?

thowell:
@Kenny-Vilella updated `update_gradient_JTDAJ`

not sure that the aloha + pot scene will simulate well with parallel linesearch and the default `iterations` and `ls_iterations` settings

mujoco_warp/_src/solver.py
thowell: the logic in this kernel has been updated. thanks!

mujoco_warp/_src/solver.py
thowell: it looks like `_efc_contact_elliptic` checks for active contacts
https://github.com/google-deepmind/mujoco_warp/blob/3bbd1f7ff8a44eb2139277305523ff0f5240b675/mujoco_warp/_src/constraint.py#L1311
and then checks
https://github.com/google-deepmind/mujoco_warp/blob/3bbd1f7ff8a44eb2139277305523ff0f5240b675/mujoco_warp/_src/constraint.py#L1327

before setting `contact_efc_address`
https://github.com/google-deepmind/mujoco_warp/blob/3bbd1f7ff8a44eb2139277305523ff0f5240b675/mujoco_warp/_src/constraint.py#L1332

---

it looks like there is a check for `conid` above. is there a scenario we can think of where the efc address from `contact_efc_address` is incorrect?

Kenny-Vilella:
@thowell What scene would you suggest to test the elliptic cone?

mujoco_warp/_src/solver.py
thowell: it makes sense to share the `_eval` functionality. added a todo for `_eval` to have an option that only computes cost.

mujoco_warp/_src/solver.py
thowell: restored

mujoco_warp/_src/solver.py
thowell: not sure about the previous version

mujoco_warp/_src/solver.py
Kenny-Vilella: Yes the problem is that if active is false here:

https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/constraint.py#L1323

then `efc_address` is never set here:

https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/constraint.py#L1332C5-L1332C28

So that means we do not know what is stored in `efc_address` for this specific conid.
As long as we iterate over the efc row, we are good and all the corresponding contact are active, but if we iterate over the contact, we do not know if this specific contact has a constraint row.

Or there is something I am missing?

mujoco_warp/_src/solver.py
thowell: if `condim=1`, the `efc_type` set in constraints.py will be `CONTACT_FRICTIONLESS` and this case will not be evaluated

mujoco_warp/_src/solver.py
thowell: comment is updated

mujoco_warp/_src/solver.py
thowell: comment is updated

mujoco_warp/_src/solver.py
thowell: ok makes sense, good catch! added a check for the contact status in `update_gradient_JTCJ`.

thowell:
scenarios with sliding friction -- locomotion or grasping an object could be good scenarios. maybe we can add keyframes to existing benchmark scenes for such scenarios.

mujoco_warp/_src/solver_test.py
```diff
@@ -128,10 +128,10 @@ def test_parallel_linesearch(self, cone):
   @parameterized.parameters(
     (ConeType.PYRAMIDAL, SolverType.CG, 5, 5, False, False),
     (ConeType.ELLIPTIC, SolverType.CG, 5, 5, False, False),
-    (ConeType.PYRAMIDAL, SolverType.NEWTON, 2, 4, False, False),
-    (ConeType.ELLIPTIC, SolverType.NEWTON, 2, 5, False, False),
-    (ConeType.PYRAMIDAL, SolverType.NEWTON, 2, 4, True, True),
-    (ConeType.ELLIPTIC, SolverType.NEWTON, 3, 16, True, True),
+    (ConeType.PYRAMIDAL, SolverType.NEWTON, 5, 10, False, False),
```
btaba, line 131: I suppose there isn't a test that was failing before this change vs after that has been added, and things were mainly checked as "WAI" on the aloha pot scene?

Kenny-Vilella:
Made a second round of review.
Spotted a few differences compared to previous implementation but all of them seems to be bugs in the previous implementation.
So we are clearly in a much better position than before.

I still got some crash when doing experiment. compute-sanitizer indicates some illegal read in `update_constraint_efc`.
After investigation there is an edge case where `efcid0 + condim > njmax`, that means efcid0 get written but some constraints of the contact are skipped.
Later on in `update_constraint_efc`, when we try to access `efcid0 + condim` we get an out of bound memory read.
But it is unrelated to your PR, so will create a different PR to address this.

There is a slight decrease in performance, but we could deal with this later.

Will still did some tests today and tomorrow, but it looks good.

mujoco_warp/_src/solver.py
```diff
@@ -134,119 +210,51 @@ def linesearch_iterative_init_gtol_p0_gauss(
...
-    opt_impratio[worldid],
-    contact_friction_in[conid],
-    efc_u_in[conid][0],
-    efc_uu_in[conid],
-    efc_uv_in[conid],
-    efc_vv_in[conid],
-    efc_jv_in[worldid, efcid],
-    efc_D_in[worldid, efcid],
-    efc_quad_in[worldid, efcid],
+  if efcid > nefc_in[worldid]:
```
Kenny-Vilella: This should be:
`efcid >= nefc_in[worldid]`

thowell: fixed

btaba (approved):
LGTM'ing for Kenny

thowell:
@Kenny-Vilella 
- contact constraints are updated to write `-1` into `contact.efc_address` if `efcid >= njmax`
- in cases that loop over contact efc variables, the solver checks that the efc address is `>0`.

Kenny-Vilella:
Made a few more tests, except CG + parallel linesearch that will be addressed in the other PR, everything works fine and is stable.

Great work !

mujoco_warp/_src/solver_test.py
thowell: the aloha + pot scene was useful for checking that things were or were not working as intended.

the tests are not failing on main. however, the updates to `constraints.xml` in #572 will lead to test failures, even if the test settings are changed to increase `iterations` and `ls_iterations`:
```
    (ConeType.PYRAMIDAL, SolverType.CG, 20, 50, False, False),
    (ConeType.ELLIPTIC, SolverType.CG, 20, 50, False, False),
    (ConeType.PYRAMIDAL, SolverType.NEWTON, 20, 50, False, False),
    (ConeType.ELLIPTIC, SolverType.NEWTON, 20, 50, False, False),
    (ConeType.PYRAMIDAL, SolverType.NEWTON, 20, 50, True, True),
    (ConeType.ELLIPTIC, SolverType.NEWTON, 20, 50, True, True),
 ```
 
 with the updates in this pr and the changes to `constraints.xml` in #572 the tests will pass with the settings above.

PR #573, opened 2025-08-05
update derivative.py

thowell:


mujoco_warp/_src/forward.py
```diff
@@ -358,7 +350,7 @@ def _euler_sparse(m: Model, d: Data):
     d.qLD_integration,
     d.qLDiagInv_integration,
     d.qacc_integration,
-    d.qfrc_integration,
+    d.efc.Ma,
```
btaba, line 353: Ma is equivalent to qfrc_smooth + qfrc_constraint, is that right?

thowell: correct. when `grad` is zero `Ma = qfrc_smooth + qfrc_constraint`

https://github.com/google-deepmind/mujoco_warp/blob/7a7bae6c1390df3f29058e90f25f9ed5caab3fb1/mujoco_warp/_src/solver.py#L1587

PR #575, opened 2025-08-05
parallel linesearch with log scale for step sizes

### LINKED ISSUE #565: Combination of CG solver and parallel linesearch is highly unstable
It seems to be a general behavior, but for instance if launching the humanoid on the viewer it directly explodes.
If changing the solver or using the iterative solver, then it behaves normally.

Command:
```
python mujoco_warp/viewer.py --mjcf=benchmark/humanoid/humanoid.xml --ls_parallel=True --njmax=59 --solver=cg
```

I modified the viewer to propagate the option as it is done for testspeed.

thowell:
log scale for step sizes

for reference: https://github.com/google-deepmind/mujoco_mpc/blob/ff572a21e7c2bf9fda62e1862a758da7e9a8719b/mjpc/utilities.cc#L819

resolves #565
```
mjwarp-viewer mujoco_warp/test_data/humanoid/humanoid.xml -o "opt.ls_parallel=True" -o "opt.solver=cg" -o "opt.iterations=64" -o "opt.ls_iterations=128"
```


mujoco_warp/_src/solver.py
```diff
@@ -893,7 +899,8 @@ def linesearch_parallel_fused(
 
   efc_quad_total_candidate = efc_quad_gauss_in[worldid]
 
-  alpha = float(alphaid) / float(nlsp - 1)
+  alpha = _log_scale(1.0e-6, 1.0, nlsp, alphaid)
```
yuvaltassa: This 1e-6 is a new, rather important constant. Can it be promoted somehow? Given more prominence?

Also I suspect it's too small, but that should really be tested empirically, we should get statistics of the winning alpha, I expect a much larger min value like 1e-3 or 1e-4 makes more sense

thowell: `min_step` is added to `Option` and a todo is added to determine a default setting for this value.

Kenny-Vilella:
I did some tests, before this change nothing is stable for the CG solver and parallel linesearch.

With this change, everything is stable for the CG solver except the `aloha_pot` benchmark (for both pyramidal and elliptic cone friction).
This change also improves stability of the Newton solver on the `aloha_pot` benchmark .

So it's a clear improvement although it is not yet working for all scenes.

Kenny-Vilella:
@thowell  In the `test_parallel_linesearch` UT, the case with elliptic cone is deactivated because not working.
With your change, I think that we can enabled it.
If it is too close to the tolerance, then we just need to increase the number of linesearch iteration.

mujoco_warp/_src/io.py
```diff
@@ -471,6 +471,7 @@ def create_nmodel_batched_array(mjm_array, dtype, expand_dim=True):
       sdf_initpoints=mjm.opt.sdf_initpoints,
       sdf_iterations=mjm.opt.sdf_iterations,
       run_collision_detection=True,
+      min_step=1.0e-6,  # TODO(team): determine good default setting
```
Kenny-Vilella: Maybe we should call it `ls_min_step`?

thowell:
@Kenny-Vilella updated `test_parallel_linesearch` to test the elliptic friction cone case

Kenny-Vilella (approved):
As mentioned in the previous comment, it's good on my side, it provides a clear improvement.

I also tried to play with the value of ls_min_step, and the value you choose seems to be good.

mujoco_warp/_src/types.py
```diff
@@ -598,6 +599,7 @@ class Option:
   sdf_initpoints: int
   sdf_iterations: int
   run_collision_detection: bool  # warp only
+  ls_min_step: float  # warp only
```
erikfrey: is this only used in parallel linesearch?  if so, should we call it `ls_parallel_min_step` ?

also, if it is only for parallel linesearch, let's move it underneath `ls_parallel`?  also, might want to annotate that `ls_parallel` is warp only

PR #579, opened 2025-08-06
Change the license field in pyproject.toml to SPDX string.

saran-t:
According to the deprecation warning, support for the old way of declaring the license will be removed from setuptools on 2026-Feb-18.


saran-t:
Replaced by https://github.com/google-deepmind/mujoco_warp/pull/580

PR #581, opened 2025-08-06
Fix JAX test with warp graph capture. Fix #319

### LINKED ISSUE #319: JAX supports conditionals in Warp CUDA graphs
#242 introduced conditional graphs for the constraint solver.

Although JAX and Warp both support conditionals, mixing them into the same CUDA graph is causing errors.  We should be able to make it such that a Warp conditional can be embedded into JAX.

btaba:


btaba:
https://github.com/google-deepmind/mujoco_warp/issues/319

thowell (approved):

PR #585, opened 2025-08-07
add step1 and step2

thowell:
mujoco reference https://github.com/google-deepmind/mujoco/blob/840108cf7cbacc6d16d47fc471f4679d214d1525/src/engine/engine_forward.c#L1223

### DIFF
--- a/mujoco_warp/_src/forward_test.py
+++ b/mujoco_warp/_src/forward_test.py
@@ -279,6 +279,126 @@ class ForwardTest(parameterized.TestCase):
 
       _assert_eq(d.actuator_force.numpy()[0], mjd.actuator_force, "actuator_force")
 
+  @parameterized.parameters(("humanoid/humanoid.xml", True), ("humanoid/humanoid.xml", False))
+  def test_step1(self, xml, energy):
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/forward_test.py
```diff
@@ -279,6 +279,126 @@ def test_tendon_actuator_force_limits(self):
...
 
+  @parameterized.parameters(("humanoid/humanoid.xml", True), ("humanoid/humanoid.xml", False))
+  def test_step1(self, xml, energy):
+    mjm, mjd, m, d = test_util.fixture(xml, kick=True, energy=energy)
+
+    # some of the fields updated by step1
+    step1_field = [
+      "xpos",
+      "xquat",
+      # TODO(team): "xmat"
```
btaba: Why TODO?  It's not clear why these fields weren't just added here to begin with

btaba (approved):
Left one comment, otherwise LGTM

mujoco_warp/_src/forward_test.py
thowell: these fields require additional processing before comparison. 

the pr is updated to add additional processing in order to test more fields. additionally, a todo is added to test more mjcfs.

PR #586, opened 2025-08-07
reset_data

thowell:
`_resetData` mujoco reference
https://github.com/google-deepmind/mujoco/blob/840108cf7cbacc6d16d47fc471f4679d214d1525/src/engine/engine_io.c#L1903

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -1648,7 +1648,6 @@ def _reset_nworld(
   # Data out:
   solver_niter_out: wp.array(dtype=int),
   ncon_out: wp.array(dtype=int),
-  ncon_world_out: wp.array(dtype=int),
   ncon_hfield_out: wp.array2d(dtype=int),
   ne_out: wp.array(dtype=int),
   ne_connect_out: wp.array(dtype=int),
@@ -1677,7 +1676,6 @@ def _reset_nworld(
   solver_niter_out[worldid] = 0
   if worldid == 0:
     ncon_out[0] = 0
-  ncon_world_out[worldid] = 0
   for i in range(nhfieldgeom):
     ncon_hfield_out[worldid, i] = 0
   ne_out[worldid] = 0
@@ -1807,7 +1805,6 @@ def reset_data(m: types.Model, d: types.Data):
     outputs=[
       d.solver_niter,
       d.ncon,
-      d.ncon_world,
       d.ncon_hfield,
       d.ne,
       d.ne_connect,


mujoco_warp/_src/io.py
```diff
@@ -407,6 +407,9 @@ def create_nmodel_batched_array(mjm_array, dtype, expand_dim=True):
   condim = np.concatenate((mjm.geom_condim, mjm.pair_dim))
   condim_max = np.max(condim) if len(condim) > 0 else 0
 
+  # height field
+  nhfieldgeom, geomgeom2hfgeom = _hfield_geom_pair(mjm)
```
btaba: geompair2hfgeom?

mujoco_warp/_src/io.py
```diff
@@ -1649,3 +1653,200 @@ def get_data_into(
...
+  contact_solreffriction_out: wp.array(dtype=wp.vec2),
+  contact_solimp_out: wp.array(dtype=types.vec5),
+  contact_dim_out: wp.array(dtype=int),
+  contact_geom_out: wp.array(dtype=wp.vec2i),
+  contact_efc_address_out: wp.array2d(dtype=int),
+  contact_worldid_out: wp.array(dtype=int),
+):
+  conid = wp.tid()
+
+  # TODO(team): only zero if conid <= ncon[0]?
```
btaba: can this be resolved in this PR? don't want to merge in out of bound writes

mujoco_warp/_src/io.py
```diff
@@ -1649,3 +1653,200 @@ def get_data_into(
...
+  d.sensordata.zero_()
+  d.qM.zero_()
+
+  # set mocap_pos/quat = body_pos/quat for mocap bodies
+  wp.launch(
+    _set_mocap, dim=(d.nworld, m.nbody), inputs=[m.body_mocapid, m.body_pos, m.body_quat], outputs=[d.mocap_pos, d.mocap_quat]
+  )
+
+  # clear contacts
+  wp.launch(
```
btaba, line 1781: OOC is there any benefit to separate kernel launches vs one that zeros everything in one go?

btaba (changes requested):
LGTM, just one comment about a TODO

mujoco_warp/_src/io.py
thowell: there is overhead for each `zero_()` https://github.com/google-deepmind/mujoco_warp/pull/452#issue-3199735355

mujoco_warp/_src/io.py

mujoco_warp/_src/io.py
thowell: added early return to this kernel

mujoco_warp/_src/io.py
```diff
@@ -1649,3 +1653,202 @@ def get_data_into(
...
+  contact_geom_out[conid] = wp.vec2i(0, 0)
+  for i in range(nefcaddress):
+    contact_efc_address_out[conid, i] = 0
+  contact_worldid_out[conid] = 0
+
+
+def reset_data(m: types.Model, d: types.Data):
+  """Clear data, set defaults."""
+  # copy qpos0 from model
+  wp.copy(d.qpos, m.qpos0)
```
erikfrey: i guess because of our stride 0 tricks this works just fine even with different `nworld`?

it probably does... but i would be a little more comfortable if the `test_reset_data` function below created test fixutres with `nworld` > 1 just to be sure

mujoco_warp/_src/io.py
```diff
@@ -1649,3 +1653,202 @@ def get_data_into(
...
+
+  # # zero out arrays that are not affected by forward
+  wp.launch(_zero_nv, dim=(d.nworld, m.nv), outputs=[d.qvel, d.qacc_warmstart, d.qfrc_applied, d.qacc])
+  wp.launch(_zero_nu, dim=(d.nworld, m.nu), inputs=[m.na], outputs=[d.act, d.ctrl, d.act_dot])
+  wp.copy(d.eq_active, m.eq_active0)
```
erikfrey: hmmm... same question here, in `make_data` we explicitly tile this, so it repeats across all worlds... note that this field is NOT batched at all in `mujoco_warp` Model

are we sure this line is doing the right thing?

erikfrey (commented):
Couple implementation questions but this generally looks on the right track.

If we like this code, I wonder if we should change `make_data` so it creates all arrays with `empty` and then we just call `reset_data` at the end?  Might be both more useful (e.g. `make_data` is not populating `qpos` from `qpos0`) and more performant.

mujoco_warp/_src/io.py
thowell: updated to set `qpos` in the `_set_nworld` kernel (`qpos[worldid, i] = qpos0[worldid, i]`) and `test_reset_data` now sets `nworld=2` for testing.

mujoco_warp/_src/io.py
thowell: updated to set this field in `_set_nworld`: `eq_active[worldid, i] = eq_active0[i]`

mujoco_warp/_src/io.py
```diff
@@ -1715,6 +1719,209 @@ def get_data_into(
...
+      d.contact.solimp,
+      d.contact.dim,
+      d.contact.geom,
+      d.contact.efc_address,
+      d.contact.worldid,
+    ],
+  )
+
+  wp.launch(
+    _set_nworld,
```
erikfrey: should this be called `_zero_nworld` to match `_zero_contact` ?

mujoco_warp/_src/io.py
```diff
@@ -1715,6 +1719,209 @@ def get_data_into(
   result.sensordata[:] = d.sensordata.numpy()
 
 
+@wp.kernel
```
erikfrey, line 1634: Should we use these kernels in `make_data` and use `wp.empty` instead of `wp.zero` in them?

erikfrey (commented):
Looks good, just one question and one nit

mujoco_warp/_src/io.py
thowell: updated kernels names by replacing `_set` or `_zero` with `_reset`.

thowell: updated kernel names: `_set`/`_zero` -> `_reset`

mujoco_warp/_src/io.py
thowell: not sure that `reset_data` can currently be called in `make_data` since `reset_data` has a `Model` input whereas `make_data` has a `mujoco.MjModel` input.

erikfrey: Ah OK fair, let's worry about that later then.

mujoco_warp/_src/io.py
```diff
@@ -1627,3 +1631,261 @@ def get_data_into(
...
+      d.qfrc_applied,
+      d.eq_active,
+      d.qacc,
+      d.act_dot,
+      d.sensordata,
+    ],
+  )
+
+
+def sample_octree_sdf(
```
erikfrey: hmmm, some kind of out of date merge?

thowell: fixed

thowell:
fixed the `Model` docstring

btaba (approved):

PR #594, opened 2025-08-09
nconmax >= 0, njmax >= 0

thowell:


kbayes:
@thowell is there a reason for this? If nconmax is set to zero I'd assume contacts should be disabled and vice versa.

adenzler-nvidia (approved):
Thanks, looks good!

thowell:
@kbayes it should be possible to set `nconmax=0` or `njmax=0`. currently, providing these values will generate a ValueError.

thowell:
@kbayes if `nconmax = 0` `collision_driver.collision` will return early
https://github.com/google-deepmind/mujoco_warp/blob/e8dadfa7640a4624e9306025e88d2d11929ec5c9/mujoco_warp/_src/collision_driver.py#L756

Issue #588, opened 2025-08-08
Fix issues indicated by compute-sanitizer


Kenny-Vilella:
Running the benchmark with compute-sanitizer highlights multiple issues that should be fixed.
This github issue would be used to list all the issues and fix them:

(1) Command:
```
python mujoco_warp/testspeed.py --function=step benchmark/aloha_pot/scene.xml --keyframe=0 --nworld=8192 --nstep=1000 --njmax=80 --nconmax=500000 -o opt.is_sparse=True -o opt.ls_parallel=False
```

Illegal read in mul_m_sparse_diag.

(2) Command:
```
python mujoco_warp/testspeed.py --function=step benchmark/apptronik_apollo/scene_flat.xml --nworld=8192 --nstep=1000 --njmax=45 --nconmax=150000 -o opt.is_sparse=True -o opt.ls_parallel=True
```

Illegal read in `_sensor_tactile_zero`

Kenny-Vilella:
Second issue solved in [this PR](https://github.com/google-deepmind/mujoco_warp/pull/589).

Kenny-Vilella:
Third issue is because `nefc` can potentially be larger than `njmax`.
In that case looping over `nefc` can cause problem.
I will make a PR to propose a new way to deal with overflow.

Kenny-Vilella:
Third issue get fixed by #598.
I can't reproduce the first issue anymore, so closing this issue.

PR #597, opened 2025-08-11
Improve handling of overflow

Kenny-Vilella:
- Created a separate kernel to handle overflow
- Reset value if above threshold

### DIFF
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -207,29 +207,12 @@ def _next_time(
   # Model:
   opt_timestep: wp.array(dtype=float),
   # Data in:
-  nconmax_in: int,
-  njmax_in: int,
-  ncon_in: wp.array(dtype=int),
-  nefc_in: wp.array(dtype=int),
   time_in: wp.array(dtype=float),
-  ncollision_in: wp.array(dtype=int),
   # Data out:
   time_out: wp.array(dtype=float),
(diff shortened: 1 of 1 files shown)


Kenny-Vilella:
Made two change:
(1) The overflow are now checked just before the solver code instead of at the end of the step.
This is because in the case where an overflow cause a crash in solver code, which is not that rare, debugging is easier as we do know immediately that there is an overflow.

(2) The value of nefc, ncon, ncollision are set to their max value in case of overflow.
This is because, as mentioned in #588, when nefc is larger than njmax, looping over nefc can cause illegal read (and potentially illegal write if we are not careful).

@adenzler-nvidia What do you think about this change?

adenzler-nvidia:
I was actually thinking of making that kernel optional such that users can decide to implement their own error reporting if they integrate things into their own applications. Sometimes it's not a good idea to use printf, but rather propagate the values to some error handler.

I do also remember that we have a few places in the code where we loop over `min(nefc, njmax)` to avoid the intermediate overflow. Would that be an option here?

Kenny-Vilella:
We can definitely use `min(nefc, njmax)`, it would just require to be more careful to make sure the pattern is used everywhere.
Using error handling is definitely better than `printf`.

So I will close this PR and implement `min(nefc, njmax)` wherever needed, we could refactor the error handling in the future.

PR #599, opened 2025-08-11
Sort efc fields before comparison in constraint_test.py

adenzler-nvidia:
the efc_id's are handed out using atomic add indexing, so the order of constraint rows is non deterministic. This sorts the fields such that a fair comparison is possible.

I had to put quite a bit of stuff into the key to get correct sorting, maybe I missed something there.


mujoco_warp/_src/constraint_test.py
```diff
@@ -37,6 +37,65 @@ def _assert_eq(a, b, name):
   np.testing.assert_allclose(a, b, err_msg=err_msg, atol=tol, rtol=tol)
 
 
+def _assert_efc_eq(d, mjd, nefc, name):
+  """Assert equality of efc fields after sorting both sides."""
+  # Get the ordering indices based on type and pos for MJWarp
```
thowell: please update the comments since the sort depends on `efc_vel` and `efc_aref` too

mujoco_warp/_src/constraint_test.py
```diff
@@ -37,6 +37,65 @@ def _assert_eq(a, b, name):
...
+  efc_vel = d.efc.vel.numpy()[0, :nefc]
+  efc_aref = d.efc.aref.numpy()[0, :nefc]
+  # Get the ordering indices based on efc_type, efc_pos, efc_vel, efc_aref for MuJoCo
+  mjd_efc_type = mjd.efc_type[:nefc]
+  mjd_efc_pos = mjd.efc_pos[:nefc]
+  mjd_efc_vel = mjd.efc_vel[:nefc]
+  mjd_efc_aref = mjd.efc_aref[:nefc]
+
+  # Create sorting keys using lexsort (more efficient for multiple keys)
+  d_sort_indices = np.lexsort((efc_pos, efc_type, efc_vel, efc_aref))
```
btaba: sometimes aref is 0 (not sure about vel), so sorting on D as well might help in those cases (which I had locally but not pushed to github yet)

btaba (approved):
Just one comment, looks good!

PR #601, opened 2025-08-11
add keyframes to aloha pot benchmark scene

thowell:
'gripper_gripper_pot'
image

'gripper_gripper'
image


thowell:
there should be contacts with these keyframes

PR #602, opened 2025-08-11
contact sensor maxforce reduction

thowell:
add maxforce reduction to contact sensor

mujoco reference: https://github.com/google-deepmind/mujoco/blob/0a02e7b7223b2e6f99f22325f179f5daa7649de6/src/engine/engine_sensor.c#L1010

### DIFF
--- a/mujoco_warp/_src/sensor_test.py
+++ b/mujoco_warp/_src/sensor_test.py
@@ -422,18 +422,18 @@ class SensorTest(parameterized.TestCase):
     datas = itertools.chain.from_iterable([itertools.combinations(field, i) for i in range(len(field))])
     datas = list(datas)
 
-    for num in [1, 2, 3, 4, 5]:
-      for geoms in [
-        'geom1="plane" geom2="geom"',
-        'geom1="geom" geom2="plane"',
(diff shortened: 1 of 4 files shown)


mujoco_warp/_src/sensor_test.py
```diff
@@ -422,18 +422,19 @@ def test_contact_sensor(self, geom):
...
-        'geom1="geom" geom2="plane"',
-        'geom1="plane" geom2="sphere"',
-        'geom1="sphere" geom2="plane"',
-        'geom1="geom" geom2="sphere"',
-        'geom1="sphere" geom2="geom"',
-      ]:
-        for data in datas:
-          data = " ".join(data)
-          contact_sensor += f'<contact {geoms} num="{num}" reduce="mindist" data="{data}"/>'
+    for geoms in [
```
btaba, line 424: this is a really big test, for MJX as well, can you pare it down a bit so that coverage is still good, but the runtime is lower?

thowell: updated `test_contact_sensor` to test fewer contact sensor instances

PR #610, opened 2025-08-12
GJK/EPA and multicontact bug bash

kbayes:
The following is a collection of commits that fix all found bugs in collision_gjk.py from importing tests from c mujoco.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -95,8 +95,9 @@ class SupportPoint:
 
 @wp.func
 def discrete_geoms(g1: int, g2: int):
-  return ((g1 == int(GeomType.MESH.value) or g1 == int(GeomType.BOX.value) or g1 == int(GeomType.HFIELD.value)) and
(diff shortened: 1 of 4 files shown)


mujoco_warp/_src/collision_gjk.py
```diff
@@ -93,6 +93,13 @@ class SupportPoint:
   vertex_index: int
 
 
+@wp.func
+def discrete_geoms(g1: int, g2: int):
+  return (g1 == int(GeomType.MESH.value) or g1 == int(GeomType.BOX.value) or g1 == int(GeomType.HFIELD.value)) and (
```
thowell, line 98: we should be able to directly utilize enums now 
`int(ENUM.value)` -> `ENUM`

thowell (approved):

mujoco_warp/_src/collision_gjk.py
kbayes: I have some style issues to fix, I'll add this as a fast follow up with that.

PR #616, opened 2025-08-14
Add keyframe selection to viewer

Kenny-Vilella:
Added keyframe selection to the viewer as it is done in testspeed.


adenzler-nvidia (approved):
LGTM!

PR #619, opened 2025-08-15
Add reporting of converged worlds in testspeed

Kenny-Vilella:
This is very useful to assess the stability of our P0 envs

### DIFF
--- a/mujoco_warp/_src/test_util.py
+++ b/mujoco_warp/_src/test_util.py
@@ -196,7 +196,7 @@ def benchmark(
   event_trace: bool = False,
   measure_alloc: bool = False,
   measure_solver_niter: bool = False,
-) -> Tuple[float, float, dict, list, list, list]:
+) -> Tuple[float, float, dict, list, list, list, int]:
   """Benchmark a function of Model and Data.
 
   Args:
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/test_util.py
```diff
@@ -217,6 +217,7 @@ def benchmark(
     list: Number of contacts.
     list: Number of constraints.
     list: Number of solver iterations.
+    int: Number of converged worlds.
```
erikfrey, line 220: don't forget to change the function annotation!  need to add one more list to `Tuple`

(it might be time to just declare a struct but we can do that later)

mujoco_warp/_src/test_util.py
```diff
@@ -263,9 +264,10 @@ def benchmark(
       if measure_solver_niter:
         solver_niter.append(d.solver_niter.numpy())
 
+    nsuccess = d.nworld - np.sum(np.any(np.isnan(d.qpos.numpy()), axis=1))
```
erikfrey: perhaps a wee bit simpler:

`nsuccess = np.sum(~np.any(np.isnan(d.qpos.numpy()), axis=1))`

erikfrey (approved):
Nice!  Just two nits

mujoco_warp/_src/test_util.py
Kenny-Vilella: Done !

mujoco_warp/_src/test_util.py
Kenny-Vilella: Done

PR #621, opened 2025-08-15
Optimization: Adding static to broadphase filter

Kenny-Vilella:
Added static to broadphase filter.
This allows the compiler to know which portion of the code will be used and optimize the code generation accordingly.

This change does not produce any measurable change for most of benchmark, except `aloha_pot`:
| This branch | Main | 
| ---------------- | ------- |
|  548.02 x     |  514.95 x |

And If I run only on the `collision` function:
| This branch | Main | 
| ---------------- | ------- |
|  5,012.54 x    |  3,822.85 x |

Command run:
```
python mujoco_warp/testspeed.py --function=step benchmark/aloha_pot/scene.xml --keyframe=0 --nworld=8192 --nstep=1000 --njmax=80 --nconmax=130000 -o opt.is_sparse=True -o opt.ls_parallel=False
```

### DIFF
--- a/mujoco_warp/_src/collision_driver.py
+++ b/mujoco_warp/_src/collision_driver.py
@@ -29,7 +29,9 @@ from .types import Data
 from .types import DisableBit
 from .types import GeomType
 from .types import Model
+from .warp_util import cache_kernel
 from .warp_util import event_scope
+from .warp_util import kernel as nested_kernel
 
 wp.set_module_options({"enable_backward": False})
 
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_driver.py
```diff
@@ -248,54 +250,56 @@ def _obb_filter(
...
+    rbound2 = geom_rbound[worldid, geom2]
+    margin1 = geom_margin[worldid, geom1]
+    margin2 = geom_margin[worldid, geom2]
+    xpos1 = geom_xpos_in[worldid, geom1]
+    xpos2 = geom_xpos_in[worldid, geom2]
+    xmat1 = geom_xmat_in[worldid, geom1]
+    xmat2 = geom_xmat_in[worldid, geom2]
+
+    if rbound1 == 0.0 or rbound2 == 0.0:
+      if wp.static(opt_broadphase_filter & int(BroadphaseFilter.PLANE.value)):
```
Kenny-Vilella, line 287: The only real change in this PR is to add `wp.static` in all this function.
We can then build the function beforehand and provide it as input to the broadphase functions.

PR #626, opened 2025-08-15
A few asv improvements.

erikfrey:
* Add `asv` geberated directory to `.gitignore`
* Make sure `asv` installs nightly mujoco and warp
* Remove the extra function call in `benchmark`, seems unnecessary now
* Default to always clearing kernel cache so `jit_duration` metric is accurate
* Document how to disable clearing the cache using `ASV_CACHE_KERNELS`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@ benchmark/mujoco_menagerie/
 benchmark/kitchen/combined_assets/
 kitchen_robot.xml
 
+## airspeed velocity working directories
+asv/
+
 # Language-specific files
 # =======================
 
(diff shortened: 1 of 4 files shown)


Kenny-Vilella:
Overall it looks good, two minor comments:
- We may want to extend the `default_benchmark_timeout`, it looks like the AlohaPot benchmark reaches the limit and failed.
- The AlohaPot benchmark misses the following line in `benchmark.py`:
```
params = mujoco_warp.BenchmarkSuite.params + ("step.euler",)
```
This is my fault, sorry for that. Not exactly sure what this is doing though ^^

PR #633, opened 2025-08-20
Use nxn_geom_pair_filtered as criteria for nxn.

erikfrey:
This uses `nxn_geom_pair_filtered` as the criteria for NXN, which is a good value as it linearly determines kernel perf.  Updating this changes the default for Aloha and ApolloTerrain from SAP to NXN:

* Aloha default (SAP_TILE): `sap_broadphase: 262.41`
* Aloha NXN: `nxn_broadphase: 84.05`
* ApolloTerrain default (SAP_SEGMENTED): `sap_broadphase: 1883.06`
* ApolloTerrain NXN: `nxn_broadphase: 712.61`


erikfrey:
I think the kitchen benchmark may benefit from SAP but I haven't looked closely, as we're saving optimizing the kitchen benchmark for after release.

Issue #640, opened 2025-08-21
Unstable simulation with specific mesh asset


jc-bao:
# The issue

Simulations involving certain mesh assets are unstable. For example, the object in the provided XML file jumps continuously instead of resting on the ground.

# Reproduce

Load the following scene: https://github.com/jc-bao/mjwarp_test/blob/978e6e623c6d4ab83b4169de6bdc4db9adc7a8a8/assets/scene_coke.xml#L1

Run the command: `mjwarp-viewer scene_coke.xml --njmax=86`

# Expected Behavior

The object should settle and remain stable on the ground plane. At least for previous mjwap and mujoco cpu, it works fine. 

# Actual Behavior

The object is unstable and jumps off the ground.

https://github.com/user-attachments/assets/086d253d-db9f-4414-b836-b5736dd69e81

# Environment

mjwarp: baeb10e73237df13802217d310788e1f02dcf92d

mujoco: 3.3.6.dev797870883

warp-lang: 1.9.0.dev20250820

# Question

I have been unable to find a stable version by rolling back. Could you advise on a specific version combination that is known to be stable or suggest a potential fix?

Thank you!

btaba:
Would you like to try with https://github.com/google-deepmind/mujoco_warp/pull/641 ?

jc-bao:
Oh yeah, that works! My convex mesh - capsule collision also looks weird, which has deeper penetration than before. Would you mind also take a look? Thanks!

jc-bao:
Oh sorry, I just pulled the latest main branch -- the mesh-plane collision error persists. Your fork works fine though.

btaba:
@jc-bao https://github.com/google-deepmind/mujoco_warp/pull/641 was merged to main though, could you clarify what you mean? Both convex-mesh <> capsule and convex-mesh <> plane are broken? Do you have a repro for the former?

kbayes:
Hi @jc-bao FYI, I just fixed capsule collisions.

thowell:
@jc-bao 

the simulation seems ok with
mujoco warp: 3f1197edd97a96a21384fcb418aa566306a22654
mujoco: 3.3.6.dev798987338
warp: 1.9.0.dev20250825

https://github.com/user-attachments/assets/e5d1442b-0ed1-4202-8e97-dfc1783de05d

please try the latest versions of mujoco warp + mujoco + warp and clear the warp cache

```
rm -r ~/.cache/warp
```

[...]

PR #613, opened 2025-08-12
plane_convex update

thowell:

PR #641, opened 2025-08-22
Fix convex-plane and box-box bugs. #640

btaba:


adenzler-nvidia:
does this one replace #613 ?

btaba:
Ah @adenzler-nvidia I wasn't aware of that PR. This PR just fixes the bug rather than completely changing the methodology. I think it's a safer change overall, as #613 is using a heuristic for "don't pick a point too close to the max-support point" rather than an actually stable manifold. We've been using this logic in MJX for quite some time.

adenzler-nvidia:
I think #613 also started out like this and then ended up in that shape.. I'm ok with submitting this, we can build on top of this if we want to change the algorithm.

adenzler-nvidia:
could you add a test that fails before and suceeds with this change?

btaba:
Ok I took a peek at the commits, I'm not sure that that PR was fixing the bug completely, which is maybe why that PR switched tack?

In any case they're different methodologies. The MJ methodology causes a lot of flicker in plane<>convex while the impl here is quite stable (I double checked this claim in the viewer with pot<>plane). Up to y'all ultimately on the change in algo. I'll look into a test

btaba:
@adenzler-nvidia ok it's taking longer than expected to come up with an isolated test. I can use the mesh and keyframe from the bug report, but hesitant to add a new asset. FWIW, we also don't have a test for this in MJX

kevinzakka:
For some context, I am still seeing (mesh, *) NaNs in my setting with Taylor's PR. Don't have a MRE just yet.

btaba:
@kevinzakka are you seeing mesh<>plane NaNs with this PR?

btaba:
Ok I reached out to Kevin but he's busy with a deadline. This is a pure bugfix, so it can more or less be independent of any subsequent overhaul to the methodology. Deferring to @thowell if he'd like to do a rewrite

Issue #445, opened 2025-07-02
Determine that all Playground environments train as expected


erikfrey:
This is a criteria for migrating MJWarp to Release.

btaba:
Locomotion: https://github.com/google-deepmind/mujoco_warp/issues/478 hfield envs NaN

btaba:
Closing this issue and will open smaller follow-ups, in light of https://github.com/google-deepmind/mujoco_playground/discussions/197

btaba:
Smaller issues:

https://github.com/google-deepmind/mujoco_warp/issues/661
https://github.com/google-deepmind/mujoco_warp/issues/662
https://github.com/google-deepmind/mujoco_warp/issues/663

PR #642, opened 2025-08-22
Fix bugs in legacy gjk.

btaba:
This algo (ported from CUDA) was used in RL and sim2real, so it's generally useful for debugging manipulation environments (which we are currently trying to get parity on).

### DIFF
--- a/mujoco_warp/_src/collision_gjk_legacy.py
+++ b/mujoco_warp/_src/collision_gjk_legacy.py
@@ -19,6 +19,7 @@ from .collision_hfield import hfield_prism_vertex
 from .collision_primitive import Geom
 from .math import gjk_normalize
 from .math import orthonormal
+from .math import orthonormal_to_z
 from .support import all_same
 from .support import any_different
 from .types import MJ_MINVAL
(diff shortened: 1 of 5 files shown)


mujoco_warp/_src/collision_gjk_legacy.py
```diff
@@ -663,7 +665,7 @@ def multicontact_legacy(
         m2 = v2[j]
         dd = (m1[0] - m2[0]) * (m1[0] - m2[0]) + (m1[1] - m2[1]) * (m1[1] - m2[1])
 
-        if i != 0 and j != 0 or dd < minDist:
+        if (i == 0 and j == 0) or (dd < minDist):
```
btaba, line 668: for those curious, this subtle bug was the real kicker

kbayes:
@btaba can you change the title since the consensus was that legacy_gjk will be removed soon, but it seems you want to keep it around as you're adding a flag for it? Also adding @nvtw as this possible decision is relevant for his collision refactor.

btaba:
I don't think any new decisions were made RE legacy_gjk, I'm just trying to debug Playground manip. policies. Y'all can delete before release and after https://github.com/google-deepmind/mujoco_warp/issues/445  (also see internal chat to pickup the convo and for more background info).

PR #645, opened 2025-08-25
Optimization: Merging kernels in camlight

Kenny-Vilella:
- `_cam_local_to_global` and `_cam_fn` are merged
- `_light_local_to_global` and  `_light_fn` are merged

Running the following benchmark:
```
python mujoco_warp/testspeed.py --function=step benchmark/humanoid/humanoid.xml --nworld=8192 --nstep=1000 --njmax=52 --nconmax=200000 -o opt.is_sparse=True -o opt.ls_parallel=True --event_trace=True -o opt.solver=newton -o opt.cone=pyramidal
```

This is what I get for trunk:

> step: 625.70
>   forward: 622.81
>     fwd_position: 104.25
>       camlight: 2.33

And this branch:
> step: 626.44
>   forward: 623.56
>     fwd_position: 103.94
>       camlight: 1.80

So a 30% speedup for the `camlight` function.
I saw similar speedup for other benchmarks.

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -513,7 +513,7 @@ def com_pos(m: Model, d: Data):
 
 
 @wp.kernel
-def _cam_fn(
+def _cam_local_to_global(
   # Model:
   cam_mode: wp.array(dtype=int),
   cam_bodyid: wp.array(dtype=int),
@@ -580,7 +580,7 @@ def _cam_fn(
 
 
 @wp.kernel
-def _light_fn(
+def _light_local_to_global(
   # Model:
   light_mode: wp.array(dtype=int),
   light_bodyid: wp.array(dtype=int),
@@ -647,7 +647,7 @@ def camlight(m: Model, d: Data):
   including special handling for tracking and target modes.
   """
   wp.launch(
-    _cam_fn,
+    _cam_local_to_global,
     dim=(d.nworld, m.ncam),
     inputs=[
       m.cam_mode,
@@ -665,7 +665,7 @@ def camlight(m: Model, d: Data):
     outputs=[d.cam_xpos, d.cam_xmat],
   )
   wp.launch(
-    _light_fn,
+    _light_local_to_global,
     dim=(d.nworld, m.nlight),
     inputs=[
       m.light_mode,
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/smooth.py
```diff
@@ -512,39 +512,20 @@ def com_pos(m: Model, d: Data):
...
-  worldid, camid = wp.tid()
-  bodyid = cam_bodyid[camid]
-  xpos = xpos_in[worldid, bodyid]
-  xquat = xquat_in[worldid, bodyid]
-  cam_xpos_out[worldid, camid] = xpos + math.rot_vec_quat(cam_pos[worldid, camid], xquat)
-  cam_xmat_out[worldid, camid] = math.quat_to_mat(math.mul_quat(xquat, cam_quat[worldid, camid]))
-
-
 @wp.kernel
 def _cam_fn(
```
erikfrey: nit: let's still call this `_cam_local_to_global` or something like that for two reasons:

1) it's more descriptive
2) save `_fn` suffix for `wp.function` (if at all)

mujoco_warp/_src/smooth.py
```diff
@@ -582,28 +571,12 @@ def _cam_fn(
...
-  light_xdir_out[worldid, lightid] = math.rot_vec_quat(light_dir[worldid, lightid], xquat)
+  else:
+    bodyid = cam_bodyid[camid]
+    xpos = xpos_in[worldid, bodyid]
+    xquat = xquat_in[worldid, bodyid]
+    cam_xpos_out[worldid, camid] = xpos + math.rot_vec_quat(cam_pos[worldid, camid], xquat)
+    cam_xmat_out[worldid, camid] = math.quat_to_mat(math.mul_quat(xquat, cam_quat[worldid, camid]))
 
 
 @wp.kernel
```
erikfrey, line 582: nit: same here, if you buy the above reasoning then this would be `_light_local_to_global`

erikfrey (approved):
Looks good just two nits

mujoco_warp/_src/smooth.py
Kenny-Vilella: I was actually not very happy with the name, so thank you !
Changed.

PR #651, opened 2025-08-26
Add SDF benchmark

quagla:


thowell:
if keyframe 1 or keyframe 2 is loaded there are errors:
```
ERROR: SDF grad type not implemented
```
```
ERROR: SDF type not implemented
```

are these errors expected?

quagla:
> if keyframe 1 or keyframe 2 is loaded there are errors:
> 
> ```
> ERROR: SDF grad type not implemented
> ```
> 
> ```
> ERROR: SDF type not implemented
> ```
> 
> are these errors expected?

Yeah, those are pending on the WIP mesh-SDF collisions #678

thowell (approved):

PR #653, opened 2025-08-26
update contact sensor matching

thowell:
currently available matching:
- geom1, geom2

add matching for:
- none
- geom1
- geom2
- body1
- body2
- body1, body2
- subtree1, subtree2
- site

mujoco documentation: https://mujoco.readthedocs.io/en/latest/XMLreference.html#sensor-contact


mujoco_warp/_src/sensor.py
```diff
@@ -1983,11 +1983,28 @@ def _sensor_tactile(
...
 
 
+@wp.func
+def _check_match(body: int, geom: int, objtype: int, id: int) -> bool:
+  """Check if a contact body/geom matches a sensor spec (objtype, id)."""
+  if objtype == int(ObjType.UNKNOWN.value):
+    return True
+  # TODO(team): site
+  if objtype == int(ObjType.GEOM.value):
+    return id == geom
```
btaba: id is a builtin function in python

mujoco_warp/_src/sensor_test.py
```diff
@@ -422,10 +422,18 @@ def test_contact_sensor(self, geom):
...
+      'geom2="plane"',
       'geom1="plane" geom2="geom"',
       'geom1="geom" geom2="plane"',
       'geom1="sphere" geom2="plane"',
       'geom1="geom" geom2="sphere"',
+      'body1="plane"',
+      'body2="plane"',
+      'body1="plane" body2="geom"',
+      'body1="geom" body2="plane"',
+      'body1="sphere" body2="plane"',
```
btaba: OOC how much time would this be adding to CI?

mujoco_warp/_src/sensor.py
```diff
@@ -2015,53 +2032,83 @@ def _contact_match(
...
+        dir = -1.0
+    elif reftype != int(ObjType.UNKNOWN.value):
+      if not match22:
+        dir = -1.0
+
+  worldid = contact_worldid_in[contactid]
+  contactmatchid = wp.atomic_add(sensor_contact_nmatch_out[worldid], contactsensorid, 1)
+
+  if contactmatchid >= MJ_MAXCONPAIR:
+    wp.printf("contact match overflow: please increase MJ_MAXCONPAIR to %u\n", contactmatchid)
```
btaba: MJ_MAXCONPAIR is typically used in CD to limit the total number of contacts for a geom-pair, I'm not really following why it should be used in the sensor logic?

btaba:
Pretty sweet Taylor :allthethings:

mujoco_warp/_src/sensor.py
thowell: fixed

mujoco_warp/_src/sensor_test.py
thowell: looking at the runs (build and test 3.12) the times looks similar

this pr:
405.80s
https://github.com/google-deepmind/mujoco_warp/actions/runs/17246729590/job/48938247432?pr=653
image

main:
423.05s
https://github.com/google-deepmind/mujoco_warp/actions/runs/17241377084/job/48919255347
image

mujoco_warp/_src/sensor.py
thowell: i think `MJ_MAXCONPAIR` made more sense when contact sensors only had geom1, geom2 matching

mujoco_warp/_src/sensor_test.py
btaba: ~15s addition? I would potentially consider not testing all combinations of "datas" x "matching criteria", but totally your call

mujoco_warp/_src/sensor.py
thowell: we need a maximum number of matches to consider for sorting

kevinzakka:
Amazing, thanks @thowell !

mujoco_warp/_src/sensor.py
btaba: Ah I see, unfortunately the print statement doesn't really lead to something that is actionable for the user to get rid of. Maybe MJ_MAXCONPAIR should be forked to Option specifically for this sensor use-case?

Also pinging @adenzler-nvidia. FWIU  there is additional overhead to launching kernels with printf even if the printf is not hit, is that right?

btaba: Maybe a better option is to fill nmatch in the sensordata with something that indicates to the user that there could've been an overflow?

Also paging @erikfrey for additional perspective, and @yuvaltassa

adenzler-nvidia: yes, printf uses a lot of registers so we should avoid it if possible. I also don't particularly like the printfs we have in next_time but I think it's super important to report njmax and nconmax overflows, especially in benchmarks.

I'm thinking about making this a bit more adaptable - maybe we need an example of how to get all that info at the end of the step and reporting it correctly. Any non-trivial app likely wants to control how/when they're reporting and potentially opt out of the default printf way. We need something that can be caught by an application such that the required actions can be taken.

thowell:
added subtree matching

thowell:
added site matching

mujoco_warp/_src/sensor.py
thowell: @btaba 
- added `contact_sensor_maxmatch` to `Option` with a default value set to `64` in `put_model`
- added todo for alternative to `wp.printf` for reporting overflows

mujoco_warp/_src/sensor_test.py
thowell: updated CI comparison (Build and test 3.10)

main https://github.com/google-deepmind/mujoco_warp/actions/runs/17464704336/job/49597440895
6min 48s

this pr https://github.com/google-deepmind/mujoco_warp/actions/runs/17464555510/job/49596941871
7m 8s

mujoco_warp/_src/sensor.py
```diff
@@ -2037,83 +2066,127 @@ def _contact_match(
...
+        dir = -1.0
+    elif reftype != int(ObjType.UNKNOWN.value):
+      if not match22:
+        dir = -1.0
+
+  contactmatchid = wp.atomic_add(sensor_contact_nmatch_out[worldid], contactsensorid, 1)
+
+  if contactmatchid >= opt_contact_sensor_maxmatch:
+    # TODO(team): alternative to wp.printf for reporting overflow?
+    wp.printf("contact match overflow: please increase Option.contact_sensor_maxmatch to %u\n", contactmatchid)
```
btaba, line 2126: how much does the presence of this printf actually affect perf on a moderate scene?

I would at least create an issue to track this. One idea is to write maxint to nfound

mujoco_warp/_src/types.py
```diff
@@ -581,6 +581,7 @@ class Option:
       contacts during the physics step (as opposed to DisableBit.CONTACT which explicitly
       zeros out the contacts at each step)
     legacy_gjk: run legacy gjk algorithm
+    contact_sensor_maxmatch: max number of contacts considered by contact sensor matching criteria
```
btaba, line 584: maybe add a snippet about what happens with overflow, "if more contacts are found for a given sensor, those contacts are ignored"

btaba (approved):
Left two small comments but generally LGTM! Thanks Taylor

mujoco_warp/_src/types.py
thowell: the docstring is updated

mujoco_warp/_src/sensor.py
thowell: added contact sensors to the humanoid
```
  <sensor>
    <contact geom1="floor" geom2="foot1_left"/>
    <contact geom1="floor" geom2="foot1_right"/>
    <contact geom1="floor" geom2="foot2_left"/>
    <contact geom1="floor" geom2="foot2_right"/>
  </sensor>
```

```
mjwarp-testspeed benchmark/humanoid/humanoid.xml --nconmax=200000 --njmax=64
```

with `wp.printf`:
```

Total JIT time: 0.30 s
Total simulation time: 9.02 s
Total steps per second: 908,597
Total realtime factor: 4,542.98 x
Total time per step: 1100.60 ns

Total JIT time: 0.29 s
Total simulation time: 9.05 s
Total steps per second: 905,513
Total realtime factor: 4,527.57 x
Total time per step: 1104.35 ns

Total JIT time: 0.29 s
Total simulation time: 9.05 s
Total steps per second: 905,148
Total realtime factor: 4,525.74 x
Total time per step: 1104.79 ns
```

without `wp.printf` (commented out):
```

Total JIT time: 0.30 s
Total simulation time: 9.00 s
Total steps per second: 910,010
Total realtime factor: 4,550.05 x
Total time per step: 1098.89 ns

Total JIT time: 0.30 s
Total simulation time: 9.05 s
Total steps per second: 904,986
Total realtime factor: 4,524.93 x
Total time per step: 1104.99 ns

Total JIT time: 0.30 s
Total simulation time: 9.05 s
Total steps per second: 905,674
Total realtime factor: 4,528.37 x
Total time per step: 1104.15 ns
```

unclear is there is much performance impact with `wp.printf`

thowell: opened an issue for `wp.printf` alternative #692

PR #665, opened 2025-08-29
Optimize sdf kernel

jvonmuralt:


mujoco_warp/_src/collision_sdf.py
```diff
@@ -636,63 +636,66 @@ def _sdf_narrowphase(
...
+    contact_worldid_out,
+  )
 
 
 @event_scope
 def sdf_narrowphase(m: Model, d: Data):
   wp.launch(
     _sdf_narrowphase,
-    dim=d.nconmax,
+    dim=(m.opt.sdf_initpoints, int(d.nconmax / m.opt.sdf_initpoints)),
```
quagla: This should be 

`dim=(m.opt.sdf_initpoints, d.nconmax),`

quagla (changes requested):
I think this can be merged (after fixing the kernel dimension).

I did a benchmark and the cow goes from 240x to 290x. We need more fine grained benchmark to see where the bottleneck is.

quagla (approved):

PR #667, opened 2025-08-29
update _geom

thowell:


kbayes:
@thowell while you're at it, maybe give _geom a better name like create_geom.

thowell:
@kbayes in #668 `_geom` -> `geom`

thowell:
note: 
the (previous) logic in `box_box_wrapper` https://github.com/thowell/mujoco_warp/blob/99336e1b2d062f4c3fddd360941c30e8702d8345/mujoco_warp/_src/collision_primitive.py#L1388 causes a segmentation fault when testing `collision_driver_test` with cpu
https://github.com/google-deepmind/mujoco_warp/actions/runs/17399418671/job/49388836982

this updated logic https://github.com/thowell/mujoco_warp/blob/efd242de89556af84002225d24f2e49b0dbfd2e1/mujoco_warp/_src/collision_primitive.py#L1386 works

not sure why this makes a difference?

PR #668, opened 2025-08-29
add gjk filtering to height field midphase

thowell:
**benchmark/apptronik_apollo/scene_hfield.xml**
```
mjwarp-testspeed benchmark/apptronik_apollo/scene_hfield.xml --nconmax=3000000 --njmax=256
```

this pr:
```

Total JIT time: 0.59 s
Total simulation time: 24.17 s
Total steps per second: 338,974
Total realtime factor: 1,694.87 x
Total time per step: 2950.08 ns
```

main:
```

Total JIT time: 0.57 s
Total simulation time: 25.64 s
Total steps per second: 319,439
Total realtime factor: 1,597.20 x
Total time per step: 3130.49 ns
```

---

reduce nconmax by 10x: 3_000_000 -> 300_000

```
mjwarp-testspeed benchmark/apptronik_apollo/scene_hfield.xml --nconmax=300000 --njmax=256
```

this pr:
```

Total JIT time: 0.58 s
Total simulation time: 22.91 s
Total steps per second: 357,610
Total realtime factor: 1,788.05 x
Total time per step: 2796.34 ns
```

main:
```
ncollision overflow - please increase nconmax to 440544
```


kbayes:
@thowell Could you separate ccd_tolerance out into another PR?

thowell:
@kbayes #673

erikfrey (commented):
Hmm, this might be the right strategy but let's talk to @btaba - in MJX, we employed a pretty simple strategy for hfield culling that was fast and didn't use GJK, so I'm curious why we need it here.

mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -131,7 +132,7 @@ def _gjk_kernel(
       x2,
     ) = ccd(
       multiccd,
-      1e-6,
+      opt_ccd_tolerance[0],
```
btaba: +1 to split, I guess you need a rebase now

mujoco_warp/_src/collision_driver.py
```diff
@@ -710,11 +715,361 @@ def nxn_broadphase(m: Model, d: Data):
   )
 
 
+@wp.kernel
+def _hfield_midphase(
```
btaba: OOC why the file move? the distinction of what belongs in `collision_hfield.py` vs. not, is not clear to me. The former setup seemed a tad cleaner from a `collision_driver` perspective

mujoco_warp/_src/collision_driver.py
```diff
@@ -710,11 +715,361 @@ def nxn_broadphase(m: Model, d: Data):
...
+        geom1.hfprism = hfield_triangle_prism(
+          geom_dataid, hfield_adr, hfield_nrow, hfield_ncol, hfield_size, hfield_data, g1, hftri_index
+        )
+
+        prism_pos = wp.vec3(0.0, 0.0, 0.0)
+        for i in range(6):
+          prism_pos += hfield_prism_vertex(geom1.hfprism, i)
+        prism_pos = geom1.rot @ (prism_pos / 6.0)
+
+        result = gjk(ccd_tolerance, opt_gjk_iterations, geom1, geom2, x_1 + prism_pos, x_2, geomtype1, geomtype2, 0.0, False)
```
btaba: considering legacy_gjk works for loco-hfield envs, can y'all first fix that NaN bug before making a change that potentially breaks the working hfield+RL codepath?  Or was it fixed already?

mujoco_warp/_src/collision_driver.py
```diff
@@ -710,11 +715,361 @@ def nxn_broadphase(m: Model, d: Data):
...
+
+        if result.dim != 0:
+          if firstprism:
+            new_pairid = pairid
+            firstprism = False
```
btaba: I'm a bit confused with this logic, you use `pairid` only once and the rest of `new_pairid` are indexed by  an `ncollision` atomic counter?

mujoco_warp/_src/collision_driver.py
btaba: if you're calling gjk here, may as well run the full collision detection too? Or does that then slow things down significantly? Maybe you can re-use the gjk result in that case instead of calling it twice?

mujoco_warp/_src/collision_driver.py
thowell: this is essentially existing logic.

broadphase adds hfield geom and then the midphase replaces the hfield geom with prism geoms. the existing memory for the hfield geom is replaced with the first prism geom.

mujoco_warp/_src/collision_driver.py
thowell: collision_gjk.py imports `hfield_prism_vertex` from collision_hfield.py so there would be a circular dependency if `gjk` was import in collision_hfield.py.

thowell: collision_hfield.py is now removed

mujoco_warp/_src/collision_driver.py
thowell: legacy_gjk should work with the latest update

mujoco_warp/_src/collision_driver.py
thowell: the updated code now more closely follows the mujoco implementation https://github.com/google-deepmind/mujoco/blob/ea660f4c13b14c6e562d2f12b3e17ea24322ad0a/src/engine/engine_collision_convex.c#L1327

mujoco_warp/_src/collision_driver.py
thowell: the refactored code now utilizes a height field narrowphase instead of a height field midphase

mujoco_warp/_src/collision_gjk_test.py
thowell: fixed -- reverted the changes in this file

thowell:
the pr has been refactored and the pr description has been updated with new timing results.

the primary change is removing the midphase routine and replacing it with a narrowphase routine.

thowell:
updated pr description with timing results for this pr after adding `MJ_MAXCONPAIR` check in `ccd_hf_kernel`

mujoco_warp/_src/collision_convex.py
```diff
@@ -155,13 +130,11 @@ def ccd_kernel(
     pair_margin: wp.array2d(dtype=float),
```
btaba: do you still need hfield fields above?

mujoco_warp/_src/collision_convex.py
```diff
@@ -391,6 +338,428 @@ def ccd_kernel(
...
+  geomtype1: int,
+  geomtype2: int,
+  gjk_iterations: int,
+  epa_iterations: int,
+  epa_exact_neg_distance: bool,
+  depth_extension: float,
+):
+  # runs convex collision on a set of geom pairs to recover contact info
+  @nested_kernel(module="unique", enable_backward=False)
+  def ccd_hf_kernel(
```
btaba: comment about code layout:

a lot of the ccd_kernel_builder code is copied a bit verbatim, and may be harder to maintain going forward. There's probably a better way to layout the code for re-use (the kernel builder can take a hfield static bool), but I'll leave that up to y'all

mujoco_warp/_src/collision_convex.py
```diff
@@ -391,6 +338,428 @@ def ccd_kernel(
...
+    x2 = geom2.pos
+
+    dx = (2.0 * size[0]) / float(ncol - 1)
+    dy = (2.0 * size[1]) / float(nrow - 1)
+    dr = wp.vec2i(1, 0)
+
+    prism = mat63()
+
+    # set zbottom value using base size
+    prism[0, 2] = -size[3]
```
btaba: thank you for simplifying the prism code! This is much easier to understand

mujoco_warp/_src/collision_convex.py
```diff
@@ -391,6 +338,428 @@ def ccd_kernel(
...
+            continue
+
+          # prism height test
+          if prism[3, 2] < zmin and prism[4, 2] < zmin and prism[5, 2] < zmin:
+            continue
+
+          geom1.hfprism = prism
+
+          # TODO(kbayes): remove legacy GJK once multicontact can be enabled
+          if wp.static(legacy_gjk):
```
btaba: nit: 3 nested for loops and nested if statements within that is a bit too gnarly for my appetite, have you considered splitting some of this out into functions?

btaba (commented):
Nice! Just a few style nits from here

mujoco_warp/_src/collision_convex.py
thowell: reorganized collision_convex.py to have one builder that has the hfield inputs

mujoco_warp/_src/collision_convex.py
thowell: reorganized collision_convex.py. there is one builder takes a static bool `is_hfield`.

mujoco_warp/_src/collision_convex.py
thowell: reorganized collision_convex.py including the introduction of two `wp.func`s: `ccd_write` and `collision_hfield.hfield_filter`.

mujoco_warp/_src/collision_convex.py
```diff
@@ -282,98 +395,154 @@ def ccd_kernel(
...
+
+            # prism center
+            x1 = geom1.pos
+            if wp.static(not legacy_gjk):
+              x1_ = wp.vec3(0.0, 0.0, 0.0)
+              for i in range(6):
+                x1_ += prism[i]
+              x1 += geom1.rot @ (x1_ / 6.0)
+
+            ncontact = ccd_write(
```
kbayes: ccd_write is a confusing name for me, as it doesn't imply that ccd was ran. Maybe borrowing ccd_penetration from mujoco c would be better [1]. What do you think?

[1] https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_collision_convex.c#L47

mujoco_warp/_src/collision_convex.py
```diff
@@ -282,98 +395,154 @@ def ccd_kernel(
...
-      if dist >= 0.0 or depth < -depth_extension:
-        return
-      sphere = int(GeomType.SPHERE.value)
-      ellipsoid = int(GeomType.ELLIPSOID.value)
-      if geom_type[g1] == sphere or geom_type[g1] == ellipsoid or geom_type[g2] == sphere or geom_type[g2] == ellipsoid:
-        count, points = multicontact_legacy(geom1, geom2, geomtype1, geomtype2, depth_extension, depth, normal, 1, 2, 1.0e-5)
-      else:
-        count, points = multicontact_legacy(geom1, geom2, geomtype1, geomtype2, depth_extension, depth, normal, 4, 8, 1.0e-1)
-      frame = make_frame(normal)
+    if wp.static(is_hfield):
```
kbayes, line 434: Could you add a comment that this is referencing the relevant part from https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_collision_convex.c#L1196?

mujoco_warp/_src/collision_convex.py
thowell: renamed `ccd_write` -> `eval_ccd_write_contact`

mujoco_warp/_src/collision_convex.py
thowell: noted `mjc_ConvexHField` in:
- `collision_hfield.hfield_filter` docstring
- `collision_convex.ccd_kernel` comment

mujoco_warp/_src/collision_convex.py
```diff
@@ -179,17 +319,16 @@ def ccd_kernel(
...
-    multiccd_endvert_in: wp.array2d(dtype=wp.vec3),
-    multiccd_face1_in: wp.array2d(dtype=wp.vec3),
-    multiccd_face2_in: wp.array2d(dtype=wp.vec3),
+    multicdd_pdist_in: wp.array2d(dtype=float),
+    multicdd_idx1_in: wp.array2d(dtype=int),
+    multicdd_idx2_in: wp.array2d(dtype=int),
+    multicdd_n1_in: wp.array2d(dtype=wp.vec3),
+    multicdd_n2_in: wp.array2d(dtype=wp.vec3),
+    multicdd_endvert_in: wp.array2d(dtype=wp.vec3),
+    multicdd_face1_in: wp.array2d(dtype=wp.vec3),
```
kbayes: Fixed now?

kbayes (approved):

mujoco_warp/_src/collision_convex.py
thowell: should be fixed now

Issue #492, opened 2025-07-10
Add back tile kernels for dense `deriv_smooth_vel`


erikfrey:
We saw a 3-5% drop in steps / second from #449 which fixed some bugs but undid using the tile API.

Let's see if we can add the tile kernel back now that the implementation is more correct.

erikfrey:
Other perf improvements have made this less necessary.

PR #671, opened 2025-09-01
update _qderiv_actuator_passive

thowell:


erikfrey (commented):
Interesting!  How did you come across this fix?  Is there a bug or a failing unit test?  Can you add something to the PR description?

thowell:
the context for this is #492. i don't think this pr is a bug fix -- the current implementation should be correct. the currently implementation (accumulating a fraction of the dof_damping contribution at each loop iteration) is less efficient than the proposed change in this pr (one dof_damping contribution).

Issue #550, opened 2025-07-28
Investigate performance of the n_humanoid scene.


Kenny-Vilella:
Following benchmark time given by #545, it looks like the scene with three humanoids is almost 10 times slower than with only one humanoid.
This scaling is not reasonable and we should understand what is happening.

adenzler-nvidia:
This is likely due to the implementation of blocked cholesky. We should rethink this approach, and compare to why it's not just as fast as the segmented cholesky in JAX. We can just call into that if it's faster.

Kenny-Vilella:
I check the nsight trace and it is correct.
48% of time spent in the blocked cholesky and 27.3% in JTDAJ.

JTDAJ is also slower than what it should be.

Kenny-Vilella:
For the blocked Cholesky, both nsight systems and nsight compute show wide variations of compute time during a single run, almost a factor 100.
From Nsight compute, I can see in slow kernels large cache miss, almost 200MB of transfer L2 cache <-> device memory and 600MB L1 cache <-> L2 cache.

adenzler-nvidia:
keeping this one open but quite closely related to #320.

thowell:
@adenzler-nvidia @Kenny-Vilella what is the status of this issue?

Kenny-Vilella:
My impression is that perf is not yet as good as what it should be.
However, the introduction of island would probably solve some of the issue we are seeing with this benchmark.
So it's maybe not worth to spend weeks optimizing this type of case.

adenzler-nvidia:
agree with Kenny - I think the obvious things have been looked at and with islanding on the horizon, I think we can close this and not let it block a release.

thowell:
@Kenny-Vilella @adenzler-nvidia ok sounds good

will remove this issue from the release roadmap and we can revisit this scene with islands + sleeping

thowell:
islands + sleeping will not affect this scene since the humanoids are actuated and actuated trees do not currently sleep. additionally, this issue is a bit out of date now.

PR #672, opened 2025-09-01
Optimization: tiled versions of JTDAJ

adenzler-nvidia:
Significant speedups, but still WIP as I need to verify the benchmarks everywhere. Thought I'd share a preview :)


erikfrey:
doop

adenzler-nvidia:
some numbers (total steps per second) using parallel linesearch

| Benchmark | main | this branch | speedup |
|-----------------|---------|----------------| ----------- |
| humanoid | 3,583,055 |3,753,086| 1.05|
| n_humanoid |358,086 | 485,103 | 1.35 |
| aloha pot |1,450,175 |1,383,222| 0.95 |
| apollo | 2,654,118 | 2,717,492 | 1.02 |
| kitchen G1 (4k worlds) | 17,338 | 21,103 | 1.21 | 
| kitchen anymal (4k worlds) | 51,271 | 67,883 | 1.32 |
| franka | 13,997,320 | 13,045,987 | 0.93 |

Will look into the regressions now. Will probably have to think about the dense version some more.

adenzler-nvidia:
This is ready for review, still looking at the test failure though, I can'r reproduce it on my system.

I think something is off with the aloha pot such that the numbers are weird, so I wouldn't trust that one too much. The franka numbers are kind of expected as that benchmark has very few actual constraints.

The numbers look great for sparse systems with big constraint counts. For dense systems, the numbers don't look that impressive  but I can see clear gains on the kernel time itself. The kernel is much less dominant which means we're not seeing a lot of the effects in total, but it's still one of the expensive ones and I think this is just a stepping stone.

mujoco_warp/_src/io.py
```diff
@@ -849,6 +849,108 @@ def create_nmodel_batched_array(mjm_array, dtype, expand_dim=True):
   return m
 
 
+def _mujoco_octree_to_warp_volume(
```
thowell: i think we don't need `_mujoco_octree_to_warp_volume` here since it was removed in #683

mujoco_warp/_src/solver.py
```diff
@@ -1364,100 +1364,151 @@ def update_gradient_set_h_qM_lower_sparse(
...
-    return
 
-  nefc = nefc_in[worldid]
+@cache_kernel
+def update_gradient_JTDAJ_sparse_tiled(tile_size: int, njmax: int):
+  TILE_SIZE = tile_size
 
-  dofi = (int(sqrt(float(1 + 8 * elementid))) - 1) // 2
-  dofj = elementid - (dofi * (dofi + 1)) // 2
+  @wp.kernel
```
thowell: should we have `@nested_kernel` here?

mujoco_warp/_src/io.py
adenzler-nvidia: that was a bad merge, fixed. Thanks.

mujoco_warp/_src/solver.py
adenzler-nvidia: indeed, and we should also disable the backwards pass.

erikfrey:
@adenzler-nvidia this looks really cool.  Before I look at the code: `aloha_pot` is an important benchmark that represents a major use case - do you think the slowdown is just noise or is something meaningful going on?  Did you run the benchmark many times and consistently see this 5% slowdown?

adenzler-nvidia:
I need to rerun things with more repeats. I remember the numbers being quite unstable, but that was before your improvements with the trajectory that lifts the lid. Will follow-up with more numbers soon.

I'm still battling CPU UT failures (GPU is fine) which might block this, but couldn't figure out the issue so far.

adenzler-nvidia:
I finally found a fix to make the UTs pass on CPU as well.

Performance stays the same. We can argue both ways for the dense version - I see very clear speedups on the kernel itself on humanoid (25% faster) but it's also about 25% slower on aloha_pot. I'll try to investigate some more once I find the time, it's starting to escape my reasoning skills.

If you want this in quicker we can also split things and just get the sparse version in, if you look at n_humanoids I think the case is clear.

adenzler-nvidia:
updated numbers: 

| Benchmark | main | this branch | 
|-----------------|---------|----------------| 
| humanoid | 3,983,425 | 4,074,985 | 
| n_humanoid | 365,992 | 509,190 | 
| aloha pot | 2,143,688 | 2,146,836 | 
| apollo | 2,752,688 | 2,803,341 | 
| kitchen G1 (4k worlds) | 21,798 | 29,940 | 
| franka | 14,411,979 | 13,957,684 |

For the ones using dense the speedups are not massive but it's clearly visible in the kernel runtimes. There is further potential now in fusing in the cholesky solve as well but I will defer that to follow-up PR as it once again requires tuning a few more thresholds, block-sizes.

adenzler-nvidia:
@erikfrey @thowell considering this ready for review now. Looking forward to your feedback.

mujoco_warp/_src/solver.py
```diff
@@ -1364,100 +1365,150 @@ def update_gradient_set_h_qM_lower_sparse(
...
-  efc_D_in: wp.array2d(dtype=float),
-  efc_state_in: wp.array2d(dtype=int),
-  efc_done_in: wp.array(dtype=bool),
-  # Data out:
-  efc_h_out: wp.array3d(dtype=float),
-):
-  worldid, elementid = wp.tid()
+@wp.func
+def state_check(D: float, state: int) -> float:
+  if state == int(types.ConstraintState.QUADRATIC.value):
```
thowell: `int(types.ConstraintState.QUADRATIC.value)` -> `types.ConstraintState.QUADRATIC`

mujoco_warp/_src/solver.py
```diff
@@ -906,6 +906,7 @@ def linesearch_jaref(
 @event_scope
 def _linesearch(m: types.Model, d: types.Data):
   # mv = qM @ search
+
```
thowell: we probably don't need a blank line here (but if it is intentional we can keep it)

mujoco_warp/_src/solver.py
```diff
@@ -1670,33 +1721,34 @@ def _update_gradient(m: types.Model, d: types.Data):
...
         inputs=[m.qM_fullm_i, m.qM_fullm_j, d.qM, d.efc.done],
         outputs=[d.efc.h],
       )
     else:
-      wp.launch(
-        update_gradient_JTDAJ_dense,
-        dim=(d.nworld, lower_triangle_dim),
+      wp.launch_tiled(
+        update_gradient_JTDAJ_dense_tiled(m.nv, d.tile_sizes.jtdaj_dense, d.njmax),
+        dim=(d.nworld),
```
thowell: nit: could be `dim=d.nworld`

thowell:
@adenzler-nvidia great work! left a few minor comments. otherwise lgtm.

will leave it to @erikfrey to approve.

mujoco_warp/_src/solver.py
adenzler-nvidia: Thanks - I also raised the minimum version of warp in pyproject.toml to make sure users have the right version for this.

mujoco_warp/_src/solver.py
adenzler-nvidia: yes that looks like it got in after I removed the debugging code. reverted.

mujoco_warp/_src/solver.py
adenzler-nvidia: fixed.

adenzler-nvidia:
not sure if this closes #550 , but at least it's getting us a bit closer.

mujoco_warp/_src/types.py
```diff
@@ -59,10 +59,18 @@ class BlockDim:
   cholesky_factorize_solve: int = 32
   # solver
   update_gradient_cholesky: int = 64
+  update_gradient_JTDAJ_sparse: int = 64
+  update_gradient_JTDAJ_dense: int = 96
   # support
   mul_m_dense: int = 32
 
 
+@dataclasses.dataclass
```
erikfrey: Couple of nits / questions:

* Curious why you put this on `Data`?  Glancing through the code, this appears to be static and invariant of `nworld` - I would have expected this on `Model`, but maybe you have some plans for this?
* Curious why this is its own class?  We did this for `TileSet` because we wanted a dynamic number of them in tuples, but here there's just one?

Let me know if you have some plans for `TileSizes` that's not obvious, but if not, I would propose just having:

```py
class Model:
   ...
   jtdaj_sparse_tile_size: int
   jtdaj_dense_tile_size: int
```

mujoco_warp/_src/io.py
```diff
@@ -902,6 +902,29 @@ def _collision_sensor_check(sensor_type, sensor_id, geom_type, err_msg):
   return m
 
 
+def get_padded_sizes(nv: int, njmax: int, nworld: int, is_sparse: bool, tile_size: int):
```
erikfrey: super ultra nits:

* name this `_get_padded_sizes` to indicate private to the module
* consider just returning `njmax_padded`, `nv_padded` so that the downstream shape definitions are a bit more legible

mujoco_warp/_src/types.py
adenzler-nvidia: Great questions. It boils down to the fact that I do not have any information that is specific to the MjWarp model available during put/make_data. So no way to get the tile size from the model. And the duplication of make/put_data also make it dangerous to have a constant defined in there as it always needs to be changed in 2 places.

I moved to a constant defined in types.py - Do you also think this is the better approach? We can move it to the model once we figured out how to solve the issue above.

mujoco_warp/_src/io.py
adenzler-nvidia: makes a lot of sense, changed. Thanks!

adenzler-nvidia:
@erikfrey Could you have another look here?

mujoco_warp/_src/types.py
erikfrey: Got it - yeah this is biting us elsewhere too.  Let's discuss separately whether `make_data` and `put_data` should take `mjw.Model` as a parameter instead of `mujoco.MjModel`

PR #673, opened 2025-09-01
add Option.ccd_tolerance

thowell:


mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -33,6 +33,7 @@ def _geom_dist(m: Model, d: Data, gid1: int, gid2: int, iterations: int, multicc
   @nested_kernel
   def _gjk_kernel(
     # Model:
+    opt_ccd_tolerance: wp.array(dtype=float),
```
kbayes: You can revert this file, these unit tests should have tolerance hardwired. See https://github.com/google-deepmind/mujoco/blob/main/test/engine/engine_collision_gjk_test.cc#L43

thowell: done

erikfrey (approved):

PR #674, opened 2025-09-01
uv

thowell:


.github/workflows/ci.yml
```diff
@@ -83,8 +82,6 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install mujoco --pre -f https://py.mujoco.org/
```
btaba, line 86: you'll want to keep a "uv pip install" here as well no?

pyproject.toml
```diff
@@ -33,6 +33,13 @@ dependencies = [
     "warp-lang",
 ]
 
+[tool.uv.pip]
```
btaba: I think it's generally recommended to use a uv-index

 https://github.com/google-deepmind/mujoco_playground/blob/229fb1f50105f771ada2c6a27045638281ce5336/pyproject.toml#L42 

to avoid having a global search path for all packages

btaba (changes requested):

.github/workflows/ci.yml
thowell: mujoco is not imported so it is not necessary to have the latest version of mujoco here

btaba: but you're running pip install right below, which has a dependency on mujoco? maybe add a separate build rule for kernel-analyzer?

thowell: updated to just pip install absl-py for kernel-analyzer

pyproject.toml
thowell: done

btaba (approved):
Nice!

adenzler-nvidia:
I'm late to the party - are we planning on having a uv.lock file in the repo? such that we can fix the dependencies?

PR #675, opened 2025-09-01
Fix tolerance bug in EPA

kbayes:


mujoco_warp/_src/collision_gjk.py
```diff
@@ -1209,7 +1209,7 @@ def _epa(tolerance: float, epa_iterations: int, pt: Polytope, geom1: Geom, geom2
   upper2 = FLOAT_MAX
   idx = int(-1)
   pidx = int(-1)
-  epsilon = wp.where(is_discrete, 1e-15, tolerance * tolerance)
+  epsilon = wp.where(is_discrete, 1e-15, tolerance)
```
erikfrey, line 1212: Should we add a cylinder test to catch this?

kbayes: Yes, but I need to overhaul the unit tests to be able to verify it. I added the exact repro conditions here to the bug, so I can pick up on it in a follow up.

thowell:
@kbayes @erikfrey from offline discussion: Kyle is going to follow up with improved tests -- approving this pr as is and leaving test updates as future work

thowell (approved):

Issue #664, opened 2025-08-29
Invalid memory access in plane_convex


adenzler-nvidia:
Running compute sanitizer on the tests gives me the following error: 

========= Invalid __global__ read of size 4 bytes
=========     at T1 wp::load<wp::vec_t<(unsigned int)3, float>>(T1 *)+0x1dd40 in array.h:908
=========     by thread (0,0,0) in block (0,0,0)
=========     Access at 0x327f6c26d4 is out of bounds
=========     and is 209 bytes after the nearest allocation at 0x327f6c2600 of size 4 bytes
=========         Device Frame: plane_convex_0(int, Geom_78ae42a3, Geom_78ae42a3, int, float, float, int, wp::vec_t<(unsigned int)5, float>, wp::vec_t<(unsigned int)2, float>, wp::vec_t<(unsigned int)2, float>, wp::vec_t<(unsigned int)5, float>, wp::vec_t<(unsigned int)2, int>, wp::array_t<int>, wp::array_t<float>, wp::array_t<wp::vec_t<(unsigned int)3, float>>, wp::array_t<wp::mat_t<(unsigned int)3, (unsigned int)3, float>>, wp::array_t<float>, wp::array_t<wp::vec_t<(unsigned int)5, float>>, wp::array_t<wp::vec_t<(unsigned int)2, float>>, wp::array_t<wp::vec_t<(unsigned int)2, float>>, wp::array_t<wp::vec_t<(unsigned int)5, float>>, wp::array_t<int>, wp::array_t<wp::vec_t<(unsigned int)2, int>>, wp::array_t<int>)+0x1dd40 in collision_primitive.py:1047
=========         Device Frame: _primitive_narrowphase_builder__locals___primitive_narrowphase_a5ae0264_cuda_kernel_forward+0x19560 in collision_primitive.py:2858

It does look like we're not guarding i for the number of vertices of the convex. Failure is in CollisionTest.test_collision19

Kenny-Vilella:
This has been solved, I do not see the error in compute-sanitizer anymore.
Closing this issue.

PR #677, opened 2025-09-02
Fix in narrowphase for illegal read

Kenny-Vilella:
As stated in #664, compute-sanitizer reports illegal read in the narrowphase when running the testsuite.
I also saw illegal read when running the Apollo benchmark.
In Newton, we also had issues for specific conditions where the G1 was totally unstable because of these illegal read.

This is a tentative fix.
I did not see measurable impact on our benchmarks.


Kenny-Vilella:
@kbayes  I am not so familiar with this part of the code.
So please let me know if the fix i s not appropriate, or if there is a better way to guard against the illegal read.

kbayes:
@Kenny-Vilella thanks for diagnosing. That's strange, I feel like there should be an underlying bug. Is there any issues here: https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/collision_gjk.py#L196

Kenny-Vilella:
@kbayes  Not sure if there is an issue there, at least I never see anything yet from there.
I only saw out of bound memory read in the `plane_convex` function.

The recent change by Tobias mitigated the consequence and we do not see explosion anymore in Newton, but there is still some out of bound memory read.

Kenny-Vilella:
@kbayes  Or maybe this change is better?
Newton seems happy, but I am really not sure about all this code.

mujoco_warp/_src/collision_primitive.py
```diff
@@ -314,17 +314,17 @@ def plane_convex(plane_normal: wp.vec3, plane_pos: wp.vec3, convex: Geom) -> Tup
         idx = convex.graph[vert_globalid + subidx]
         support = wp.dot(plane_pos_local - convex.vert[convex.vertadr + idx], n)
         dist_mask = wp.where(support > threshold, 0.0, -_HUGE_VAL)
-        ap = a - convex.vert[convex.vertadr + i]
```
adenzler-nvidia, line 317: This really just looks like a type compared to the other loops, so I think we're good here.

thowell: @Kenny-Vilella great find!

kbayes:
Yeah, I think this looks good now. Was there ever a repro we could use as a test?

adenzler-nvidia:
This came up in one of the newton examples with a Unitree G1. Not sure if we can extract a repro out of that.. will merge this to now to fix the newton CI. Thanks all!

PR #678, opened 2025-09-02
Add SDF-mesh collision

quagla:


mujoco_warp/_src/collision_sdf.py
```diff
@@ -22,6 +22,7 @@
 from .collision_primitive import contact_params
 from .collision_primitive import write_contact
 from .math import make_frame
+from .ray import _ray_mesh
```
thowell: maybe change `_ray_mesh` to `ray_mesh`?

mujoco_warp/_src/collision_sdf.py
```diff
@@ -628,14 +731,38 @@ def _sdf_narrowphase(
...
+  attr1, g1_plugin_id, volume_data1, mesh_data1 = get_sdf_params(
     volume_ids, oct_aabb, plugin, plugin_attr, type1, geom1.size, g1_plugin, geom_dataid[g1]
   )
 
-  attr2, g2_plugin_id, volume_data2 = get_sdf_params(
+  attr2, g2_plugin_id, volume_data2, mesh_data2 = get_sdf_params(
     volume_ids, oct_aabb, plugin, plugin_attr, type2, geom2.size, g2_plugin, geom_dataid[g2]
   )
 
+  mesh_data1.nmeshface = nmeshface
```
thowell, line 742: would it makes sense to instantiate `MeshData` in `get_sdf_params` instead of returning an uninitialized `MeshData`?

thowell: alternatively, would it makes sense to have a method `get_mesh_data` that returns an initialized `MeshData` and not have `get_sdf_params` return `MeshData`?

quagla: Maybe but this is intentional since the sensor calls `get_sdf_params` and to match mujoco's behavior we don't want to use `mesh_data` inside `sdf`, so the easy way is to leave it uninitialized and if so, don't use it in the `sdf` function.

PR #681, opened 2025-09-03
Replace wp.config.enable_backward calls with module config calls

adenzler-nvidia:
Changing the global option is bad practice, this switch does this on a per-module basis.


thowell:
would it make sense to set the global setting `wp.config.enable_backward = False` once somewhere instead of multiple module level settings?

adenzler-nvidia:
the problem with the global setting is that it changes everything else as well - say you have an application that imports MjWarp. Once the global statement is hit in MjWarp, all the future compilations from the parent app will not generate code for the backwards pass as well. That might not be desired or expected behavior.

We can do the global config one in the benchmarks or tests, as we don't expect anyone to integrate those into a larger application.

PR #684, opened 2025-09-03
Follow-up: compile time fixes

adenzler-nvidia:
Looks like we had a big regressions on the time to run unittests because the per-module backward pass disable did not disable it for the module=unique kernels.

This fixes the issue, but is rather verbose. I'll look into providing a cleaner solution with the warp team.


adenzler-nvidia:
thanks - quick follow-up: pytest on a cold warp cache is now 3 minutes again vs 36 minutes before.. Sorry about that.

PR #695, opened 2025-09-05
fix _contact_match

thowell:
add missing early return

thanks @kevinzakka!


kevinzakka:
Confirming this fixes my issue with the sensor.

btaba (approved):

PR #697, opened 2025-09-08
Add in array memory for multiccd

kbayes:


mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -170,6 +192,17 @@ def _gjk_kernel(
   face_index = wp.array(shape=(6 * iterations,), dtype=int)
   face_map = wp.array(shape=(6 * iterations,), dtype=int)
   horizon = wp.array(shape=(6 * iterations,), dtype=int)
+  polygon = wp.array(shape=(150,), dtype=wp.vec3)
```
thowell, line 195: should we make the value `150` here and below a constant or an option?

mujoco_warp/_src/collision_gjk.py
```diff
@@ -1845,59 +1863,83 @@ def _polygon_clip(
...
       if inside2:
-        clipped[nclipped] = Q
+        clipped_out[nclipped] = Q
         nclipped += 1
 
     # swap clipped and polygon
-    tmp = polygon
-    polygon = clipped
-    clipped = tmp
+    tmp = polygon_out
```
thowell, line 1882: do we have a test that this swap works correctly?

mujoco_warp/_src/collision_gjk.py
```diff
@@ -1845,59 +1863,83 @@ def _polygon_clip(
...
   for i in range(npolygon):
-    witness2[i] = polygon[i]
+    witness2[i] = polygon_out[i]
     witness1[i] = witness2[i] - dir
   return npolygon, witness1, witness2
 
 
+@wp.func
+def _set_edge(
+  vert1: wp.array(dtype=(wp.vec3)), vert2: wp.array(dtype=(wp.vec3)), start: int, end: int, face_out: wp.array(dtype=(wp.vec3))
```
thowell: is it intentional to have `(` `)` around `wp.vec3` in the arrays here?

thowell:
do we have testspeed results for before and after these changes to evaluate the impact on performance?

mujoco_warp/_src/collision_gjk.py
kbayes: Fixed.

mujoco_warp/_src/collision_gjk.py
kbayes: Yes, there's tests in place already.

mujoco_warp/_src/collision_gjk_test.py
kbayes: Actually, I think this is throwaway code. I'll be deleted once I refactor the unit tests.

kbayes:
@thowell This is still behind a disabled flag, so no testspeed results yet. The goal is to compare results to box-box.

PR #700, opened 2025-09-09
add collision sensors

thowell:
add collision sensors
- distance
- normal
- fromto

mujoco documentation: https://mujoco.readthedocs.io/en/latest/XMLreference.html#collision-sensors
mujoco reference: https://github.com/google-deepmind/mujoco/blob/ea660f4c13b14c6e562d2f12b3e17ea24322ad0a/src/engine/engine_sensor.c#L550

notes:
- sensors with collisions including height fields are not implemented for mujoco or in this pr for mujoco warp
- sensors with collisions including planes are not implemented for mujoco warp
- `geom_distance` currently does not utilize margins with GJK for spheres (points) or capsules (lines)

### DIFF
--- a/mujoco_warp/_src/sensor.py
+++ b/mujoco_warp/_src/sensor.py
@@ -21,8 +21,10 @@ from . import math
 from . import ray
 from . import smooth
 from . import support
+from .collision_primitive import _geom
 from .collision_sdf import get_sdf_params
 from .collision_sdf import sdf
+from .support import geom_distance
 from .types import MJ_MINVAL
 from .types import ConeType
 from .types import ConstraintType
@@ -50,6 +52,7 @@ wp.set_module_options({"enable_backward": False})
 @wp.func
 def _write_scalar(
   # Model:
+  sensor_type: wp.array(dtype=int),
   sensor_datatype: wp.array(dtype=int),
   sensor_adr: wp.array(dtype=int),
   sensor_cutoff: wp.array(dtype=float),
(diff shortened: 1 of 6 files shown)


kbayes:
@thowell just waiting for you to fix the tests

mujoco_warp/_src/sensor.py
```diff
@@ -576,10 +631,151 @@ def _sensor_pos(
...
+          epa_vert2_in[tid],
+          epa_vert_index1_in[tid],
+          epa_vert_index2_in[tid],
+          epa_face_in[tid],
+          epa_pr_in[tid],
+          epa_norm2_in[tid],
+          epa_index_in[tid],
+          epa_map_in[tid],
+          epa_horizon_in[tid],
+          multiccd_polygon_in[tid],
```
kbayes, line 747: I think multiccd will always be off for the sensor, so I think you can get away with putting empty arrays here.

mujoco_warp/_src/sensor.py
```diff
@@ -576,10 +631,151 @@ def _sensor_pos(
...
+          mesh_polymapnum,
+          mesh_polymap,
+          pos2,
+          geom_xmat_in[worldid, geom2id],
+        )
+
+        dist_new, _, witness1_new, witness2_new = ccd(
+          False,  # no multiccd
+          tolerance,
+          margin,
```
kbayes: This should be called cutoff. Margin is a different concept.

thowell: this variable name is from the mujoco c implementation https://github.com/google-deepmind/mujoco/blob/9a5acbd662e6e4525bfc9422e93eb5002a6c6a68/src/engine/engine_sensor.c#L555

should we update c code?

mujoco_warp/_src/sensor.py
thowell: added a todo for now

mujoco_warp/_src/sensor.py
thowell: updated `margin` -> `cutoff`

kbayes (approved):

PR #702, opened 2025-09-09
Draft: Delete the core collision functions and use those from newton

nvtw:
Delete the core collision functions and use those from newton (collision_primitive_core.py was moved into newton)

ATTENTION: This introduces a new dependency for mujoco_warp: The new dependency targets the main branch of the newton repo


adenzler-nvidia:
should we close this? I think the implementations diverged a lot in recent times..

nvtw:
Yes, let's delete it

PR #703, opened 2025-09-09
Avoid slow pickle in physics loop in viewer.

erikfrey:
I looked closely at the simulate UI components and they seem to exclusively modify `mjm.opt`, so this change is safe for general use.  Tested it out on my local desktop and the viewer is indeed much smoother with this change.

### DIFF
--- a/mujoco_warp/viewer.py
+++ b/mujoco_warp/viewer.py
@@ -22,9 +22,9 @@ Example:
 """
 
 import ast
+import copy
 import enum
 import logging
-import pickle
 import sys
 import time
 from typing import Sequence, Union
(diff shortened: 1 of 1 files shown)


mujoco_warp/viewer.py
```diff
@@ -203,9 +204,9 @@ def _main(argv: Sequence[str]) -> None:
         wp.copy(d.qvel, wp.array([mjd.qvel.astype(np.float32)]))
         wp.copy(d.time, wp.array([mjd.time], dtype=wp.float32))
 
-        hash = pickle.dumps(mjm)
-        if hash != mjm_hash:
-          mjm_hash = hash
+        # if the user changed an option in the MuJoCo Simulate UI, go ahead and recompile the step
+        if mjm.opt != opt:
```
btaba, line 208: does this comparison work for realz?

erikfrey: for realz!

PR #706, opened 2025-09-10
io.put_model check for implicit integrators and fluid model

thowell:
currently missing and implementation of `mjd_inertiaBoxFluid`
https://github.com/google-deepmind/mujoco/blob/4b18009f47559a23e2c90b662769e0118b77c61e/src/engine/engine_derivative.c#L1270 

this pr provides a check in `io.put_model` for an `mjModel` with an implicit integrator and fluid model.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -122,6 +122,12 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   if mjm.opt.noslip_iterations > 0:
     raise NotImplementedError(f"noslip solver not implemented.")
 
+  if (mjm.opt.viscosity > 0 or mjm.opt.density > 0) and mjm.opt.integrator in [
+    mujoco.mjtIntegrator.mjINT_IMPLICITFAST,
+    mujoco.mjtIntegrator.mjINT_IMPLICIT,
+  ]:
+    raise NotImplementedError(f"Implicit integrators and fluid model not implemented.")
+
   # TODO(team): remove after _update_gradient for Newton uses tile operations for islands
   nv_max = 60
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/io.py
```diff
@@ -122,6 +122,12 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   if mjm.opt.noslip_iterations > 0:
     raise NotImplementedError(f"noslip solver not implemented.")
 
+  if (mjm.opt.viscosity > 0 or mjm.opt.density > 0) and mjm.opt.integrator in [
```
erikfrey: nit: for literal sequences prefer tuple, e.g. `( ... , ...)` vs `[ ... , ...]`

mujoco_warp/_src/io.py
```diff
@@ -122,6 +122,12 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
...
+    mujoco.mjtIntegrator.mjINT_IMPLICITFAST,
+    mujoco.mjtIntegrator.mjINT_IMPLICIT,
+  ]:
+    raise NotImplementedError(f"Implicit integrators and fluid model not implemented.")
```
erikfrey, line 129: nit: f-string without any substitution (interestingly this warns in my vscode) - you can remove the `f`

mujoco_warp/_src/io.py
thowell: changed

mujoco_warp/_src/io.py
thowell: removed here and also fixed a similar case above

PR #708, opened 2025-09-12
disablebit.spring and disablebit.damper updates

thowell:
following an update to mujoco: mjDSBL_PASSIVE -> mjDSBL_SPRING, mjDSBL_DAMPER
https://github.com/google-deepmind/mujoco/commit/b092563cab47ee765dc4292c788e39a109e7fdfe#diff-bdecef5b553097863be51390f5babf464b32bae3e94c0b35d5aee1b17c7457de

the implementation is updated to utilize the separate disablebits


mujoco_warp/_src/derivative.py
```diff
@@ -139,40 +139,40 @@ def deriv_smooth_vel(m: Model, d: Data, flg_forward: bool = True):
...
-        qMi,
-        qMj,
-        actuation_enabled,
-        passive_enabled,
-      ],
-      outputs=[d.qM_integration],
-    )
 
-  if passive_enabled:
+  wp.launch(
```
erikfrey: OOC why did you remove the if guard?  This might yield an extra unneeded kernel launch no?

mujoco_warp/_src/forward.py
```diff
@@ -393,7 +393,7 @@ def euler(m: Model, d: Data):
   """Euler integrator, semi-implicit in velocity."""
 
   # integrate damping implicitly
-  if not m.opt.disableflags & DisableBit.EULERDAMP.value:
+  if not m.opt.disableflags & DisableBit.EULERDAMP and not m.opt.disableflags & DisableBit.DAMPER:
```
erikfrey:
```
if not m.opt.disableflags & (DisableBit.EULERDAMP | DisableBit.DAMPER):
```

mujoco_warp/_src/forward.py
```diff
@@ -522,10 +522,11 @@ def implicit(m: Model, d: Data):
...
 
   # compile-time constants
-  passive_enabled = not m.opt.disableflags & (DisableBit.SPRING.value | DisableBit.DAMPER.value)
-  actuation_enabled = (not m.opt.disableflags & DisableBit.ACTUATION.value) and m.actuator_affine_bias_gain
+  dsbl_actuation = m.opt.disableflags & DisableBit.ACTUATION.value
+  dsbl_spring = m.opt.disableflags & DisableBit.SPRING.value
+  dsbl_damper = m.opt.disableflags & DisableBit.DAMPER.value
 
-  if passive_enabled or actuation_enabled:
+  if not dsbl_actuation or not dsbl_spring or not dsbl_damper:
```
erikfrey: OK this one is optional      :-)

```
if ~(x | ~(DisableBit.ACTUATION | DisableBit.SPRING | DisableBit.DAMPER)):
```

Great book btw: https://en.wikipedia.org/wiki/Hacker%27s_Delight

mujoco_warp/_src/passive.py
```diff
@@ -39,6 +39,9 @@ def _spring_damper_dof_passive(
   # Data in:
   qpos_in: wp.array2d(dtype=float),
   qvel_in: wp.array2d(dtype=float),
+  # In:
```
erikfrey: nit: if you just pass in disabelflags then this can live under `# Model:`

mujoco_warp/_src/derivative.py
thowell: restored the guard

we do need to set `qM_integration`, so this performed if actuation and damper are both turned off. 

added a todo for directly utilizing `qM` if both actuation and damper settings are turned off

mujoco_warp/_src/forward.py
thowell: done

mujoco_warp/_src/forward.py
thowell: cool! -- done

mujoco_warp/_src/passive.py
thowell: done

PR #709, opened 2025-09-12
modify `_factor_solve_i_dense` to store factorization

### LINKED ISSUE #656: NaNs when using CG solver
Hello,

I'm using mujoco playground in dev mode, [which currently depends directly on the main branch of MJX on github for `mujoco-mjx`](https://github.com/google-deepmind/mujoco_playground/blob/63c863df65a5bf59647eb54d21a6f1f54bb91cb4/pyproject.toml#L55). Currently, stepping scenes with the CG solver and `impl="warp"` leads to NaNs, no such problem is seen when using Newton, non-warp implementations or the 3.3.5 version of warp.

Naturally pre-release versions are expected to be incomplete, but wanted to report a heads up based on the recommendation of @thowell.

Reproduction:

Install mujoco-mjx from the main branch of the mujoco repo.

``` python

import mujoco
from mujoco import mjx
import jax
print(f"Current backend: {jax.default_backend()}")

xml="""
<mujoco>
  <worldbody>
    <body>
      <joint type="hinge"  axis="0 1 0"/>
      <geom type="cylinder" pos="1 0 0" size=".1 .05"/>
    </body>
  </worldbody>
</mujoco>

"""
 
solver=mujoco.mjtSolver.mjSOL_CG

mj_model = mujoco.MjModel.from_xml_string(xml)

mj_model.opt.solver = solver

mjx_model = mjx.put_model(mj_model, impl="warp")
mjx_data = mjx.make_data(mj_model, impl="warp", nconmax=30)

jit_step = jax.jit(mjx.step)

mjx_data = jit_step(mjx_model, mjx_data)
print(mjx_data.qpos)
mjx_data = jit_step(mjx_model, mjx_data)
print(mjx_data.qpos)

```
Swap out solver for Newton, stepping is resolved without Nans.

thowell:
currently the factorization of the inertia matrix `qLD` is not stored in the dense case (`is_sparse=False`). this is problematic since the `CG` solver assumes there is a valid factorization when `smooth.solve_m` is called in `solve._update_gradient`.

this pr updates `smooth._factor_solve_i_dense` to store the factorization computed in the `smooth.cholesky_factorize_solve` kernel.

additionally, `smooth_test.test_factor_solve_i` is updated.

should resolve #656


mujoco_warp/_src/smooth_test.py
```diff
@@ -361,17 +361,34 @@ def test_factor_solve_i(self, sparse):
...
+      d.qLDiagInv.fill_(wp.inf)
 
     res = wp.zeros((1, mjm.nv), dtype=float)
     vec = wp.ones((1, mjm.nv), dtype=float)
 
     mjwarp._src.smooth.factor_solve_i(m, d, d.qM, d.qLD, d.qLDiagInv, res, vec)
-
     _assert_eq(res.numpy()[0], np.linalg.solve(qM, vec.numpy()[0]), "qM \\ 1")
 
+    if sparse:
```
erikfrey, line 371: Nice - this is a great fix but I'm a bit confused by this test!

It seems like we are (correctly) updating our expectation that `factor_solve_i` (both dense and sparse) now should write to `L` - why are you explicitly running `factor_i_sparse` and `factor_i_dense` - shouldn't you just check `qLD` (and `qLDiagInv` if sparse) after runnning `factor_solve_i` above?

erikfrey (commented):
Great find - just one question about your test design

mujoco_warp/_src/smooth_test.py
thowell: updated the test:
- in the sparse case `qLD` and `qLDiagInv` are compared with the fields in `mjd`
- in the dense case `qLD` is compared to the result from `np.linalg.cholesky(qM)`

erikfrey (approved):
Looks great thanks.

PR #711, opened 2025-09-12
New unroll_test and some test/import cleanup.

### LINKED ISSUE #100: Improve test coverage of functions and models
There are a few gaps in our testing that aren't addressed well by typical coverage statistics.

First, we have a few parts of the code that are prone to error, for example the functions in `io.py` can break when C MuJoCo data structures change.  We should beef up testing for `io.py` and possibly `constraint.py` which have important shared fields that are being actively futzed with right now on the C MuJoCo side.

Second, we should introduce more test fixture models that exercise parts of the code that are currently not well tested, e.g. some collision functions, diverse constraints in diverse dense/sparse configurations, etc.

erikfrey:
This adds a new `unroll_test` that unrolls trajectories within a scene and checks for some coarse expected final state.

These tests are meant to catch subtle numerical / physics stability regressions.  MuJoCo has many complex algorithms that interact in surprising ways.  Some physics bugs only arise over long horizons of hundreds or thousands of physics steps.  These tests ought to catch some of those regressions.

Along the way this also became a bit of a test / import cleanup PR as well:

* Refactor `test_util.py` which was becoming a bit of a dumping ground:
  * Single function for model parameter overriding now shared by unit tests, `viewer` and `testspeed`.
  * `fixture` moved to `test_data` and API to create/fetch test fixtures cleaned up
  * benchmark functions removed from `test_util` into a new file `benchmark.py`
  * `test_util` deleted and removed from public API
* Some mild test import cleanup: prefer public import for public functions/types, otherwise private import
* `ApptronikApolloTerrain` added to `asv` benchmarks now that hfields are working better
* Standardize import shortcut: `import mujoco_warp as mjw`
* Clone `aloha_pot` into `test_data` for `unroll_test`: this won't grow the repo but it'll add a 1-2MB to the pip package

This PR closes #100 - although we don't have a good way to measure test coverage (Warp kernels aren't easy to instrument for coverage), with `unroll_test` and the effort over the past month or two to improve test coverage in general, I think we are in a pretty good place.

### DIFF
--- a/mujoco_warp/_src/benchmark.py
+++ b/mujoco_warp/_src/benchmark.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
-"""Utilities for becnchmarking MuJoCo Warp."""
+"""Utilities for benchmarking MuJoCo Warp."""
 
 import importlib
 import os
--- a/mujoco_warp/test_data/__init__.py
+++ b/mujoco_warp/test_data/__init__.py
@@ -40,8 +40,8 @@ def fixture(
   """Loads MuJoCo MjModel / MjData and corresponding mjw.Model / mjw.Data.
 
   Args:
-    path: path to the XML file to load, relative to the `test_data` root
-    xml: XML string to load (instead of fname)
+    path: path to the XML file to load, relative to the `test_data` root.
+    xml: XML string to load (instead of fname).
     keyframe: if provided, the data will be reset to the specified keyframe.
     qpos_noise: initialize `qpos` with uniform noise with this magnitude.
     qvel_noise: initialize `qvel` with uniform noise with this magnitude.


mujoco_warp/_src/benchmark.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
-"""Utilities for testing."""
+"""Utilities for becnchmarking MuJoCo Warp."""
```
thowell: `becnchmarking` -> `benchmarking`

mujoco_warp/test_data/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+  ctrl_noise: Optional[float] = None,
+  qfrc_noise: Optional[float] = None,
+  xfrc_noise: Optional[float] = None,
+  mocap_noise: Optional[float] = None,
+  overrides: Union[dict[str, Any], Sequence[str]] = tuple(),
+) -> Tuple[mujoco.MjModel, mujoco.MjData, mjw.Model, mjw.Data]:
+  """Loads MuJoCo MjModel / MjData and corresponding mjw.Model / mjw.Data.
+
+  Args:
+    path: path to the XML file to load, relative to the `test_data` root
```
thowell: nit: maybe add a period at the end of the description since the other variable descriptions end with a period.

mujoco_warp/test_data/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+    mjd.mocap_pos[:] = np.random.uniform(-mocap_noise, mocap_noise, size=(mjm.nmocap, 3))
+    mjd.mocap_quat[:] = np.random.uniform(-mocap_noise, mocap_noise, size=(mjm.nmocap, 4))
+
+  if qpos_noise or qvel_noise or ctrl_noise or qfrc_noise or xfrc_noise:
+    mujoco.mj_step(mjm, mjd, 10)  # let dynamics get state significantly non-zero, non-identity rotations etc
+
+  mujoco.mj_forward(mjm, mjd)
+  mjd.qacc_warmstart = mjd.qacc
+  m = mjw.put_model(mjm)
+  override_model(m, overrides)
```
thowell, line 100: should we also apply the overrides to mjm, and before forward/step?

erikfrey: Yes, we do!  Check line 71.

PR #712, opened 2025-09-15
Change children order in the SDF octree

quagla:
https://github.com/google-deepmind/mujoco/commit/4fc69fa64d0a623ccdc4f811a7b9bdb626ae5159 changed the ordering of the octree node children in the compiler. This PR propagates the change to mjWarp.


thowell:
@quagla did the upstream mujoco change result in any mujoco warp test failures?

quagla:
> @quagla did the upstream mujoco change result in any mujoco warp test failures?

Added check that would have reported the error.

PR #713, opened 2025-09-15
Add margin support to convex collisions in GJK/EPA.

kbayes:


mujoco_warp/_src/collision_gjk.py
```diff
@@ -89,37 +89,25 @@ class SupportPoint:
 
 
 @wp.func
-def _discrete_geoms(g1: int, g2: int) -> bool:
+def _discrete_geoms(geom1: Geom, geom2: Geom, g1: int, g2: int) -> bool:
```
thowell: maybe arguments `margin1` and `margin2` instead of `geom1` and `geom2`?

mujoco_warp/_src/collision_gjk.py
```diff
@@ -89,37 +89,25 @@ class SupportPoint:
...
 
 @wp.func
-def _support_margin(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
+def _support(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
   sp = SupportPoint()
   sp.cached_index = -1
   sp.vertex_index = -1
   if geomtype == int(GeomType.SPHERE.value):
-    sp.point = geom.pos
+    sp.point = geom.pos + (0.5 * geom.margin) * geom.size[0] * dir
```
thowell, line 104: it might be helpful to add a comment noting that `geom.margin` and `geom.size` have been modified in `gjk`?

(for an unmodified sphere `geom` i would have expected the margin to be added to the radius (size[0]))

mujoco_warp/_src/collision_gjk.py
```diff
@@ -205,6 +193,8 @@ def _support(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
         sp.point = vert
     sp.point = geom.rot @ sp.point + geom.pos
 
+  if geom.margin > 0.0:
```
thowell, line 194: similar comment to the one above about noting the modified `geom`.

mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -487,6 +489,26 @@ def test_box_box_ccd2(self):
     _, ncon, _, _ = _geom_dist(m, d, 0, 1, MAX_ITERATIONS, multiccd=True)
     self.assertEqual(ncon, 4)
 
+  def test_sphere_mesh_margin(self):
```
thowell, line 491: should we add tests for cases where geom1 or geom2 have margin, but not both?

mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -487,6 +489,26 @@ def test_box_box_ccd2(self):
...
+    """Test sphere-mesh margin."""
+    _, _, m, d = test_util.fixture(
+      xml=f"""
```
thowell, line 495: don't need an f-string here

mujoco_warp/_src/collision_gjk.py
kbayes: I just precompute this in the ccd function now.

mujoco_warp/_src/collision_gjk.py
kbayes: Actually, I did it this way so GJK/EPA doesn't know if margin / size has been modified.

mujoco_warp/_src/collision_gjk.py
kbayes: See other comment.

mujoco_warp/_src/collision_gjk_test.py
kbayes: The same margin gets applied to both actually. See [here](https://github.com/google-deepmind/mujoco_warp/blob/8335d9baea84c802578d354ea83303b90981d57c/mujoco_warp/_src/collision_primitive.py#L498) for more details

mujoco_warp/_src/collision_gjk_test.py
kbayes: Thanks, they're all f-strings, so I'll just leave it for another day.

thowell (approved):

PR #714, opened 2025-09-16
constraint.py updates

thowell:
- variable names changes
- add early return to `_efc_friction_tendon`
- remove size checks for kernel launches
- read `opt_timestep` after early return
- formatting
- remove `wp.static`
- update `_efc_friction_dof` to move `wp.atomic(nf_out, worldid, 1)` before early return


erikfrey:
@thowell if I'm reading this correctly, it seems like you are changing the behavior such that `ne_*` counters no longer reflect overflow, but `nefc` still correctly reports overflow.

Can you update the PR description with what problem you're solving?  Are there any implications to changing the meaning of `ne_*` counters?  This seems more than just cleanup.

thowell:
reverted the changes that moved `wp.atomic_add` operations after early returns

updated the pr description

PR #715, opened 2025-09-17
Use model.opt.ccd_iterations instead of bespoke value

kbayes:


thowell:
@kbayes since the option `ccd_iterations` is utilized to allocate memory and this setting can be modified in the viewer without reallocating memory for `Data` fields, can we add a todo to the visualizer code to check that there is enough memory for the specified setting value?

mujoco_warp/_src/io.py
```diff
@@ -1440,17 +1436,17 @@ def padtile(x, length, dtype=None):
...
-    epa_vert2=wp.zeros(shape=(nconmax, 5 + MJ_CCD_ITERATIONS), dtype=wp.vec3),
-    epa_vert_index1=wp.zeros(shape=(nconmax, 5 + MJ_CCD_ITERATIONS), dtype=int),
-    epa_vert_index2=wp.zeros(shape=(nconmax, 5 + MJ_CCD_ITERATIONS), dtype=int),
-    epa_face=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=wp.vec3i),
-    epa_pr=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=wp.vec3),
-    epa_norm2=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=float),
-    epa_index=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=int),
-    epa_map=wp.zeros(shape=(nconmax, 6 + 6 * MJ_CCD_ITERATIONS), dtype=int),
-    epa_horizon=wp.zeros(shape=(nconmax, 6 * MJ_CCD_ITERATIONS), dtype=int),
+    epa_vert=wp.zeros(shape=(nconmax, 5 + mjm.opt.ccd_iterations), dtype=wp.vec3),
```
erikfrey, line 1436: whew, each of these vec3 arrays is ~200MB on GPU if we assume an nconmax of 300,000, which some of our becnhmarks do - so this is hitting north of 1GB, and it looks like we allocate it even for models that don't use CCD right?

@thowell  We ought to consider whether `io.put_data` / `io.make_data` puts fields if we know we'll use them, and whether doing that breaks any other use cases.

Don't need to do that in this PR but let's at least check this PR isn't slowing down or breaking any of our benchmarks with OOM: you can do `asv run ccd_flag2^!` to confirm?

`asv` will also report a `device_memory_allocated` which you can check didn't get too big.

erikfrey (approved):
OK in the abstract, but I'm worried about these big allocations so let's double check it's OK.

mujoco_warp/_src/io.py
kbayes: @erikfrey I think a reasonable thing to do is to lower these estimates and if the user hits an insufficient memory error with EPA, then we ask them to bump up opt.ccd_iterations as this will indirectly increase the preallocated memory.

kbayes:
@thowell @erikfrey OK, I was able to reduce some of the parameters down. It's still an increase of memory, but half of what it would have been.

PR #718, opened 2025-09-19
add forward tests

thowell:
adds tests from mujoco's engine_forward_tests.cc https://github.com/google-deepmind/mujoco/blob/main/test/engine/engine_forward_test.cc

fixes `_next_activation` launch in `_rk_perturb_state`

### DIFF
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -433,7 +433,17 @@ def _rk_perturb_state(m: Model, d: Data, scale: float):
     wp.launch(
       _next_activation,
       dim=(d.nworld, m.na),
-      inputs=[m.opt.timestep, d.act_t0, d.act_dot, scale, False],
+      inputs=[
+        m.opt.timestep,
+        m.actuator_dyntype,
+        m.actuator_actlimited,
+        m.actuator_dynprm,
+        m.actuator_actrange,
+        d.act_t0,
+        d.act_dot,
+        scale,
+        False,
+      ],
       outputs=[d.act],
     )
 
--- a/mujoco_warp/_src/forward_test.py
+++ b/mujoco_warp/_src/forward_test.py
@@ -29,7 +29,7 @@ from mujoco_warp import GainType
 from mujoco_warp import IntegratorType
 from mujoco_warp import test_data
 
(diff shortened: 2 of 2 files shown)


mujoco_warp/_src/forward.py
```diff
@@ -433,7 +433,17 @@ def _rk_perturb_state(m: Model, d: Data, scale: float):
     wp.launch(
       _next_activation,
       dim=(d.nworld, m.na),
-      inputs=[m.opt.timestep, d.act_t0, d.act_dot, scale, False],
+      inputs=[
```
erikfrey: nit: since we're here, we can delete the `if m.na:` yeah?

mujoco_warp/_src/forward_test.py
```diff
@@ -488,6 +488,646 @@ def _getattr(arr):
         d_arr = d_arr[: d.nefc.numpy()[0]]
       _assert_eq(d_arr, getattr(mjd, arr), arr)
 
+  @absltest.skipIf(not wp.get_device().is_cuda, "Skipping test that requires GPU.")
```
erikfrey, line 491: LOVE seeing these tests, thank you for doing this.

Let's think of using `ScopedCapture` inside a test as a last resort... I'm definitely guilty of using it, but only where my imagination comes up short on how to test expected behavior with only a few steps.  It would be nice to keep the property that our test coverage on CPU only is still pretty good.

Some of these unit tests only step 10 times or so - you can check but I think for 10 steps, the test time is not too bad if we step via CPU.

For the tests that are stepping hundreds of times, can you take a look and decide if there's anyway to reduce the number of steps and still have a meaningful test?

If after reviewing these two things, we still have some tests with the `absltest.skipIf` at the top, that's OK.

mujoco_warp/_src/forward.py
thowell: removed instances of `if m.na`

erikfrey:
@thowell is this PR still relevant?

thowell:
yes, the tests are still relevant. this pr isn't blocking release so i haven't prioritized it. to enable many of these tests to run in a reasonable time on cpu we probably need to generate keyframes that can be loaded that enable skipping most steps.

the gpu only tests are still potentially useful for locally debugging (with gpu). we could add todos for now and then merge.

PR #719, opened 2025-09-19
Organize model fields according to mjmodel, clean up put_model.

erikfrey:
Organize fields so the match MuJoCo MjModel / MjData order.

Reduce lines of code in io functions by iterating over fields that have simple/default assignments.


erikfrey:
Closing this as I'm splitting it up into a bunch of smaller PRs

PR #720, opened 2025-09-19
per-world reset

thowell:
adds `reset` field to `Data`. the values in this array determine if the corresponding world is reset when `reset_data` is called. by default values are set to `True`.

### DIFF
--- a/mujoco_warp/_src/io_test.py
+++ b/mujoco_warp/_src/io_test.py
@@ -216,6 +216,35 @@ class IOTest(parameterized.TestCase):
     for arr in d.contact.__dataclass_fields__:
       _assert_eq(getattr(d.contact, arr).numpy(), 0.0, arr)
 
+  def test_reset_data_world(self):
+    """Tests per-world reset."""
+    _MJCF = """
+    <mujoco>
+      <worldbody>
+        <body>
+          <geom type="sphere" size="1"/>
+          <joint type="slide"/>
+        </body>
+      </worldbody>
+    </mujoco>
+    """
+    mjm = mujoco.MjModel.from_xml_string(_MJCF)
+    m = mjwarp.put_model(mjm)
+    d = mjwarp.make_data(mjm, nworld=2)
+
+    # set to nonzero values
+    d.qvel = wp.array(np.array([[1.0], [2.0]]), dtype=float)
+
+    # don't reset second world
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/io_test.py
```diff
@@ -216,6 +216,35 @@ def test_reset_data(self, xml):
...
+    m = mjwarp.put_model(mjm)
+    d = mjwarp.make_data(mjm, nworld=2)
+
+    # set to nonzero values
+    d.qvel = wp.array(np.array([[1.0], [2.0]]), dtype=float)
+
+    # don't reset second world
+    d.reset = wp.array(np.array([True, False]), dtype=bool)
+
+    mjwarp.reset_data(m, d)
```
erikfrey, line 241: i would not expect such an API

I could imagine an optional parameter to `reset_data` that is such a bit mask, but I would not put it directly on `Data` itself.  Is there a reason for it to live on `Data`?

erikfrey (commented):
API question

mujoco_warp/_src/io_test.py
kevinzakka: +1 to Erik

thowell: `reset_data` is updated to take an optional argument `reset`.

currently, there is still a field `Data.reset` that has default values of 1. If the argument `reset` is None, then this array is provided to the kernels.

alternatives could be:
- if `reset` is None, create a zero stride wp.array in reset_data to pass to the kernels
- create two version of each reset kernel, one that always resets and one that takes a bit mask argument. the value of `reset` would determine which kernel to call.

is there an option here or an alternative proposal that we prefer?

erikfrey: Yeah good question - I know it's not pretty but for now I would make two separate kernels.

We don't support allocating arrays during an API call that might end up in a graph capture.  It might actually work now with our recent improvements to XLA/FFI, but we should save that for a future investigation.

thowell: updated:
- two kernels approach: one kernel resets all worlds and the other kernel resets worlds based on a bit mask
- removed `Data.reset`

thowell: added a todo to have shared `wp.func` for `_reset` kernel

erikfrey (approved):
Looks great!

PR #723, opened 2025-09-23
update collision_gjk_test.py

### LINKED ISSUE #722: Deprecation Warning in `test_box_box_contact`
Hi,

In `mujoco_warp/_src/collision_gjk_test.py`, the test `test_box_box_contact` calls `wp.normalize(x1 - x2)`, which triggers this warning:

```
Warp DeprecationWarning: Support for built-in functions called with non-Warp array types, such as lists, tuples, NumPy arrays, and others, will be dropped in the future. Use a Warp type such as `wp.vec`, `wp.mat`, `wp.quat`, or `wp.transform`.
```

For context, this warning was introduced in January 2024 (Warp `v0.11.0`), and we plan to remove this functionality in the next release `v1.10.0`. Once that happens, this test will fail.

The proposed fix is to wrap these values in `wp.vec3`, as shown below:

```diff
--- a/mujoco_warp/_src/collision_gjk_test.py
+++ b/mujoco_warp/_src/collision_gjk_test.py
@@ -257,7 +257,7 @@ def _geom_dist(m: Model, d: Data, gid1: int, gid2: int, iterations: int, multicc
       pos_out,
     ],
   )
-  return dist_out.numpy()[0], ncon_out.numpy()[0], pos_out.numpy()[0], pos_out.numpy()[1]
+  return dist_out.numpy()[0], ncon_out.numpy()[0], wp.vec3(pos_out.numpy()[0]), wp.vec3(pos_out.numpy()[1])
 
 
 class GJKTest(absltest.TestCase):
```

thowell:
- should fix #722
- remove f-strings

### DIFF
--- a/mujoco_warp/_src/collision_gjk_test.py
+++ b/mujoco_warp/_src/collision_gjk_test.py
@@ -357,8 +357,8 @@ class GJKTest(absltest.TestCase):
       """
     )
     dist, _, x1, x2 = _geom_dist(m, d, 0, 1, MAX_ITERATIONS)
-    dif = x1 - x2
-    normal = dif / np.linalg.norm(dif)
+    diff = x1 - x2
+    normal = diff / np.linalg.norm(diff)
 
     self.assertAlmostEqual(-1, dist)
     self.assertAlmostEqual(normal[0], 1)


mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -356,8 +357,10 @@ def test_box_box_contact(self):
       """
     )
     dist, _, x1, x2 = _geom_dist(m, d, 0, 1, MAX_ITERATIONS)
+    dif = x1 - x2
```
kbayes: nit: diff

thowell: done

PR #724, opened 2025-09-23
update geom input types

### LINKED ISSUE #721: Segfault in `collision_primitive._primitive_narrowphase_builder()` When Model Has No Meshes
Hi,

While running [a test for Newton](https://github.com/newton-physics/newton/blob/main/newton/tests/test_anymal_reset.py), @shi-eric found that it can hit a segfault in some cases.

After investigation, the issue occurs when the `collision_primitive._primitive_narrowphase_builder()` function receives a model with no meshes. In this situation, the `geom()` function correctly checks for `geom_dataid >= 0` before attempting to set mesh data, but this check is not enforced at the call site of `geom()`, where `mesh_*` arrays are still accessed with an index of `-1`, even though they are empty ([see code](https://github.com/google-deepmind/mujoco_warp/blob/8f870c87272888f36dcec4e337285de55e0eaa30/mujoco_warp/_src/collision_primitive.py#L1549-L1593)).

Running this test with `wp.config.mode = "debug"` helps exposing this issue through assertions.

thowell:
provide all mesh data to `geom` as arrays.

should resolve #721


kbayes:
LGTM, feel free to merge.

kbayes (approved):

PR #730, opened 2025-09-26
Return early from non-colliding configurations to fix invalid access

camevor:
In the exhaustive-search arm of plane-convex, the threshold is set up to generate contacts between a mesh and a plane only when the the separation is below a constant. This change adds an early exit to avoid searching for an infeasible contact and accessing element -1 of the vertex array.

The issue can be reproduced by viewing the below mjcf with `mjwarp-viewer` with warp debug mode and `wp.config.verify_cuda = True`.
```
<mujoco>
  <option timestep="0.01" integrator="implicitfast" ls_iterations="20"/>
  <asset>
    <mesh name="meshcube" vertex="-1 -1 1 1 -1 1 -1 1 1 1 1 1 -1 -1 -1 1 -1 -1 -1 1 -1 1 1 -1" face="3 2 0 6 7 5 3 0 1 5 1 0 5 0 4 5 3 1 5 7 3 6 0 2 6 4 0 6 2 3 6 3 7 6 5 4"/>
  </asset>
  <worldbody>
    <geom pos="0 0 0" size="5 5 5" type="plane" />
    <body>
      <geom pos="0 0 2" type="mesh" mesh="meshcube"/>
      <joint type="free"/>
    </body>
  </worldbody>
</mujoco>
```

### DIFF
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -24,10 +24,9 @@ import mujoco_warp as mjw
 from mujoco_warp import BroadphaseType
 from mujoco_warp import DisableBit
 from mujoco_warp import test_data
+from mujoco_warp._src.collision_primitive import Geom
+from mujoco_warp._src.collision_primitive import plane_convex
 from mujoco_warp.test_data.collision_sdf.utils import register_sdf_plugins
-
-
-from mujoco_warp._src.collision_primitive import Geom, plane_convex
 
 from . import types
 


nvtw (approved):
Thanks for catching this

mujoco_warp/_src/collision_primitive.py
```diff
@@ -176,7 +176,9 @@ def plane_convex(plane_normal: wp.vec3, plane_pos: wp.vec3, convex: Geom) -> Tup
       support = wp.dot(plane_pos_local - convex.vert[convex.vertadr + i], n)
       max_support = wp.max(support, max_support)
 
-    threshold = wp.max(0.0, max_support - 1e-3)
+    threshold = max_support - 1e-3
+    if max_support < 0:
+      return
```
thowell: should this return `contact_dist`, `contact_pos`, and `plane_normal`?

camevor: Yes, thanks! I'll fix it to return non-colliding values.

thowell:
@camevor great fix!

if possible, let's add a test that fails before this fix and now passes. thanks!

camevor:
@thowell @nvtw Thanks for your reviews!
I added a test for the warp func. Note that as Warp 1.10 will support negative indexing for arrays, the original code will no longer cause invalid accesses with recent Warp nightlies, but using e.g. `1.10.0.dev20250914` the invalid access will cause a crash in the test.

I also went ahead and reused the point found in the first loop to avoid the second loop.

nvtw (approved):

thowell (approved):

PR #741, opened 2025-10-01
Add more cache_kernel to avoid fetching from disk in CPU mode

### LINKED ISSUE #737: Performance issue: MuJoCo kernels continuously reload from disk on macOS
We're experiencing significant performance degradation when running the `mjlab` demo on macOS. 

**Expected behavior:**
Kernels should be loaded from cache once at program startup (as observed on Linux).

**Actual behavior:**
On macOS, kernels are continuously reloaded from disk throughout execution, causing console spam and performance issues.

**Steps to reproduce:**
1. Clone mjlab: `git clone https://github.com/mujocolab/mjlab.git`
2. Run the demo with uv: `uv run demo`
3. Observe continuous kernel loading messages in console

**Environment:**
- OS: macOS
- Comparison: Issue does not occur on Linux

**Video demonstration:** (courtesy of @brentyi)

https://github.com/user-attachments/assets/0c439ebb-c09d-4357-b863-b7a43b2f04fa

adenzler-nvidia:
Seems like we missed 2 more, fixes #737

### DIFF
--- a/mujoco_warp/_src/collision_primitive.py
+++ b/mujoco_warp/_src/collision_primitive.py
@@ -1449,7 +1449,6 @@ assert _check_primitive_collisions(), "_PRIMITIVE_COLLISIONS is in invalid order
 
 @cache_kernel
 def _create_narrowphase_kernel(primitive_collisions_types, primitive_collisions_func):
-
   # AD: no unique here:
   # * we expect this generator to be called only once per model, so no repeated compilation
   # * module="unique" is generating problems because it uses the function name as the key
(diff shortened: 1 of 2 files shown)


adenzler-nvidia:
converting to draft - looks like issues with the primitive narrowphase builder in tests. I tried using the input arguments for the caching and it seems to work but it looks like something with the warp kernel caching itself isn't working.

adenzler-nvidia:
Should be ready to go now. Using module="unique" turned out to be problematic if the module has different version based on static input. Likely never going to be an issue outside of the tests where we aren't clearing the internal kernel cache.

thowell (approved):

thowell:
@adenzler-nvidia thank you!

PR #751, opened 2025-10-06
update testspeed ctrlnoise

thowell:
- update `benchmark.ctrl_noise` to more closely match the [mujoco implementation](https://github.com/google-deepmind/mujoco/blob/9e5c5b110dc8bf578034f0843a82c4e6ab2711f6/sample/testspeed.cc#L63)
- testspeed: if not replay mode and a keyframe is provided, set `ctrls` to a trajectory of keyframe `ctrl`s (`benchmark.ctrl_noise` will utilize the keyframe `ctrl` as `midpoint`)

### DIFF
--- a/mujoco_warp/_src/benchmark.py
+++ b/mujoco_warp/_src/benchmark.py
@@ -45,28 +45,49 @@ def _sum(stack1, stack2):
 @wp.kernel
 def ctrl_noise(
   # Model:
+  opt_timestep: wp.array(dtype=float),
   actuator_ctrllimited: wp.array(dtype=bool),
   actuator_ctrlrange: wp.array2d(dtype=wp.vec2),
+  # Data in:
+  ctrl_in: wp.array2d(dtype=float),
   # In:
   ctrl_center: wp.array1d(dtype=float),
   step: int,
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/benchmark.py
```diff
@@ -45,28 +45,49 @@ def _sum(stack1, stack2):
...
-  elif actuator_ctrllimited[actid]:
-    center = (ctrlrange[1] + ctrlrange[0]) / 2.0
-    radius = (ctrlrange[1] - ctrlrange[0]) / 2.0
-  radius *= ctrlnoise
-  noise = 2.0 * halton((step + 1) * (worldid + 1), actid + 2) - 1.0
-  ctrl_out[worldid, actid] = center + radius * noise
+    midpoint = ctrl_center[actid]
+
+  # exponential convergence to midpoint at ctrlnoiserate
+  if step > 0:
```
erikfrey: I'm not totally following this bit - why do you ignore `ctrl_in` for the first step?

thowell: i don't think we need this -- replaced the if/else with the first case

PR #752, opened 2025-10-07
Merge primitive collision functions into convex collision detection (ccd).

kbayes:


adenzler-nvidia:
@kbayes could you provide some context as to why this PR is necessary/your decision making? I fear this might make it even harder to have a separate collision library.

kbayes:
@adenzler-nvidia yes, please take a closer look in case that I'm misunderstanding something. I was hoping that this PR would make things easier actually. How we're currently doing things is by having two endpoints for calling narrowphase: primitive and general. I think this simplifies things by having one endpoint: convex_narrowphase. And have its logic be to use a primitive test otherwise fallback to the general test. This removes a lot of redundant code, and the core functions aren't altered so I didn't think it would impact Newton.

adenzler-nvidia:
understood - yes looking closer at this it doesn't change any of the core functions.

My separate feedback is that I'm unsure whether it's the right thing to launch a kernel per pair type for all the possible pair types - leading to n x naconmax threads being launched over the course of this, where before it was just naconmax threads for the entirety of the primitive collisions. There is a downside to this as well due to the branching involved, so the solution is probably somewhere in the middle?

kbayes:
Yes the difference is that the primitive kernel runs a batch of naconmax and branches on each geom type pair while the convex kernel runs a batch of n x naconmax where n is statically the number of unique geom type pairs. I think we should do one or the other but not both.

erikfrey:
@kbayes I'm a bit confused by the organization scheme here... it looks like you're putting all the narrowphase functions into `collision_convex.py` - but then should we call it `narrowphase.py`?

Stepping back, one of the nice properties of the various implementations of MuJoCo is that we (kind of, sort of) have parity in function and file names.  I can look in `engine_collision_driver.c` and see more or less the same functionality as here in `collision_driver.py` - it would be nice to maintain that property.

Is there a way to maintain that property and still achieve what you're going for here?  Feel free to grab me in person.

kbayes:
@adenzler-nvidia @erikfrey did some benchmarking but found no difference in timing (that includes the geom diverse [mixed.xml](https://github.com/google-deepmind/mujoco/blob/e75c5aa83c5c2a7d500e61c7710cebb98bee9178/test/engine/testdata/collision_convex/perf/mixed.xml) model). 

In terms of where I'm going with this as this is a first step. This PR better aligns with the [table](https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_collision_driver.c#L41) logic in engine_collision_driver.c. The next steps will be to delete the legacy GJK logic, and move the kernel setup out of collision_convex and into collision_driver.py, and keep collision_convex.py has a shared util for the primitive and gjk code.

mujoco_warp/_src/collision_convex.py
```diff
@@ -44,41 +43,204 @@
...
-    idx = upper_trid_index(len(GeomType), pair[0].value, pair[1].value)
-    if pair[1] < pair[0] or idx <= prev_idx:
-      return False
-    prev_idx = idx
-  return True
-
-
-assert _check_convex_collision_pairs(), "_CONVEX_COLLISION_PAIRS is in invalid order."
+
+@wp.func
```
thowell, line 47: would it makes sense to move this function to collision driver? (we could move `ccd_kernel_builder` to collision driver too).

mujoco_warp/_src/collision_driver.py
```diff
@@ -35,6 +50,56 @@
 wp.set_module_options({"enable_backward": False})
 
 
+_CONVEX_COLLISION_PAIRS = {
```
thowell: `_COLLISION_PAIRS`?

mujoco_warp/_src/collision_driver.py
```diff
@@ -35,6 +50,56 @@
...
+  (GeomType.CYLINDER, GeomType.CYLINDER): None,
+  (GeomType.CYLINDER, GeomType.BOX): None,
+  (GeomType.CYLINDER, GeomType.MESH): None,
+  (GeomType.BOX, GeomType.BOX): box_box_wrapper,
+  (GeomType.BOX, GeomType.MESH): None,
+  (GeomType.MESH, GeomType.MESH): None,
+}
+
+
+def _check_convex_collision_pairs():
```
thowell: `_check_collision_pairs`?

mujoco_warp/_src/collision_sdf.py
```diff
@@ -18,7 +18,7 @@
 import warp as wp
 
 from . import math
-from .collision_primitive import contact_params
+from .collision_convex import contact_params
```
thowell: would it make sense to move `sdf_narrowphase` to collision_driver and then not need to import `contact_params` here?

mujoco_warp/_src/collision_driver.py
```diff
@@ -35,6 +50,56 @@
...
+  prev_idx = -1
+  for pair in _CONVEX_COLLISION_PAIRS.keys():
+    idx = upper_trid_index(len(GeomType), pair[0].value, pair[1].value)
+    if pair[1] < pair[0] or idx <= prev_idx:
+      return False
+    prev_idx = idx
+  return True
+
+
+assert _check_convex_collision_pairs(), "_CONVEX_COLLISION_PAIRS is in invalid order."
```
erikfrey: This is cool, I like it a lot.  How about we make it a unit test instead?  The main advantage is just to save users from running this on module import.  I have similar unit tests that test static properties of the code, so we have precedent - check types_test.py for example.

mujoco_warp/_src/collision_driver.py
```diff
@@ -680,11 +745,178 @@ def nxn_broadphase(m: Model, d: Data):
...
+  The convex geom types handled by this function are SPHERE, CAPSULE, ELLIPSOID, CYLINDER,
+  BOX, MESH, HFIELD.
+
+  To optimize performance, this function dynamically builds and launches a specialized
+  kernel for each type of convex collision pair present in the model, avoiding unnecessary
+  computations for non-existent pair types.
+  """
+
+  nonprimitives = {key: value for key, value in _CONVEX_COLLISION_PAIRS.items() if value == None}
+  need_mem = any(m.geom_pair_type_count[upper_trid_index(len(GeomType), g[0].value, g[1].value)] for g in nonprimitives.keys())
```
erikfrey: hmmm, why not just early return as in the original code?

e.g., say:

```py
need_ccd = any(m.geom_pair_type_count[upper_trid_index(...

if not need_cdd:
  return
```

erikfrey (changes requested):
OK I think I see what's going on here, I think this is directionally good but still has a few architectural issues.

Probably the best starting point for discussion is: I think it might still be better to have distinct narrowphase helpers for functionally distinct parts of the code, e.g. primitive vs. convex vs sdf... not sure where hfield lives in that hierarchy, if it should be grouped with convex or not.

My main argument for this is that, apparently, kernel launches are somewhat sensitive to the number of parameters, and I've been warned by folks on the Warp team that some of our kernel launches are starting to break older CUDA versions that only supported smaller stack sizes.

So in general, prefer fewer inputs/outputs where possible/practical.

Besides that, any big change like this, please do run the benchmarks before and after and let us know if you're seeing changes eitehr to step time or JIT time.  You can compare head to yours by doing:

to test your branch:

```asv run ccd_refactor^!```

against main:

```asv run```

Thanks for your patience - this is a big refactor, please grab me in chat if you think it'll save more thrashing.

mujoco_warp/_src/collision_driver.py
kbayes: MJ_CONVEX_COLLISION_FUNC to match mujoco c.

mujoco_warp/_src/collision_driver.py
kbayes: Moved to tests.

mujoco_warp/_src/collision_driver.py
kbayes: Now in unit test.

mujoco_warp/_src/collision_sdf.py
kbayes: sdf_narrowphase is kinda it's own monolith block so I don't want to touch it in this PR.

mujoco_warp/_src/collision_convex.py
kbayes: sdf_narrowphase depends on it.

kbayes:
From my perspective the user shouldn't care if a convex pair was handled by a primitive function or ccd (as long as it was handled in a performant and accurate manner). The current way duplicates logic to segregate geom pairs into two categories which doesn't necessary align with anything (i.e. box-box pairs can use ccd or the box-box primitive code). And it makes more sense to unify these into one collision function table just like in mujoco c. We could separate out the primitives into their own kernel launch but that doesn't change the size of input / outputs nor give us any improvements in performance  (actually, there might be a slight edge to unify them), so why not just merge them? SDFs are fundamentally different and should be handled separately.

erikfrey:
@kbayes ah you're right! I had assumed that unioning the two kernels would add parameters but it doesn't!  OK, that's nice.

I think I'm still leaning towards keeping them separate, actually as I was thinking of your characterization of the user - we have many kinds of users, but here's two I can think of:

1) the library user that just wants to call `mjwarp.step(...` - yes, at the limit they don't care about how anything is handled, and just want the library to simulate fast.

2) the reader of the code, for whom I hold special affection.  They will take a model (like `humanoid.xml`) and read the code to figure out what's going on, tracing down through the call graph to understand each step.  You are a formidable programmer and GJK is quite impressive but intimidating to newcomers.  It would be nice if the base case of that code path is simple and bypasses GJK's imposing variable declarations etc.

I also realized something by reading your reply, which is that you use the term "convex" to mean both convex meshes and primitive shapes (which are all convex).  This made me stop and think for a moment!  GJK is cool because you can swap in any shape that has a support function and replace the primitive collider, and I can see how that promotes the idea that everything is convex.

I think in practice, especially once the code is done and laid to rest, we usually think of these two categories of collider as quite distinct.  Sure, we could do sphere<>sphere within GJK, but we won't.

Curious what @yuvaltassa thinks - Kyle, I can see how you arrived at this POV.

erikfrey:
@kbayes what's the status of this PR?  Is it still valid or are parts of it now implemented elsewhere?

kbayes:
Closing as #1004 combines the collision tables.

Issue #747, opened 2025-10-04
GitHub Issue: Severe Performance Regression in Sensor Module Compilation (158x slower)


kevinzakka:
# Severe Performance Regression in Sensor Module Compilation (158x slower)

## Description
Critical performance regression in `mujoco_warp._src.sensor` module compilation time introduced between commits `5a1b755` and `815ae38`. The sensor module compilation time increased from **3 seconds to 477 seconds** (158x slower), causing total compilation time to increase from 25.55 seconds to 500 seconds (~20x slower).

## Environment Information
```
Warp 1.10.0.dev20251004 initialized:
   Git commit: 9804f384b05dbe00019eaa2e2530d6b47333bce8
   CUDA Toolkit 12.8, Driver 13.0
   Devices:
     "cpu"      : "x86_64"
     "cuda:0"   : "NVIDIA GeForce RTX 5070 Ti Laptop GPU" (11 GiB, sm_120, mempool enabled)
   Kernel cache:
     /home/kevin/.cache/warp/1.10.0.dev20251004
```

## Reproduction Steps

Using `mjlab`:

```bash
rm -rf ~/.cache/warp/
uv run play --task Mjlab-Tracking-Flat-Unitree-G1-Play --wandb-run-path gcbc_researchers/mjlab_alpha/m98s35n7
```

Note I'm attaching a self-contained zip file containing the G1 robot with a plane and all sensors used to reproduce the mjlab setup.

## Performance Comparison

| Module | 5a1b755 (ms) | 815ae38 (ms) | Δ Time (ms) | Slowdown |
|--------|--------------|--------------|-------------|----------|
| mujoco_warp._src.smooth | 2,446.66 | 2,565.71 | +119.05 | 1.05x |
| mujoco_warp._src.collision_driver | 173.37 | 174.47 | +1.10 | 1.01x |
| _nxn_broadphase | 348.85 | 346.68 | -2.17 | 1.00x |
| _primitive_narrowphase | 446.34 | 459.47 | +13.13 | 1.03x |
| mujoco_warp._src.constraint | 997.42 | 1,076.07 | +78.65 | 1.08x |
| mujoco_warp._src.forward | 444.84 | 508.71 | +63.87 | 1.14x |
| actuator_velocity | 175.25 | 180.47 | +5.22 | 1.03x |
| mujoco_warp._src.passive | 521.98 | 548.04 | +26.06 | 1.05x |
[...]

PR #755, opened 2025-10-08
refactor collision sensor

thowell:
refactor collision sensor to follow @erikfrey's suggestion to utilize the collision driver

- ~~augments `nxn_pairid`/`nxn_pairid_filtered`/`collision_pairid` with two new elements [1]: unique collision id corresponding to row in new field `Data.sensor_collision`, [2]: unique geom collision id since collision sensors can be specified with body semantics and potentially have multiple geoms that will need to be evaluted and compared~~
- ~~add field `sensor_collision` to `Data`. this field will be computed by the collision driver and `_sensor_pos` will read from this field and utilize the information to compute values for `sensordata`~~
- augments `nxn_pairid`/`nxn_pairid_filtered`/`collision_pairid` with a new element (making the array type `wp.vec2i`): a unique collision id ~~corresponding to an element in the new field `Data.collision`~~.
- ~~add field `collision` to `Data`. this field is computed by the collision driver and `_sensor_pos` will read from this field and utilize the information to compute collision sensor values for `sensordata`. each element in this field is a `vec7`. each element is `distance` (1) and `fromto` (6).~~
- adds type `ContactType` for a new bitflag field `Contact.type` that is utilized by constraints and sensors.

notes:
- each collision sensor currently iterates over all contacts in order to compute values for `sensordata`. a TODO has been added for improving performance here by iterating over all contacts once for all of the collision sensors
- `Data.contact` and `Data.nacon` can now include and account for contacts that don't contribute to the constraints

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -1467,7 +1467,9 @@ class Contact:
     efc_address: address in efc; -1: not included                    (naconmax, ncondim)
     worldid: world id                                                (naconmax,)
     type: ContactType                                                (naconmax,)
-    geomcollisionid: id for collision                                (naconmax,)
+    geomcollisionid: i-th contact generated for geom                 (naconmax,)
+                     helps uniquely identity contact when multiple
+                     contacts are generated for geom pair
   """
 
   dist: wp.array(dtype=float)


mujoco_warp/_src/collision_convex.py
```diff
@@ -178,10 +181,16 @@ def eval_ccd_write_contact(
       points = mat3c()
       geom1.margin = margin
       geom2.margin = margin
+      if pairid[1] >= 0:
+        cutoff = (
```
erikfrey: nit: unusual formatting

maybe just put the comment above the `cutoff = 1.0e32` bit?

mujoco_warp/_src/types.py
```diff
@@ -951,10 +956,12 @@ class Model:
     nxn_geom_pair: collision pair geom ids [-2, ngeom-1]     (<= ngeom * (ngeom - 1) // 2,)
     nxn_geom_pair_filtered: valid collision pair geom ids    (<= ngeom * (ngeom - 1) // 2,)
                             [-1, ngeom - 1]
-    nxn_pairid: predefined pair id, -1 if not predefined,    (<= ngeom * (ngeom - 1) // 2,)
-                -2 if skipped
-    nxn_pairid_filtered: predefined pair id, -1 if not       (<= ngeom * (ngeom - 1) // 2,)
-                         predefined
+    nxn_pairid: predefined pair id                           (<= ngeom * (ngeom - 1) // 2, 3)
```
erikfrey: this is starting to get a little tricky to reason about... what if we introduce an enum that's like:

```py
class NxNType(enum.IntEnum):
  DEFAULT = 0
  SKIP = 1
  PAIR = 2
  SENSOR = 3
```

Then separately we have an `nxn_objid` that corresponds to the collision id or sensor id?

If want to keep it as a single array that's fine, but this way it can be a `vec2i` (type, objid) instead of a `vec3i` - WDYT?

mujoco_warp/_src/collision_convex.py
thowell: updated this comment and its formatting

mujoco_warp/_src/types.py
thowell: not sure if the proposed `enum.IntEnum` will work in all cases we need. for example:
- contact pair and distance sensor have the same geoms
- exclude contact and distance sensor have the same geoms

we could enumerate all possibilities, for example adding `SKIPSENSOR` (this would skip contact but write collision sensor), etc or maybe utilize an `enum.IntFlag` instead?

---

i think we probably need three numbers in some form in order to write contacts and write collision sensor data during the narrowphase

- [0]: contact pair id, -1 if no pair id, -2 if contact is excluded/skipped
- [1]: sensor id, -1 if no sensor

[2]: collision sensors have both geom and body semantics. we need some way of organizing geom data by body when we write collision sensor data. the current implementation writes to `sensor_collision[worldid, [1], [2]]` using the second and third numbers. an alternative approach might be to have a counter and to utilize atomic adds to increment it. another approach might be to have the third number be the body id and then search over `sensor_collision` for body id matches in the sensor function.

---

do we think there is an approach that only requires two numbers (ie, type and objid)?

erikfrey: Sure, we can use IntFlag.  Not quite sure why you need to store the bodyid in `nxn_pairid` - can't you get it from`geom_bodyid`?

kevinzakka:
Can confirm this compiles much much faster and in principle resolves #747.

thowell:
@erikfrey the pr and pr description are updated.

mujoco_warp/_src/io.py
```diff
@@ -393,47 +368,97 @@ def create_nmodel_batched_array(mjm_array, dtype, expand_dim=True):
...
+  geom_type_pair_count = np.bincount(
+    [
+      math.upper_trid_index(len(types.GeomType), int(mjm.geom_type[geom1[i]]), int(mjm.geom_type[geom2[i]]))
+      for i in np.arange(len(geom1))
+      if nxn_pairid_contact[i] > -2 or nxn_pairid_collision[i] > -1
+    ],
+    minlength=len(types.GeomType) * (len(types.GeomType) + 1) // 2,
+  )
+
+  # Disable collisions if there are no potentially colliding pairs
```
erikfrey: since you're touching this code...

it's a little weird that we're modifying the input MjModel.... can we either skip this (tbh I could argue it's not necessary / possibly harmful) or at least only modify the resulting `mjw.Model` opt?

erikfrey (approved):
OK this is lovely and moves the resulting code is easier to understand!

I know you had some concerns that sweeping across all contacts to find just the sensor contacts might be expensive.  Did you run any benchmarks?

Assuming you're satisfied with the perf, I think this code looks nice.  Just had one nit.

mujoco_warp/_src/io.py
thowell: introduced a local `disableflags` variable that is modified instead of `mjm.opt.disableflags`

thowell:
benchmarking with `humanoid/humanoid.xml`

```
mjwarp-testspeed benchmark/humanoid/humanoid.xml --nconmax=24 --njmax=64 --event_trace
```

```

Total JIT time: 0.33 s
Total simulation time: 3.04 s
Total steps per second: 2,692,632
Total realtime factor: 13,463.16 x
Total time per step: 371.38 ns

    sensor_pos: 0.17
```

**adding a distance sensor**

```
  <sensor>
    <distance geom1="hand_left" geom2="hand_right"/>
  </sensor>
```

note: needed to increase `nconmax=24` (benchmark) -> `nconmax=25` to prevent overflow

```
mjwarp-testspeed benchmark/humanoid/humanoid.xml --nconmax=25 --njmax=64 --event_trace
```

```

Total JIT time: 0.34 s
Total simulation time: 8.46 s
Total steps per second: 968,807
Total realtime factor: 4,844.04 x
Total time per step: 1032.20 ns

    sensor_pos: 660.98
```

---

to improve performance we can add a kernel launch over contacts that writes to a temporary array. will follow up with new timing results when the updated implementation is ready.

thowell:
the updated implementation aggregates contact information for collision sensors before writing to `sensordata`
- adds field `Contact.geomcollisionid` to identify unique geom pair collisions -- this information is important because colliders can generate more than one contact per geom pair (eg, plane<>capsule: 2)
- adds a new kernel `_sensor_collision` that is launched over contacts by `sensor_pos`. this organizes and writes distance and fromto contact information by world, collision sensor, and geom collision id into a temporary array `sensor_collision`. this array is utilized to write into `sensordata` for distance, normal, and fromto sensors.

---

**humanoid + distance sensor**

```
mjwarp-testspeed benchmark/humanoid/humanoid.xml --nconmax=25 --njmax=64 --event_trace
```

```

Total JIT time: 0.34 s
Total simulation time: 3.06 s
Total steps per second: 2,675,142
Total realtime factor: 13,375.71 x
Total time per step: 373.81 ns

    sensor_pos: 2.41
```

mujoco_warp/_src/types.py
```diff
@@ -1448,6 +1466,8 @@ class Contact:
     geom: geom ids; -1 for flex                                      (naconmax, 2)
     efc_address: address in efc; -1: not included                    (naconmax, ncondim)
     worldid: world id                                                (naconmax,)
+    type: ContactType                                                (naconmax,)
+    geomcollisionid: id for collision                                (naconmax,)
```
erikfrey: can you expand the field description on this a bit... I think I understand it, but I had to read the PR comments.  would be nice if this was a bit more self-explanatory?

thowell: updated the docstring

PR #732, opened 2025-09-26
Make nconmax match njmax semantics in make_data/put_data.

erikfrey:
This PR does two things:

1) Makes make_data / put_data `nconmax` parameter **per-world** to match `njmax` semantics.

Besides being (hopefully) a bit more intuitive for users, this change it possible to respect `nconmax` and `njmax` in the XML in a future update - still TBD, this might be useful at some point.

2) Renames `Data.ncon` to `Data.nacon`, `Data.nconmax` to `Data.naconmax`

The idea is to correctly distinguish that `ncon` is a term reserved for **per-world** contact counts.  `nacon` will be used for counts of contacts across all worlds.

Closes #652

PR #756, opened 2025-10-09
viewer update for nconmax

thowell:
the viewer currently does not work at head.

there is a print that depends on `Data.nconmax` and this field was replaced with `Data.naconmax` in #732. 
```
    print(f"Data\n  nworld: {d.nworld} nconmax: {d.nconmax} njmax: {d.njmax}\n")
                                                 ^^^^^^^^^
AttributeError: 'Data' object has no attribute 'nconmax'. Did you mean: 'naconmax'?
```
---

this pr computes `nconmax = naconmax / nworld` and then prints `nconmax`.

### DIFF
--- a/mujoco_warp/viewer.py
+++ b/mujoco_warp/viewer.py
@@ -150,7 +150,7 @@ def _main(argv: Sequence[str]) -> None:
         f"  integrator: {integrator} graph_conditional: {m.opt.graph_conditional}"
       )
       d = mjw.put_data(mjm, mjd, nconmax=_NCONMAX.value, njmax=_NJMAX.value)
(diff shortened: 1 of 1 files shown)


thowell:
@erikfrey should we add a test for the viewer, and maybe also for testspeed, to help prevent future breakages?

erikfrey:
@thowell whew thanks for catching.  I'm ambivalent - I suppose we could test that it starts up?  But certainly wouldn't want to test any kind of correctness per se.  On the flip side, I certainly wouldn't want a release with a broken viewer!

In an ideal world we would have QA support.  Hmm, I think the most reasonable comporomise is we put together a small release checklist of steps we take to verify expected e2e behavior before any official package release.  WDYT?

PR #757, opened 2025-10-09
update `io.get_data_into`

thowell:
updates for `io.get_data_into`

- address TODOs about setting efc fields
- compute and enforce ordering for contact efc fields that is compatible with mujoco (eg, mj_contactForce expects `efc_force` elements for the contact force are contiguous. the result from mujoco warp is not necessarily contiguous so an ordering is enforced now in `get_data_into`)
- added `io_test.test_get_data_into`

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -610,7 +610,7 @@ class Option:
   ls_iterations: int
   disableflags: int
   enableflags: int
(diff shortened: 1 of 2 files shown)


thowell:
forces look reasonable

https://github.com/user-attachments/assets/97a4ded4-f7ef-4353-9b83-a234278d54c3

PR #765, opened 2025-10-11
add: `types.State`, `support.get_state`, `support.set_state`

### LINKED ISSUE #717: mj_getState, mj_setState analog for specific worlds
Often it is desirable to reset some worlds to a certain state, for example I want to reset to a stable grasp for simulating in-hand manipulation.
MuJoCo provides `mj_getState` and `mj_setState` for this use case (https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_support.c#L188).
It would be great to have an "indexed" versions of these functions in mujoco_warp.

thowell:
adds two functions and one class to the api: `get_state`, `set_state`, and `State`

`get_state` and `set_state` provide functionality similar to MuJoCo's [mj_getState](https://github.com/google-deepmind/mujoco/blob/9e0885a4fab5665bdeb3ca7c54bab5f51db5c6dc/src/engine/engine_support.c#L184) and [mj_setState](https://github.com/google-deepmind/mujoco/blob/9e0885a4fab5665bdeb3ca7c54bab5f51db5c6dc/src/engine/engine_support.c#L215). additionally, these functions take an optional argument `active` (`wp.array(dtype=bool)`) than enables per-world get/set (resolves #717)

`State` mirrors MuJoCo's [mjtState](https://mujoco.readthedocs.io/en/latest/APIreference/APItypes.html#mjtstate) struct and provides bitflags for substates.

tests are added: `support_test.test_get_state` and `support_test.test_set_state`

### DIFF
--- a/mujoco_warp/_src/support.py
+++ b/mujoco_warp/_src/support.py
@@ -668,7 +668,7 @@ def get_state(m: Model, d: Data, state: wp.array2d(dtype=float), sig: int, activ
       d.mocap_pos,
       d.mocap_quat,
       int(sig),
-      active if active is not None else wp.ones(d.nworld, dtype=bool),
+      active or wp.ones(d.nworld, dtype=bool),
     ],
     outputs=[state],
   )
@@ -801,7 +801,7 @@ def set_state(m: Model, d: Data, state: wp.array2d(dtype=float), sig: int, activ
       m.neq,
       m.nmocap,
       int(sig),
-      active if active is not None else wp.ones(d.nworld, dtype=bool),
+      active or wp.ones(d.nworld, dtype=bool),
       state,
     ],
     outputs=[


mujoco_warp/_src/support.py
```diff
@@ -532,3 +533,515 @@ def jac_dot(
...
+  mocap_pos_in: wp.array2d(dtype=wp.vec3),
+  mocap_quat_in: wp.array2d(dtype=wp.quat),
+  # In:
+  sig_in: int,
+  # Out:
+  state_out: wp.array2d(dtype=float),
+):
+  worldid = wp.tid()
+
+  _get_state_fn(
```
erikfrey: I totally get why you made two kernels that both call a `wp.func`, but maybe this is an opportunity to try out generic kernels where `active_in` is of type `Any` - WDYT?  It might let us save some code, avoid declaring a separate func and also have one kernel instead of two.

https://nvidia.github.io/warp/modules/generics.html#generic-kernels

mujoco_warp/_src/types.py
```diff
@@ -1732,3 +1732,45 @@ class Data:
 
   # actuator
   actuator_trntype_body_ncon: wp.array2d(dtype=int)  # warp only
+
+
+class State(enum.IntEnum):
```
erikfrey: nit: let's keep the enums all grouped together up at the top

mujoco_warp/_src/support.py
```diff
@@ -532,3 +533,515 @@ def jac_dot(
...
+    ctrl_out,
+    qfrc_applied_out,
+    xfrc_applied_out,
+    eq_active_out,
+    mocap_pos_out,
+    mocap_quat_out,
+  )
+
+
+def set_state(m: Model, d: Data, state: wp.array2d(dtype=float), sig: int, active: Optional = None):
```
erikfrey: let's try to maintain the property that any public API function has a full docstring with args

mujoco_warp/_src/support.py
```diff
@@ -532,3 +533,515 @@ def jac_dot(
...
+    eq_active_in,
+    mocap_pos_in,
+    mocap_quat_in,
+    worldid,
+    sig_in,
+    state_out,
+  )
+
+
+def get_state(m: Model, d: Data, state: wp.array2d(dtype=float), sig: int, active: Optional = None):
```
erikfrey: let's try to maintain the property that any public API function has a full docstring with args

mujoco_warp/_src/types.py

mujoco_warp/_src/support.py
thowell: added a docstring

mujoco_warp/_src/support.py
thowell: added a docstring

mujoco_warp/_src/support.py
thowell: updated `get_state`. please take a look and let me know if this is what you have in mind. once we confirm we like a pattern i can update `set_state`.

thowell: updated

erikfrey (approved):
Lovely!

erikfrey:
Please merge once you've sorted out whatever the CI failure is.

mujoco_warp/_src/support.py
```diff
@@ -668,7 +668,7 @@ def _get_state(
       d.mocap_pos,
       d.mocap_quat,
       int(sig),
-      active if active is not None else 0,
+      active if active is not None else wp.ones(d.nworld, dtype=bool),
```
erikfrey: nit: `active or wp.ones(...`

mujoco_warp/_src/support.py
```diff
@@ -792,7 +792,18 @@ def _set_state(
...
+    inputs=[
+      m.nq,
+      m.nv,
+      m.nu,
+      m.na,
+      m.nbody,
+      m.neq,
+      m.nmocap,
+      int(sig),
+      active if active is not None else wp.ones(d.nworld, dtype=bool),
```
erikfrey: nit: `active or wp.ones(...`

mujoco_warp/_src/support.py
thowell: done

mujoco_warp/_src/support.py
thowell: done

PR #768, opened 2025-10-13
cuda toolkit and driver version

thowell:
adds a check to `io.put_model` for compatible cuda toolkit and driver versions. 

updates the readme with recommended and minimum cuda toolkit versions.

### DIFF
--- a/mujoco_warp/_src/warp_util.py
+++ b/mujoco_warp/_src/warp_util.py
@@ -220,14 +220,6 @@ def cache_kernel(func):
   return wrapper
 
 
-def conditional_graph_supported():
-  try:
-    assert_conditional_graph_support()
-  except Exception:
-    return False
-  return True
-
-
 def check_toolkit_driver():
   if wp.context.runtime is None:
     wp.context.init()
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -530,7 +530,7 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
       ccd_iterations=mjm.opt.ccd_iterations,
       broadphase=int(broadphase),
       broadphase_filter=int(types.BroadphaseFilter.PLANE | types.BroadphaseFilter.SPHERE | types.BroadphaseFilter.OBB),
-      graph_conditional=True and warp_util.conditional_graph_supported(),
+      graph_conditional=True,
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -26,6 +26,10 @@ MJWarp is maintained by [Google DeepMind](https://deepmind.google/) and [NVIDIA]
 
 MuJoCo Warp is currently supported on Windows or Linux on x86-64 architecture (to be expanded to more platforms and architectures soon).
 
+**CUDA**
+
+The recommended CUDA version is `>=12.4` and the minimum version is `12.0`.
```
adenzler-nvidia: Looks like my AI tools lied to me and module reloading during graph capture is only supported from 12.3 upwards. So we should have 12.3 be the minspec - or we just go for only 12.4+ support.

mujoco_warp/_src/warp_util.py
```diff
@@ -226,3 +227,11 @@ def conditional_graph_supported():
...
     return False
   return True
+
+
+def check_toolkit_driver():
+  runtime = wp.context.Runtime()
+  if runtime.toolkit_version < (12, 0) or runtime.driver_version < (12, 0):
+    raise RuntimeError("Minimum CUDA version: 12.0")
+  if runtime.toolkit_version < (12, 4) or runtime.driver_version < (12, 4):
+    warnings.warn("Recommended CUDA version: >=12.4")
```
adenzler-nvidia: We should check this with the warp team, I think initializing the runtime here is quite expensive. Maybe there is a better option?

Or if we just go for 12.4+, we can reuse the assert_conditional_graph_support function.

README.md
thowell: updated minimum version to 12.4

mujoco_warp/_src/warp_util.py
thowell: updated this function to check for `wp.context.runtime` before initializing a runtime. 

i think this should now only create one runtime since `wp.contex.init` checks for a runtime https://github.com/NVIDIA/warp/blob/979c9575a2d0ba38e764aaec67183da1fb023135/warp/_src/context.py#L8153 before creating a runtime?

adenzler-nvidia: yes, looks good like this.

thowell:
removed `warp_util.conditional_graph_supported` and the check that called this function when setting `opt.graph_conditional`.

PR #771, opened 2025-10-13
Make field order match C MuJoCo.

erikfrey:
Make Option/Model/Data field order match C MuJoCo, and introduce a few tests to ensure field and docstring order match.

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -1423,12 +1423,12 @@ class Data:
   """Dynamic state that updates each step.
 
   Attributes:
+    solver_niter: number of solver iterations                   (nworld,)
     ne: number of equality constraints                          (nworld,)
     nf: number of friction constraints                          (nworld,)
     nl: number of limit constraints                             (nworld,)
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/types.py
```diff
@@ -1397,45 +1423,33 @@ class Data:
...
-    ne_connect: number of equality connect constraints          (nworld,)
-    ne_weld: number of equality weld constraints                (nworld,)
-    ne_jnt: number of equality joint constraints                (nworld,)
-    ne_ten: number of equality tendon constraints               (nworld,)
     nf: number of friction constraints                          (nworld,)
     nl: number of limit constraints                             (nworld,)
     nefc: number of constraints                                 (nworld,)
-    nsolving: number of unconverged worlds                      (1,)
     time: simulation time                                       (nworld,)
+    solver_niter: number of solver iterations                   (nworld,)
```
thowell: `solver_niter` should precede `ne`?

erikfrey: Wow, amazing catch!  OK, fixed!

PR #772, opened 2025-10-14
Bring back the margin parameter for box box collisions

nvtw:
Bring back the margin parameter for box box collisions. It got lost during the API finding process but it is needed to allow to generate box box contacts before the boxes start to overlap.

### DIFF
--- a/mujoco_warp/_src/collision_primitive_core.py
+++ b/mujoco_warp/_src/collision_primitive_core.py
@@ -589,7 +589,7 @@ def box_box(
   box2_pos: wp.vec3,
   box2_rot: wp.mat33,
   box2_size: wp.vec3,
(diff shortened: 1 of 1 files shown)


nvtw:
The kernel analyzer writes:
Error: "margin" default not allowed.
But without a default value, this is an API change. Is it ok to adjust the API @thowell ?

thowell:
think we can add `# kernel_analyzer: off` to the line that adds the `margin` input. 

probably also ok to modify the api for this function. the mujoco implementation has a `margin` argument https://github.com/google-deepmind/mujoco/blob/f7708954cb8838a2b4514168b42a00a66c3bde58/src/engine/engine_collision_box.c#L603

PR #773, opened 2025-10-14
updates for parallel linesearch

thowell:
updates for the parallel linesearch

- `nlsp ->` ~~`opt.ls_nparallel`~~ `opt.ls_iterations`
- `efc.cost_candidate` ~~-> `efc.ls_parallel_cost`~~ replaced with an inline allocation
- custom numeric field ~~`ls_nparallel`~~ `ls_parallel` is parsed from xml and sets ~~`opt.nparallel`~~ `opt.ls_parallel` in `io.put_model` ~~and the second dimension of `efc.ls_parallel_cost` in `io.make_data`/`io.put_data`~~.

add test ~~`io_test.test_ls_nparallel`~~ `io_test.test_ls_parallel` to check that ~~`ls_nparallel`~~ `ls_parallel` is correctly parsed from the xml ~~and that the relevant fields are set correctly~~.

updates for tests in solver_test.py ~~to correctly allocate `efc.ls_parallel_cost`~~

updates for viewer and testspeed to ~~correctly allocate `efc.ls_parallel_cost` and~~ print parallel linesearch information

remove todos from solver.py related to parallel linesearch that were a bit outdated

### DIFF
--- a/mujoco_warp/testspeed.py
+++ b/mujoco_warp/testspeed.py
@@ -145,15 +145,12 @@ def _main(argv: Sequence[str]):
     broadphase, filter = mjw.BroadphaseType(m.opt.broadphase).name, mjw.BroadphaseFilter(m.opt.broadphase_filter).name
     solver, cone = mjw.SolverType(m.opt.solver).name, mjw.ConeType(m.opt.cone).name
     integrator = mjw.IntegratorType(m.opt.integrator).name
-    iterations, ls_iterations, ls_parallel = m.opt.iterations, m.opt.ls_iterations, m.opt.ls_parallel
-    if ls_parallel:
-      ls_str = f"parallel steps: {ls_iterations}"
-    else:
-      ls_str = f"iterative ls_iterations: {ls_iterations}"
+    iterations, ls_iterations = m.opt.iterations, m.opt.ls_iterations
+    ls_str = f"{'parallel' if m.opt.ls_parallel else 'iterative'} linesearch iterations: {ls_iterations}"
     print(
       f"  nbody: {m.nbody} nv: {m.nv} ngeom: {m.ngeom} nu: {m.nu} is_sparse: {m.opt.is_sparse}\n"
       f"  broadphase: {broadphase} broadphase_filter: {filter}\n"
-      f"  solver: {solver} cone: {cone} iterations: {iterations} ls: {ls_str}\n"
+      f"  solver: {solver} cone: {cone} iterations: {iterations} {ls_str}\n"
       f"  integrator: {integrator} graph_conditional: {m.opt.graph_conditional}"
     )
     d = mjw.put_data(mjm, mjd, nworld=_NWORLD.value, nconmax=_NCONMAX.value, njmax=_NJMAX.value)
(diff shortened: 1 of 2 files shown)


thowell:
updated to utilize an inline array allocation

erikfrey (commented):
Hmmm - but how do you specify `ls_parallel` in the XML in the first place?

What if we just made it that there's a way to specify iterative vs. parallel in the XML, and if it's parallel, it just uses `ls_iterations` ?

thowell:
@erikfrey the pr and pr description are updated. 

`ls_parallel` is parsed from the xml and determines if parallel linesearch is utilized instead of iterative linesearch and `ls_iterations` set the number of steps sizes to evaluate by the parallel linesearch.

mujoco_warp/_src/io.py
```diff
@@ -27,6 +27,14 @@
 _TOLERANCE_F32 = 1.0e-6
 
 
+def _numeric(mjm: mujoco.MjModel, name: str, default: int = -1) -> int:
```
erikfrey: could you use mj_name2id with `mjtObj. mjOBJ_NUMERIC` instead?

mujoco_warp/_src/solver.py
```diff
@@ -516,15 +515,17 @@ def linesearch_parallel_fused(
   efc_done_in: wp.array(dtype=bool),
   njmax_in: int,
   nacon_in: wp.array(dtype=int),
-  # Data out:
-  efc_cost_candidate_out: wp.array2d(dtype=float),
+  # In:
+  nstep: int,
```
erikfrey: this should be named `opt_ls_iterations` no?

mujoco_warp/_src/solver.py
```diff
@@ -613,17 +614,18 @@ def linesearch_parallel_fused(
...
 def linesearch_parallel_best_alpha(
   # Model:
   opt_ls_parallel_min_step: float,
-  nlsp: int,
   # Data in:
   efc_done_in: wp.array(dtype=bool),
-  efc_cost_candidate_in: wp.array2d(dtype=float),
+  # In:
+  cost_in: wp.array2d(dtype=float),
+  nstep: int,
```
erikfrey: same - `opt_ls_iterations`?

mujoco_warp/testspeed.py
```diff
@@ -146,10 +146,14 @@ def _main(argv: Sequence[str]):
     solver, cone = mjw.SolverType(m.opt.solver).name, mjw.ConeType(m.opt.cone).name
     integrator = mjw.IntegratorType(m.opt.integrator).name
     iterations, ls_iterations, ls_parallel = m.opt.iterations, m.opt.ls_iterations, m.opt.ls_parallel
+    if ls_parallel:
```
erikfrey: nit: "steps" is a bit overloaded term, i would avoid using it in this context

there's plenty of options but perhaps I'd keep it simple and just use "iterations" whether it's the parallel or iterative linesearch

mujoco_warp/_src/io.py

mujoco_warp/_src/solver.py

mujoco_warp/_src/solver.py

mujoco_warp/testspeed.py
thowell: updated. "iterations" now used instead of "steps".

PR #774, opened 2025-10-14
Remove MAX_WORLDS, allow diverse model batch sizes.

erikfrey:
For batched `Model` arrays, users can pass an array of any size, and kernels will modulo-index into the batch using `worldid`.

The cost of extra modulos is within noise:

| scene                  | HEAD  | this PR | delta  |
|------------------------|-------|---------|--------|
| AlohaPot               | 567   | 566     | -0.2%  |
| AlohaSdf               | 2568  | 2520    | -1.9%  |
| ApptronikApolloFlat    | 639   | 631     | -1.3%  |
| ApptronikApolloHfield  | 967   | 946     | -2.2%  |
| ApptronikApolloTerrain | 1532  | 1555    | +1.5% |
| Cloth                  | 10638 | 10726   | +0.8%  |
| FrankaEmikaPanda       | 61    | 61      | +0%    |
| Humanoid               | 409   | 416     | +1.7%  |
| ThreeHumanoids         | 6743  | 6654    | -1.3%  |

Interestingly, I did find one place in the code where the extra modulos impact performance - `nxn_broadphase` in `ApptronikApolloTerrain` is a bottleneck because the scene has >5000 geoms.  I added some `wp.static` special casing for the filter in f2053b4624a397e3d3f9030b5f0dc277656e887d and now there's no significant perf impact for that scene.

### DIFF
--- a/mujoco_warp/_src/collision_primitive.py
+++ b/mujoco_warp/_src/collision_primitive.py
@@ -1556,6 +1556,9 @@ def _create_narrowphase_kernel(primitive_collisions_types, primitive_collisions_
     )
 
     geom1_dataid = geom_dataid[g1]
+    geom_xpos_id = worldid % geom_xpos_in.shape[0]
+    geom_xmat_id = worldid % geom_xmat_in.shape[0]
+
     geom1 = geom(
       type1,
       geom1_dataid,
@@ -1574,8 +1577,8 @@ def _create_narrowphase_kernel(primitive_collisions_types, primitive_collisions_
       mesh_polymapadr,
       mesh_polymapnum,
       mesh_polymap,
-      geom_xpos_in[worldid % geom_xpos_in.shape[0], g1],
-      geom_xmat_in[worldid % geom_xmat_in.shape[0], g1],
+      geom_xpos_in[geom_xpos_id, g1],
+      geom_xmat_in[geom_xmat_id, g1],
     )
 
     geom2_dataid = geom_dataid[g2]
@@ -1597,8 +1600,8 @@ def _create_narrowphase_kernel(primitive_collisions_types, primitive_collisions_
       mesh_polymapadr,
       mesh_polymapnum,
       mesh_polymap,
-      geom_xpos_in[worldid % geom_xpos_in.shape[0], g2],
-      geom_xmat_in[worldid % geom_xmat_in.shape[0], g2],
+      geom_xpos_in[geom_xpos_id, g2],
+      geom_xmat_in[geom_xmat_id, g2],
     )
 
     for i in range(wp.static(len(primitive_collisions_func))):
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -95,7 +95,6 @@ def _kinematics_level(
   body_pos_id = worldid % body_pos.shape[0]
   body_quat_id = worldid % body_quat.shape[0]
   jnt_axis_id = worldid % jnt_axis.shape[0]
-  jnt_pos_id = worldid % jnt_pos.shape[0]
 
   if jntnum == 0:
     # no joints - apply fixed translation and rotation relative to parent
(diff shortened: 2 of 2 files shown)


mujoco_warp/_src/collision_primitive.py
```diff
@@ -1555,7 +1559,7 @@ def _primitive_narrowphase(
     geom1 = geom(
       type1,
       geom1_dataid,
-      geom_size[worldid, g1],
+      geom_size[worldid % geom_size.shape[0], g1],
```
thowell, line 1565: we could compute `worldid % geom_size.shape[0]` once? same for `worldid % geom_xpos_in.shape[0]` and `worldid % geom_xmat_in.shape[0]` below.

mujoco_warp/_src/smooth.py
```diff
@@ -130,15 +134,15 @@ def _kinematics_level(
         qloc = wp.normalize(qloc)
         xquat = math.mul_quat(xquat, qloc)
         # correct for off-center rotation
-        xpos = xanchor - math.rot_vec_quat(jnt_pos[worldid, jntadr], xquat)
+        xpos = xanchor - math.rot_vec_quat(jnt_pos[jnt_pos_id, jntadr], xquat)
       elif jnt_type_ == JointType.SLIDE:
-        xpos += xaxis * (qpos[qadr] - qpos0[worldid, qadr])
+        xpos += xaxis * (qpos[qadr] - qpos0[worldid % qpos0.shape[0], qadr])
```
thowell: since we are looping over joints, would it makes sense to compute `worldid % qpos0.shape[0]` once before the loop?

mujoco_warp/_src/smooth.py
```diff
@@ -258,7 +262,7 @@ def _mocap(
   xpos_out: wp.array2d(dtype=wp.vec3),
   xquat_out: wp.array2d(dtype=wp.quat),
   xmat_out: wp.array2d(dtype=wp.mat33),
-  xipos_out: wp.array2d(dtype=wp.vec3),
+  xipos_out: wp.array2d(dtype=wp.vec3),  # this is a typo in the original, should be mocap_quat_out
```
thowell: should we add a todo somewhere related to this comment?

mujoco_warp/_src/smooth.py
```diff
@@ -740,7 +752,7 @@ def _qM_dense(
 ):
   worldid, dofid = wp.tid()
   bodyid = dof_bodyid[dofid]
-  # init M(i,i) with armature inertia
+  # init M(i,i) with armature inertia. dof_armature is batched
   M = dof_armature[worldid, dofid]
```
thowell: the comment above says the field is batched, but the indexing here is using `worldid` directly. should it be `worldid % dof_armature.shape[0]`?

mujoco_warp/_src/collision_primitive.py
erikfrey: Yes, fixed!

mujoco_warp/_src/smooth.py
erikfrey: Certainly - done.

mujoco_warp/_src/smooth.py
erikfrey: Huh, yeah - spurious code comment has revealed my hand - I used Gemini to write this PR!

mujoco_warp/_src/smooth.py
erikfrey: Fixed!

erikfrey:
@adenzler-nvidia is confrming this doesn't hurt any Newton benchmarks - let's wait to hear from him before merging.

adenzler-nvidia (approved):
Hacky- dirty numbers for Newton look stable, so no concerns there.

I don't particularly like how this reduces the readability of the code and can be a source of nasty bugs, but I also don't see another way to solve this.

Thanks a lot!

erikfrey:
@adenzler-nvidia I hear you - there is a regular pattern accessing batched model fields, so we can consider either a helper function or we could even add something to kernel analyzer that specifically recognizes model batched fields and enforces an access pattern.  Let's wait for our first bug and then visit these options after that happens        :-)

PR #777, opened 2025-10-15
Inline allocation of GJK/EPA arrays.

erikfrey:
With the advent of `GraphMode.WARP` we know have a way of doing inline allocation of arrays that is JAX compatible.

To kick things off, this PR moves allocation of GJK/EPA arrays from `Data` to inline, so they are allocated as needed.

The drop in device memory is dramatic (for scenes that don't use this feature), and the inline allocs seem to have no impact on step time:

| scene                  | HEAD device mem (MB) | this PR | HEAD steps ns | this PR |
|------------------------|----------------------|---------|---------------|---------|
| AlohaPot               | 3,723                | 606     | 562           | 561     |
| AlohaSdf               | 4,800                | 673     | 2,515         | 2,520   |
| ApptronikApolloFlat    | 1,612                | 471     | 643           | 641     |
| ApptronikApolloHfield  | 2,921                | 573     | 957           | 962     |
| ApptronikApolloTerrain | 7,484                | 3,995   | 1,471         | 1,500   |
| Cloth                  | 10,672               | 9,129   | 10,206        | 10,250  |
| FrankaEmikaPanda       | 807                  | 438     | 64            | 65      |
| Humanoid               | 2,183                | 465     | 402           | 416     |
| ThreeHumanoids         | 1,177                | 270     | 6,743         | 6,091   |

Note some of the figures above are misleading as they don't report the inline allocs, just the outlay of `Data` on device.  Still, for any of the scenes that don't use GJK/EPA (e.g. the Apptronik scenes), these memory gains are real.

### DIFF
--- a/mujoco_warp/_src/collision_gjk_legacy.py
+++ b/mujoco_warp/_src/collision_gjk_legacy.py
@@ -448,7 +448,7 @@ def multicontact_legacy(
   depth: float,
   normal: wp.vec3,
   ncontact: int,
-  npoly: int,
+  npolygon: int,
   perturbation_angle: float,
 ):
   # Calculates multiple contact points given the normal from EPA.
@@ -466,7 +466,7 @@ def multicontact_legacy(
   #    they do not intersect, the closest points of both polygons are found.
 
   assert ncontact <= MULTI_CONTACT_COUNT
-  assert npoly <= MULTI_POLYGON_COUNT
+  assert npolygon <= MULTI_POLYGON_COUNT
 
   if depth < -depth_extension:
     return 0, mat3c()
@@ -488,9 +488,9 @@ def multicontact_legacy(
   # in the basis of the contact frame.
   v1count = int(0)
   v2count = int(0)
-  angle_ratio = wp.static(2.0 * wp.pi) / float(npoly)
+  angle_ratio = wp.static(2.0 * wp.pi) / float(npolygon)
 
-  for i in range(npoly):
+  for i in range(npolygon):
     angle = angle_ratio * float(i)
     axis = wp.cos(angle) * dir + wp.sin(angle) * dir2
 


mujoco_warp/_src/collision_gjk.py
```diff
@@ -1314,41 +1314,41 @@ def _next(n: int, i: int) -> int:
 
 
 @wp.func
-def _polygon_quad(polygon: wp.array(dtype=wp.vec3), npolygon: int) -> wp.vec4i:
+def _polygon_quad(polygon: wp.array(dtype=wp.vec3), npoly: int) -> wp.vec4i:
```
kbayes: Could you revert npoly back? npolygon is named to reflect its association with polygon. Poly is also bad as it can be mistaken for polytope, which is also a name being used.

erikfrey: @kbayes yes for sure, happy to - can you help me with a name collision then.  Look here:

https://github.com/google-deepmind/mujoco_warp/blob/fe8f76c28ae81dcf57d73262f3c0bd933ae0deb3/mujoco_warp/_src/types.py#L1333-L1334

What should we name these two?  Ideally they conform to other size variables, e.g. start with "n" and so on.  Once I rename those, then the `npolygon` name collision won't exist anymore and we can undo the changes above.

kbayes: nmaxpolygon and nmaxmeshdeg would be ideal. Here's some comments / explanations (feel free to reword)
nmaxpolygon # max number of sides of the polygonal faces of all meshes in the model
nmaxmeshdeg # max degree of all meshes in the model

The degree deg(M) of a mesh M is defined as the following:

For any vertex v of M, let deg(v) = # of edges with v as endpoint, 

then deg(M) = max deg(v) for all v vertex of M.

PR #780, opened 2025-10-16
update `pyproject.toml` and add `uv.lock`

thowell:
pins mujoco and warp packages to the latest release versions and adds `uv.lock`

sets:
- `mujoco==3.3.7`
- `warp-lang==1.9.1`

mujoco warp can now be installed with:

```
uv sync
```
add `--extra dev --extra cuda` to install additional dependencies for testing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,14 +28,10 @@ requires-python = ">=3.9"
 dependencies = [
     "absl-py",
     "etils[epath]",
-    "mujoco==3.3.7",
+    "mujoco>=3.3.7",
     "numpy",
-    "warp-lang==1.9.1",
+    "warp-lang>=1.9.1",
 ]
-
-[[tool.uv.index]]
-name = "pypi"
-url = "https://pypi.org/simple"
 
 [[tool.uv.index]]
 name = "nvidia"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -28,11 +28,15 @@ requires-python = ">=3.9"
 dependencies = [
     "absl-py",
     "etils[epath]",
-    "mujoco>=3.3.6.dev802089588",
+    "mujoco==3.3.7",
     "numpy",
-    "warp-lang>=1.9.0",
+    "warp-lang==1.9.1",
```
shi-eric: Personally, I find the `==` too restrictive and would leave the `>=` as before (e.g. we release Warp 1.10). The `uv.lock` file can be used for version pinning.

btaba (approved):

pyproject.toml
btaba: +1

shi-eric: I should also note that using == for the warp-lang dependency will cause issues with Newton, as we would be forced to use the same version as MJ Warp, preventing the use of nightlies and the ability to use a newer Warp without coordinating with MJ Warp

thowell: updated: `==` -> `>=`

PR #781, opened 2025-10-16
Make geom_aabb randomizable

adenzler-nvidia:
With geom_rbound and geom_size being randomizable, I think we also need this for the AABB.


thowell:
makes sense to add a batch dimension for this field.

maybe we land https://github.com/google-deepmind/mujoco_warp/pull/774 and then update + merge this pr?

adenzler-nvidia:
sure this can wait for #774

adenzler-nvidia:
This is updated to reflect the changes in #774.

thowell (approved):

PR #785, opened 2025-10-21
`mul_m`: make `skip` argument optional

thowell:
`mul_m` argument `skip` is made optional and `wp.zeros(d.nworld, dtype=bool)` is allocated inline if the argument is not provided. additionally, `Data.energy_vel_mul_m_skip` and `Data.inverse_mul_m_skip` are removed.

### DIFF
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -65,7 +65,9 @@ from ._src.smooth import tendon as tendon
 from ._src.smooth import transmission as transmission
 from ._src.solver import solve as solve
 from ._src.support import contact_force as contact_force
+from ._src.support import get_state as get_state
 from ._src.support import mul_m as mul_m
+from ._src.support import set_state as set_state
 from ._src.support import xfrc_accumulate as xfrc_accumulate
 from ._src.types import BiasType as BiasType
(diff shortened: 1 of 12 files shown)


adenzler-nvidia (approved):
Thank you, looks like the cases that need the inline allocations are not very performance-sensitive? Otherwise it probably makes sense to have a quick look as a wp.zeros also does a memset.

erikfrey:
+1 to @adenzler-nvidia - `mul_m` is called every every linesearch iter, so potentially hundreds of times per physics step.  I'd benchmark this one just to confirm we're OK.  Probably are... but let's double check!

thowell:
updated the implementation to create an array with `wp.empty` (instead of `wp.zeros`) and then guard the skip early return with `wp.static`.

---

here are benchmark results with asv

**Humanoid**

```
asv run -b Humanoid
```

main (1981e9e8)
image

```
asv run mul_m^! -b Humanoid
```

this pr (f2d3e1ab)
image

**Aloha Pot**

```
asv run -b AlohaPot
```

main (1981e9e8)
image

```
asv run mul_m^! -b AlohaPot
```

this pr (f2d3e1ab)
image

PR #786, opened 2025-10-21
inline allocation for `qacc_discrete`

thowell:
allocate temporary array `qacc_discrete` inline and remove field from `Data`.

### DIFF
--- a/mujoco_warp/_src/inverse.py
+++ b/mujoco_warp/_src/inverse.py
@@ -122,8 +122,7 @@ def inverse(m: Model, d: Data):
   invdiscrete = m.opt.enableflags & EnableBit.INVDISCRETE
   if invdiscrete:
     # save discrete-time qacc and compute continuous-time qacc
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/inverse.py
```diff
@@ -122,7 +122,8 @@ def inverse(m: Model, d: Data):
   invdiscrete = m.opt.enableflags & EnableBit.INVDISCRETE
   if invdiscrete:
     # save discrete-time qacc and compute continuous-time qacc
-    wp.copy(d.qacc_discrete, d.qacc)
+    qacc_discrete = wp.empty((d.nworld, m.nv), dtype=float)
+    wp.copy(qacc_discrete, d.qacc)
```
adenzler-nvidia: same point as before about potentially using wp.clone here.

thowell: done

PR #787, opened 2025-10-21
inline allocations for sap broadphase arrays

thowell:
allocate arrays for sap broadphase inline and remove the corresponding fields from `Data`. additionally:
- update `segmented_sort` kernel to have output arrays (previously was updating the input arrays)
- for the call to `wp.utils.segmented_sort_pairs` provide start and end indices as inputs instead of start indices with num_segments + 1 to include the last index) warp reference: https://github.com/NVIDIA/warp/blob/898ef18b3dac8f733909ce91fdea6de20ea5a946/warp/_src/utils.py#L198


adenzler-nvidia:
@nvtw could you have a look please?

PR #794, opened 2025-10-24
contact sensor updates

thowell:
- allocate contact sensor arrays inline and remove the corresponding fields from `Data`
- add the number of contact sensors `nsensorcontact` to `Model`
- fix `sensor_contact_{matchid/direction/criteria}` shape last dimension `MJ_MAXCONPAIR` -> `m.opt.contact_sensor_maxmatch`

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -774,6 +774,7 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
     M_colind=wp.array(mjm.M_colind, dtype=int),
     mapM2M=wp.array(mjm.mapM2M, dtype=int),
     # warp only fields:
+    nacttrnbody=np.sum(mjm.actuator_trntype == mujoco.mjtTrn.mjTRN_BODY),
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/io.py
```diff
@@ -742,6 +742,7 @@ def _collision_sensor_check(sensor_type, sensor_id, geom_type, err_msg):
     mapM2M=wp.array(mjm.mapM2M, dtype=int),
     # warp only fields:
     nsensortaxel=sum(mjm.mesh_vertnum[mjm.sensor_objid[mjm.sensor_type == mujoco.mjtSensor.mjSENS_TACTILE]]),
+    nsensorcontact=np.sum(mjm.sensor_type == mujoco.mjtSensor.mjSENS_CONTACT),
```
erikfrey, line 780: can we now annotate `sensor_contact_adr` as having size `nsensorcontact`?

erikfrey (approved):
looks great, just one nit

mujoco_warp/_src/io.py
thowell: done

PR #796, opened 2025-10-24
update ray

thowell:
inline allocations for ray

### DIFF
--- a/mujoco_warp/_src/ray.py
+++ b/mujoco_warp/_src/ray.py
@@ -833,18 +833,20 @@ def ray(
       wp.array2d(dtype=int): IDs of intersected geoms (-1 if none).
   """
 
+  assert pnt.shape[0] == 1
   assert pnt.shape[0] == vec.shape[0]
-  assert d.ray_dist.shape[1] == d.ray_geomid.shape[1]
-  assert pnt.shape[0] == d.ray_dist.shape[1]
 
   if geomgroup is None:
     geomgroup = vec6(-1, -1, -1, -1, -1, -1)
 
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/ray.py
```diff
@@ -833,18 +833,20 @@ def ray(
...
   assert pnt.shape[0] == vec.shape[0]
-  assert d.ray_dist.shape[1] == d.ray_geomid.shape[1]
-  assert pnt.shape[0] == d.ray_dist.shape[1]
 
   if geomgroup is None:
     geomgroup = vec6(-1, -1, -1, -1, -1, -1)
 
-  d.ray_bodyexclude.fill_(bodyexclude)
+  ray_bodyexclude = wp.empty(1, dtype=int)
+  ray_bodyexclude.fill_(bodyexclude)
```
adenzler-nvidia, line 843: fill_ is unfortunately not capturable in a CUDA graph so I don't think this will work.

adenzler-nvidia: I see now that this was already there. Feel free to make a ticket and follow up for this then.

adenzler-nvidia (approved):
Thanks, see the comment about _fill. I guess that one needs a general audit of the codebase.

mujoco_warp/_src/ray.py
adenzler-nvidia: I'm being told that there is a fix for fill_ that makes it capturable coming in today so should be in the nightly soon. So we don't need to care about this.

PR #798, opened 2025-10-27
Add a benchmark scene with multiple rigid objects

Kenny-Vilella:
- Added a new Apollo scene with a box of rigid objects (cubes)

### DIFF
--- a/benchmark/apptronik_apollo/scene_objects.xml
+++ b/benchmark/apptronik_apollo/scene_objects.xml
@@ -11,6 +11,7 @@
     <texture type="skybox" builtin="gradient" rgb1="1 1 1" rgb2="1 1 1" width="800" height="800" />
     <texture name="grid" type="2d" builtin="checker" width="512" height="512" rgb1=".1 .2 .3" rgb2=".2 .3 .4"/>
     <material name="grid" texture="grid" texrepeat="1 1" texuniform="true" reflectance=".2"/>
+    <model name="apptronik_apollo" file="apptronik_apollo.xml"/>
   </asset>
 
   <worldbody>
@@ -69,5 +70,8 @@
             <geom name="geom2" type="box" size="0.03 0.03 0.03" contype="1" conaffinity="1" rgba="0 1 1 1"/>
         </body>
     </replicate>
+    
+    <!-- attach Apptronik Apollo -->
+    <attach model="apptronik_apollo" body="base_link" prefix="_"/>
   </worldbody>
 </mujoco>


Kenny-Vilella:
@erikfrey I have been told by @adenzler-nvidia that we should add benchmark scenes that are important for us.
This is one of the representative scene that are currently critical for us.

There is some jittering in the cubes collision, but not sure how to make it consistently stable.

erikfrey:
@Kenny-Vilella this is awesome, thanks.  Yes let's add it.  Let me study the XML a bit before merging just to make sure the defaults are sensible, etc.  Will get to it soon.

benchmark/apptronik_apollo/scene_objects.xml
```diff
@@ -0,0 +1,105 @@
...
+    <body name="minicube_2">
+      <joint name="joint_minicube_2" type="free"/>
+      <geom name="minicube_2" size="0.03 0.03 0.03" pos="0.6 0.0 1.0" type="box" contype="1" conaffinity="1"/>
+    </body>
+    <body name="minicube_3">
+      <joint name="joint_minicube_3" type="free"/>
+      <geom name="minicube_3" size="0.03 0.03 0.03" pos="0.7 0.0 1.0" type="box" contype="1" conaffinity="1"/>
+    </body>
+
+    <body name="cube_4">
```
hartikainen: Would it make sense to include other primitives (`sphere`, `capsule`, `ellipsoid`, `cylinder`) here too?

Kenny-Vilella: It would be great.
I tried with spheres and boxes but didn't manage to get stable simulation.
If someone more experienced than me in tuning mujoco simulation can make it work with more primitive types, I would gladly implement the change.

Kenny-Vilella:
After the merge, I cannot include the apollo anymore.
It seems that the replicate clash with the keyframe in the apollo model.
If I remove the replicate and add individual objects, then I guess the keyframe from the apollo overwrite the objects body pose and everything start at the origin.
If I do not put a position in the body but at the geom level, then it is highly unstable.

What is the proper way to deal with this?

yuvaltassa:
@Kenny-Vilella Apologies for the delay. Can you please describe more precisely what you are seeing?

> keyframe from the apollo overwrite the objects body pose and everything start at the origin

Are you saying that pose information from the robot is "leaking" into the poses of objects?

Kenny-Vilella:
@yuvaltassa I didn't track deep into the code, but it looks like something from the robot is leaking into the poses of objects.
I should have some time today or tomorrow to look deeper in what is happening.

Summary of my tests:
(1) Adding to the new xml
`<include file="apptronik_apollo.xml" />`
leads to the following error:
```
File "/home/kvilella/Documents/Codes/mujoco_warp/mujoco_warp/viewer.py", line 83, in _load_model
    spec = mujoco.MjSpec.from_file(path.as_posix())
ValueError: XML Error: Error: Keyframe 'stand' has invalid qpos size, got 26, should be 33
Element 'replicate', line 29
```
26 qpos are 7 for the root + 19 joints from the apollo.
It looks like it expects the object pose to be present in the keyframe of the robot.
If I remove the keyframe from the robot everything works as expected.

(2) Same as (1) except that I replaced all the replicate by adding directly the objects:
Just adding the 4 spheres is enough:
```
    <body name="sphere_1" pos="0.6 -0.1 0.81">
        <freejoint name="sphere_1"/>
        <geom name="geom_sphere_1" type="sphere" size="0.05" contype="1" conaffinity="1" rgba="1 0 0 1"/>
    </body>
    <body name="sphere_2" pos="0.675 -0.025 0.81">
        <freejoint name="sphere_2"/>
        <geom name="geom_sphere_2" type="sphere" size="0.05" contype="1" conaffinity="1" rgba="1 0 0 1"/>
    </body>
    <body name="sphere_3" pos="0.75 0.05 0.81">
        <freejoint name="sphere_3"/>
        <geom name="geom_sphere_3" type="sphere" size="0.05" contype="1" conaffinity="1" rgba="1 0 0 1"/>
    </body>
    <body name="sphere_4" pos="0.825 0.125 0.81">
        <freejoint name="sphere_4"/>
        <geom name="geom_sphere_4" type="sphere" size="0.05" contype="1" conaffinity="1" rgba="1 0 0 1"/>
    </body>
```

With this, I can load the model, but all the spheres start at the origin, instead of the indicated body pose.
If I remove the keyframe from the apollo model, again everything works as expected.
So yes I think that the keyframe information from the robot is leaking into the scene.

Kenny-Vilella:
For the first case, it looks like that when we replicate bodies, we use the function `mjs_attach` ([here](https://github.com/google-deepmind/mujoco/blob/main/src/user/user_api.cc#L248)), which ultimately call [this operator](https://github.com/google-deepmind/mujoco/blob/main/src/user/user_objects.cc#L1383). And in this operator we call `StoreKeyframes`, which throw an error because we did not first expand the keyframe. Does it looks like an edge case that was not accounted for?

For the second case, it looks like that the behavior is "expected". The `ExpandKeyframe` function is just filling 0.0 for missing qpos in the keyframe. I expected that the body position in the xml would be used in case of missing qpos in the keyframe, but it's probably not that easy ^^

thowell:
@Kenny-Vilella i submitted a pr to your branch that should resolve this issue https://github.com/Kenny-Vilella/mujoco_warp/pull/2

the fix was adding apptronik_apollo.xml as a model asset and then adding it to the scene with attach.

https://github.com/user-attachments/assets/3cee6102-cd3c-4f4b-a129-ce77c36a8884

thowell:
@Kenny-Vilella @hartikainen this scene helped identity an issue with attach + replicate + keyframe w/ ctrl. thanks!

Kenny-Vilella:
It's working perfectly now, thanks @thowell !

PR #803, opened 2025-10-28
tendon: inline allocations

thowell:
inline allocations for tendon arrays

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -1041,6 +1041,7 @@ class Model:
                       geom distance sensors
     nsensortaxel: number of taxels in all tactile sensors
     nsensorcontact: number of contact sensors
+    nrangefinder: number of rangefinder sensors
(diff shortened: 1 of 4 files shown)


mujoco_warp/_src/types.py
```diff
@@ -1598,10 +1598,6 @@ class Data:
     collision_pairid: ids from broadphase                       (naconmax, 2)
     collision_worldid: collision world ids from broadphase      (naconmax,)
     ncollision: collision count from broadphase
-    ten_Jdot: time derivative of tendon Jacobian                (nworld, ntendon, nv)
```
erikfrey, line 1603: these are very nice descriptions and i wouldn't be sad if you kept them as comments above where you moved them to become variable declarations

(optional comment, feel free to skip if you disagree)

thowell: added comments above the inline allocations

PR #804, opened 2025-10-28
Add GJK Intersect as fallback for numerical stability (draft only; do not merge)

kbayes:


mujoco_warp/_src/collision_gjk.py
```diff
@@ -591,6 +592,7 @@ def gjk(
   n = int(0)
   coordinates = wp.vec4()  # barycentric coordinates
   epsilon = wp.where(is_discrete, 0.0, 0.5 * tolerance * tolerance)
+  backup_gjk = cutoff == 0.0
```
thowell, line 585: is this condition necessary? for collision sensors we want to be able to run the routine with `cutoff > 0`.

kbayes: Yes, the fallback doesn't work for geom distances.

thowell:
@kbayes with #875 merged, do you recommend to close this pr?

PR #805, opened 2025-10-28
update docstrings and typing

thowell:


mujoco_warp/_src/benchmark.py
```diff
@@ -93,35 +93,31 @@ def benchmark(
   d: Data,
   nstep: int,
   ctrls: Optional[np.ndarray] = None,
-  event_trace: bool = False,
-  measure_alloc: bool = False,
-  measure_solver_niter: bool = False,
+  event_trace: Optional[bool] = False,
```
erikfrey: why add optional?  i could understand Optional if you explicitly want to consider three states, but that's not the case here?

thowell: didn't realize that `Optional` indicates that a type might be `None` -- was incorrectly thinking it meant there is a default and that the argument could optionally be provided -- will fix and check for other incorrect usage of `Optional`

thowell: fixed

PR #809, opened 2025-10-29
Fix race condition in geom_skip.

erikfrey:
This is an alternative to #808 - both fix a critical bug, but this PR doesn't sacrifice performance on `ApolloTerrain` benchmark.

| branch | ApolloTerrain benchmark step (ns) |
|--------|--------|
| main | 1629 |
| this | 1538 | 
| #808 | 1883 |

On the flip side, this PR maintains the property that `geom_xpos` and `geom_xmat` are stateful, which may induce future footguns.

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -176,7 +176,7 @@ def _geom_local_to_global(
   bodyid = geom_bodyid[geomid]
 
   if body_weldid[bodyid] == 0 and body_mocapid[body_rootid[bodyid]] == -1:
-    # geoms attached to the world are static (unless they are attached to mcocap bodies)
+    # geoms attached to the world are static (unless they are descended from mcocap bodies)
     # for such static geoms, geom_xpos and geom_xquat are computed only once during make_data
     return
 


mujoco_warp/_src/smooth.py
```diff
@@ -165,24 +165,22 @@ def _geom_local_to_global(
...
-    # Calculate only if necessary
-    xpos = xpos_in[worldid, bodyid]
-    xquat = xquat_in[worldid, bodyid]
-    geom_xpos_out[worldid, geomid] = xpos + math.rot_vec_quat(geom_pos[worldid % geom_pos.shape[0], geomid], xquat)
-    geom_xmat_out[worldid, geomid] = math.quat_to_mat(math.mul_quat(xquat, geom_quat[worldid % geom_quat.shape[0], geomid]))
 
-    if bodyid == 0:
-      # static geom pose are calculated only once
-      geom_skip_out[geomid] = True
+  if bodyid == 0:
```
yuvaltassa: This is not comprehensive, ya? What we want is `body_weldid[bodyid] == 0` which will also cover geoms in static bodies that are descendants of the world, not just direct world geoms.

This is not hypothetical hair splitting either. E.g. @kevinzakka, for reasons that I do not condone but are also not illegitimate, puts all his very fancy terrains in a static body under the world, which this optimization will miss.

yuvaltassa (commented):
LGTM modulo my comment.

The important thing is that we got rid of `geom_skip`, nicely done 🙂

mujoco_warp/_src/io.py
```diff
@@ -1016,6 +1016,14 @@ def make_data(
 
   njmax_padded, nv_padded = _get_padded_sizes(mjm.nv, njmax, nworld, mujoco.mj_isSparse(mjm), tile_size)
 
+  # static geoms (attached to the world) have their poses calculated once during make_data instead
+  # of during each physics step.  this speeds up scenes with many static geoms (e.g. terrains)
+  # TODO(team): remove this when we introduce dof islands + sleeping
+  mjd = mujoco.MjData(mjm)
```
yuvaltassa, line 1022: Pedagogical twist of fate: in a future version of MuJoCo after sleeping has landed, if the model has sleeping enabled, then the kinematics of static geoms will be untouched by `mj_kinematics` but rather come pre-cooked in the constructor on this line, basically the exact same thing as this PR 🙂

thowell:
we should update the early return check as @yuvaltassa suggests, then lgtm

thowell (approved):

mujoco_warp/_src/smooth.py
```diff
@@ -172,8 +175,8 @@ def _geom_local_to_global(
   worldid, geomid = wp.tid()
   bodyid = geom_bodyid[geomid]
 
-  if bodyid == 0:
-    # geoms attached to the world are static - they will never move
+  if body_weldid[bodyid] == 0 and body_mocapid[body_rootid[bodyid]] == -1:
+    # geoms attached to the world are static (unless they are attached to mcocap bodies)
```
yuvaltassa: descended from mocap bodies

?

erikfrey: Nice, makes sense.

mujoco_warp/_src/io.py
erikfrey: Looking forward to it!

PR #811, opened 2025-10-30
update `reset_data`

thowell:


mujoco_warp/_src/io.py
```diff
@@ -22,6 +22,7 @@
 from . import math
 from . import types
 from . import warp_util
+from .warp_util import kernel as nested_kernel
```
erikfrey, line 25: at some point we can just directly rename `warp_util.kernel` -> `warp_util.nested_kernel`

(no action for now, just thinking out loud)

mujoco_warp/_src/io.py
```diff
@@ -1684,462 +1685,252 @@ def get_data_into(
...
-  contact_dim_out[conid] = 0
-  contact_geom_out[conid] = wp.vec2i(0, 0)
-  for i in range(nefcaddress):
-    contact_efc_address_out[conid, i] = 0
-  contact_worldid_out[conid] = 0
-  contact_type_out[conid] = 0
-  contact_geomcollisionid_out[conid] = 0
-
+    worldid = contact_worldid_out[conid]
+    if wp.static(reset is not None):
```
erikfrey, line 1814: could turn these nested ifs to ands?

mujoco_warp/_src/io.py
thowell: http://github.com/google-deepmind/mujoco_warp/pull/826

mujoco_warp/_src/io.py
thowell: i think we want the nested structure since the generated cuda code will not include the early return when `reset is None`.

confirmed with the generated code for the two cases:

**nested**

```
from typing import Optional

import warp as wp

def kernel_builder(input: Optional[bool] = None):
  @wp.kernel
  def kernel(arr: wp.array(dtype=bool)):
    if wp.static(input is not None):
      if arr[0]:
        print("input is not None")

  arr = input or wp.ones(1, dtype=bool)
  wp.launch(kernel, dim=1, inputs=[arr])

kernel_builder()
```

generated .cu

```
...
    {
            // reset shared memory allocator
        wp::tile_shared_storage_t::init();

        //---------
        // primal vars
        const bool var_0 = false;
        //---------
        // forward
        // def kernel(arr: wp.array(dtype=bool)):                                                 <L 7>
        // if wp.static(input is not None):                                                       <L 8>
    }
...
```

**and**

```
from typing import Optional

import warp as wp

def kernel_builder(input: Optional[bool] = None):
  @wp.kernel
  def kernel(arr: wp.array(dtype=bool)):
    if wp.static(input is not None) and arr[0]:
      print("input is not None")

  arr = input or wp.ones(1, dtype=bool)
  wp.launch(kernel, dim=1, inputs=[arr])

kernel_builder()
```

generated .cu
```
...
   {
            // reset shared memory allocator
        wp::tile_shared_storage_t::init();

        //---------
        // primal vars
        const bool var_0 = false;
        const wp::int32 var_1 = 0;
        bool* var_2;
        bool var_3;
        bool var_4;
        const wp::str var_5 = "input is not None";
        //---------
        // forward
        // def kernel(arr: wp.array(dtype=bool)):                                                 <L 7>
        // if wp.static(input is not None) and arr[0]:                                            <L 8>
        var_2 = wp::address(var_arr, var_1);
        var_3 = wp::load(var_2);
        var_4 = var_0 && var_3;
        if (var_4) {
            // print("input is not None")                                                         <L 9>
            wp::print(var_5);
        }
    }
...
```

PR #817, opened 2025-10-30
collection of updates that should be no ops

thowell:


erikfrey (approved):
OK, I like it!  Maybe coordinate with @kbayes as I think this has some overlap with #752

kbayes (approved):

PR #822, opened 2025-10-31
Fix an off-by-factor-2 bug in box box collision code

nvtw:
Fix an off-by-factor-2 bug in box box collision code


thowell:
great fix, thanks!

nvtw:
Something is still off with the test

nvtw:
@thowell @yuvaltassa could you check if the CPU Mujoco Code also has this off-by-factor-2 problem in some box box collision configurations?

thowell:
@nvtw yes this is a bug in the implementation. great catch, thanks for reporting this bug!

we should be able to land a fix in mujoco soon.

PR #824, opened 2025-10-31
fix `Data` field indexing

thowell:


adenzler-nvidia (approved):
Thanks! this change keeps on giving :-)

Issue #736, opened 2025-09-29
Heightfield terrain is very slow to compile and has very weird contact behavior


kevinzakka:
Attaching a big terrain composed of a mixture of sub-terrains. Some of the sub-terrains are heightfields. These heightfields make the compilation take forever (~5min) and the quadruped sinks when walking on it.

[rough_terrain.xml](https://github.com/user-attachments/files/22587706/rough_terrain.xml)

surajvh-ga:
There seems to be an indexing issue in the height field collision implementation. I wish I could take credit for this but this was pointed out by ChatGPT

https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/collision_convex.py#L463C7-L465

The indexing probably needs to change from
`nrow = hfield_nrow[g1]`
`ncol = hfield_ncol[g1]`
`size = hfield_size[g1]`
to 
`nrow = hfield_nrow[geom1_dataid]`
`ncol = hfield_ncol[geom1_dataid]`
`size = hfield_size[geom1_dataid]`

Tagging the issue #815 here as well

This fixes the sinking of the robot into the ground that @kevinzakka is mentioning

thowell:
@surajvh-ga will you please submit a pr with these changes? thanks!

surajvh-ga:
@thowell I've raised a pull request

Abban-Fahim:
@kevinzakka I came across the same issue you described when trying to use the rough-terrain environment in playground. In my case it just crashes (killed by OOM, no python backtracre) when using `impl=warp`, while the flat-terrain works perfectly fine. This is despite my XML having way fewer bodies in the terrain. I suspect that this is because of my limited RAM (<= 16GB), and my process gets killed immediately. Is there any progress in speeding up the compilation step / reducing the memory footprint? I'm using the latest version of this repo, e605c406363805d949b75d50dd5ffb3d79bc3f70.

thowell:
@kevinzakka is this resolved?

kevinzakka:
So far resolved, will re-open if I encounter further issues.

PR #825, opened 2025-10-31
Incorrect indexing of hfield data when checking for collisions with height field

surajvh-ga:
Geom id was being used to index hfield data instead of the geom data id for hfield. This could have been overlooked because usually the first geom is the terrain geom which is indexed as 0 and there is usually only one height field. As soon as you add more than 1 height field or more terrain geoms, you start to see inconsistent behavior during collision with height maps, specifically sinking.

Partially related to #736 . This does not relate to the slow compilation issue, which hasn't been the case in my experience at HEAD


surajvh-ga:
@thowell Update on #736 here

thowell:
thank you for the fixes!

PR #826, opened 2025-11-01
warp_util: kernel -> nested_kernel

thowell:
instead of `from .warp_util import kernel as nested_kernel` throughout the code `warp_util.kernel` is renamed to `warp.util.nested_kernel`


thowell:
updated the kernel analyzer to version 0.5.0

erikfrey (approved):

PR #830, opened 2025-11-03
`make_data`, `put_data`: check nworld * nconmax != naconmax

thowell:
updates `io.make_data` and `io.put_data` to check if `nworld * nconmax != naconmax` when `nconmax` and `naconmax` are both set. 

adds `io_test.test_nconmax_naconmax` to test that `ValueError`s are raised.

### DIFF
--- a/mujoco_warp/_src/io_test.py
+++ b/mujoco_warp/_src/io_test.py
@@ -538,6 +538,16 @@ class IOTest(parameterized.TestCase):
 
     self.assertEqual(m.opt.ls_parallel, True)
 
+  def test_nconmax_naconmax(self):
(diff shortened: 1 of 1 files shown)


erikfrey:
The current semantics (which are not documented well) are that if `naconmax` is provided, then `nconmax` is ignored.

So another solution would be to better document this expectation.  WDYT?

thowell:
it might be confusing that both `nconmax` and `naconmax` can be provided and then one might be ignored silently.

if we prefer the existing implementation, maybe we should consider adding a warning when both arguments are provided? in this case, we can also update the [documentation](https://mujoco.readthedocs.io/en/latest/mjwarp/index.html#batch-sizes)

thowell:
let's update the docs to describe the current implementation -- closing this pr

PR #831, opened 2025-11-03
Fix issue where when a geom penetrates a height field and sinks, it stays sunk and there is no force exerted to bring it out

surajvh-ga:
1. Mujoco has code to handle the flipping of the normal direction in the above specified case: [link](https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_collision_convex.c#L1346)
2. MJX Jax backend also has this implemented: [link](https://github.com/google-deepmind/mujoco/blob/main/mjx/mujoco/mjx/_src/collision_convex.py#L1036)

Here is the [test.xml](https://github.com/user-attachments/files/23306347/test.xml) used to reproduce this issue.
You can run this with `mjwarp-viewer test.xml`

Before:

https://github.com/user-attachments/assets/34d7c76b-8fd5-42f4-b565-c9182831262c

After:

https://github.com/user-attachments/assets/ca4e67a4-73d7-4f7f-a97b-037c76fd798e

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -246,18 +246,18 @@ def ccd_kernel_builder(
         # Hfield "up" (world)
         up = geom1.rot[:, 2]
         if wp.dot(top_n, up) < 0.0:
-            top_n = -top_n
+          top_n = -top_n
 
         # Start from the CCD/GJK normal but replace it if it's not in the up half-space
         n = wp.normalize(normal)
         if wp.dot(n, up) < 1.0e-6:
-            n = top_n
+          n = top_n
 
         # (Optional but helpful) Re-project contact points onto the top face along n,
         # so pos/normal are coherent for the solver:
         for i in range(ncontact):
-            d = wp.dot((points[i] - v0), top_n)
-            points[i] = points[i] - d * top_n
+          d = wp.dot((points[i] - v0), top_n)
+          points[i] = points[i] - d * top_n
 
         normal = n
       frame = make_frame(normal)


surajvh-ga:
This probably fixes the sinking issue reported in #736

surajvh-ga:
@thowell would appreciate you checking it out

thowell:
@surajvh-ga thanks for submitting a pr!

i think we also need to consider additional cases, for example where collisions occurs on the opposite side of the height field. something like this scene:

```
<mujoco>
  <option gravity="0 0 1"/>
  <asset>
    <hfield name="hfield" nrow="16" ncol="16" size="1 1 1 0.1"/>
  </asset>
  <worldbody>
    <light type="directional" pos="0 0 1"/>
    <geom type="hfield" hfield="hfield"/>
    <body>
      <geom type="box" size=".1 .2 .3" rgba="1 0 0 1"/>
      <freejoint/>  
    </body>
  </worldbody>
  <keyframe>
    <key qpos="0 0 -1 1 0 0 0"/>
  </keyframe>
</mujoco>
```

initialized at keyframe 0.

kbayes:
The mjc_fixNormal function doesn't have to do with height fields. That's related to fixing issues with libccd and is obsolete. I'm not really a fan of doing hacks like this, even if it improves the behavior a little.  The way we currently handle height fields is flawed and the contacts we produce for deep penetrations are nonsense. I'm actively looking into how to fix this at a lower level.

thowell:
this pr is superseded by #892. 

@surajvh-ga please let us know if you encounter any issues with height fields after updating to get the latest fixes. thanks!

Issue #816, opened 2025-10-30
Optimize GJK device memory usage


erikfrey:
Franka Pick Cube overflows a 4090 at 8k worlds.

thowell:
aa6c0fb86a33ff56e32cd193205f10478ba62174

`nworld=8192`

```
mjwarp-testspeed mujoco_playground/_src/manipulation/franka_emika_panda/xmls/mjx_single_cube.xml --nworld=8192 --nconmax=24 --njmax=128 --memory
```

```
Loading model from: mujoco_playground/_src/manipulation/franka_emika_panda/xmls/mjx_single_cube.xml...
  nbody: 14 nv: 15 ngeom: 83 nu: 8 is_sparse: False graph_conditional: True
  broadphase: NXN broadphase_filter: PLANE|SPHERE|OBB
  solver: NEWTON iterations: 5 linesearch: iterative ls_iterations: 8
  cone: PYRAMIDAL integrator: IMPLICITFAST
Data
  nworld: 8192 naconmax: 196608 njmax: 128

Rolling out 1000 steps at dt = 0.005...

Summary for 8192 parallel rollouts

Total JIT time: 0.74 s
Total simulation time: 0.95 s
Total steps per second: 8,643,320
Total realtime factor: 43,216.60 x
Total time per step: 115.70 ns
Total converged worlds: 8192 / 8192

Model memory 19.35 MB (5.04% of used memory):
 mesh_vert: 4.54 MB (1.18%)
 mesh_normal: 4.54 MB (1.18%)
Data memory 268.52 MB (69.93% of used memory):
 xmat: 3.94 MB (1.03%)
 ximat: 3.94 MB (1.03%)
 geom_xpos: 7.78 MB (2.03%)
 geom_xmat: 23.34 MB (6.08%)
 cinert: 4.38 MB (1.14%)
 crb: 4.38 MB (1.14%)
 qM: 8.00 MB (2.08%)
 qLD: 7.03 MB (1.83%)
 contact.frame: 6.75 MB (1.76%)
 efc.type: 4.00 MB (1.04%)
 efc.id: 4.00 MB (1.04%)
 efc.J: 64.00 MB (16.67%)
 efc.pos: 4.00 MB (1.04%)
 efc.margin: 4.00 MB (1.04%)
 efc.D: 4.00 MB (1.04%)
 efc.vel: 4.00 MB (1.04%)
 efc.aref: 4.00 MB (1.04%)
[...]

PR #833, opened 2025-11-04
Remove multiccd along with memory when not needed

kbayes:
Multicontact is also pulled out as a separate call, which helps reduce the logic in the ccd call.

Note: This still allocates multiccd memory for box-box.

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -693,7 +693,7 @@ def convex_narrowphase(m: Model, d: Data):
   # TODO(team): fix early return?
   if not any(m.geom_pair_type_count[upper_trid_index(len(GeomType), g[0].value, g[1].value)] for g in _CONVEX_COLLISION_PAIRS):
     return
-  
+
   # set to true to enable multiccd
   use_multiccd = False
   nmaxpolygon = m.nmaxpolygon if use_multiccd else 0


thowell:
with #834 and #835 i think we can set the minimum value to 0 instead of 1 when there are no multicontact geom pairs. additionally, since we currently do not use ccd for box<>box i think we should only consider `nboxmesh` for `box_factor`. we should add a todo to restore `nboxbox` when ccd provides box<>box.

kbayes:
@thowell this is ready to review.

mujoco_warp/_src/collision_convex.py
```diff
@@ -717,7 +744,7 @@ def convex_narrowphase(m: Model, d: Data):
     g2 = geom_pair[1].value
     if m.geom_pair_type_count[upper_trid_index(len(GeomType), g1, g2)]:
       wp.launch(
-        ccd_kernel_builder(m.opt.legacy_gjk, g1, g2, m.opt.ccd_iterations, True, 1e9, g1 == GeomType.HFIELD),
+        ccd_kernel_builder(m.opt.legacy_gjk, g1, g2, m.opt.ccd_iterations, True, 1e9, g1 == GeomType.HFIELD, False),
```
thowell: if box<>box, use_multiccd should be True?

mujoco_warp/_src/io.py
```diff
@@ -28,6 +28,46 @@
...
+  nboxmesh = geom_pair_type_count[
+    math.upper_trid_index(len(types.GeomType), types.GeomType.BOX.value, types.GeomType.MESH.value)
+  ]
+  nmeshmesh = geom_pair_type_count[
+    math.upper_trid_index(len(types.GeomType), types.GeomType.MESH.value, types.GeomType.MESH.value)
+  ]
+
+  # need at least 4 (square sides) if there's a box collision needing multiccd
+  # TODO(kbayes): remove nboxbox or enable ccd for box-box collisions
+  box_factor = 4 if nboxbox + nboxmesh > 0 else 0
```
thowell, line 43: we shouldn't consider nboxbox here since ccd currently is not backing the box<>box collider

mujoco_warp/_src/io.py
```diff
@@ -28,6 +28,46 @@
...
+  nboxmesh = geom_pair_type_count[
+    math.upper_trid_index(len(types.GeomType), types.GeomType.BOX.value, types.GeomType.MESH.value)
+  ]
+  nmeshmesh = geom_pair_type_count[
+    math.upper_trid_index(len(types.GeomType), types.GeomType.MESH.value, types.GeomType.MESH.value)
+  ]
+
+  # need at least 3 (3 edges per vertex) if there's a box collision needing multiccd
+  # TODO(kbayes): remove nboxbox or enable ccd for box-box collisions
+  box_factor = 3 if nboxbox + nboxmesh > 0 else 0
```
thowell, line 63: same comment as above about nboxbox

thowell:
do we want to make the changes introduced in #835 in this pr?

mujoco_warp/_src/collision_convex.py
kbayes: Done.

mujoco_warp/_src/io.py
kbayes: I want to keep this because there's a high chance that box-box gets replaced and I don't want to have to introduce the memory later.

thowell:
@kbayes we really should not allocate the multicontact memory for box<>box collisions until this collider is provided by the general convex collision pipeline. these unnecessary memory allocations can end up being significant for applications like reinforcement learning where `nworld` can be 2048, 4096, 8192, ...

we have a todo in place, so let's remove the nboxbox from consideration for now. thanks!

thowell:
for additional context: the franka pick cube environment has box<>box collisions and we probably resolve the specific mention in #816 by not allocating unnecessary memory for multicontact arrays for when there are box<>box collisions but no collisions that currently utilize the general convex collision multicontact.

kbayes:
@thowell I don't follow your reasoning, because if it's a OOM problem now then it'll be a problem later. And, this is just one of multiple PRs to reduce the memory footprint. My argument is that it's useful to switch between the collision functions for debugging, especially before the release. Aggressively removing memory makes these switches difficult. It'll be easy to remove the memory right before release (if it's not needed) then it is to add it back (if it's needed).

kbayes:
What I'm mainly saying is that there's a high possibility of replacing the box-box collider in the near future, and I would like to get the benchmarks working with the memory budget now instead of introducing a memory regression in the future.

kbayes:
@thowell resolved offline

PR #836, opened 2025-11-04
Render

StafaH:


hartikainen:
Hey Mustafa! This is super neat. What do I need to test this? What [`warp`](https://github.com/NVIDIA/warp) version should I use?

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,240 @@
...
+  enabled_geom_ids: wp.array(dtype=int),
+  geom_type: wp.array(dtype=int),
+  geom_dataid: wp.array(dtype=int),
+  geom_size: wp.array2d(dtype=wp.vec3),
+  geom_pos: wp.array2d(dtype=wp.vec3),
+  geom_rot: wp.array2d(dtype=wp.mat33),
+  mesh_bounds_size: wp.array(dtype=wp.vec3),
+  lowers: wp.array(dtype=wp.vec3),
+  uppers: wp.array(dtype=wp.vec3),
+  groups: wp.array(dtype=wp.int32),
```
thowell: wp.int32 -> int

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,240 @@
...
+
+  lowers[world_id * bvh_ngeom + bvh_geom_local] = lower
+  uppers[world_id * bvh_ngeom + bvh_geom_local] = upper
+  groups[world_id * bvh_ngeom + bvh_geom_local] = world_id
+
+
+@wp.kernel
+def compute_bvh_group_roots(
+  bvh_id: wp.uint64,
+  group_roots: wp.array(dtype=wp.int32),
```
thowell: wp.int32 -> int

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,240 @@
...
+  root = wp.bvh_get_group_root(bvh_id, tid)
+  group_roots[tid] = root
+
+
+def build_warp_bvh(m: Model, d: Data, rc: RenderContext):
+  """Build a Warp BVH for all geometries in all worlds."""
+
+  wp.launch(
+    kernel=compute_bvh_bounds,
+    dim=(rc.nworld * rc.bvh_ngeom),
```
thowell: since the launch dimension is 1 we don't need the parentheses

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,240 @@
...
+    inputs=[
+      rc.bvh_ngeom,
+      rc.nworld,
```
thowell: can `rc.nworld != d.nworld`?

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,240 @@
...
+      d.geom_xpos,
+      d.geom_xmat,
+      rc.mesh_bounds_size,
+      rc.lowers,
+      rc.uppers,
+      rc.groups,
+    ],
+  )
+
+  bvh = wp.Bvh(
```
thowell: this could be one line

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,240 @@
...
+    kernel=compute_bvh_group_roots,
+    dim=rc.nworld,
+    inputs=[bvh.id, rc.group_roots],
+  )
+
+
+def refit_warp_bvh(m: Model, d: Data, rc: RenderContext):
+  wp.launch(
+    kernel=compute_bvh_bounds,
+    dim=(rc.nworld * rc.bvh_ngeom),
```
thowell: same comment as above regarding parentheses

mujoco_warp/_src/ray.py
```diff
@@ -206,6 +206,24 @@ def _ray_plane(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec: wp
...
+  size: wp.vec3,
+  pnt: wp.vec3,
+  vec: wp.vec3,
+) -> Tuple[bool, wp.float32, wp.vec3]:
+  """Returns distance and normal at which a ray intersects with a plane."""
+  x = _ray_plane(pos, mat, size, pnt, vec)
+  if x == wp.inf:
+    return False, wp.inf, wp.vec3(0.0, 0.0, 0.0)
+  # Local plane normal is +Z; rotate to world space
+  normal_world = mat @ wp.vec3(0.0, 0.0, 1.0)
```
thowell: does `wp.vec3(mat[0, 2], mat[1, 2], mat[2, 2])` or `mat[:, 2]` work?

mujoco_warp/_src/ray.py
```diff
@@ -243,8 +276,8 @@ def _ray_capsule(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec:
   # solve a * x^2 + 2 * b * x + c = 0
   sol, xx = _ray_quad(a, b, c)
 
-  # make sure round solution is between flat sides
-  if sol >= 0.0 and wp.abs(lpnt[2] + sol * vec[2]) <= size[1]:
+  # make sure round solution is between flat sides (must use local z component)
+  if sol >= 0.0 and wp.abs(lpnt[2] + sol * lvec[2]) <= size[1]:
```
thowell, line 281: is there a test that we should add that would have caught this?

mujoco_warp/_src/ray.py
```diff
@@ -276,6 +309,30 @@ def _ray_capsule(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec:
...
+  """Returns distance and normal at which a ray intersects with a capsule."""
+  x = _ray_capsule(pos, mat, size, pnt, vec)
+  if x == wp.inf:
+    return False, wp.inf, wp.vec3(0.0, 0.0, 0.0)
+  # Compute continuous normal: vector from closest point on axis segment to the hit point
+  lpnt, lvec = _ray_map(pos, mat, pnt, vec)
+  hit_local = lpnt + x * lvec
+  z_clamped = wp.min(size[1], wp.max(-size[1], hit_local[2]))
+  axis_point = wp.vec3(0.0, 0.0, z_clamped)
+  normal_local = wp.normalize(hit_local - axis_point)
```
thowell: instead of introducing `axis_point` can we do something like `hit_local[2] - z_clamped`?

mujoco_warp/_src/ray.py
```diff
@@ -390,6 +447,38 @@ def _ray_box(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec: wp.v
...
+  size: wp.vec3,
+  pnt: wp.vec3,
+  vec: wp.vec3,
+) -> Tuple[bool, wp.float32, wp.vec3]:
+  """Returns distance and normal at which a ray intersects with a box."""
+  x, all = _ray_box(pos, mat, size, pnt, vec)
+  if x == wp.inf:
+    return False, wp.inf, wp.vec3(0.0, 0.0, 0.0)
+
+  # Select the face by matching the closest intersection time among the 6 faces
```
thowell: "time"?

mujoco_warp/_src/ray.py
```diff
@@ -603,6 +692,38 @@ def ray_mesh(
...
+  mesh_geom_id: int,
+  pos: wp.vec3,
+  mat: wp.mat33,
+  pnt: wp.vec3,
+  vec: wp.vec3,
+  max_t: wp.float32,
+) -> Tuple[bool, wp.float32, wp.vec3, wp.float32, wp.float32, int, int]:
+  """Returns intersection information at which a ray intersects with a mesh.
+  
+  Requires wp.Mesh be constructed and their ids to be passed"""
```
thowell: `...passed."""`

mujoco_warp/_src/ray.py
```diff
@@ -603,6 +692,38 @@ def ray_mesh(
...
+  t = wp.float32(wp.inf)
```
thowell: wp.float32 -> float

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+
+from .types import Model
+from .types import Data
+from .types import GeomType
+from . import bvh
+
+wp.set_module_options({"enable_backward": False})
+
+
+def create_render_context(
```
thowell: this function is part of the public api. please add a docstring with `Args` and `Returns`

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+  d: Data,
+  nworld: int,
+  width: int,
+  height: int,
+  use_textures: bool,
+  use_shadows: bool,
+  render_rgb: bool,
+  render_depth: bool,
+  enabled_geom_groups = [0, 1, 2],
+):
```
thowell: please add a return type

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+  m: Model,
+  d: Data,
+  nworld: int,
+  width: int,
+  height: int,
+  use_textures: bool,
+  use_shadows: bool,
+  render_rgb: bool,
+  render_depth: bool,
+  enabled_geom_groups = [0, 1, 2],
```
thowell: please add a type

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+):
+  rc = RenderContext(
```
thowell: we could directly return the output from `RenderContext`

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+  tid = wp.tid()
+  total = img_w * img_h
+  if tid >= total:
+    return
+  px = tid % img_w
+  py = tid // img_w
+  aspect_ratio = float(img_w) / float(img_h)
+  u = (float(px) + 0.5) / float(img_w) - 0.5
+  v = (float(py) + 0.5) / float(img_h) - 0.5
+  h = wp.tan(fov_rad / 2.0)
```
thowell: could be `* 0.5`

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+
+@wp.kernel
+def build_primary_rays(img_w: int, img_h: int, fov_rad: float, rays_cam: wp.array(dtype=wp.vec3)):
+  tid = wp.tid()
+  total = img_w * img_h
+  if tid >= total:
+    return
+  px = tid % img_w
+  py = tid // img_w
+  aspect_ratio = float(img_w) / float(img_h)
```
thowell: we could compute `1 / float(img_h)` once and then utilize below to compute `v`

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+
+  tex_data_packed = wp.zeros((total_size,), dtype=wp.uint32)
+  tex_adr_packed = []
+
+  for i in range(mjm.ntex):
+    tex_adr_packed.append(mjm.tex_adr[i] // mjm.tex_nchannel[i])
+
+  wp.launch(
+    convert_texture_to_packed,
+    dim=(total_size,),
```
thowell: `dim=total_size`

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+    inputs=[total_size, mjm.tex_nchannel[0], wp.array(mjm.tex_data, dtype=wp.uint8), tex_data_packed]
+  )
+
+  return tex_data_packed, wp.array(tex_adr_packed, dtype=int)
+
+
+@dataclasses.dataclass
+class RenderContext:
```
thowell, line 31: maybe move this class to the top of the file?

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+  lowers: wp.array(dtype=wp.vec3)
+  uppers: wp.array(dtype=wp.vec3)
+  groups: wp.array(dtype=wp.int32)
+  group_roots: wp.array(dtype=wp.int32)
+  pixels: wp.array3d(dtype=wp.uint32)
+  depth: wp.array3d(dtype=wp.float32)
+
+  def __init__(
+    self,
+    mjm: mujoco.MjModel,
```
thowell, line 62: do we need both `mujoco.MjModel` and `Model` as inputs?

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+    m: Model,
+    d: Data,
+    nworld,
```
thowell: can `nworld` be different from `Data.nworld`?

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+    self.mesh_bvh_ids=wp.array(mesh_bvh_ids, dtype=wp.uint64)
+    self.mesh_bounds_size=wp.array(mesh_bounds_size, dtype=wp.vec3)
+    self.mesh_texcoord=wp.array(mjm.mesh_texcoord, dtype=wp.vec2)
+    self.mesh_texcoord_offsets=wp.array(mjm.mesh_texcoordadr, dtype=int)
+    self.mesh_texcoord_num=wp.array(mjm.mesh_texcoordnum, dtype=int)
+    self.tex_adr=tex_adr_packed
+    self.tex_data=tex_data_packed
+    self.tex_height=wp.array(mjm.tex_height, dtype=int)
+    self.tex_width=wp.array(mjm.tex_width, dtype=int)
+    self.lowers = wp.zeros((nworld * bvh_ngeom,), dtype=wp.vec3)
```
thowell: maybe no spaces around "=" to match the formatting above?

thowell: `(nworld * bvh_ngeom,)` -> `nworld * bvh_ngeom`

there are similar instances we can update below

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+    self.rays_cam = wp.zeros((width * height), dtype=wp.vec3)
+    wp.launch(
+      kernel=build_primary_rays,
+      dim=(width * height),
+      inputs=[width, height, self.fov_rad, self.rays_cam],
+    )
+
+    self.bvh = None
+    self.bvh_id = None
+    bvh.build_warp_bvh(m, d, self)
```
thowell: please add a new line at the end of the file

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,248 @@
...
+    self.lowers = wp.zeros((nworld * bvh_ngeom,), dtype=wp.vec3)
+    self.uppers = wp.zeros((nworld * bvh_ngeom,), dtype=wp.vec3)
+    self.groups = wp.zeros((nworld * bvh_ngeom,), dtype=wp.int32)
+    self.group_roots = wp.zeros((nworld,), dtype=wp.int32)
+    self.pixels = wp.zeros((nworld, mjm.ncam, width * height), dtype=wp.uint32)
+    self.depth = wp.zeros((nworld, mjm.ncam, width * height), dtype=wp.float32)
+    self.rays_cam = wp.zeros((width * height), dtype=wp.vec3)
+    wp.launch(
+      kernel=build_primary_rays,
+      dim=(width * height),
```
thowell: `width * height`

mujoco_warp/render.py
```diff
@@ -0,0 +1,188 @@
...
+_CLEAR_KERNEL_CACHE = flags.DEFINE_bool("clear_kernel_cache", False, "clear Warp kernel cache before rendering")
+_OVERRIDE = flags.DEFINE_multi_string("override", [], "Model overrides (notation: foo.bar = baz)", short_name="o")
+_OUTPUT_RGB = flags.DEFINE_string("output_rgb", "debug.png", "output path for RGB image")
+_OUTPUT_DEPTH = flags.DEFINE_string("output_depth", "debug_depth.png", "output path for depth image")
+_DEPTH_SCALE = flags.DEFINE_float("depth_scale", 5.0, "scale factor to map depth to 0..255 for preview")
+_ROLL = flags.DEFINE_bool("roll_kinematics", False, "advance simulation before rendering")
+_ROLL_STEPS = flags.DEFINE_integer("roll_steps", 50, "number of steps to advance when rolling")
+
+
+def _load_model(path: epath.Path) -> mujoco.MjModel:
```
thowell, line 54: it looks like this file is not formatted

mujoco_warp/render.py
```diff
@@ -0,0 +1,188 @@
...
+
+    with wp.ScopedDevice(_DEVICE.value):
+        m = mjw.put_model(mjm)
+
+        # apply CLI overrides first
+        if _OVERRIDE.value:
+            override_model(m, _OVERRIDE.value)
+
+        # prepare Data from current MuJoCo state so camera/light transforms are valid
+        d = mjw.put_data(
```
thowell: could be one line

mujoco_warp/render.py
```diff
@@ -0,0 +1,188 @@
...
+_RENDER_DEPTH = flags.DEFINE_bool("depth", True, "render depth image")
+_USE_TEXTURES = flags.DEFINE_bool("textures", True, "use textures")
+_USE_SHADOWS = flags.DEFINE_bool("shadows", True, "use shadows")
+_DEVICE = flags.DEFINE_string("device", None, "override the default Warp device")
+_CLEAR_KERNEL_CACHE = flags.DEFINE_bool("clear_kernel_cache", False, "clear Warp kernel cache before rendering")
+_OVERRIDE = flags.DEFINE_multi_string("override", [], "Model overrides (notation: foo.bar = baz)", short_name="o")
+_OUTPUT_RGB = flags.DEFINE_string("output_rgb", "debug.png", "output path for RGB image")
+_OUTPUT_DEPTH = flags.DEFINE_string("output_depth", "debug_depth.png", "output path for depth image")
+_DEPTH_SCALE = flags.DEFINE_float("depth_scale", 5.0, "scale factor to map depth to 0..255 for preview")
+_ROLL = flags.DEFINE_bool("roll_kinematics", False, "advance simulation before rendering")
```
thowell: maybe rename to something like kinematics instead of roll if that is the utility? it looks like there is not logic below that just computes the kinematics?

mujoco_warp/render.py
```diff
@@ -0,0 +1,188 @@
...
+_ROLL_STEPS = flags.DEFINE_integer("roll_steps", 50, "number of steps to advance when rolling")
```
thowell: this could be nstep (similar to testspeed)?

mujoco_warp/render.py
```diff
@@ -0,0 +1,188 @@
...
+            depth = rc.depth.numpy()
+            row = depth[world, cam]
+            if row.shape[0] != num_pixels:
+                raise RuntimeError("unexpected depth buffer size")
+            _save_depth(row, width, height, _DEPTH_SCALE.value, _OUTPUT_DEPTH.value)
+            print(f"Saved depth to: {_OUTPUT_DEPTH.value}")
+
+
+def main():
+    sys.argv[0] = "mujoco_warp.render"
```
thowell, line 163: we can add mjwarp-render to pyproject.toml under [project.scripts]

mujoco_warp/testspeed.py
```diff
@@ -155,10 +166,24 @@ def _main(argv: Sequence[str]):
...
+        mjm,
+        m,
+        d,
+        _NWORLD.value,
+        _WIDTH.value,
+        _HEIGHT.value,
+        _USE_TEXTURES.value,
+        _USE_SHADOWS.value,
+        _RENDER_RGB.value,
+        _RENDER_DEPTH.value)
```
thowell, line 179: this doesn't look formatted

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,240 @@
...
+from .types import GeomType
+from .types import Model
+from .types import Data
+from .render_context import RenderContext
+
+wp.set_module_options({"enable_backward": False})
+
+
+@wp.func
+def compute_box_bounds(
```
thowell: we can add an `_` prefix to this function and others in this module that are not part of the public api or will not be imported by other modules

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+
+import warp as wp
+
+from . import bvh
+from .ray import ray_plane_with_normal
+from .ray import ray_sphere_with_normal
+from .ray import ray_capsule_with_normal
+from .ray import ray_box_with_normal
+from .ray import ray_mesh_with_bvh
+from .ray import _ray_plane
```
thowell: let's remove the `_` prefix from this function and the ones below since they are being imported by this module

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  tile_x = (tile_id % tiles_x) * TILE_W
+  tile_y = (tile_id // tiles_x) * TILE_H
+
+  i = tile_x + u
+  j = tile_y + v
+  return i, j
+
+
+@event_scope
+def render(m: Model, d: Data, rc: RenderContext):
```
thowell, line 87: this is a public api function, please add a docstring with `Args`

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+THREADS_PER_TILE: int = TILE_W * TILE_H
+
+@wp.func
+def ceil_div(a: int, b: int):
+  return (a + b - 1) // b
+
+
+# Map linear thread id (per image) -> (px, py) using TILE_W x TILE_H tiles
+@wp.func
+def tile_coords(tid: int, W: int, H: int):
```
thowell: we can add `_` prefix to functions in this module that are not part of the public api or will not be imported by other modules

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+@wp.func
+def compute_camera_ray(
+  width: int,
+  height: int,
+  fov_rad: float,
+  px: int,
+  py: int,
+  cam_xpos: wp.vec3,
+  cam_xmat:wp.mat33,
+):
```
thowell: please add a return type

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  aspect_ratio = float(width) / float(height)
```
thowell: we could compute `1 / float(height)` once and then utilize here and below

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  u = (float(px) + 0.5) / float(width) - 0.5
+  v = (float(py) + 0.5) / float(height) - 0.5
+  h = wp.tan(fov_rad / 2.0)
```
thowell: `* 0.5`

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  ray_dir_cam_space_x = u * 2.0 * h
+  ray_dir_cam_space_y = -v * 2.0 * h / aspect_ratio
+  ray_dir_cam_space_z = -1.0
+  ray_dir_local_cam = wp.normalize(
```
thowell: could probably be one line

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  width: int,
+  height: int,
+  tex_adr: int,
+  tex_data: wp.array(dtype=wp.uint32),
+) -> wp.vec3:
+  ix = wp.min(width - 1, wp.int32(uv[0] * wp.float32(width)))
+  iy = wp.min(height - 1, wp.int32(uv[1] * wp.float32(height)))
+  linear_idx = tex_adr + (iy * width + ix)
+  packed_rgba = tex_data[linear_idx]
+  r = wp.float32((packed_rgba >> wp.uint32(16)) & wp.uint32(0xFF)) / 255.0
```
thowell: we could precompute 1.0 / 255.0 once and utilize here and below

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+
+@wp.func
+def sample_texture_2d(
+  uv: wp.vec2,
+  width: int,
+  height: int,
+  tex_adr: int,
+  tex_data: wp.array(dtype=wp.uint32),
+) -> wp.vec3:
+  ix = wp.min(width - 1, wp.int32(uv[0] * wp.float32(width)))
```
thowell: wp.int32 -> int and wp.float32 -> float

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  v: wp.float32,
+  f: wp.int32,
+  mesh_id: wp.int32,
+) -> wp.vec3:
+  tex_color = wp.vec3(1.0, 1.0, 1.0)
+
+  if geom_matid == -1 or mat_texid == -1:
+    return tex_color
+
+  if geom_type[geom_id] == int(GeomType.PLANE.value):
```
thowell: `int(GeomType.PLANE.value)` -> `GeomType.PLANE`

(similar updates for similar instances below)

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  enabled_geom_ids: wp.array(dtype=int),
+  geom_type: wp.array(dtype=int),
+  geom_dataid: wp.array(dtype=int),
+  geom_size: wp.array2d(dtype=wp.vec3),
+  mesh_bvh_ids: wp.array(dtype=wp.uint64),
+  geom_xpos: wp.array2d(dtype=wp.vec3),
+  geom_xmat: wp.array2d(dtype=wp.mat33),
+  ray_origin_world: wp.vec3,
+  ray_dir_world: wp.vec3,
+) -> Tuple[wp.int32, wp.float32, wp.vec3, wp.float32, wp.float32, wp.int32, wp.int32]:
```
thowell: wp.int32 > int and wp.float32 -> float

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  dist = wp.float32(wp.inf)
```
thowell: float(wp.inf)

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  normal = wp.vec3(0.0, 0.0, 0.0)
+  geom_id = wp.int32(-1)
+  bary_u = wp.float32(0.0)
+  bary_v = wp.float32(0.0)
+  face_idx = wp.int32(-1)
+  geom_mesh_id = wp.int32(-1)
+
+  query = wp.bvh_query_ray(bvh_id, ray_origin_world, ray_dir_world, group_root)
+  bounds_nr = wp.int32(0)
```
thowell: `int(0)`

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+
+  while wp.bvh_query_next(query, bounds_nr, dist):
+    gi_global = bounds_nr
+    gi_bvh_local = gi_global - (world_id * bvh_ngeom)
+    gi = enabled_geom_ids[gi_bvh_local]
+
+    if geom_type[gi] == GeomType.PLANE:
```
thowell, line 285: wondering if we can implement branch elimination with static loop unrolling here -- maybe add a todo to investigate performance optimizations in the future?

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  mesh_bvh_ids: wp.array(dtype=wp.uint64),
+  geom_xpos: wp.array2d(dtype=wp.vec3),
+  geom_xmat: wp.array2d(dtype=wp.mat33),
+  ray_origin_world: wp.vec3,
+  ray_dir_world: wp.vec3,
+  max_dist: wp.float32,
+) -> bool:
+  """ A simpler version of cast_ray_first_hit that only checks for the first hit."""
+  query = wp.bvh_query_ray(bvh_id, ray_origin_world, ray_dir_world, group_root)
+  bounds_nr = wp.int32(0)
```
thowell: int(0)

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  light_xdir: wp.vec3,
+  normal: wp.vec3,
+  geom_type: wp.array(dtype=int),
+  geom_dataid: wp.array(dtype=int),
+  geom_size: wp.array2d(dtype=wp.vec3),
+  mesh_bvh_ids: wp.array(dtype=wp.uint64),
+  geom_xpos: wp.array2d(dtype=wp.vec3),
+  geom_xmat: wp.array2d(dtype=wp.mat33),
+  hit_point: wp.vec3,
+) -> wp.float32:
```
thowell: float

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+
+  light_contribution = wp.float32(0.0)
```
thowell: float

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+
+  if not light_active:
```
thowell, line 440: how common is the case `not light_active`? we could potentially utilize `wp.static`

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  L = wp.vec3(0.0, 0.0, 0.0)
+  dist_to_light = wp.float32(wp.inf)
+  attenuation = wp.float32(1.0)
+
+  if light_type == 1: # directional light
+    L = wp.normalize(-light_xdir)
+  else:
+    to_light = light_xpos - hit_point
+    dist_to_light = wp.length(to_light)
+    L = wp.normalize(to_light)
```
thowell: `math.normalize_with_norm`

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+      outer = SPOT_OUTER_COS
+      spot_factor = wp.min(1.0, wp.max(0.0, (cos_theta - outer) / (inner - outer)))
+      attenuation = attenuation * spot_factor
+
+  ndotl = wp.max(0.0, wp.dot(normal, L))
+  if ndotl == 0.0:
+    return light_contribution
+
+  visible = wp.float32(1.0)
+  shadow_min_visibility = wp.float32(0.3) # reduce shadow darkness (0: full black, 1: no shadow)
```
thowell: could this value be a constant defined at the top of the file?

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+
+  if use_shadows and light_castshadow:
+    # Nudge the origin slightly along the surface normal to avoid
+    # self-intersection when casting shadow rays
+    eps = 1.0e-4
+    shadow_origin = hit_point + normal * eps
+    # Distance-limited shadows: cap by dist_to_light (for non-directional)
+    max_t = wp.float32(dist_to_light - 1.0e-3)
```
thowell: float

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+  if num_view_groups == 0:
+    return
+
+  if rc.render_rgb:
+    rc.pixels.fill_(wp.uint32(BACKGROUND_COLOR))
+
+  if rc.render_depth:
+    rc.depth.fill_(wp.float32(0.0))
+
+  @wp.kernel
```
thowell: @nested_kernel(module="unique", enable_backward="False")

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,797 @@
...
+          base_color[1] * tex_color[1],
+          base_color[2] * tex_color[2],
+        )
+
+      up = wp.vec3(0.0, 0.0, 1.0)
+      len_n = wp.length(normal)
+      n = normal if len_n > 0.0 else up
+      n = wp.normalize(n)
+      hemispheric = 0.5 * (wp.dot(n, up) + 1.0)
+      sky = wp.vec3(0.4, 0.4, 0.45)
```
thowell: should these defaults be constants at the top of this file?

thowell:
@StafaH awesome work!!

mujoco_warp/_src/bvh.py
StafaH: removing rc.nworld entirely, it was an artifact from a time when we didn't have data as input to creating rc, but this isn't the case anymore

mujoco_warp/_src/ray.py
StafaH: mat[:, 2] works, good catch.

mujoco_warp/_src/ray.py
StafaH: Not sure, maybe we can add something to capsule test. The ray + normal will need new tests, but I'm waiting to see how C MuJoCo implements everything (and MJX?). I remember this fix was when I saw it in a render being off.

mujoco_warp/_src/ray.py
StafaH: Fixed for all

btaba:
There is a circular import between render_context and bvh, should be resolved. Ownership between the two modules can also be improved

mujoco_warp/_src/render_context.py
StafaH: Yes, we're going to need stuff not in Model (tex data being the most important).

mujoco_warp/_src/render_context.py
StafaH: Fixed now by removal of rc.nworld

mujoco_warp/render.py
StafaH: Forgot to run ruff check, all files should now be formatted.

mujoco_warp/render.py
StafaH: Removed from now. It was an artifact from previous tests. I might add this back later when I modify this script to create tiled rollouts.

mujoco_warp/render.py
StafaH: We can do this later depending on if it stays here or goes into contrib

mujoco_warp/_src/bvh.py
StafaH: Done. Everything in this module will not be public so added _ to all

mujoco_warp/_src/render.py
StafaH: Removing the function altogether. We now pre-compute rays inside the render_context (so that we can eventually support raycaster + camera flexibly)

mujoco_warp/_src/render.py
StafaH: Removing the function altogether as per previous comment.

mujoco_warp/_src/render.py
StafaH: Removing the function altogether as per previous comment.

mujoco_warp/_src/render.py
StafaH: Removing the function altogether as per previous comment.

mujoco_warp/_src/render.py
StafaH: Yep, will leave a TODO, i think theres other similar TODO's in the repo so at some point someone can sweep across and update this

mujoco_warp/_src/render.py
StafaH: Leaving a TODO here instead, because why loop over all lights, maybe we should be looping over active lights similar to how we only iterate through enabled geoms. Something to think about later.

mujoco_warp/_src/render.py
StafaH: We might eventually want to move some stuff into render context, but for now constants at top makes sense

StafaH:
Thanks @thowell, I went through all the comments and cleaned everything up.

@btaba I can hack it in place with an if TYPE_CHECKING, but the correct way here would probably be to move the create_render_context to either render or io, since that's where the circularity comes from. Open to either, @thowell should we keep rendering contained to one render.py or have the create function be called make_render_context and live alongside make_data etc?

@hartikainen there are warp changes that will make the renderer possible, currently the only way to test is to compile from source using the PR/fork found in the warp repo

mujoco_warp/_src/ray.py
thowell: will you add a todo below this comment to add a test for this case? thanks!

mujoco_warp/_src/render_context.py
thowell: what are the fields in `mujoco.MjModel` but not in `Model` that are needed for the renderer? @erikfrey should we add these fields to `Model`?

thowell:
@erikfrey what is your preference for where `make_render_context` is located?

thowell:
@StafaH 
- we should add `render_test.py` with a few tests. we can also add todos for tests we think would be valuable to eventually implement.
- we should add something to the readme, maybe under features: `Renderer: ray tracing`? or maybe a separate table for renderer under 'Compatibility' that lists current features? @erikfrey what are your thoughts/recommendations?
- it would be great to add a benchmark with rendering to benchmark/benchmark.py and TODOs for additional benchmarks that we think would be valuable.

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,781 @@
...
+
+
+@event_scope
+def render(m: Model, d: Data, rc: RenderContext):
+  """Render the current frame.
+
+  Outputs are stored in buffers within the render context.
+
+  Args:
+    m (Model): The model on device.
```
thowell: we don't need the type after the variable in the docstrings. will you remove the types? thanks!

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,260 @@
...
+  use_textures: Optional[bool] = True,
+  use_shadows: Optional[bool] = False,
+  render_rgb: Optional[bool] = True,
+  render_depth: Optional[bool] = True,
+  enabled_geom_groups: Optional[list[int]] = [0, 1, 2],
+) -> RenderContext:
+  """Creates a render context on device.
+
+    Args:
+      mjm (mujoco.MjModel): The model containing kinematic and dynamic information (host).
```
thowell: we can remove the types and `Defaults to ...` for the arguments.

mujoco_warp/_src/render_context.py
```diff
@@ -0,0 +1,260 @@
...
+      width (int): The width to render every camera image.
+      height (int): The height to render every camera image.
+      use_textures (bool, optional): Whether to use textures. Defaults to True.
+      use_shadows (bool, optional): Whether to use shadows. Defaults to False.
+      render_rgb (bool, optional): Whether to render RGB images. Defaults to True.
+      render_depth (bool, optional): Whether to render depth images. Defaults to True.
+      enabled_geom_groups (list[int], optional): The geom groups to render. Defaults to [0, 1, 2].
+
+    Returns:
+      RenderContext: The render context containing rendering fields and output arrays (device).
```
thowell: we can remove the type here and just have a description of what is being returned

mujoco_warp/_src/ray.py
StafaH: Done

mujoco_warp/render.py
StafaH: This was removed to simplify the script (for now)

erikfrey:
@thowell my first instinct would be that `create_render_context` live in io.py, but that's not a blocker to get this PR merged.  LGTM for now.

hartikainen:
For what it's worth, I played around with the renderer a bit. I'm noticing some weird behavior from it from a simple multi-object scene.

Here's the model I'm using:
<details><summary>Details</summary>

```console
<mujoco model="cubes">

  <visual>
    <map znear="0.01" zfar="200" />
    <quality shadowsize="8192" />
    <headlight diffuse=".8 .8 .8" ambient=".2 .2 .2" specular="1 1 1" />
    <global azimuth="140" elevation="-20" />
  </visual>

  <asset>
    <texture type="skybox" builtin="gradient" rgb1="1 1 1" rgb2="1 1 1" width="800" height="800" />
    <texture name="grid" type="2d" builtin="checker" width="512" height="512" rgb1=".1 .2 .3" rgb2=".2 .3 .4"/>
    <material name="grid" texture="grid" texrepeat="1 1" texuniform="true" reflectance=".2"/>
  </asset>

  <worldbody>
    <geom name="floor" size="0 0 .05" type="plane" material="grid" contype="1" conaffinity="1"/>
    <camera pos="3 0 2" xyaxes="0 1 0 -0.6 0 1"/>
    <light pos="0 0 3" dir="0 0 -1" directional="true" />

    <geom name="table" pos="0.8 0.0 0.5" type="box" size="0.5 1.0 0.01" quat="1.0 0.0 0.0 0.0" contype="1" conaffinity="1"/>

    <replicate count="5" offset="0.1 0 0">
        <replicate count="5" offset="0 0.2 0">
            <body name="sphere" pos="0.6 -0.4 0.8">
                <freejoint name="sphere"/>
                <geom name="geom_sphere" type="sphere" size="0.05" contype="1" conaffinity="1" rgba="1 0 0 1"/>
            </body>
        </replicate>
    </replicate>

    <replicate count="5" offset="0.1 0 0">
        <replicate count="5" offset="0 0.2 0">
            <body name="capsule" pos="0.6 -0.4 0.95">
                <freejoint name="capsule"/>
                <geom name="geom_capsule" type="capsule" size="0.04 0.05" contype="1" conaffinity="1" rgba="0 1 0 1"/>
            </body>
        </replicate>
    </replicate>

    <replicate count="5" offset="0.1 0 0">
        <replicate count="5" offset="0 0.2 0">
            <body name="ellipsoid" pos="0.6 -0.4 1.1">
                <freejoint name="ellipsoid"/>
                <geom name="geom_ellipsoid" type="ellipsoid" size="0.05 0.04 0.03" contype="1" conaffinity="1" rgba="0 0 1 1"/>
            </body>
        </replicate>
    </replicate>

    <replicate count="5" offset="0.1 0 0">
        <replicate count="5" offset="0 0.2 0">
            <body name="cylinder" pos="0.6 -0.4 1.2">
                <freejoint name="cylinder"/>
                <geom name="geom_cylinder" type="cylinder" size="0.04 0.04" contype="1" conaffinity="1" rgba="1 1 0 1"/>
            </body>
        </replicate>
    </replicate>

    <replicate count="5" offset="0.1 0 0">
        <replicate count="5" offset="0 0.2 0">
            <body name="box1" pos="0.6 -0.4 1.4">
                <freejoint name="joint1"/>
                <geom name="geom1" type="box" size="0.03 0.03 0.03" contype="1" conaffinity="1" rgba="0 1 1 1"/>
            </body>
        </replicate>
    </replicate>
  </worldbody>
</mujoco>
```

</details> 


Here's what the scene looks when rendered on the standard MJC viewer:
image

And here's with when renderer with the warp renderer. First, with `width=160` and `height=120`:

<details><summary>Details</summary>
```console
python -m mujoco_warp.render \
  --nworld=4 \
  --width=160 \
  --height=120 \
  --rgb=true \
  --depth=true \
  --textures=true \
  --shadows=false \
  --cam=0 \
  --depth_scale=8.0 \
  ./benchmark/100_cubes/scene.xml
Loading model from: benchmark/100_cubes/scene.xml...
Model: ncam=1 nlight=1 ngeom=127
Render: 640x480 rgb=True depth=True
Rendering...
Saved RGB to: debug.png
Saved depth to: debug_depth.png
```
</details> 

debug_depth

With `width=640` and `height=480`:

<details><summary>Details</summary>
```console
python -m mujoco_warp.render \
  --nworld=4 \
  --width=640 \
  --height=480 \
  --rgb=true \
  --depth=true \
  --textures=true \
  --shadows=false \
  --cam=0 \
  --depth_scale=8.0 \
  ./benchmark/100_cubes/scene.xml
Loading model from: benchmark/100_cubes/scene.xml...
Model: ncam=1 nlight=1 ngeom=127
Render: 640x480 rgb=True depth=True
Rendering...
Saved RGB to: debug.png
Saved depth to: debug_depth.png
```
</details> 

debug_depth

There are two, I believe, separate issues in the above images. First, it seems as if some black, noisy plane is being rendered right in front of the camera. Second, the bottom-right corner has a completely black strip to it. The reason I believe these are two separate issues is that I see the black strips on other models too, even when the depth rendering otherwise looks good. For example, if I remove all the objects except the table from the above scene, I get:

debug_depth

Here's some system information in case that matters:
```console
$ uv pip list | grep "mujoco\|warp"
mujoco                  3.3.8.dev831015711
mujoco-mjx              3.3.8
mujoco-warp             0.0.1              /home/user/google-deepmind/mujoco_warp
warp-lang               1.11.0.dev20251111
$ /usr/bin/lspci | grep "NVIDIA"
00:03.0 3D controller: NVIDIA Corporation AD104GL [L4] (rev a1)
00:04.0 3D controller: NVIDIA Corporation AD104GL [L4] (rev a1)
```

hartikainen:
Digging a bit deeper into my issue above, I think the weird noise is caused by `ellipsoid` geoms. If I modify the above scene by commenting out the ellipsoids, things seem fine:
<details><summary>Details</summary>

```diff
modified   benchmark/100_cubes/scene.xml
@@ -38,14 +38,14 @@
         </replicate>
     </replicate>
 
-    <replicate count="5" offset="0.1 0 0">
-        <replicate count="5" offset="0 0.2 0">
-            <body name="ellipsoid" pos="0.6 -0.4 1.1">
-                <freejoint name="ellipsoid"/>
-                <geom name="geom_ellipsoid" type="ellipsoid" size="0.05 0.04 0.03" contype="1" conaffinity="1" rgba="0 0 1 1"/>
-            </body>
-        </replicate>
-    </replicate>
+    <!-- <replicate count="5" offset="0.1 0 0"> -->
+    <!--     <replicate count="5" offset="0 0.2 0"> -->
+    <!--         <body name="ellipsoid" pos="0.6 -0.4 1.1"> -->
+    <!--             <freejoint name="ellipsoid"/> -->
+    <!--             <geom name="geom_ellipsoid" type="ellipsoid" size="0.05 0.04 0.03" contype="1" conaffinity="1" rgba="0 0 1 1"/> -->
+    <!--         </body> -->
+    <!--     </replicate> -->
+    <!-- </replicate> -->
 
     <replicate count="5" offset="0.1 0 0">
         <replicate count="5" offset="0 0.2 0">
```
</details>

debug_depth

The weird black strip still remains.

StafaH:
Hi @hartikainen, Thanks for doing some early testing! Yes you are correct, the issues you are seeing are because ellipsoids haven't been implemented yet in the primitives, I'll be adding that soon. The black strip is because of the patch tiles we use, but this won't be an issue in the next batch of changes since we stop using those patch tiles altogether. For now the workaround would be to use a square image size until that fix is in.

PR #840, opened 2025-11-06
Replace geom function with geom_collision_pair

kbayes:
This refactor reduces some of repeated logic in collision_convex, collision_primitive, and collision_sdf.


erikfrey (approved):
Looks great, thanks Kyle.

PR #845, opened 2025-11-12
Refactor hill climbing support function

kbayes:
No change in behavior. There might be a minor performance speedup.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -157,7 +157,6 @@ def _support(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
       edge_localid = geom.graphadr + 2 + 2 * numvert
       prev = int(-1)
       imax = wp.where(geom.index > -1, geom.index, 0)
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_gjk.py
```diff
@@ -155,26 +155,23 @@ def _support(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
...
       vert_globalid = geom.graphadr + 2 + numvert
       edge_localid = geom.graphadr + 2 + 2 * numvert
-      # hillclimb until no change
       prev = int(-1)
-      imax = int(0)
-      if geom.index > -1:
-        imax = geom.index
-        sp.cached_index = geom.index
+      imax = wp.where(geom.index > -1, geom.index, 0)
+      sp.cached_index = geom.index
```
erikfrey: you are setting `cached_index` here and then setting it again on line 175.  Seems like you don't need this line?

kbayes: Yeah, I don't know why this is here.

PR #851, opened 2025-11-14
add matrix types to types.py

thowell:
adds `mat23`, `mat43`, and `mat63` to types.py

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -596,11 +596,26 @@ class vec11f(wp.types.vector(length=11, dtype=float)):
   pass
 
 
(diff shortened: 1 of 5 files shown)


mujoco_warp/_src/types.py
```diff
@@ -596,11 +596,26 @@ class vec11f(wp.types.vector(length=11, dtype=float)):
   pass
 
 
+class mat23f(wp.types.matrix(shape=(2, 3), dtype=float)):
```
erikfrey, line 599: I still feel ambivalent about these type aliases, but this PR is still an improvement so let's merge it.

PR #856, opened 2025-11-19
parse `Option.contact_sensor_maxmatch` from XML

thowell:


mujoco_warp/_src/io.py
```diff
@@ -170,7 +170,10 @@ def not_implemented(objtype, objid, geomtype):
   opt.broadphase_filter = types.BroadphaseFilter.PLANE | types.BroadphaseFilter.SPHERE | types.BroadphaseFilter.OBB
   opt.graph_conditional = True
   opt.run_collision_detection = True
-  opt.contact_sensor_maxmatch = 64
+  contact_sensor_maxmatch_id = mujoco.mj_name2id(mjm, mujoco.mjtObj.mjOBJ_NUMERIC, "contact_sensor_maxmatch")
+  opt.contact_sensor_maxmatch = (
```
erikfrey: total style nit, the nittiest of nits:

once your ternary operator is split across multiple lines like this, maybe it's cleaner to just write a typical multiline if/else block

erikfrey (approved):
LGTM, just one tiny style nit

mujoco_warp/_src/io.py
thowell: updated to if/else block

PR #858, opened 2025-11-20
Fix kinematics for mocap bodies with child bodies

adenzler-nvidia:
The sequence of kinematics -> mocap was wrong because if a mocap body had descendant bodies, the kinematics for these bodies were calculated based on the body_pos/quat values instead of mocap_pos/mocap_quat.

I took the liberty of just fusing in the mocap update into the kinematics kernel because it made sense to me, and it looks like there is a slight edge to this performance-wise. It will definitely help for scenes with a lot of mocap bodies.

### DIFF
--- a/mujoco_warp/_src/smooth_test.py
+++ b/mujoco_warp/_src/smooth_test.py
@@ -48,7 +48,7 @@ class SmoothTest(parameterized.TestCase):
     xml = """
     <mujoco>
       <worldbody>
-        <body name="mocap_parent" mocap="true" pos="0 0 0">
+        <body name="mocap_parent" mocap="true">
           <geom type="sphere" size="0.1"/>
           <body name="child" pos="1 0 0">
             <geom type="box" size="0.1 0.1 0.1"/>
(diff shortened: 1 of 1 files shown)


thowell:
confirmed that this is a bug with the MJX implementation, thanks for reporting and fixing this bug with MJWarp @adenzler-nvidia!

mujoco_warp/_src/smooth_test.py
```diff
@@ -38,6 +38,83 @@ def _assert_eq(a, b, name):
...
+    """Tests that mocap bodies and child bodies are correctly updated after mocap_pos changes.
+
+    This is a regression test for a bug where mocap positions were updated after kinematics,
+    causing child bodies of mocap bodies to have incorrect positions based on stale mocap data.
+    """
+    # Create a simple scene with a mocap body that has a child body attached
+    xml = """
+    <mujoco>
+      <worldbody>
+        <body name="mocap_parent" mocap="true" pos="0 0 0">
```
thowell: we don't need to specify `pos="0 0 0"` since this is the default

mujoco_warp/_src/smooth_test.py
```diff
@@ -38,6 +38,83 @@ def _assert_eq(a, b, name):
...
+    mjw.kinematics(m, d)
+    initial_mocap_xpos = d.xpos.numpy()[0, mocap_body_id]
+    initial_child_xpos = d.xpos.numpy()[0, child_body_id]
+    initial_site_xpos = d.site_xpos.numpy()[0, child_site_id].copy()
+
+    # Verify initial positions match MuJoCo
+    _assert_eq(d.xpos.numpy()[0], mjd.xpos, "initial xpos")
+
+    # Expected: child should be at mocap_pos + (1, 0, 0) relative offset
+    np.testing.assert_allclose(initial_child_xpos, [1.0, 0.0, 0.0], atol=1e-5)
```
thowell: can we use `_assert_eq` here and below?

mujoco_warp/_src/smooth_test.py
adenzler-nvidia: removed, thanks for the pointer.

mujoco_warp/_src/smooth_test.py
adenzler-nvidia: done, thanks,

PR #860, opened 2025-11-21
Add flex edge equality

quagla:


thowell:
it would great if we could add a few tests and update the 'Compatibility' section of the readme. thanks!

mujoco_warp/_src/smooth_test.py
```diff
@@ -468,6 +468,33 @@ def test_tendon_armature(self):
     mjw._src.smooth.tendon_bias(m, d, d.qfrc_bias)
     _assert_eq(d.qfrc_bias.numpy()[0], mjd.qfrc_bias, "qfrc_bias")
 
+  def test_flex(self):
+    mjm, mjd, m, d = test_data.fixture("flex/floppy.xml")
+
+    mjw._src.smooth.kinematics(m, d)
```
thowell: mjw.kinematics

mujoco_warp/_src/smooth_test.py
```diff
@@ -468,6 +468,33 @@ def test_tendon_armature(self):
...
+    mjw._src.smooth.com_pos(m, d)
```
thowell: mjw.com_pos

mujoco_warp/_src/smooth_test.py
```diff
@@ -468,6 +468,33 @@ def test_tendon_armature(self):
...
+    mjw._src.smooth.flex(m, d)
```
thowell: since `mj_flex` is part of the MuJoCo API, let's add `flex` to MJWarp's API and then use `mjw.flex` here.

mujoco_warp/_src/smooth_test.py
```diff
@@ -468,6 +468,33 @@ def test_tendon_armature(self):
...
+    mujoco.mj_kinematics(mjm, mjd)
+    mujoco.mj_comPos(mjm, mjd)
+    mujoco.mj_flex(mjm, mjd)
+
+    if mujoco.mj_isSparse(mjm):
```
thowell: since we are only testing one case here we probably don't need the if/else?

mujoco_warp/_src/smooth_test.py
```diff
@@ -468,6 +468,33 @@ def test_tendon_armature(self):
...
+      flexedge_J = np.zeros((mjm.nflexedge, mjm.nv), dtype=float)
+      mujoco.mju_sparse2dense(
+        flexedge_J, mjd.flexedge_J.ravel(), mjd.flexedge_J_rownnz, mjd.flexedge_J_rowadr, mjd.flexedge_J_colind.ravel()
+      )
+    else:
+      flexedge_J = mjd.flexedge_J
+
+    _assert_eq(m.dof_bodyid.numpy(), mjm.dof_bodyid, "flexedge_invweight0")
+    _assert_eq(d.cdof.numpy()[0], mjd.cdof, "cdof")
```
thowell: can we fill the relevant Data arrays above with wp.inf (for example, something like `d.cdof.fill_(wp.inf)`) after `test_data.fixture` and before calling the api functions?

PR #864, opened 2025-11-24
Prune box-box contacts down to 4 in multiccd

kbayes:
Historically in C mujoco, the box-box collider returned all 8 contacts even if 4 sufficed. With GPU, let's stick to 4 as contact memory is limited.


thowell (approved):

thowell:
nice! thanks Kyle

PR #870, opened 2025-11-28
Optimization: improve implementation of block  cholesky

adenzler-nvidia:
The main changes are moving the padding initialization and handling out of the core kernel, and optimizing shared memory reuse as well as dependency distances.

Some numbers (RTX 6000 Pro Blackwell):

n_humanoids

main

```

Total JIT time: 0.50 s
Total simulation time: 18.28 s
Total steps per second: 448,078
Total realtime factor: 2,240.39 x
Total time per step: 2231.75 ns

```
this PR

```

Total JIT time: 0.46 s
Total simulation time: 16.22 s
Total steps per second: 505,070
Total realtime factor: 2,525.35 x
Total time per step: 1979.93 ns
```

G1 with hands

main

```

Total JIT time: 1.16 s
Total simulation time: 21.12 s
Total steps per second: 387,965
Total realtime factor: 775.93 x
Total time per step: 2577.56 ns
```

this PR

```

Total JIT time: 1.13 s
Total simulation time: 19.54 s
Total steps per second: 419,299
Total realtime factor: 838.60 x
Total time per step: 2384.93 ns
```

G1 Kitchen

2048 envs - only total time spent in this kernel due to other issues that affect runtime

main

`update_gradient_cholesky_blocked: 881ms`

this PR

`update_gradient_cholesky_blocked: 135ms`

### DIFF
--- a/mujoco_warp/_src/derivative.py
+++ b/mujoco_warp/_src/derivative.py
@@ -153,37 +153,41 @@ def deriv_smooth_vel(m: Model, d: Data, out: wp.array2d(dtype=float)):
   """
   qMi = m.qM_fullm_i
   qMj = m.qM_fullm_j
+
   # TODO(team): implicit requires different sparsity structure
 
-  if ~(m.opt.disableflags & (DisableBit.ACTUATION | DisableBit.DAMPER)):
-    wp.launch(
-      _qderiv_actuator_passive,
-      dim=(d.nworld, qMi.size),
-      inputs=[
-        m.nu,
-        m.opt.timestep,
-        m.opt.disableflags,
-        m.opt.is_sparse,
-        m.dof_damping,
-        m.actuator_dyntype,
-        m.actuator_gaintype,
-        m.actuator_biastype,
-        m.actuator_actadr,
-        m.actuator_actnum,
-        m.actuator_gainprm,
-        m.actuator_biasprm,
-        d.act,
-        d.ctrl,
-        d.actuator_moment,
-        d.qM,
-        qMi,
-        qMj,
-      ],
-      outputs=[out],
-    )
-  else:
-    # TODO(team): directly utilize qM for these settings
-    wp.copy(out, d.qM)
+  if not m.opt.is_sparse:
+    # TODO(team): only clear elements not set by _qderiv_actuator_passive
+    out.zero_()
+
+  wp.launch(
+    _qderiv_actuator_passive,
+    dim=(d.nworld, qMi.size),
+    inputs=[
+      m.nu,
+      m.opt.timestep,
+      m.opt.disableflags,
+      m.opt.is_sparse,
+      m.dof_damping,
+      m.actuator_dyntype,
+      m.actuator_gaintype,
+      m.actuator_biastype,
+      m.actuator_actadr,
+      m.actuator_actnum,
+      m.actuator_gainprm,
+      m.actuator_biasprm,
+      d.act,
+      d.ctrl,
+      d.actuator_moment,
+      d.qM,
+      qMi,
+      qMj,
+    ],
+    outputs=[out],
+  )
+
+  # TODO(team): if ~(m.opt.disableflags & (DisableBit.ACTUATION | DisableBit.DAMPER)) use qM
+  # directly instead of computing out
 
   if not m.opt.disableflags & DisableBit.DAMPER:
     wp.launch(


mujoco_warp/_src/support_test.py
```diff
@@ -177,6 +177,176 @@ def test_set_state(self):
...
+  def test_block_cholesky(self):
+    """Tests block Cholesky decomposition and solve against numpy using n_humanoid model."""
+    from pathlib import Path
+
+    from mujoco_warp._src.block_cholesky import create_blocked_cholesky_func
+    from mujoco_warp._src.block_cholesky import create_blocked_cholesky_solve_func
+    from mujoco_warp._src.support import nested_kernel
+
+    # Load n_humanoid from benchmark directory (nv=81, larger than simple humanoid nv=27)
+    n_humanoid_path = Path(__file__).parent.parent.parent / "benchmark" / "humanoid" / "n_humanoid.xml"
```
erikfrey: nit: let's copy n_humanoid into `test_data` if we plan to use it here.

I'd like to allow for some future where we exclude benchmark scenes from the default library package, leave them as extras.

if you make that change, then you can load it here with `test_data.fixture`

mujoco_warp/_src/block_cholesky.py
```diff
@@ -23,49 +23,26 @@ def create_blocked_cholesky_func(block_size: int):
...
-          A_kk_tile[row, col] = value
-
-      if k > 0:
-        for j in range(0, k, block_size):
-          L_block = wp.tile_load(L, shape=(block_size, block_size), offset=(k, j))
-          L_block_T = wp.tile_transpose(L_block)
-          L_L_T_block = wp.tile_matmul(L_block, L_block_T)
-          A_kk_tile -= L_L_T_block
+
+      for j in range(0, k, block_size):
```
erikfrey, line 43: wow lovely, much simpler!

mujoco_warp/_src/block_cholesky.py
```diff
@@ -74,48 +51,26 @@ def blocked_cholesky_func(
...
+
+        wp.tile_lower_solve_inplace(L_kk_tile, wp.tile_transpose(A_ik_tile))
+        wp.tile_store(L, A_ik_tile, offset=(i, k))
 
   return blocked_cholesky_func
 
 
 @lru_cache(maxsize=None)
-def create_blocked_cholesky_solve_func(block_size: int):
+def create_blocked_cholesky_solve_func(block_size: int, N: int):
```
erikfrey: do we need for `create_blocked_cholesky_solve_func` to take N as input and for the wp.func that it returns to also take `matrix_size` as input?

also generally, can we clean up `n` vs `N` vs `matrix_size`?

mujoco_warp/_src/solver.py
```diff
@@ -1676,33 +1676,49 @@ def kernel(
...
+    wp.static(create_blocked_cholesky_func(TILE_SIZE))(efc_h_in[worldid], matrix_size, cholesky_L_tmp[worldid])
+    wp.static(create_blocked_cholesky_solve_func(TILE_SIZE, matrix_size))(
+      cholesky_L_tmp[worldid], efc_grad_in[worldid], matrix_size, efc_Mgrad_out[worldid]
     )
 
   return kernel
 
 
+@wp.kernel
+def padding_efc_h(
```
erikfrey: optional nit: short kernels like this, you can put the params on one line and remove the comment separates (e.g. Model, Data in, Data out)

mujoco_warp/_src/types.py
```diff
@@ -1517,11 +1517,10 @@ class Constraint:
   force: array("nworld", "njmax", float)
   Jaref: array("nworld", "njmax", float)
   Ma: array("nworld", "nv", float)
-  grad: array("nworld", "nv", float)
-  cholesky_L_tmp: array("nworld", "nv", "nv", float)
-  cholesky_y_tmp: array("nworld", "nv", float)
+  grad: array("nworld", "nv_pad", float)
```
erikfrey, line 1520: you know i'm a bit surprised this didn't break a test!

i think you will have to change `io.put_data` to clip `nv_pad` back to `nv`.  there's a snippet of code for clipping `efc_pad` that you can emulate:

https://github.com/google-deepmind/mujoco_warp/blob/c5306fd1d5a24e09687bb13fd043d457da8cb019/mujoco_warp/_src/io.py#L748-L749

mujoco_warp/_src/support_test.py
adenzler-nvidia: Makes a lot of sense, copied it to test_data. In theory the test could even be done without an asset at all, but I think it's worthwhile to have one to test whether the padding stuff works correctly.

mujoco_warp/_src/block_cholesky.py
adenzler-nvidia: This is a weird one - I added a comment in the code to explain things.

Initially I wanted to make matrix_size only be a static input, but that made warp unroll the loop unconditionally. Which means all the benefits of blocking are gone, and we run out of shared memory soon. So passing it as a function argument solved that. On the other hand, we still need it as a static input because we need it for tile sizing.

I changed all of n/N to matrix_size, and named the static one matrix_size_static. Hope that's fine.

mujoco_warp/_src/solver.py
adenzler-nvidia: Thanks, done.

mujoco_warp/_src/types.py
adenzler-nvidia: You're scaring me now!

I'm trying to follow which efc fields need to be copied during put_data. So I went back a bit to somewhere before your change (https://github.com/google-deepmind/mujoco_warp/blob/467e31923f178f0df3ecb6c8aa6ce0fb69b832e8/mujoco_warp/_src/io.py) - looks like the gradient was never copied over?

We need the padding to be zero there, but I think this was taken care of by initializing with np.zeros, compared to having a wp.empty before.

Or was this just missing before and we should copy all of them?

mujoco_warp/_src/support_test.py
erikfrey: OK great!

mujoco_warp/_src/block_cholesky.py
erikfrey: Interesting - OK!

mujoco_warp/_src/types.py
erikfrey: Oh wait, you know what, you're right.  In the old code we were not copying these efc fields in `put_data`, and in the new code we are not either, just filling with zeros.

At some point I will fix this properly so `grad` and other fields are properly copied over, but not now.

PR #875, opened 2025-12-03
Change some constants in collision_gjk.py for numerical stability

kbayes:


thowell:
great change @kbayes!! thanks!

PR #879, opened 2025-12-04
update deriv_smooth_vel

thowell:
this pr should fix reported test failures for forward_test.test_step2 with the implicitfast integrator.

the changes in #866 failed to zero elements in `out` that are not part of `qM`'s sparsity patten for the dense case leading to  uninitialized elements. this pr initializes all elements in `out` with zeros to fix this issue.

additionally, the existing `else` case was incorrect in the dense case when `nv != nv_pad` because `qM` has shape `(nworld, nv_pad, nv_pad)` whereas `out` is constructed with shape `(nworld, nv, nv)` in this case. to fix this issue, instead of guarding and utilizing `wp.copy`, `_qderiv_actuator_passive` is always launched.

### DIFF
--- a/mujoco_warp/_src/derivative.py
+++ b/mujoco_warp/_src/derivative.py
@@ -153,37 +153,41 @@ def deriv_smooth_vel(m: Model, d: Data, out: wp.array2d(dtype=float)):
   """
   qMi = m.qM_fullm_i
   qMj = m.qM_fullm_j
+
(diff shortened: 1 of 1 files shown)


adenzler-nvidia:
Thanks for catching this - wouldn't it make sense to initialize qderiv with wp.zeros and have it use the padding sizes as well?

PR #881, opened 2025-12-04
readme: add colab button

thowell:
image


erikfrey:
My feeling about these colab buttons that show up everywhere in github:

ehh

PR #883, opened 2025-12-05
Precompute inverse square root of impratio

adenzler-nvidia:
I don't need this change in 100%. I just came across this, tried it and I can see a slight edge in benchmarks/kernel timings. Nothing compared to the bigger things we still have to optimize though.

That being said, apart from the interface change I think this has only benefits, as it removes runtime square root instructions in some kernel that run very often (linesearch) and sqrt throughput is generally lower than multiplication. My intuition and a very quick experiment tell me that a non-default impratio that doesn't have a trivial solution to sqrt makes these gains more visible.

Numbers for humanoid on an RTX Pro 6000 Blackwell.

impratio == 1:

main:

```

Total JIT time: 0.31 s
Total simulation time: 2.29 s
Total steps per second: 3,580,815
Total realtime factor: 17,904.08 x
Total time per step: 279.27 ns
```

this pr:

```

Total JIT time: 0.31 s
Total simulation time: 2.28 s
Total steps per second: 3,591,947
Total realtime factor: 17,959.73 x
Total time per step: 278.40 ns
```

Impratio == 7:

main:

```

Total JIT time: 0.32 s
Total simulation time: 2.62 s
Total steps per second: 3,128,554
Total realtime factor: 15,642.77 x
Total time per step: 319.64 ns
```

this pr:

```

Total JIT time: 0.30 s
Total simulation time: 2.60 s
Total steps per second: 3,145,472
Total realtime factor: 15,727.36 x
Total time per step: 317.92 ns
```

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -460,10 +460,10 @@ def _linesearch_iterative(m: types.Model, d: types.Data):
     dim=d.nworld,
     inputs=[
       m.nv,
-      m.opt.impratio_invsqrt,
       m.opt.tolerance,
       m.opt.ls_tolerance,
       m.opt.ls_iterations,
+      m.opt.impratio_invsqrt,
       m.stat.meaninertia,
       d.ne,
       d.nf,
@@ -648,8 +648,8 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
     linesearch_parallel_fused,
     dim=(d.nworld, m.opt.ls_iterations),
     inputs=[
-      m.opt.impratio_invsqrt,
       m.opt.ls_iterations,
+      m.opt.impratio_invsqrt,
       m.opt.ls_parallel_min_step,
       d.ne,
       d.nf,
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/io.py
```diff
@@ -157,7 +157,10 @@ def not_implemented(objtype, objid, geomtype):
       raise NotImplementedError(f"Collision sensors with box-box collisions are not implemented.")
 
   # create opt
-  opt = types.Option(**{f.name: getattr(mjm.opt, f.name, None) for f in dataclasses.fields(types.Option)})
+  opt_kwargs = {f.name: getattr(mjm.opt, f.name, None) for f in dataclasses.fields(types.Option)}
+  if hasattr(mjm.opt, "impratio"):
+    opt_kwargs["impratio_invsqrt"] = 1.0 / np.sqrt(mjm.opt.impratio)
```
thowell: can we have a safe division here, maybe something like `1.0 / np.sqrt(np.maximum(mjm.opt.impratio, mujoco.mjMINVAL))`?

adenzler-nvidia: makes sense. I hardcoded instead of using safe_div from math.py because it's quite slow to call a wp.func in python. Hope that's fine.

PR #894, opened 2025-12-09
Fix wp.full crash caused by NumPy float promotion in eq_active sum

### LINKED ISSUE #890: mujoco_warp: wp.full crashes because eq_active sum promotes to float
https://github.com/google-deepmind/mujoco/blob/main/mjx/mujoco/mjx/third_party/mujoco_warp/_src/io.py#L811-L812

```
  d.ne_connect = wp.full(nworld, 3 * np.sum((mjm.eq_type == mujoco.mjtEq.mjEQ_CONNECT) & mjd.eq_active), dtype=int)
  d.ne_weld = wp.full(nworld, 6 * np.sum((mjm.eq_type == mujoco.mjtEq.mjEQ_WELD) & mjd.eq_active), dtype=int)
```
`mjd.eq_active` is type array of uint8, and `(mjm.eq_type == mujoco.mjtEq.mjEQ_CONNECT)` is array of bool, np.sum will return a `uint`
multiplying a python `integer` with numpy `unsigned int` promotes it to `float64`.
```
type(np.sum((mjm.eq_type == mujoco.mjtEq.mjEQ_CONNECT) & mjd.eq_active))
<class 'numpy.uint64'>

type(3 * np.sum((mjm.eq_type == mujoco.mjtEq.mjEQ_CONNECT) & mjd.eq_active))
<class 'numpy.float64'>
```
but `np.full` will not be happy with float

```
wp.full(nworld, 3 * np.sum((mjm.eq_type == mujoco.mjtEq.mjEQ_CONNECT) & mjd.eq_active), dtype=int)
Traceback (most recent call last):
  File "/home/zhengyuz/miniconda3/envs/newton_isaaclab/lib/python3.11/site-packages/isaacsim/extscache/omni.warp.core-1.11/warp/_src/types.py", line 3326, in fill_
    cvalue = self.dtype._type_(value)
             ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'numpy.float64' object cannot be interpreted as an integer

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/zhengyuz/miniconda3/envs/newton_isaaclab/lib/python3.11/site-packages/isaacsim/extscache/omni.warp.core-1.11/warp/_src/context.py", line 5887, in full
    arr.fill_(value)
  File "/home/zhengyuz/miniconda3/envs/newton_isaaclab/lib/python3.11/site-packages/isaacsim/extscache/omni.warp.core-1.11/warp/_src/types.py", line 3328, in fill_
    raise ValueError(f"Failed to convert the value to the array data type: {e}") from e
ValueError: Failed to convert the value to the array data type: 'numpy.float64' object cannot be interpreted as an integer
```

the fix seems to just be casting to int to revert the promotion. Im happy to submit a pr if that seems the right fix

```
  d.ne_connect = wp.full(nworld, 3 * int(np.sum((mjm.eq_type == mujoco.mjtEq.mjEQ_CONNECT) & mjd.eq_active)), dtype=int)
  d.ne_weld = wp.full(nworld, 6 * int(np.sum((mjm.eq_type == mujoco.mjtEq.mjEQ_WELD) & mjd.eq_active)), dtype=int)
```

swankystark:
This PR fixes an issue where wp.full receives a float64 due to NumPy's type
promotion when multiplying Python integers with uint64 results from np.sum().
Warp cannot convert float inputs to integer arrays, causing a crash.

The fix explicitly casts the sum to int before multiplication to ensure the
value passed to wp.full is always an integer.

A regression test (test_make_data_eq_type) has been added to ensure the bug
never returns. The test checks that make_data executes without crashing and
that all ne_* fields contain valid integer values.

Fixes #890


mujoco_warp/_src/io_test.py
```diff
@@ -567,6 +572,23 @@ def test_contact_sensor_maxmatch(self):
 
     self.assertEqual(m.opt.contact_sensor_maxmatch, 5)
 
+  def test_make_data_eq_type(self):
+    mjm, _, _, _ = test_data.fixture(path="constraints.xml")
+
+    d = mjwarp.make_data(mjm)
```
erikfrey: The fix seems to be in `put_data` but here you are calling `make_data` - this is not exercising your fix.

swankystark: I will update this asap!

swankystark: I reproduced the issue using an environment with NumPy < 2.0, as suggested.

Before the fix:
- On Linux, running `pytest` crashes during test setup inside `put_data`
- Error: `'numpy.float64' object cannot be interpreted as an integer`
- The failure occurs when `wp.full` receives a float64 due to NumPy type
  promotion from `np.sum(...) * int` involving `eq_active`
- I was not able to reproduce this crash on Windows with NumPy < 2.0;
  the issue appears to be Linux-specific

After applying the fix:
- The sum is explicitly cast to `int` before multiplication
- Running `pytest` on Linux now completes successfully
- Result: 940 passed, 4 skipped

This demonstrates that the fix resolves the NumPy < 2.0 float promotion issue
without requiring an additional regression test. I will remove the added test
from this PR.

mujoco_warp/_src/io_test.py
```diff
@@ -15,6 +15,9 @@
 
 """Tests for io functions."""
 
+import sys
+sys.path.insert(0, ".")
```
erikfrey: can you revert `io_test.py` completely from this PR?

erikfrey:
@swankystark I'm still seeing changes to `io_test.py` in the PR - perhaps you need to merge main into your branch?

swankystark:
@erikfrey  Done! I've merged main into my branch. The PR should now only show the fix in

erikfrey (approved):
Thank you!

PR #895, opened 2025-12-09
add memory report to testspeed

thowell:
run testspeed with `--memory=True` to print a memory report

**humanoid**
```
mjwarp-testspeed ./benchmark/humanoid/humanoid.xml --nconmax=24 --njmax=64 --nstep=1 --nworld=128 --memory=True
```

```
Loading model from: benchmark/humanoid/humanoid.xml...
  nbody: 17 nv: 27 ngeom: 20 nu: 21 is_sparse: False
  broadphase: NXN broadphase_filter: PLANE|SPHERE|OBB
  solver: NEWTON cone: PYRAMIDAL iterations: 100 iterative linesearch iterations: 50
  integrator: EULER graph_conditional: True
Data
  nworld: 128 naconmax: 3072 njmax: 64

Rolling out 1 steps at dt = 0.005...

Summary for 128 parallel rollouts

Total JIT time: 0.33 s
Total simulation time: 0.01 s
Total steps per second: 25,193
Total realtime factor: 125.96 x
Total time per step: 39694.24 ns
Total converged worlds: 128 / 128

Total device memory: 48640.12 MB
Total utilized memory: 5.17 MB (0.01%)

Model memory (0.58%):
 (no field >= 1% of utilized memory)
Data memory (99.37%):
 xmat: 0.07 MB (1.45%)
 ximat: 0.07 MB (1.45%)
 geom_xmat: 0.09 MB (1.70%)
 cdof: 0.08 MB (1.53%)
 cinert: 0.08 MB (1.61%)
 actuator_moment: 0.28 MB (5.36%)
 crb: 0.08 MB (1.61%)
 qM: 0.38 MB (7.41%)
 qLD: 0.36 MB (6.89%)
 cdof_dot: 0.08 MB (1.53%)
 contact.frame: 0.11 MB (2.04%)
 contact.friction: 0.06 MB (1.13%)
 contact.solimp: 0.06 MB (1.13%)
 efc.J: 0.88 MB (16.93%)
 efc.cholesky_L_tmp: 0.38 MB (7.41%)
 efc.quad: 0.09 MB (1.81%)
 efc.h: 0.38 MB (7.41%)
```
 
 **aloha pot**
 ```
 mjwarp-testspeed benchmark/aloha_pot/scene.xml --nconmax=32 --njmax=128 --nstep=1 --nworld=128 --memory=True
 ```
 
```
Loading model from: benchmark/aloha_pot/scene.xml...
[...]

PR #896, opened 2025-12-09
allocate newton solver hessian and factor in solver._solve

thowell:
removes `Constraint.h` (newton solver hessian) and `Constraint.cholesky_L_tmp` (newton solver hessian factor) and instead allocates these arrays as necessary in `solver._solve`.

this change saves memory when using the cg solver.

---

the memory savings can be considerable:

**aloha cloth**

using memory report from #895:
```
mjwarp-testspeed benchmark/aloha_cloth/scene.xml --nconmax=920 --njmax=8000 --nstep=1 --nworld=256 --memory=True
```

previously, `efc.h` and `efc.cholesky_L_tmp` were each 16.8% of the total memory utilization

with this pr:
```
Loading model from: benchmark/aloha_cloth/scene.xml...
  nbody: 922 nv: 2716 ngeom: 995 nu: 14 is_sparse: True
  broadphase: SAP_TILE broadphase_filter: PLANE|SPHERE|OBB
  solver: CG cone: PYRAMIDAL iterations: 100 iterative linesearch iterations: 50
  integrator: EULER graph_conditional: True
Data
  nworld: 256 naconmax: 235520 njmax: 8000

Rolling out 1 steps at dt = 0.002...

Summary for 256 parallel rollouts

Total JIT time: 1.69 s
Total simulation time: 0.66 s
Total steps per second: 387
Total realtime factor: 0.77 x
Total time per step: 2581917.70 ns
Total converged worlds: 256 / 256

Total device memory: 48640.12 MB
Total utilized memory: 28558.24 MB (58.71%)

Model memory (0.16%):
 (no field >= 1% of utilized memory)
Data memory (99.84%):
 flexedge_J: 6845.70 MB (23.97%)
 efc.J: 21250.00 MB (74.41%)
```

the solver is cg and memory is not allocated for a newton solver hessian and hessian factor arrays (ie, both are `wp.empty((d.nworld, 0, 0), dtype=float)`)

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -1351,7 +1351,7 @@ def update_gradient_set_h_qM_lower_sparse(
   qM_in: wp.array3d(dtype=float),
   efc_done_in: wp.array(dtype=bool),
   # Out:
-  h: wp.array3d(dtype=float),
+  h_out: wp.array3d(dtype=float),
 ):
   worldid, elementid = wp.tid()
 
@@ -1360,7 +1360,7 @@ def update_gradient_set_h_qM_lower_sparse(
 
   i = qM_fullm_i[elementid]
   j = qM_fullm_j[elementid]
-  h[worldid, i, j] += qM_in[worldid, 0, elementid]
+  h_out[worldid, i, j] += qM_in[worldid, 0, elementid]
 
 
 @wp.func
@@ -1392,7 +1392,7 @@ def update_gradient_JTDAJ_sparse_tiled(tile_size: int, njmax: int):
     efc_state_in: wp.array2d(dtype=int),
     efc_done_in: wp.array(dtype=bool),
     # Out:
-    h: wp.array3d(dtype=float),
+    h_out: wp.array3d(dtype=float),
   ):
     worldid, elementid = wp.tid()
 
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/io.py
```diff
@@ -199,6 +199,9 @@ def not_implemented(objtype, objid, geomtype):
   m.opt = opt
   m.stat = stat
 
+  m.nv_pad = _get_padded_sizes(
+    mjm.nv, 0, is_sparse(mjm), types.TILE_SIZE_JTDAJ_SPARSE if is_sparse(mjm) else types.TILE_SIZE_JTDAJ_DENSE
+  )[1]
```
adenzler-nvidia, line 204: The alternative to adding this to the model would be to just look up the right array dimension from another one that uses these dims - J or grad, for example.

mujoco_warp/_src/solver.py
```diff
@@ -2073,8 +2074,14 @@ def _solve(m: types.Model, d: types.Data):
   else:
     wp.copy(d.qacc, d.qacc_smooth)
 
+  # Newton solver Hessian
+  h_nv = m.nv_pad if m.opt.solver == types.SolverType.NEWTON else 0
+  h = wp.empty((d.nworld, h_nv, h_nv), dtype=float)
+  hfactor_nv = m.nv_pad if (m.opt.solver == types.SolverType.NEWTON) and (m.nv > _BLOCK_CHOLESKY_DIM) else 0
+  hfactor = wp.empty((d.nworld, hfactor_nv, hfactor_nv), dtype=float)
```
adenzler-nvidia: I think you need wp.zeros here to make sure the padding is 0.

adenzler-nvidia (approved):
2 comment, but no objections, this looks good. Thank you!

mujoco_warp/_src/solver.py
thowell: fixed. thanks!

mujoco_warp/_src/io.py
thowell: will leave `Model.nv_pad` for now. happy to revisit this in the future if we want to infer the dimension from array shapes.

PR #899, opened 2025-12-10
Move narrowphase kernel compilation to separate conftest.py that is included in the package

adenzler-nvidia:
running the tests from a clean pip install is broken because conftest.py is not included in the package. This change adds another conftest.py inside the package that sets up the kernel cache for the tests.

I was hoping to find a better way to include the main conftest.py into the package, but I can't do it the same way as test_data because it's a .py file and will not be considered by setuptools. Happy to consider another approach though, I'm not a setuptools expert.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -17,6 +17,7 @@ import warp as wp
 
 collect_ignore = ["benchmark/mujoco_menagerie"]
 
+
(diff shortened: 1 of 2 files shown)


adenzler-nvidia:
closing in favor of #901

PR #909, opened 2025-12-11
Optimization: disable bounds_check in mul_m_dense

adenzler-nvidia:
This PR:

```

Total JIT time: 0.32 s
Total simulation time: 2.28 s
Total steps per second: 3,593,219
Total realtime factor: 17,966.09 x
Total time per step: 278.30 ns
```

main:

```

Total JIT time: 0.32 s
Total simulation time: 2.30 s
Total steps per second: 3,569,207
Total realtime factor: 17,846.03 x
Total time per step: 280.17 ns
```

### DIFF
--- a/mujoco_warp/_src/support.py
+++ b/mujoco_warp/_src/support.py
@@ -116,10 +116,10 @@ def mul_m_dense(tile: TileSet, check_skip: bool):
         return
 
     dofid = adr[nodeid]
-    qM_tile = wp.tile_load(qM_in[worldid], shape=(TILE_SIZE, TILE_SIZE), offset=(dofid, dofid))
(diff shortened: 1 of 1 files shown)


thowell:
do we expect to be able to skip most of the tile operation bounds checks throughout the code for performance improvements?

thowell (approved):

adenzler-nvidia:
It's tricky. In terms of functionality, the codebase is safe because we never read out of bounds intentionally and rely on the oob accesses being zeroed.

However for many of the 2d tiles that are 16B aligned on purpose, I can see some slowdowns if I disable the bounds check. I'm in conversations with the warp team about that one. We should definitely audit the codebase a bit and see where we can have similar benefits.

So far the only places where I looked at this very closely is JTDAJ and cholesky/block cholesky. And now this.

PR #910, opened 2025-12-11
kinematics: update _kinematics_level and test_kinematics

thowell:


mujoco_warp/_src/smooth.py
```diff
@@ -105,15 +105,19 @@ def _kinematics_level(
 
     # mocap bodies have world body as parent
     mocapid = body_mocapid[bodyid]
-    if pid == 0 and mocapid != -1:
+    if mocapid >= 0:
```
erikfrey, line 108: perhaps with this, we can now afford a simplifcation and maybe even save a few registers:

```py
mocapid = body_mocapid[bodyid]
if mocapid >= 0:
  xpos = mocap_pos_in[worldid, mocapid]
  xquat = mocap_quat_in[worldid, mocapid]
else:
  xpos = body_pos[body_pos_id, bodyid]
  xquat = body_quat[body_quat_id, bodyid]

if pid >= 0:
  xpos = xmat_in[worldid, pid] @ xpos + xpos_in[worldid, pid]
  xquat = math.mul_quat(xquat_in[worldid, pid], xquat)
```

WDYT?

mujoco_warp/_src/smooth.py
```diff
@@ -142,7 +146,8 @@ def _kinematics_level(
       jntadr += 1
 
   xpos_out[worldid, bodyid] = xpos
-  xquat_out[worldid, bodyid] = wp.normalize(xquat)
+  xquat = wp.normalize(xquat)
```
thowell, line 146: i think it is important to normalize the quaternion before computing `xmat` and `xipos`

mujoco_warp/_src/smooth_test.py
```diff
@@ -114,29 +114,46 @@ def test_mocap_kinematics(self):
...
     if make_data:
-      d = mjw.make_data(mjm)
+      mjd = mujoco.MjData(mjm)
+      d = mjw.make_data(mjm, nworld=nworld)
 
     for arr in (d.xpos, d.xipos, d.xquat, d.xmat, d.ximat, d.xanchor, d.xaxis, d.site_xpos, d.site_xmat):
       arr_view = arr[:, 1:]  # skip world body
       arr_view.fill_(wp.inf)
 
+    qpos = np.random.uniform(size=mjm.nq)
```
erikfrey: how about instead of random qpos and quat, we put in a keyframe with some non-init state?  you can put mocap pos and quat into a keyframe too.  it makes this code a bit cleaner, and i'm always suspicious of random init, it can be the source of flakiness if we rely on it too much.

mujoco_warp/_src/smooth_test.py
```diff
@@ -114,29 +114,46 @@ def test_mocap_kinematics(self):
   @parameterized.parameters(True, False)
   def test_kinematics(self, make_data):
     """Tests kinematics."""
-    mjm, mjd, m, d = test_data.fixture("pendula.xml")
+    nworld = 2
```
erikfrey, line 118: I'm sure there's some good reason, but why are you testing nworld = 2 for kinematics?  Consider adding a comment if this is needed.

mujoco_warp/_src/smooth.py
thowell: done

mujoco_warp/_src/smooth_test.py
thowell: done

mujoco_warp/_src/smooth_test.py
thowell: `nworld > 1` is to make the test a bit more similar to the mjx vmap tests for kinematics
https://github.com/google-deepmind/mujoco/blob/a0a56065e0d3b7ea475d6120a122e516784c47eb/mjx/mujoco/mjx/warp/smooth_test.py#L98
https://github.com/google-deepmind/mujoco/blob/a0a56065e0d3b7ea475d6120a122e516784c47eb/mjx/mujoco/mjx/warp/smooth_test.py#L189

also in general we should probably improve most of the tests by testing `nworld > 1` scenarios to check that `Model` batched field indexing is working correctly and avoid future issues like those addressed in #821, #824, #865

thowell: added a todo for batched `Model` field testing that is related to `nworld=2`

PR #917, opened 2025-12-13
GitHub CI: separate release and dev tests.

thowell:
update to ci.yml to test release and dev versions of warp-lang and mujoco

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,9 +54,9 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10", "3.12"]
-        warp-version: ["nightly", "release"]
+        dep-version: ["release", "dev"]
 
-    continue-on-error: ${{ matrix.warp-version == 'release' }}
+    continue-on-error: ${{ matrix.dep-version == 'release' }}
 
     steps:
       - uses: actions/checkout@v3
@@ -64,16 +64,19 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Install pip and uv
         run: |
           python -m pip install --upgrade pip
           pip install uv
-          uv pip install --system -e .[dev,cpu]
-      - name: Uninstall nightly warp-lang and install release warp-lang
-        if: matrix.warp-version == 'release'
+      - name: Install release
+        if: matrix.dep-version == 'release'
         run: |
-          pip uninstall -y warp-lang
-          pip install warp-lang
+          pip install pytest pytest-xdist
+          uv pip install --system -e .
+      - name: Install dev
+        if: matrix.dep-version == 'dev'
+        run: |
+          uv pip install --system -e .[dev]
       - name: Test with pytest
         run: |
           pytest


.github/workflows/ci.yml
```diff
@@ -69,6 +69,13 @@ jobs:
       - name: Test with pytest
         run: |
           pytest
+      - name: Uninstall dev warp-lang and install release warp-lang
```
erikfrey: Instead of running the tests serially, let's add another variable to the matrix above so that it runs in parallel.  So it means we should end up with 4 job runs.

Also can you see if for now, you can make the CI warn instead of error if we have a test failure against warp release?

so:

3.10, nightly => error on fail
3.12, nightly => error on fail
3.10, release => warn on fail
3.12, release => warn on fail

thowell: added 
- 'warp-version' strategy: release testing is distinct from nightly testing
- 'continue-on-error': test failures for release warp-lang do not result in ci failure

.github/workflows/ci.yml
```diff
@@ -66,6 +69,11 @@ jobs:
           python -m pip install --upgrade pip
           pip install uv
           uv pip install --system -e .[dev,cpu]
+      - name: Uninstall nightly warp-lang and install release warp-lang
+        if: matrix.warp-version == 'release'
+        run: |
+          pip uninstall -y warp-lang
```
erikfrey: OK but we also want to use the release version of mujoco, right?  Basically any non-nightlies?

Rather than explicitly uninstalling/reinstalling, which might be a source of error, can you rely on the deps in pyproject.toml?  it may require you to tweak the optional dependencies so that we have a way of installing dev with nightly deps and dev with release deps, so it might look something like:

```
- name: Install dependencies
  run: |
    pip install uv
    uv pip install --system -e .[${{ matrix.deps-version == 'release' && 'dev-release' || 'dev' }},cpu]
```

thowell: the pr is updated to test the `pip install -e .` version (release: mujoco, warp) and the `pip install -e .[dev]` (nightly: mujoco, warp)

.github/workflows/ci.yml
```diff
@@ -54,18 +54,33 @@ jobs:
...
+        run: |
+          uv pip install --system -e .
+      - name: Install dev
+        if: matrix.dep-version == 'dev'
+        run: |
+          uv pip install --system -e .[dev]
+      - name: Install pytest
+        if: matrix.dep-version == 'release'
+        run: |
+          pip install pytest
```
erikfrey: nit: why not just put these two above, so:

```
      - name: Install release
        if: matrix.dep-version == 'release'
        run: |
          pip install pytest pytest-xdist
          uv pip install --system -e .
      - name: Install dev
        if: matrix.dep-version == 'dev'
        run: |
          uv pip install --system -e .[dev]
```

btw why are you using `pip` for some commands and `uv pip` for others?  I might be missing some details but if it's easy to just do one or the other, let's consolidate.

erikfrey (approved):
LGTM, just one nit

.github/workflows/ci.yml
thowell: moved pip install pytest pytest-xdist as suggested

thowell: the pip / uv pip usage here is to follow what we have for installation steps in the readme.

thowell: we should consider moving everything to uv. will leave this as potential future work for now, relatedly see #973 for proposed changes with the ci and uv.

PR #919, opened 2025-12-13
Better defaults for nconmax, njmax.

### LINKED ISSUE #411: Find better defaults for `njmax` and `nconmax`
For many MuJoCo models, `io.put_data` picks a default `njmax` and `nconmax` that is too low and yields overflows during simulation.  We should design better heuristics for picking `njmax` and `nconmax`.

erikfrey:
These defaults were checked against our benchmark and test_data models, a linear regression followed by a few heuristics for sim features that are more demanding for contacts / constraints (sdf, flex).

Closes #411

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -560,6 +560,32 @@ def _get_padded_sizes(nv: int, njmax: int, is_sparse: bool, tile_size: int):
   return njmax_padded, nv_padded
 
 
+def _default_nconmax(mjm: mujoco.MjModel, mjd: Optional[mujoco.MjData] = None) -> int:
+  """Returns a default guess for an ideal nconmax given a Model and optional Data.
+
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/io.py
```diff
@@ -582,17 +582,23 @@ def make_data(
     The data object containing the current state and output arrays (device).
   """
   # TODO(team): move nconmax, njmax to Model?
-  # TODO(team): improve heuristic for nconmax and njmax
-  nconmax = nconmax or 20
-  njmax = njmax or nconmax * 6
+  valid_sizes = (2 + (np.arange(19) % 2)) * (2 ** (np.arange(19) // 2 + 3))  # 16, 24, 32, 48, ... 8192
```
thowell: since the heuristic is the same for `make_data` and `put_data` we could create and call a utility function for computing `nconmax` and `njmax`?

PR #921, opened 2025-12-14
Enable margin trick for heightfield collisions.

kevinzakka:
Implement the margin trick optimization for hfield-sphere/capsule collisions (previously marked TODO). Also fix EPA witness point to select the correct prism face based on vertex indices.

The support function tags vertices as -2 (bottom) or -3 (top) based on search direction. When EPA produces a face with mixed vertex indices, we use majority voting to select the appropriate prism face for witness point calculation.


thowell:
thanks @kevinzakka!

kbayes:
@kevinzakka unfortunately your logic is flawed here as the indices can only identify that the sides of the prisms are in contact with the other geom not the orientation of the true contact normal. Also the TODOs and the margin disablement are recent additions and should not be removed.

PR #922, opened 2025-12-14
Avoid possible NaNs when witness points are too close together

kbayes:


kevinzakka:
We're still seeing NaNs with the G1 capsule / hfield task with this branch.

thowell:
@kbayes are these changes superseded by #928 and should we close this pr?

kbayes:
@thowell actually no it's not superseded. Let me test it out, as I believe it might be worthwhile to keep.

thowell:
@kbayes what is the status of this pr?

PR #923, opened 2025-12-15
Optimization: switching tree traversal per branch

Kenny-Vilella:
~~This PR changes the way we do tree traversal depending on tree topology.
In case where the kinematic tree is deep and narrow, the tree traversal is switched to a per branch traversal.~~

This PR changes the way we do tree traversal from a level approach to a per branch traversal.

Perf improvement on my RTX4080 (average of 5 runs):

| Environment               | Step time (main) | Step time (this PR) | Δ %     |
| ----------------------------- | --------------------- | ------------------------- | --------- |
AlohaPot                       |  691.5                 |   681                      |  1.5      |
ApptronikApolloFlat      |  674.6                 |   632                      |  6.3      |
ApptronikApolloHfield   |  3057.5               |  3032                     |  0.8      |
ApptronikApolloTerrain  |  1940.6              |  1889.4                   |  2.6     |
FrankaEmikaPanda      |  82.0                   |  72.1                      |  12.0   |
Humanoid                     |  474.7                 |  462.7                    |  2.5      |
ThreeHumanoids          |  5951.5               |  5883.7                  |  1.1      |

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -259,18 +259,18 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   branches = [ancestor_chain(l) for l in np.where(children_count[1:] == 0)[0] + 1]
   m.nbranch = len(branches)
 
-  branch_bodies_flat = []
-  branch_start = []
+  body_branches = []
+  body_branch_start = []
   offset = 0
 
   for branch in branches:
-    branch_start.append(offset)
-    branch_bodies_flat.extend(branch)
+    body_branches.extend(branch)
+    body_branch_start.append(offset)
     offset += len(branch)
-  branch_start.append(offset)
+  body_branch_start.append(offset)
 
-  m.branch_bodies = np.array(branch_bodies_flat, dtype=int)
-  m.branch_start = np.array(branch_start, dtype=int)
+  m.body_branches = np.array(body_branches, dtype=int)
+  m.body_branch_start = np.array(body_branch_start, dtype=int)
 
   m.mocap_bodyid = np.arange(mjm.nbody)[mjm.body_mocapid >= 0]
   m.mocap_bodyid = m.mocap_bodyid[mjm.body_mocapid[mjm.body_mocapid >= 0].argsort()]
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -54,8 +54,8 @@ def _kinematics_branch(
   jnt_qposadr: wp.array(dtype=int),
   jnt_pos: wp.array2d(dtype=wp.vec3),
   jnt_axis: wp.array2d(dtype=wp.vec3),
-  branch_bodies: wp.array(dtype=int),
-  branch_start: wp.array(dtype=int),
+  body_branches: wp.array(dtype=int),
+  body_branch_start: wp.array(dtype=int),
   # Data in:
   qpos_in: wp.array2d(dtype=float),
   mocap_pos_in: wp.array2d(dtype=wp.vec3),
@@ -68,13 +68,13 @@ def _kinematics_branch(
 ):
   worldid, branchid = wp.tid()
 
-  start = branch_start[branchid]
-  end = branch_start[branchid + 1]
+  start = body_branch_start[branchid]
+  end = body_branch_start[branchid + 1]
 
   qpos = qpos_in[worldid]
 
   for i in range(start, end):
-    bodyid = branch_bodies[i]
+    bodyid = body_branches[i]
     pid = body_parentid[bodyid]
     jntadr = body_jntadr[bodyid]
     jntnum = body_jntnum[bodyid]
@@ -313,8 +313,8 @@ def kinematics(m: Model, d: Data):
       m.jnt_qposadr,
       m.jnt_pos,
       m.jnt_axis,
-      m.branch_bodies,
-      m.branch_start,
+      m.body_branches,
+      m.body_branch_start,
       d.qpos,
       d.mocap_pos,
       d.mocap_quat,
@@ -1022,8 +1022,8 @@ def _cacc_branch(
   body_parentid: wp.array(dtype=int),
   body_dofnum: wp.array(dtype=int),
   body_dofadr: wp.array(dtype=int),
-  branch_bodies: wp.array(dtype=int),
-  branch_start: wp.array(dtype=int),
+  body_branches: wp.array(dtype=int),
+  body_branch_start: wp.array(dtype=int),
   # Data in:
   qvel_in: wp.array2d(dtype=float),
   qacc_in: wp.array2d(dtype=float),
@@ -1036,14 +1036,14 @@ def _cacc_branch(
 ):
   worldid, branchid = wp.tid()
 
-  start = branch_start[branchid]
-  end = branch_start[branchid + 1]
+  start = body_branch_start[branchid]
+  end = body_branch_start[branchid + 1]
 
-  bodyid = branch_bodies[start]
+  bodyid = body_branches[start]
   pid = body_parentid[bodyid]
   local_cacc = cacc_out[worldid, pid]
   for i in range(start, end):
-    bodyid = branch_bodies[i]
+    bodyid = body_branches[i]
     dofnum = body_dofnum[bodyid]
     dofadr = body_dofadr[bodyid]
     for j in range(dofnum):
@@ -1061,8 +1061,8 @@ def _rne_cacc_forward(m: Model, d: Data, flg_acc: bool = False):
       m.body_parentid,
       m.body_dofnum,
       m.body_dofadr,
-      m.branch_bodies,
-      m.branch_start,
+      m.body_branches,
+      m.body_branch_start,
       d.qvel,
       d.qacc,
       d.cdof,
@@ -1716,8 +1716,8 @@ def _comvel_branch(
   body_jntadr: wp.array(dtype=int),
   body_dofadr: wp.array(dtype=int),
   jnt_type: wp.array(dtype=int),
-  branch_bodies: wp.array(dtype=int),
-  branch_start: wp.array(dtype=int),
+  body_branches: wp.array(dtype=int),
+  body_branch_start: wp.array(dtype=int),
   # Data in:
   qvel_in: wp.array2d(dtype=float),
   cdof_in: wp.array2d(dtype=wp.spatial_vector),
@@ -1727,14 +1727,14 @@ def _comvel_branch(
 ):
   worldid, branchid = wp.tid()
 
-  start = branch_start[branchid]
-  end = branch_start[branchid + 1]
+  start = body_branch_start[branchid]
+  end = body_branch_start[branchid + 1]
 
   qvel = qvel_in[worldid]
   cdof = cdof_in[worldid]
 
   for i in range(start, end):
-    bodyid = branch_bodies[i]
+    bodyid = body_branches[i]
     pid = body_parentid[bodyid]
     cvel = cvel_out[worldid, pid]
     dofid = body_dofadr[bodyid]
@@ -1799,8 +1799,8 @@ def com_vel(m: Model, d: Data):
       m.body_jntadr,
       m.body_dofadr,
       m.jnt_type,
-      m.branch_bodies,
-      m.branch_start,
+      m.body_branches,
+      m.body_branch_start,
       d.qvel,
       d.cdof,
     ],
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -1023,8 +1023,8 @@ class Model:
     has_sdf_geom: whether the model contains SDF geoms
     block_dim: block dim options
     body_tree: list of body ids by tree level
-    branch_bodies: flattened body ids for all branches
-    branch_start: start index in branch_bodies for each branch   (nbranch + 1,)
+    body_branches: flattened body ids for all branches
+    body_branch_start: start index in body_branches for each branch   (nbranch + 1,)
     mocap_bodyid: id of body for mocap                       (nmocap,)
     body_fluid_ellipsoid: does body use ellipsoid fluid      (nbody,)
     jnt_limited_slide_hinge_adr: limited/slide/hinge jntadr
@@ -1372,8 +1372,8 @@ class Model:
   has_sdf_geom: bool
   block_dim: BlockDim
   body_tree: tuple[wp.array(dtype=int), ...]
-  branch_bodies: wp.array(dtype=int)
-  branch_start: wp.array(dtype=int)
+  body_branches: wp.array(dtype=int)
+  body_branch_start: wp.array(dtype=int)
   mocap_bodyid: array("nmocap", int)
   body_fluid_ellipsoid: array("nbody", bool)
   jnt_limited_slide_hinge_adr: wp.array(dtype=int)


Kenny-Vilella:
@erikfrey @thowell  This optimization adds some complexity to the codebase (one more switch), so I just wonder if in principle you would be in favor of such a change.

One thing is that I am not yet totally convinced about the backward tree traversal using segments.
In theory, it should work great, but in practice I do not see significant speedup, maybe less than 1% for the franka benchmark.

This PR should work well with Alain's PR about quat/mat33, so I will wait for his PR to be merged and then update this one.
I will also reevaluate the usefulness of the backward tree traversal after the change.

thowell:
nice optimizations @Kenny-Vilella!

this would primarily improve performance for environments that contain chains with many dofs? 

if that is the case, it would be interesting to see comparison benchmark results for:
- a world with 1 chain and many dofs
- a world with many chains each with many dofs.

erikfrey:
@Kenny-Vilella your PRs are always a treat!

You’re right that this adds complexity - do we need both approaches?  Did you find actually meaningful benchmark scenes for which your approach is a lot slower?

Basically I’m trying to justify whether we really need both approaches, or just fully switch to yours, or keep the current.

Kenny-Vilella:
I made some tests using a script that creates an articulation with N chains with every chain having M DoFs. This type of very orderly system introduce some bias, but let's say is good enough for our purpose.

First thing to mention is that separating the calculation of xipos and ximat has sometimes a huge impact for both cases.
For example, if I have 2 chains of 15 DoF, with the per-level approach the total time spent on kinematic level goes from 420ms on main to 255ms + 23ms on my branch.
Because of this, it only makes sense to compare the two approaches within this PR and not with main.

For a chain of 64 DoFs, I can only see perf improvement in Nsight because qLD_acc totally dominate the compute time (another optimization potential but quite challenging...). 
Chain of 2 DoFs => 0% perf improvement
Chain of 4 DoFs => 10% perf improvement
Chain of 8 DoFs => 4% perf improvement
Chain of 16 DoFs => 7% perf improvement
Chain of 30 DoFs => 6% perf improvement

Surprisingly, even for 16 chains of 1 DoF, I see a negligible 1% perf improvement.
Only cases where I can measure significantly worse perf is for cases with 1-2 branches and 1-2 DoFs.
For example -12% for 2 branches, 1 DoF.

For X chains of X Dofs, these are the results:
2 => 3% perf decrease
3 => 3% perf improvement
4 => 1% perf improvement
5 => 2% perf improvement
6 => 4% perf improvement
7 => 4% perf improvement
8 => 3% perf improvement

I think it could make sense to totally switch to this new approach.
I didn't expect that it performs so well for most cases.

erikfrey:
@Kenny-Vilella OK!  If we aren't seeing any reason to keep both, can you take a stab at changing this PR to only do per branch traversal?  Please focus on simplicity and legibility as `smooth.py` is the first place many curious newcomers land when they are trying to understand the code.

Once you've done that let us know and we'll do a proper review.  Thanks for finding this speedup!

Kenny-Vilella:
@erikfrey Thanks !
I am waiting for this [PR](https://github.com/google-deepmind/mujoco_warp/pull/902) to be merged and then will update this one.

Kenny-Vilella:
@erikfrey @thowell  Updated this PR, since the mat33 <-> quat PR is on pause.

Latest number on my RTX4080, median of 5 runs of the benchmark suite:

| Environment               | Step time (main) | Step time (this PR)  |  Δ %      |
| ------------------------- | ------------------- | ---------------------- | --------- |
aloha_cloth                    |  3784938            |   3802677               |   -0.5      |
aloha_pot                       |  684.8                  |   663.9                     |   3.2     |
aloha_sdf                        |  4365                   |   4371                     |   -0.1    |
apollo_flat                      |  651.1                  |   608.9                    |   6.9      |
apollo_hfield                  |  1092                   |   1052                     |   3.8      |
apollo_terrain                |  1931                   |   1878                     |   2.8      |
cloth                               |  3209902            |    3192023              |  0.6      |
franka_emika_panda    |  82.7                     |   72.0                       |   14.8   |
humanoid                      |  478.7                   |   463.7                     |   3.2    |
three_humanoids         |   5752                   |   5795                      |   -0.7    |

Negative number are within standard deviation.

mujoco_warp/_src/io.py
```diff
@@ -177,6 +269,35 @@ def _check_friction(name: str, id_: int, condim: int, friction, checks):
     opt_kwargs["impratio_invsqrt"] = 1.0 / np.sqrt(np.maximum(mjm.opt.impratio, mujoco.mjMINVAL))
   opt = types.Option(**opt_kwargs)
 
+  # create model
```
erikfrey: is there a reason you moved the model creation up here?

trying to group the code here so that it's:

- opt
- stat
- model

can we retain that order?

mujoco_warp/_src/io.py
```diff
@@ -26,6 +26,98 @@
...
+    # Add first independent bodies and then chains, order is important
+    if parallel_bodies:
+      segments.append((parallel_bodies, False))
+    for chain in chain_bodies:
+      segments.append((chain, True))
+
+  return segments
+
+
+def _compute_branches(mjm: mujoco.MjModel) -> list[list[int]]:
```
erikfrey: why not something like:

```py
def _compute_branches(mjm: mujoco.MjModel) -> list[list[int]]:
  """Identify branches in kinematic tree."""
  cc = np.bincount(mjm.body_parentid[1:], minlength=mjm.nbody)
  anc = lambda b: anc(mjm.body_parentid[b]) + [b] if b else []
  return [anc(l) for l in np.where(cc[1:] == 0)[0] + 1]
  ```

btw it's short enough that maybe doesn't need to be a function anymore?  could go either way

File: mujoco_warp/_src/io.py [hunk 3]
```diff
@@ -26,6 +26,98 @@
 from .warp_util import nested_kernel
 
 
+def _compute_bottom_up_segments(mjm: mujoco.MjModel) -> list[tuple[list[int], bool]]:
```
erikfrey: i think it might be a bit more legible if you just made it two separate functions:

```py
def _compute_parallel_segments(mjm: mujoco.MjModel) -> list[list[int]]:
  """Compute parallel segments - groups of leaves at each depth level."""
  if mjm.nbody <= 1:
    return []
  depth = np.zeros(mjm.nbody, dtype=int)
  cc = np.bincount(mjm.body_parentid[1:], minlength=mjm.nbody).copy()
  for i in range(1, mjm.nbody):
    depth[i] = depth[mjm.body_parentid[i]] + 1

  segments = []
  for d in range(depth.max(), 0, -1):
    leaves = [b for b in np.where(depth == d)[0] if cc[b] == 0]
    if leaves:
      segments.append(leaves)
      for b in leaves:
        cc[mjm.body_parentid[b]] -= 1
  return segments


def _compute_chain_segments(mjm: mujoco.MjModel) -> list[list[int]]:
  """Compute chain segments - linear paths through single-child bodies."""
  if mjm.nbody <= 1:
    return []
  cc = np.bincount(mjm.body_parentid[1:], minlength=mjm.nbody)

  segments = []
  for leaf in np.where(cc[1:] == 0)[0] + 1:
    chain, current = [], leaf
    while current > 0 and cc[current] <= 1:
      chain.append(current)
      current = mjm.body_parentid[current]
    if len(chain) > 1:
      segments.append(chain)
  return segments
```

haven't fully tested it but something like that

mujoco_warp/_src/smooth.py
```diff
@@ -496,14 +593,21 @@ def com_pos(m: Model, d: Data):
...
 
-  for i in reversed(range(len(m.body_tree))):
-    body_tree = m.body_tree[i]
-    wp.launch(
-      _subtree_com_acc,
-      dim=(d.nworld, body_tree.size),
-      inputs=[m.body_parentid, d.subtree_com, body_tree],
-      outputs=[d.subtree_com],
-    )
+  for segment_bodies, is_chain in zip(m.bottom_up_segment_bodies, m.bottom_up_segment_is_chain):
```
erikfrey: rather than a zip maybe just have two arrays, one for the chain and one for the not chain

erikfrey (commented):
Apologies in advance for the nitpickiness - this is touching the very early parts of our code that people are likely to read first, so trying to keep it as simple as possible.

One high level question: this seems to be changing two functions, kinematics and crb.  Did you verify that they both lead to the speedup or is one more important than the other?

Want to be 100% sure we are only adding complexity here if it's helpful

Kenny-Vilella:
I do enjoy some nitpickiness !

Let's start with the meaty comment.
Give you some numbers from nsight on the two benchmarks that significantly improved, Apollo flat and Franka.
The kernel changed are:
- `_kinematics_level`  => to a branch approach
- `_cacc`  => to a branch approach
- `_comvel_level`  => to a branch approach
- `_subtree_com_acc`  => to a segment approach
- `_crb_accumulate`  => to a segment approach
- `_cfrc_backward`  => to a segment approach

On the Franka:
|   kernel                   | main  (%)    |  main (avg us)   |   HEAD (%)   |   HEAD  (avg us)   |
| --------------------- | ------------ | ----------------- | --------------- | -------------------- |
`_kinematics_level`    |   16.3         |   413                 |   8.7               |  196.2
`_comvel_level`          |   6.3            |  161.6              |   4.8                |  107.3
`_cacc`                        |   3.9            |   100.4              |  2.7                |  61.4
`_crb_accumulate`     |   3.0            |   78.0                |   2.5                |  57.4
`_cfrc_backward`       |   2.2            |   57.1                |   1.8                |   41.6
`_subtree_com_acc` |   1.4            |   35.3                 |  1.0                |  22.3

On the apollo flat:
| kernel                     | main  (%)    |  main (avg us)   |   HEAD (%)   |   HEAD  (avg us)   |
| --------------------- | ------------ | ----------------- | --------------- | -------------------- |
`_kinematics_level`    |    6.2          |    325.1             |    2.4              |     115.8
`_cacc`                        |    3.6          |    190.4              |    2.1              |     103.7
`_comvel_level`          |    2.2          |    116.8             |    1.4              |     68.4 
`_cfrc_backward`       |    1.9          |    97.8                |    1.7             |     82.8
`_crb_accumulate`     |   1.4           |    73.3                |    1.3             |     61.6
`_subtree_com_acc`  |    0.7           |   34.3                 |   0.5              |     26.6

I think that it is better for the user to limit as much as possible the number of tree traversal methods we are using.
So let's group the kernels into two, forward propagation (branch based) and backward propagation (segment based).
In both cases, forward propagation group accounts for ~85% of the perf improvement.

So if you want to limit the amount of change, I can revert the change to the segment kernels, i.e., `_cfrc_backward,` `_crb_accumulate`, and `_subtree_com_acc`.

mujoco_warp/_src/io.py
Kenny-Vilella: My apologies for that.
I moved it back down.

mujoco_warp/_src/io.py
Kenny-Vilella: For non-performance critical part of the code, it's a bit subjective to choose between simplicity and efficiency.

I changed following your suggestion.

mujoco_warp/_src/smooth.py
```diff
@@ -60,92 +58,171 @@ def _kinematics_level(
...
+  xanchor_out: wp.array2d(dtype=wp.vec3),
+  xaxis_out: wp.array2d(dtype=wp.vec3),
+):
+  worldid, branchid = wp.tid()
+
+  start = branch_start[branchid]
+  length = branch_length[branchid]
+
+  for i in range(length):
+    bodyid = branch_bodies[start + i]
```
erikfrey: Dumb question just for my own education:

because branches can share bodies (e.g. if two branches have the same ancestor), does that mean we have multiple threads writing to the same array indices?

e.g. we have two threads both writing to xpos_out[1] if they both share bodyid=1 in their respective branches?

this seems like excessive array writing, no?  did you play with a version where we record an "owner" thread for each body, so that information only gets written once?

mujoco_warp/_src/smooth.py
```diff
@@ -60,92 +58,171 @@ def _kinematics_level(
...
+    pid = body_parentid[bodyid]
+    xpos = xpos_out[worldid, pid]
+    xquat = xquat_out[worldid, pid]
+    xmat = xmat_out[worldid, pid]
+    _compute_body_kinematics(
```
erikfrey: nit: unless we plan to use `_compute_body_kinematics` elsewhere (probably not) let's just inline the logic into the kernel, keep it all in one place - the user has to do less mental bookkeeping on the list of arguments.

mujoco_warp/_src/types.py
```diff
@@ -1022,6 +1022,12 @@ class Model:
     has_sdf_geom: whether the model contains SDF geoms
     block_dim: block dim options
     body_tree: list of body ids by tree level
+    num_branches: number of branches (leaf-to-root paths)
```
erikfrey: nit: let's use the style for expressing a size which is `nsomething` and put it up near the other sizes (e.g. after `nmaxmeshdeg`)

leave it to you to fill in the `something` with a name you like          :-)

mujoco_warp/_src/types.py
```diff
@@ -1022,6 +1022,12 @@ class Model:
...
+    branch_bodies: flattened body ids for all branches
```
erikfrey: nit: let's use the style where the first word communicates the element we are dealing with, so maybe something like `body_branches`?

mujoco_warp/_src/smooth.py
```diff
@@ -60,92 +58,171 @@ def _kinematics_level(
...
+  xpos_out: wp.array2d(dtype=wp.vec3),
+  xquat_out: wp.array2d(dtype=wp.quat),
+  xmat_out: wp.array2d(dtype=wp.mat33),
+  xanchor_out: wp.array2d(dtype=wp.vec3),
+  xaxis_out: wp.array2d(dtype=wp.vec3),
+):
+  worldid, branchid = wp.tid()
+
+  start = branch_start[branchid]
+  length = branch_length[branchid]
```
erikfrey: do you need length?  could you just do `end = branch_start[branchid + 1]` or so?

erikfrey (commented):
Next round of comments - I promise we are converging, thanks for your patience!

mujoco_warp/_src/smooth.py
Kenny-Vilella: You are totally right, we are writing multiple times the same value to the same array.
The benefit of parallelizing the work is much higher than the cost of these excessive array writing.

I did not try but I can do some experiment. We will basically trade more read for less write.
We will need to be smart about what thread write what.

mujoco_warp/_src/smooth.py
Kenny-Vilella: Oups forgot to revert it when removing the legacy method.

Since we are not moving forward with the mat33 -> quat PR, I also split out the xmat calculation in a single kernel.
We can squeeze 1-2% gain out of this. Will update the benchmark number later.

Similarly, removed the `_compute_body_comvel` function.
Note that we could also be smart and split out the calculation of `cdof_dot`.
But it will have no perf gain for most cases.

mujoco_warp/_src/types.py
Kenny-Vilella: Good suggestion.
Changed to nbranch.

mujoco_warp/_src/types.py
Kenny-Vilella: Are you sure about this?
It feels a bit weird to have `body_branches` and then `branch_start`.
Or we should rename `branch_start` to `body_branch_start`?

mujoco_warp/_src/smooth.py
Kenny-Vilella: That's clever !
The only thing is we need to make sure the end value for the last branch is correct.

mujoco_warp/_src/smooth.py
erikfrey: Yes agreed- more read but for nbody size array vs more write for nworld * nbody size array!

So I think it would win but then again, my intuition hit rate is not super high for cuda code.    :-)

mujoco_warp/_src/io.py
Kenny-Vilella: Deprecated as I removed the segment optimization

Kenny-Vilella (commented):
Still need to do the experiment with removing unnecessary write and run again the benchmark.

I reverted the optimization with the segment_chain.
It increases the complexity significantly with a minor perf gain, so let's focus this PR on the most obvious gain.

I will keep investigating the segment_chain optimization after this PR is merged and see if I can get significant enough perf gain to justify the change.

mujoco_warp/_src/smooth.py
Kenny-Vilella: Spent an hour or two on this and didn't manage to get any perf out of it.
Only tried on the kinematic_branch kernel though.

At the end of the day we are as slow as the slowest threads.
My guess is that the write was not the bottleneck because we were overlapping it with other operation, so the change only had thread divergence, which makes the code slightly slower.

Tried three load balancing strategy:
- Shortest branch has priority for writing
- First come - First served, i.e. following the branch creation
- Tried to equalize writing between threads

Kenny-Vilella:
The last benchmark,

On the Franka:
|   kernel                   | main  (%)    |  main (avg us)   |   HEAD (%)   |   HEAD  (avg us)   |
| --------------------- | ------------ | ----------------- | --------------- | -------------------- |
`_kinematics_level`    |   16.3         |   413                 |   7.8               |  171.3
`_comvel_level`          |   6.3            |  161.6              |   4.8                |  107.3
`_cacc`                        |   3.9            |   100.4              |   2.8              |    61.4

On the apollo flat:
| kernel                     | main  (%)    |  main (avg us)   |   HEAD (%)   |   HEAD  (avg us)   |
| --------------------- | ------------ | ----------------- | --------------- | -------------------- |
`_kinematics_level`    |    6.2          |    325.1             |    2.4              |     119.7
`_cacc`                        |    3.6          |    190.4              |    2.1             |     103.8
`_comvel_level`          |    2.2          |    116.8             |     1.2             |     60.8

Note that `_kinematics_level` also include `_compute_body_inertial_frames` and `_compute_body_matrices`.
My previous post did not include `_compute_body_inertial_frames`.

mujoco_warp/_src/smooth.py
Kenny-Vilella: Deprecated as I removed the segment optimization

mujoco_warp/_src/types.py
erikfrey: that's correct, in general we try to make the first word reflect the id type, e.g. `dof` or `jnt` or `body` - so `body_branch_start` works!

mujoco_warp/_src/smooth.py
erikfrey: Interesting, thanks for trying.  Warp perf is still so surprising sometimes!

mujoco_warp/_src/types.py
Kenny-Vilella: So changed both !

Kenny-Vilella:
@erikfrey  Is it ready to be merged
I don't have the sacred power to merge^^

erikfrey:
Looks great, thanks Kenny!

PR #925, opened 2025-12-16
CCD dataclass for ccd memory

thowell:
the current testspeed memory report does not include memory allocated for convex collision detection (ccd)

in order to provide memory information for ccd this pr:

- adds dataclass `CCD` containing memory for ccd
- updates testspeed memory report: change utilized memory to be the sum of `Model` and `Data` memory since `wp.get_mempool_used_mem_current` and `wp.get_mempool_used_mem_high` do not include ccd memory allocated in `collision_convex.convex_narrowphase` (i think the memory is released after the function call?)

this update retains the property that memory is not allocated for ccd if `collision_convex.convex_narrowphase` is never called. however, if the function is called the ccd memory is no longer released after the call but is persistent in `Data.ccd`.

---

**aloha + pot**

```
mjwarp-testspeed ./benchmark/aloha_pot/scene.xml --nconmax=24 --njmax=128 --memory=True
```

this pr:
```
Rolling out 1000 steps at dt = 0.002...

Summary for 8192 parallel rollouts

Total JIT time: 0.68 s
Total simulation time: 4.95 s
Total steps per second: 1,656,290
Total realtime factor: 3,312.58 x
Total time per step: 603.76 ns
Total converged worlds: 8192 / 8192

Total memory: 2002.56 MB / 48640.12 MB (4.12%)
Model memory (0.27%):
 (no field >= 1% of utilized memory)
Data memory (99.73%):
 geom_xmat: 57.38 MB (2.87%)
 efc.J: 96.00 MB (4.79%)
 ccd.epa_vert: 90.00 MB (4.49%)
 ccd.epa_vert1: 90.00 MB (4.49%)
 ccd.epa_vert2: 90.00 MB (4.49%)
 ccd.epa_vert_index1: 30.00 MB (1.50%)
 ccd.epa_vert_index2: 30.00 MB (1.50%)
 ccd.epa_face: 407.25 MB (20.34%)
 ccd.epa_pr: 407.25 MB (20.34%)
 ccd.epa_norm2: 135.75 MB (6.78%)
 ccd.epa_index: 135.75 MB (6.78%)
 ccd.epa_map: 135.75 MB (6.78%)
```

main (528695a43fd2cac8680eb21b9b30f7ac434a9fe3):

```
Rolling out 1000 steps at dt = 0.002...

Summary for 8192 parallel rollouts

Total JIT time: 0.69 s
Total simulation time: 4.95 s
Total steps per second: 1,654,766
Total realtime factor: 3,309.53 x
[...]

PR #927, opened 2025-12-16
NewtonSolver dataclass for newton solver memory

thowell:
similar to #925, this PR creates a dataclass for Newton solver data

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -32,6 +32,9 @@ MJ_MAX_EPAFACES = 5
 TILE_SIZE_JTDAJ_SPARSE = 16
 TILE_SIZE_JTDAJ_DENSE = 16
 
+# dof threshold for blocked cholesky
+BLOCK_CHOLESKY_DIM = 32
+
 
 # TODO(team): add check that all wp.launch_tiled 'block_dim' settings are configurable
 @dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/types.py
```diff
@@ -735,6 +738,31 @@ class TileSet:
...
+    hessian_factor: Hessian factorization  (nworld, nv_pad, nv_pad)
+  """
+
+  hessian: wp.array3d(dtype=float)
+  hessian_factor: wp.array3d(dtype=float)
+
+  def __init__(self):
+    self.hessian = wp.empty((0, 0, 0), dtype=float)
+    self.hessian_factor = wp.empty((0, 0, 0), dtype=float)
+
```
hartikainen, line 756: I think a more pythonic way for these dataclasses is to not define `__init__`-method but instead use default values or `__post_init__`.

```suggestion
  hessian: ArrayLike = field(
      default_factory=lambda: wp.empty((0, 0, 0), dtype=float)
  )
  hessian_factor: ArrayLike = field(
      default_factory=lambda: wp.empty((0, 0, 0), dtype=float)
  )

```

erikfrey:
Much like #925 this is pending some internal discussion about how to best manage inline allocation.

thowell:
closing in favor of arena memory #966

PR #934, opened 2025-12-18
sparse `efc_J`

thowell:
this pr introduces a sparse representation for `efc_J` as part of the sparse jacobian effort #88

- adds sparsity representation fields `J_rownnz`, `J_rowadr`, and `J_colind` to `Constraint`
- adds sparse code paths to kernels that construct constraints
- updates solver sparse code path to utilize sparse constraint jacobian
- this pr introduces the sparsity pattern, but the memory remains dense for now (ie, rownnz == nv always) -- a follow-up pr addresses #936

follow-up prs will
- integrate sparse jacobians `flexedge_J` and `ten_J` during constraint construction #935 
- add a parameter for reducing the memory required by the constraint jacobian #936

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -915,13 +915,12 @@ def put_data(
     )
     efc.J_colind = wp.array(np.tile(np.arange(mjm.nv), (nworld, njmax)).reshape((nworld, 1, -1)), dtype=int)
 
-    if mjd.nefc == 0:
-      mj_efc_J = np.zeros((0, mjm.nv))
-    elif mujoco.mj_isSparse(mjm):
-      mj_efc_J = np.zeros((mjd.nefc, mjm.nv))
-      mujoco.mju_sparse2dense(mj_efc_J, mjd.efc_J, mjd.efc_J_rownnz, mjd.efc_J_rowadr, mjd.efc_J_colind)
-    else:
-      mj_efc_J = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
+    mj_efc_J = np.zeros((mjd.nefc, mjm.nv))
+    if mjd.nefc:
+      if mujoco.mj_isSparse(mjm):
+        mujoco.mju_sparse2dense(mj_efc_J, mjd.efc_J, mjd.efc_J_rownnz, mjd.efc_J_rowadr, mjd.efc_J_colind)
+      else:
+        mj_efc_J = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
     efc_J = np.zeros((njmax, mjm.nv), dtype=float)
     efc_J[: mjd.nefc, : mjm.nv] = mj_efc_J
     efc.J = wp.array(np.tile(efc_J.reshape(-1), (nworld, 1, 1)).reshape((nworld, 1, -1)), dtype=float)
@@ -930,11 +929,12 @@ def put_data(
     efc.J_rowadr = wp.zeros((nworld, 0), dtype=int)
     efc.J_colind = wp.zeros((nworld, 0, 0), dtype=int)
 
-    if mujoco.mj_isSparse(mjm):
-      mj_efc_J = np.zeros((mjd.nefc, mjm.nv))
-      mujoco.mju_sparse2dense(mj_efc_J, mjd.efc_J, mjd.efc_J_rownnz, mjd.efc_J_rowadr, mjd.efc_J_colind)
-    else:
-      mj_efc_J = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
+    mj_efc_J = np.zeros((mjd.nefc, mjm.nv))
+    if mjd.nefc:
+      if mujoco.mj_isSparse(mjm):
+        mujoco.mju_sparse2dense(mj_efc_J, mjd.efc_J, mjd.efc_J_rownnz, mjd.efc_J_rowadr, mjd.efc_J_colind)
+      else:
+        mj_efc_J = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
     efc_J = np.zeros((nworld, sizes["njmax_pad"], sizes["nv_pad"]), dtype=float)
     efc_J[:, : mjd.nefc, : mjm.nv] = np.tile(mj_efc_J, (nworld, 1, 1))
     efc.J = wp.array(efc_J, dtype=float)


mujoco_warp/_src/types.py
```diff
@@ -1561,7 +1566,10 @@ class Constraint:
 
   type: array("nworld", "njmax", int)
   id: array("nworld", "njmax", int)
-  J: array("nworld", "njmax_pad", "nv_pad", float)
+  J_rownnz: wp.array2d(dtype=int)
+  J_rowadr: wp.array2d(dtype=int)
+  J_colind: wp.array3d(dtype=int)
+  J: wp.array3d(dtype=float)
```
btaba, line 1576: why doesn't this using the new annotation style `array("nworld", ...` ? is it because it needs to support both sparse and dense, or something like that?

mujoco_warp/_src/types.py
```diff
@@ -1665,6 +1673,7 @@ class Data:
     naconmax: maximum number of contacts (shared across all worlds)
     naccdmax: maximum number of contacts for CCD (all worlds)
     njmax: maximum number of constraints per world
+    njmax_pad: maximum number of constraints per world +
```
btaba: nit: + at the end? it's not clear how this differs compared to njmax

edit: I dug a bit and it seems like njmax_pad is a derived field right? should expound a bit

mujoco_warp/_src/constraint_test.py
```diff
@@ -98,29 +119,15 @@ def _assert_efc_eq(d, mjd, nefc, name, nv):
...
-    (ConeType.ELLIPTIC, 1, 6),
-    (ConeType.ELLIPTIC, 3, 3),
-    (ConeType.ELLIPTIC, 3, 4),
-    (ConeType.ELLIPTIC, 3, 6),
-    (ConeType.ELLIPTIC, 4, 4),
-    (ConeType.ELLIPTIC, 4, 6),
-    (ConeType.ELLIPTIC, 6, 6),
+    *itertools.product(
+      (ConeType.PYRAMIDAL, ConeType.ELLIPTIC),
+      ((1, 1), (1, 3), (1, 4), (1, 6), (3, 3), (3, 4), (3, 6), (4, 4), (4, 6), (6, 6)),
```
btaba: this is a combination? if you're using itertools already

mujoco_warp/_src/io.py
```diff
@@ -715,6 +715,17 @@ def make_data(
   contact = types.Contact(**{f.name: _create_array(None, f.type, sizes) for f in dataclasses.fields(types.Contact)})
   efc = types.Constraint(**{f.name: _create_array(None, f.type, sizes) for f in dataclasses.fields(types.Constraint)})
 
+  if is_sparse(mjm):
+    efc.J_rownnz = wp.zeros((nworld, njmax), dtype=int)
+    efc.J_rowadr = wp.zeros((nworld, njmax), dtype=int)
+    efc.J_colind = wp.zeros((nworld, 1, njmax * mjm.nv), dtype=int)
+    efc.J = wp.zeros((nworld, 1, njmax * mjm.nv), dtype=float)
+  else:
+    efc.J_rownnz = wp.zeros((nworld, njmax), dtype=int)
```
btaba: why do you need to allocate memory here if using dense jacobian?

mujoco_warp/_src/io.py
```diff
@@ -901,14 +913,41 @@ def put_data(
...
+      mj_efc_J = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
+    efc_J = np.zeros((njmax, mjm.nv), dtype=float)
+    efc_J[: mjd.nefc, : mjm.nv] = mj_efc_J
+    efc.J = wp.array(np.tile(efc_J.reshape(-1), (nworld, 1, 1)).reshape((nworld, 1, -1)), dtype=float)
   else:
-    efc_j = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
-  efc.J = np.zeros((nworld, sizes["njmax_pad"], sizes["nv_pad"]), dtype=f.type.dtype)
-  efc.J[:, : mjd.nefc, : mjm.nv] = np.tile(efc_j, (nworld, 1, 1))
-  efc.J = wp.array(efc.J, dtype=float)
+    # TODO(team): don't need to allocate sparsity arrays if not sparse
```
btaba: why not fix this in this PR?

mujoco_warp/_src/io.py
```diff
@@ -901,14 +913,41 @@ def put_data(
 
   efc = types.Constraint(**efc_kwargs)
 
-  if mujoco.mj_isSparse(mjm):
-    efc_j = np.zeros((mjd.nefc, mjm.nv))
-    mujoco.mju_sparse2dense(efc_j, mjd.efc_J, mjd.efc_J_rownnz, mjd.efc_J_rowadr, mjd.efc_J_colind)
+  if is_sparse(mjm):
+    # TODO(team): mujoco sparsity pattern instead of dense
```
btaba: this is a bit confusing, the kernels use the sparsity pattern but the allocation is dense, could the comment be validated a bit more? (I know I'm also *not* looking at the downstream PRs so I'm missing context in this review)

mujoco_warp/_src/io.py
```diff
@@ -1125,9 +1165,21 @@ def get_data_into(
     mujoco.mj_factorM(mjm, result)
 
   if nefc > 0:
-    if mujoco.mj_isSparse(mjm):
-      efc_J = d.efc.J.numpy()[world_id, efc_idx, : mjm.nv]
-      mujoco.mju_dense2sparse(result.efc_J, efc_J, result.efc_J_rownnz, result.efc_J_rowadr, result.efc_J_colind)
+    if is_sparse(mjm):
+      efc_J = np.zeros((nefc, mjm.nv))
+      if is_sparse(mjm):
```
btaba: two nested if statements with the same condition

isn't the right condition to check `mujoco.mj_isSparse(mjm)`? then you can branch and check `is_sparse(mjm)` to see how to extract out the efc_J from warp?

mujoco_warp/_src/smooth.py
```diff
@@ -2354,13 +2262,148 @@ def transmission(m: Model, d: Data):
   )
 
   if m.nacttrnbody:
+
+    @wp.kernel(module="unique", enable_backward=False)
+    def transmission_body_moment(
```
btaba: Why not have one kernel that works for sparse, and one for dense, rather than combining them here? The current diff is hard to read and error prone as it mixes sparse/dense logic into a single kernel. I don't feel strongly either way...it's just the diff!!

mujoco_warp/_src/constraint.py
```diff
@@ -44,959 +44,410 @@ def _zero_constraint_counts(
...
-  solimp = eq_solimp[worldid % eq_solimp.shape[0], eqid]
-  timestep = opt_timestep[worldid % opt_timestep.shape[0]]
+@event_scope
+def make_constraint(m: types.Model, d: types.Data):
+  """Creates constraint jacobians and other supporting data."""
 
-  for i in range(3):
-    efcidi = efcid + i
+  @wp.func
+  def efc_row(
```
btaba: OOC does this kernel really need to be nested inside `make_constraint`, what's the rationale, rather than creating a closure outside of `make_constraint`?

generally the git diff makes it hard to review for subtle bugs that may be introduced in all the code movement, what kind of validation strategy do you have for PRs like this beyond unit tests?

mujoco_warp/_src/constraint.py
```diff
@@ -44,959 +44,410 @@ def _zero_constraint_counts(
...
-    J = float(dist_min < dist_max) * 2.0 - 1.0
-    efc_J_out[worldid, efcid, dofadr] = J
-    Jqvel = J * qvel_in[worldid, dofadr]
+    if wp.static(m.is_sparse):
+      if jntid_2 > -1:
+        rownnz = 2
+      else:
+        rownnz = 1
+      efc_J_rownnz_out[worldid, efcid] = rownnz
+      rowadr = efcid * wp.static(m.nv)
```
btaba: OOC does this solution for "striding" sparsity change in downstream PRs?

mujoco_warp/_src/solver.py
```diff
@@ -1410,16 +1413,36 @@ def kernel(
     jv_out = float(0.0)
 
     if wp.static(dofs_per_thread >= nv):
-      for i in range(wp.static(min(dofs_per_thread, nv))):
-        jv_out += efc_J_in[worldid, efcid, i] * ctx_search_in[worldid, i]
+      if wp.static(opt_is_sparse):
+        # Sparse: iterate over non-zero entries in the row
+        rownnz = efc_J_rownnz_in[worldid, efcid]
+        rowadr = efc_J_rowadr_in[worldid, efcid]
+        for k in range(rownnz):
```
btaba, line 1428: very nice

mujoco_warp/_src/solver.py
```diff
@@ -1689,22 +1716,30 @@ def kernel(
...
+        colind = efc_J_colind_in[worldid, 0, sparseid]
+        jaref += efc_J_in[worldid, 0, sparseid] * qacc_in[worldid, colind]
       ctx_Jaref_out[worldid, efcid] = jaref - efc_aref_in[worldid, efcid]
-
     else:
-      for i in range(wp.static(dofs_per_thread)):
-        ii = dofstart * wp.static(dofs_per_thread) + i
-        if ii < nv:
-          jaref += efc_J_in[worldid, efcid, ii] * qacc_in[worldid, ii]
+      if wp.static(dofs_per_thread >= nv):
```
btaba, line 1736: is this some kind of hidden optimization as part of this PR, to avoid the atomic add?

mujoco_warp/_src/solver.py
```diff
@@ -2202,7 +2279,175 @@ def kernel(
...
+
+    h = float(0.0)
+
+    for dim1id in range(condim):
+      if dim1id == 0:
+        efcid1 = efcid0
+      else:
+        efcid1 = contact_efc_address_in[conid, dim1id]
+
+      # TODO(team): more efficient sparsity
```
btaba: what does this mean? I guess it's immediately addressed in downstream PRs?

mujoco_warp/_src/solver_test.py
btaba, line 34: which of these tests is now hitting the sparse codepath? it's not really clear

mujoco_warp/_src/types.py
thowell: these arrays have sizes determined by dense/sparse in io

mujoco_warp/_src/constraint.py
thowell: yes, please see #936

mujoco_warp/_src/types.py
thowell: updated the docstring

mujoco_warp/_src/constraint_test.py
thowell: updated

mujoco_warp/_src/io.py
thowell: updated shapes to use zeros for sparsity structure fields when dense

mujoco_warp/_src/io.py
thowell: done in this pr now

mujoco_warp/_src/io.py
thowell: updated comment -- essentially the code currently assumes each constraint row has `nv` elements (ie rownnz[i] is always nv), would need to process the mujoco sparsity structure that can have rownnz <= nv to create a valid sparsity structure for mujoco warp. future prs will change this assumption on rownnz

mujoco_warp/_src/io.py
thowell: mujoco warp and mujoco have different dof thresholds for dense or sparse when jacobian is auto so we need this nested check.

mujoco_warp/_src/smooth.py
thowell: this is updated to 2 kernels -- eventually it probably makes sense to combine kernels to avoid code duplication

mujoco_warp/_src/constraint.py
thowell: efc_row and the constraint kernels are moved to the module level now instead of being in the make_constraint scope

mujoco_warp/_src/solver.py
thowell: this dofs_per_threads check is from existing code

mujoco_warp/_src/solver.py
thowell: updated the comment -- we should optimize performance after some follow up prs are merged

mujoco_warp/_src/solver_test.py
thowell: solver_test.py has been updated with additional testing for sparsity

mujoco_warp/_src/constraint.py
```diff
@@ -1007,94 +463,119 @@ def _efc_limit_slide_hinge(
       efc_frictionloss_out,
     )
 
+  return equality_joint
+
+
+def _equality_tendon(is_sparse, nv, njmax, refsafe):
```
erikfrey: How did you decide that these variables should be static?

These paramaters grow the number of kernel compilations.

mujoco_warp/_src/solver_test.py
btaba: awesome

mujoco_warp/_src/io.py
```diff
@@ -901,14 +913,35 @@ def put_data(
...
+      mj_efc_J = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
+    efc_J = np.zeros((njmax, mjm.nv), dtype=float)
+    efc_J[: mjd.nefc, : mjm.nv] = mj_efc_J
+    efc.J = wp.array(np.tile(efc_J.reshape(-1), (nworld, 1, 1)).reshape((nworld, 1, -1)), dtype=float)
   else:
-    efc_j = mjd.efc_J.reshape((mjd.nefc, mjm.nv))
-  efc.J = np.zeros((nworld, sizes["njmax_pad"], sizes["nv_pad"]), dtype=f.type.dtype)
-  efc.J[:, : mjd.nefc, : mjm.nv] = np.tile(efc_j, (nworld, 1, 1))
-  efc.J = wp.array(efc.J, dtype=float)
+    efc.J_rownnz = wp.zeros((nworld, 0), dtype=int)
```
btaba, line 928: nice

mujoco_warp/_src/smooth.py
```diff
@@ -2198,96 +2198,223 @@ def _transmission(
...
+      # project Jacobian along the normal of the contact frame
+      wp.atomic_add(actuator_moment_out[worldid, actid], dofid, wp.dot(normal, jacdif))
 
-    else:
-      npyramid = contact_dim - 1  # number of frictional directions
-      efc_force = 0.5 / float(npyramid)
+  return transmission_body_moment_dense
+
+
+def _transmission_body_moment_sparse(is_elliptic):
```
btaba: ok I admit that my suggestion to split these impls may have been a bad one. I'm not sure this split is better :| sorry I was just trying to see if the diff would get better (which it did, but the actual code may be worse). Ultimately it's your call / style-choice

mujoco_warp/_src/smooth.py
btaba: yeah I think my comment here was wrong...

mujoco_warp/_src/constraint.py
thowell: previously all of the constraint kernels were in the `make_constraint` scope and these variables were passed to `wp.static`. this has been changed, so the constraint kernel builders are updated to just have `is_sparse` arguments.

mujoco_warp/_src/smooth.py
thowell: merged the dense and sparse kernels into 1 kernel

mujoco_warp/_src/constraint.py
erikfrey: Nice thanks - but going further:

It seems like you've already done a pretty good job of isolating the `is_sparse` branch so there's only a few ops on either side of it in the kernels.  Are you sure `is_sparse` shouldn't just be a normal kernel param too?

Feel free to back up your investigations here with some benchmarks.  Just trying to make sure we do the absolute simplest thing possible.

thowell:
```
mjwarp-testspeed benchmarks/humanoid/humanoid.xml --nconmax=24 --njmax=64 -o "opt.jacobian="sparse""
```

the current implementation with `wp.static(is_sparse)`

```

Model
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: True
  ls_parallel: False
Data

Total JIT time: 10.31 s
Total simulation time: 7.10 s
Total steps per second: 1,153,004
Total realtime factor: 5,765.02 x
Total time per step: 867.30 ns
```

removing `wp.static(is_sparse)` and using `is_sparse` as a kernel argument

```

Model
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: True
  ls_parallel: False
Data

Total JIT time: 14.76 s
Total simulation time: 7.12 s
Total steps per second: 1,150,365
Total realtime factor: 5,751.83 x
Total time per step: 869.29 ns
```

running this for both a few times, it seems like any potential difference in sps is really in the noise

mujoco_warp/_src/constraint.py
```diff
@@ -80,8 +82,7 @@ def _update_efc_row(
   mid = solimp[3]
   power = solimp[4]
 
-  # TODO(team): wp.static?
-  if not refsafe:
+  if not (opt_disableflags & DisableBit.REFSAFE):
```
erikfrey, line 85: yes!!!!

mujoco_warp/_src/smooth.py
```diff
@@ -2198,96 +2198,146 @@ def _transmission(
...
-):
-  trnbodyid, conid, dofid = wp.tid()
-  actid = actuator_trntype_body_adr[trnbodyid]
-  bodyid = actuator_trnid[actid][0]
-
-  if conid >= nacon_in[0]:
-    return
-
-  worldid = contact_worldid_in[conid]
+def _transmission_body_moment(is_sparse, is_elliptic):
```
erikfrey: shall we simplify this kernel too?

mujoco_warp/_src/types.py
```diff
@@ -1665,6 +1675,7 @@ class Data:
     naconmax: maximum number of contacts (shared across all worlds)
     naccdmax: maximum number of contacts for CCD (all worlds)
     njmax: maximum number of constraints per world
+    njmax_pad: njmax rounded up to the nearest multiple of TILE_SIZE_JTDAJ (16)
```
erikfrey: nit: the 16 is confusing here as it looks like how we annotate size.  i'd probably just remove it.

mujoco_warp/_src/smooth.py
thowell: done -- `is_sparse` is now a kernel argument and `is_elliptic` has been replaced with the kernel argument `opt_cone` that is then compared to `ConeType.ELLIPTIC` in the kernel

mujoco_warp/_src/types.py
thowell: removed

adenzler-nvidia:
Are you guys aware that this PR introduces a massive performance regression? SPS on G1 with hands went from 670k FPS to 180K.

Will have a closer look at what exactly the problem is, and report back.

thowell:
@adenzler-nvidia the performance regression for jacobian=sparse is expected at this point since there may be kernels that are not optimized for sparsity

here is what i see with the three humanoids:

```
mjwarp-testspeed benchmarks/humanoid/three_humanoids.xml --nconmax=100 --njmax=192 --event_trace
```

current
```
Loading model from: benchmarks/humanoid/three_humanoids.xml...

Model
  nq: 84 nv: 81 nu: 63 nbody: 49 ngeom: 58
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: True
  ls_parallel: False
Data
  nworld: 8192 naconmax: 819200 njmax: 192

Total JIT time: 53.41 s
Total simulation time: 121.71 s
Total steps per second: 67,309
Total realtime factor: 336.54 x
Total time per step: 14856.96 ns
Total converged worlds: 0 / 8192

Event trace:

step: 14853.34
  forward: 14752.03
    fwd_position: 127.91
      kinematics: 24.57
      com_pos: 14.79
      camlight: 3.52
      flex: 0.18
      crb: 24.62
      tendon_armature: 0.17
      collision: 14.86
        nxn_broadphase: 11.89
        convex_narrowphase: 0.18
        primitive_narrowphase: 1.86
      make_constraint: 10.43
      transmission: 32.94
    sensor_pos: 0.18
    fwd_velocity: 71.22
      com_vel: 15.92
      passive: 2.21
      rne: 16.73
      tendon_bias: 0.18
    sensor_vel: 0.18
    fwd_actuation: 32.65
    fwd_acceleration: 97.94
      xfrc_accumulate: 6.01
    solve: 14419.86
      mul_m: 7.59
    sensor_acc: 0.18
  euler: 100.74
  ```
  
  previous commit e28c6038cdf8a353b4146974e4cf37e74dda809a
  
  ```
  Loading model from: benchmarks/humanoid/three_humanoids.xml...

Model
  nq: 84 nv: 81 nu: 63 nbody: 49 ngeom: 58
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: True
  ls_parallel: False
Data
  nworld: 8192 naconmax: 819200 njmax: 192

Total JIT time: 23.83 s
Total simulation time: 16.75 s
Total steps per second: 489,182
Total realtime factor: 2,445.91 x
Total time per step: 2044.23 ns

Event trace:

step: 2042.25
  forward: 1944.34
    fwd_position: 377.45
      kinematics: 23.19
      com_pos: 14.54
      camlight: 3.49
      flex: 0.18
      crb: 23.96
      tendon_armature: 0.18
      collision: 30.20
        nxn_broadphase: 18.29
        convex_narrowphase: 0.18
        primitive_narrowphase: 10.81
      make_constraint: 244.34
      transmission: 35.59
    sensor_pos: 0.17
    fwd_velocity: 69.48
      com_vel: 15.56
      passive: 2.14
      rne: 16.38
      tendon_bias: 0.18
    sensor_vel: 0.18
    fwd_actuation: 32.60
    fwd_acceleration: 94.98
      xfrc_accumulate: 5.84
    solve: 1367.46
      mul_m: 6.42
    sensor_acc: 0.18
  euler: 97.37
  ```

PR #935, opened 2025-12-18
sparse contact and flex constraints

thowell:
this pr depends on #934 and implements sparse contact and flex constraints


thowell:
updated `test_constraints` to test `flex/floppy.xml` for the flexedge equality constraints. contact constraints should be tested by the existing test xml `constraints.xml`

PR #939, opened 2025-12-20
update testspeed measure_alloc print: ncon -> nacon

thowell:
testspeed with `--measure_alloc=True` prints `ncon` when it should print `nacon`

### DIFF
--- a/mujoco_warp/testspeed.py
+++ b/mujoco_warp/testspeed.py
@@ -201,16 +201,16 @@ Total converged worlds: {nsuccess} / {d.nworld}""")
 
     if nacon and nefc:
       idx = 0
-      ncon_matrix, nefc_matrix = [], []
+      nacon_matrix, nefc_matrix = [], []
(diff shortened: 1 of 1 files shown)


mujoco_warp/testspeed.py
```diff
@@ -210,7 +210,7 @@ def _main(argv: Sequence[str]):
         nefc_matrix.append([np.mean(nefc_arr), np.std(nefc_arr), np.min(nefc_arr), np.max(nefc_arr)])
         idx += size
 
-      _print_table(ncon_matrix, ("mean", "std", "min", "max"), "ncon alloc")
+      _print_table(ncon_matrix, ("mean", "std", "min", "max"), "nacon alloc")
```
erikfrey: ah good catch, then let's also rename this variable `nacon_matrix`

thowell: done. also renamed `ncon_arr` -> `nacon_arr`.

PR #940, opened 2025-12-21
Add ray with normal

StafaH:
Modify ray functions to return normal information as well.

Follows the C MuJoCo implementation: https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_ray.c

Currently waiting on the C MuJoCo mju_rayNormal to become enabled, which will allow ray_test to compare normals with the engine implementation.

### DIFF
--- a/mujoco_warp/_src/ray.py
+++ b/mujoco_warp/_src/ray.py
@@ -105,7 +105,7 @@ def _ray_quad(a: float, b: float, c: float) -> Tuple[float, wp.vec2]:
   """Compute solutions from quadratic: a*x^2 + 2*b*x + c = 0."""
   det = b * b - a * c
   if det < MJ_MINVAL:
-    return wp.inf, wp.vec2(wp.inf, wp.inf)
+    return -1.0, wp.vec2(-1.0, -1.0)
   det = wp.sqrt(det)
 
   # compute the two solutions
@@ -120,14 +120,14 @@ def _ray_quad(a: float, b: float, c: float) -> Tuple[float, wp.vec2]:
   elif x1 >= 0.0:
     return x1, x
   else:
-    return wp.inf, x
+    return -1.0, x
 
 
 @wp.func
 def _ray_triangle(
   v0: wp.vec3, v1: wp.vec3, v2: wp.vec3, pnt: wp.vec3, vec: wp.vec3, b0: wp.vec3, b1: wp.vec3
 ) -> Tuple[float, wp.vec3]:
-  """Returns the distance at which a ray intersects with a triangle."""
+  """Returns the distance and normal at which a ray intersects with a triangle."""
   dif0 = v0 - pnt
   dif1 = v1 - pnt
   dif2 = v2 - pnt
@@ -147,7 +147,7 @@ def _ray_triangle(
     or (planar_01 > 0.0 and planar_11 > 0.0 and planar_21 > 0.0)
     or (planar_01 < 0.0 and planar_11 < 0.0 and planar_21 < 0.0)
   ):
-    return float(wp.inf), wp.vec3(0.0, 0.0, 0.0)
+    return -1.0, wp.vec3()
 
   # determine if origin is inside planar projection of triangle
   # A = (p0-p2, p1-p2), b = -p2, solve A*t = b
@@ -160,14 +160,14 @@ def _ray_triangle(
 
   det = A00 * A11 - A10 * A01
   if wp.abs(det) < MJ_MINVAL:
-    return float(wp.inf), wp.vec3(0.0, 0.0, 0.0)
+    return -1.0, wp.vec3()
 
   t0 = (A11 * b[0] - A10 * b[1]) / det
   t1 = (-A01 * b[0] + A00 * b[1]) / det
 
   # check if outside
   if t0 < 0.0 or t1 < 0.0 or t0 + t1 > 1.0:
-    return float(wp.inf), wp.vec3(0.0, 0.0, 0.0)
+    return -1.0, wp.vec3()
 
   # intersect ray with plane of triangle
   dif0 = v0 - v2
@@ -176,26 +176,26 @@ def _ray_triangle(
   nrm = wp.cross(dif0, dif1)  # normal to triangle plane
   denom = wp.dot(vec, nrm)
   if wp.abs(denom) < MJ_MINVAL:
-    return float(wp.inf), wp.vec3(0.0, 0.0, 0.0)
+    return -1.0, wp.vec3()
 
   dist = -wp.dot(dif2, nrm) / denom
-  return wp.where(dist >= 0.0, dist, float(wp.inf)), wp.normalize(nrm)
+  return wp.where(dist >= 0.0, dist, -1.0), wp.normalize(nrm)
 
 
 @wp.func
 def _ray_plane(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec: wp.vec3) -> Tuple[float, wp.vec3]:
-  """Returns the distance at which a ray intersects with a plane."""
+  """Returns the distance and normal at which a ray intersects with a plane."""
   # map to local frame
   lpnt, lvec = _ray_map(pos, mat, pnt, vec)
 
   # z-vec not pointing towards front face: reject
   if lvec[2] > -MJ_MINVAL:
-    return wp.inf, wp.vec3(0.0, 0.0, 0.0)
+    return -1.0, wp.vec3()
 
   # intersection with plane
   x = -lpnt[2] / lvec[2]
   if x < 0.0:
-    return wp.inf, wp.vec3(0.0, 0.0, 0.0)
+    return -1.0, wp.vec3()
 
   p = wp.vec2(lpnt[0] + x * lvec[0], lpnt[1] + x * lvec[1])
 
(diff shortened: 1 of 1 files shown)


thowell:
thanks @StafaH!

mujoco_warp/_src/ray.py
```diff
@@ -528,7 +593,19 @@ def _ray_hfield(
       if z < z0 * (y0 + 1.0 - y) + z1 * (y - y0):
         x = all[i]
 
-  return x
+        normal_local = wp.vec3(0.0, 0.0, 0.0)
```
erikfrey: how about:

```py
normal_local = wp.vec3(float(i == 1) - float(i == 0), float(i == 3) - float(i == 2), 0.0)
```

mujoco_warp/_src/ray.py
```diff
@@ -145,7 +147,7 @@ def _ray_triangle(v0: wp.vec3, v1: wp.vec3, v2: wp.vec3, pnt: wp.vec3, vec: wp.v
     or (planar_01 > 0.0 and planar_11 > 0.0 and planar_21 > 0.0)
     or (planar_01 < 0.0 and planar_11 < 0.0 and planar_21 < 0.0)
   ):
-    return float(wp.inf)
+    return float(wp.inf), wp.vec3(0.0, 0.0, 0.0)
```
erikfrey: nit: `wp.vec3()` already inits to zeros

(comment applies here and elsewhere)

mujoco_warp/_src/ray.py
```diff
@@ -124,7 +124,9 @@ def _ray_quad(a: float, b: float, c: float) -> Tuple[float, wp.vec2]:
 
 
 @wp.func
-def _ray_triangle(v0: wp.vec3, v1: wp.vec3, v2: wp.vec3, pnt: wp.vec3, vec: wp.vec3, b0: wp.vec3, b1: wp.vec3) -> float:
+def _ray_triangle(
+  v0: wp.vec3, v1: wp.vec3, v2: wp.vec3, pnt: wp.vec3, vec: wp.vec3, b0: wp.vec3, b1: wp.vec3
+) -> Tuple[float, wp.vec3]:
   """Returns the distance at which a ray intersects with a triangle."""
```
thowell: let's update the docstring to mention that the normal (?) is also returned

mujoco_warp/_src/ray.py
```diff
@@ -174,38 +176,38 @@ def _ray_triangle(v0: wp.vec3, v1: wp.vec3, v2: wp.vec3, pnt: wp.vec3, vec: wp.v
...
 
   dist = -wp.dot(dif2, nrm) / denom
-  return wp.where(dist >= 0.0, dist, float(wp.inf))
+  return wp.where(dist >= 0.0, dist, float(wp.inf)), wp.normalize(nrm)
 
 
 @wp.func
-def _ray_plane(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec: wp.vec3) -> float:
+def _ray_plane(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec: wp.vec3) -> Tuple[float, wp.vec3]:
   """Returns the distance at which a ray intersects with a plane."""
```
thowell: let's update the docstring to mention that the normal is returned

mujoco_warp/_src/ray.py
```diff
@@ -174,38 +176,38 @@ def _ray_triangle(v0: wp.vec3, v1: wp.vec3, v2: wp.vec3, pnt: wp.vec3, vec: wp.v
...
+    return x, wp.vec3(mat[0, 2], mat[1, 2], mat[2, 2])
   else:
-    return wp.inf
+    return wp.inf, wp.vec3(0.0, 0.0, 0.0)
 
 
 @wp.func
-def _ray_sphere(pos: wp.vec3, dist_sqr: float, pnt: wp.vec3, vec: wp.vec3) -> float:
+def _ray_sphere(pos: wp.vec3, dist_sqr: float, pnt: wp.vec3, vec: wp.vec3) -> Tuple[float, wp.vec3]:
   """Returns the distance at which a ray intersects with a sphere."""
```
thowell: update docstring

mujoco_warp/_src/ray.py
StafaH: That is a very nifty one-liner! Beautiful. Replaced.

mujoco_warp/_src/ray.py
StafaH: Done

mujoco_warp/_src/ray.py
StafaH: Fixed

mujoco_warp/_src/ray.py
StafaH: Fixed

mujoco_warp/_src/ray.py
StafaH: Fixed

StafaH:
@thowell @erikfrey Fixed up the comments. Also moved to returning -1 for non-hits to match MuJoCo implementation. Just waiting now on C MuJoCo to enable the ray + normal and we can add tests to make sure the normals match the engine impl.

mujoco_warp/_src/ray_test.py
```diff
@@ -42,12 +42,14 @@ def test_ray_nothing(self):
...
     vec = wp.array([wp.vec3(0.0, 0.0, -1.0)], dtype=wp.vec3).reshape((1, 1))
-    dist, geomid = mjw.ray(m, d, pnt, vec)
+    dist, geomid, normal = mjw.ray(m, d, pnt, vec)
     wp.synchronize()
     geomid_np = geomid.numpy()[0, 0]  # Extract from [[-1]]
     dist_np = dist.numpy()[0, 0]  # Extract from [[-1.]]
+    normal_np = normal.numpy()[0, 0]
     _assert_eq(geomid_np, -1, "geom_id")
     _assert_eq(dist_np, -1, "dist")
+    _assert_eq(normal_np, wp.vec3(0.0, 0.0, 0.0), "normal")
```
thowell, line 52: wp.vec3 -> 0.0

mujoco_warp/_src/ray.py
```diff
@@ -780,24 +857,30 @@ def _ray(
...
 
     tile_geomid = wp.tile(geomid)
+    tile_normal = wp.tile(normal, preserve_type=True)
 
     if local_min_dist < min_dist:
       min_dist = local_min_dist
       min_geomid = tile_geomid[local_min_geomid[0]]
+      min_normal = tile_normal[local_min_geomid[0]]
 
   if wp.isinf(min_dist):
```
thowell, line 878: is it possible to refactor in order to use -1 for no intersection everywhere for rays?

mujoco_warp/_src/ray.py
```diff
@@ -780,24 +857,30 @@ def _ray(
...
     else:
       dist = wp.inf
+      normal = wp.vec3()
 
     tile_dist = wp.tile(dist)
     local_min_geomid = wp.tile_argmin(tile_dist)
     local_min_dist = tile_dist[local_min_geomid[0]]
 
     tile_geomid = wp.tile(geomid)
+    tile_normal = wp.tile(normal, preserve_type=True)
```
thowell, line 871: is `preserve_type` necessary?

mujoco_warp/_src/ray.py
```diff
@@ -739,18 +814,20 @@ def _ray(
   # Out:
   dist_out: wp.array(dtype=float, ndim=2),
   geomid_out: wp.array(dtype=int, ndim=2),
+  normal_out: wp.array(dtype=wp.vec3, ndim=2),
```
thowell, line 817: we should follow the convention used throughout the code: `wp.array(..., ndim=2)` -> `wp.array2d(...)`. let's update this line and the 2 above?

mujoco_warp/_src/ray.py
```diff
@@ -335,28 +372,43 @@ def _ray_cylinder(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec:
...
 
 
 _IFACE = wp.types.matrix((3, 2), dtype=int)(1, 2, 0, 2, 0, 1)
 
 
 @wp.func
-def _ray_box(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec: wp.vec3) -> Tuple[float, vec6]:
-  """Returns the distance at which a ray intersects with a box."""
+def _ray_box(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, pnt: wp.vec3, vec: wp.vec3) -> Tuple[float, vec6, wp.vec3]:
+  """Returns the distance and normal at which a ray intersects with a box."""
```
thowell, line 396: we should mention in the docstring that `all` is returned. maybe refer to it as 'intersection information'? reference: https://github.com/google-deepmind/mujoco/blob/b9fa8ee6e945f27dab076d20529f15d269477c06/src/engine/engine_ray.c#L493

thowell (approved):
added a few more comments. otherwise lgtm. 

great work and thank you for this contribution @StafaH!

mujoco_warp/_src/ray.py
StafaH: Yes, without it  `wp.tile` will create a tile of shape (3, N) rather than (N,): https://nvidia.github.io/warp/language_reference/_generated/warp._src.lang.tile.html#warp._src.lang.tile

mujoco_warp/_src/ray.py
StafaH: I considered it here but since we want to use tile_argmin, its necessary here to use wp.inf so that the argmin works, and then convert back to -1. The alternative here would be to use tile_map to create a temperary array that converts -1 to inf, find the argmin, and then use the index.

PR #947, opened 2025-12-27
feat: Add `impratio` to testspeed results

### LINKED ISSUE #944: add more info to testspeed results
add more information to the results printed by testspeed:

- [x] `Option` fields
  - [x] all fields #1055 
  - [x] `impratio` @zer-art #947 
- [x] collision routines required: primitive, ccd, sdf #1055 
- [x] number of contacts per collider type #1055

zer-art:
Fixes #944

### Summary
This PR updates `mujoco_warp/testspeed.py` to include additional simulation details in the final output report.

**Changes:**
- Added logging for `impratio` (checked via `mjm.opt.impratio`).
- Added logic to determine and report required collision routines:
  - **Primitive:** Checked against standard primitive collision pairs.
  - **CCD:** Checked against convex collision pairs.
  - **SDF:** Checked via `m.has_sdf_geom`.

### Verification
I tested the changes locally using the `aloha_pot` scene.

**Command:**
```bash
python mujoco_warp/testspeed.py mujoco_warp/test_data/aloha_pot/scene.xml --nworld 10 --nstep 10

```

**Output:**
```
Data
  nworld: 10 naconmax: 480 njmax: 64

  impratio: 10.0
  collision routines required: primitive, ccd
  ```
  ### Checklist
- [x] I have signed the CLA.
- [x] My code follows the style guidelines of this project.

### CODE REVIEW

**Comment by thowell** [comment]
@zer-art thank you for contributing to mujoco warp!

lets focus on just adding `impratio` information in this pr and we should print the value from the mujoco warp model option `m.opt.impratio` instead of the mujoco mjmodel option.

additionally lets add a flag `--extra_info` that is false by default and needs to be true in order to print `impratio`.

**Comment by zer-art** [comment]
## Summary of Changes

### 1. Feature: impratio Reporting `testspeed.py`

Added the --extra_info flag (defaults to False).

Removed the collision routine calculation section to keep the PR atomic and focused.

The script now prints m.opt.impratio only when the --extra_info flag is enabled.

### 2. Core Update `_src/types.py & _src/io.py`

Context: I found that the Warp model options (m.opt) previously only stored impratio_invsqrt, making m.opt.impratio inaccessible by default.

Change: I added impratio to the Option struct and populated it in put_model to ensure the benchmark script can access the correct value.

### 3. Fix: CPU/Mac Compatibility `_src/benchmark.py`

Context: While verifying these changes locally on macOS (CPU), I encountered a crash because wp.ScopedCapture and wp.ScopedStream were being used unconditionally.

Change: I added a check for wp.get_device().is_cuda to ensure these CUDA-specific context managers are only active when a CUDA device is present.

Verified locally with python -m mujoco_warp.testspeed ... --extra_info. Ready for review!

**Comment by thowell** [comment]
@zer-art thanks for the update! 

- lets remove changes to all files except testspeed.py
- lets compute `impratio` (`1.0 / (impratio_invsqrt * impratio_invsqrt)`) in the f-string

**Comment by zer-art** [comment]
**@thowell** Done! I've reverted all changes to **_src/types.py, _src/io.py**, and **_src/benchmark.py.**

I also updated **testspeed.py** to calculate the ratio in-place using the formula: **1.0 / (impratio_invsqrt * impratio_invsqrt).**

Heads up on a Mac/CPU Bug: While verifying the revert, I confirmed that the current main branch crashes on macOS because benchmark.py calls CUDA functions unconditionally.

The Error:
```
File "warp/_src/context.py", line 7304, in capture_begin
    if runtime.driver_version >= (12, 3):
TypeError: '>=' not supported between instances of 'NoneType' and 'tuple'
```

_(This happens because runtime.driver_version is None on CPU devices)._

I plan to open a separate Issue to track this bug so we can keep this PR focused strictly on the feature. Let me know if you'd prefer I handle it differently.

File: mujoco_warp/testspeed.py [hunk 1]
```diff
@@ -179,6 +180,12 @@ def _main(argv: Sequence[str]):
     d = mjw.put_data(mjm, mjd, nworld=_NWORLD.value, nconmax=_NCONMAX.value, njmax=_NJMAX.value)
     print(f"Data\n  nworld: {d.nworld} naconmax: {d.naconmax} njmax: {d.njmax}\n")
 
+    # print impratio
```
thowell: we can remove this comment

mujoco_warp/testspeed.py
```diff
@@ -179,6 +180,12 @@ def _main(argv: Sequence[str]):
...
+
```
thowell: lets add a todo here: `# TODO(team): https://github.com/google-deepmind/mujoco_warp/issues/944`

thowell:
@zer-art thank you for the update!

if you are interested in contributing an update to testspeed related to not working with cpu:
- create a separate pr
- add a check to testspeed.py for cpu, something like

```
if "cpu" in (_DEVICE.value, wp.get_device()):
  raise ValueError("testspeed available for gpu only")
```

thanks!

zer-art:
@thowell Updated! Removed the print comment and added the TODO reference as requested.

thowell:
thinking that we can simplify and report all of the options together, so lets:
- completely remove the `--extra_info` flag (apologies for the detour with this)
- move `impratio` to be reported on a new line below this [line](https://github.com/google-deepmind/mujoco_warp/blob/aa6c0fb86a33ff56e32cd193205f10478ba62174/mujoco_warp/testspeed.py#L163). we can compute `impratio` in the f-string using something like: `1.0 / np.square(m.opt.impratio_invsqrt)`

thanks!

zer-art:
@thowell Done! I've removed the extra_info flag and updated the print statement to calculate impratio using the formula you provided.

mujoco_warp/testspeed.py
```diff
@@ -60,6 +60,7 @@
 _MEMORY = flags.DEFINE_bool("memory", False, "print memory report")
 
 
+
```
thowell: we can remove this new line

mujoco_warp/testspeed.py
```diff
@@ -174,11 +175,13 @@ def _main(argv: Sequence[str]):
...
       f"  broadphase: {broadphase} broadphase_filter: {filter}\n"
       f"  solver: {solver} cone: {cone} iterations: {iterations} {ls_str}\n"
-      f"  integrator: {integrator} graph_conditional: {m.opt.graph_conditional}"
+      f"  integrator: {integrator} graph_conditional: {m.opt.graph_conditional}\n"
+      f"  impratio: {1.0 / np.square(m.opt.impratio_invsqrt.numpy()[0]):g}"
     )
     d = mjw.put_data(mjm, mjd, nworld=_NWORLD.value, nconmax=_NCONMAX.value, njmax=_NJMAX.value)
     print(f"Data\n  nworld: {d.nworld} naconmax: {d.naconmax} njmax: {d.njmax}\n")
 
+
```
thowell: we can remove this new line

mujoco_warp/testspeed.py
```diff
@@ -174,11 +175,13 @@ def _main(argv: Sequence[str]):
       f"  nbody: {m.nbody} nv: {m.nv} ngeom: {m.ngeom} nu: {m.nu} is_sparse: {m.opt.is_sparse}\n"
       f"  broadphase: {broadphase} broadphase_filter: {filter}\n"
       f"  solver: {solver} cone: {cone} iterations: {iterations} {ls_str}\n"
-      f"  integrator: {integrator} graph_conditional: {m.opt.graph_conditional}"
+      f"  integrator: {integrator} graph_conditional: {m.opt.graph_conditional}\n"
+      f"  impratio: {1.0 / np.square(m.opt.impratio_invsqrt.numpy()[0]):g}"
```
thowell: what is the `:g` for?

thowell:
left some minor comments and a question, otherwise lgtm. after that all that is left to do is resolve any branch conflicts.

thanks!

mujoco_warp/testspeed.py
zer-art: It stands for "general" format. I used it because impratio values can vary significantly in magnitude. :g automatically switches between fixed-point and scientific notation and removes insignificant trailing zeros, ensuring the output stays compact and readable.

thowell:
thanks @zer-art!

PR #948, opened 2025-12-28
add benchmarks to ci and update benchmarks

thowell:
- adds option `graph_capture` to `benchmark`
- adds option 'device' to 'benchmark'
- run benchmarks with github actions ci (without graph capture in order to work on cpu)
- update AlohaCloth and Cloth benchmarks: increase `njmax` and add `nstep = 100`
- add setup for AlohaCloth

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
         if: matrix.task == 'benchmarks'
         run: |
           asv machine --yes
-          ASV_GRAPH_CAPTURE=false ASV_DEVICE=cpu asv run --quick --show-stderr --python=same
+          ASV_GRAPH_CAPTURE=false ASV_DEVICE=cpu ASV_EVENT_TRACE=false asv run --quick --show-stderr --python=same
 
   kernel_analyzer:
     name: Kernel analyzer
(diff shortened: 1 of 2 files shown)


Tanisha-Sharma005:
@thowell 

CLA signed with **tanishaasharma20@gmail.com** at https://cla.developers.google.com
Git email updated to match. Re-triggered checks.

Please verify CLA status.

Thanks!

thowell:
current nworld for AlohaCloth requires too much memory for github ci

erikfrey:
I'm hesitant to put run benchmarks directly in our CI.  Here's why:

* It's pretty slow!  15 mins is a big bump for every PR
* I think the right time for benchmarks is post-submit, not pre-submit.  So e.g. just run them nightly.
* I'm considering removing `asv` - it's not really doing much for us right now.  mjlab has a really nice html layout we could steal when the time is right, and it's much nicer than `asv` output: https://mujocolab.github.io/mjlab/nightly/
* If there's some component of the benchmarks that we think is contributing to them failing, then we should just test that component.  e.g. if we are inadvertently emitting too many constraints, then we should create unit tests that confirm the right number of constraints are emitted in some scenario.

thowell:
benchmark updates now in #988 and abandoning this approach for testing benchmarks

PR #955, opened 2026-01-01
update `_epa_witness`

thowell:
- transform height field prism vertices into global frame
- update support evaluation geom type


kbayes:
Oh jeez, thanks for the fix!

PR #957, opened 2026-01-01
collision_gjk: precompute face_pr / lower

thowell:


mujoco_warp/_src/collision_gjk.py
```diff
@@ -1266,13 +1266,14 @@ def _epa(
     # compute support point w from the closest face's normal
     lower = wp.sqrt(lower2)
     wi = pt.nvert
-    i1, i2 = _epa_support(pt, wi, geom1, geom2, geomtype1, geomtype2, pt.face_pr[idx] / lower)
+    face_pr_lower = pt.face_pr[idx] / lower
```
kbayes: face_pr_normalized = pt.face_pr[idx] / lower 

as lower = sqrt(dot(pt.face_pr[idx], pt.face_pr[idx]))

kbayes (approved):

PR #932, opened 2025-12-18
Fix index issue in _efc_equality_joint

Kenny-Vilella:

PR #962, opened 2026-01-02
improve testing for batched `Model` fields

thowell:
this pr aims to improve testing for batched `Model` fields. hopefully we can prevent bugs like the ones fixed in #865, #932, #821, #819, #807, #800

notes:
- `--debug_mode` does not work with `--cpu` so the changes to ci.yml will not currently work with the github actions cpu-based ci

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-    types: [ opened, reopened, synchronize, edited ]
+    types: [opened, reopened, synchronize, edited]
 
 jobs:
   format:
@@ -15,17 +15,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
         with:
+          version: "0.9.27"
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install ruff
+      - name: Set up Python ${{ matrix.python-version }}
+        run: uv python install
       - name: Format with Ruff
         run: |
-          ruff format --check .
+          uvx ruff format --check .
 
   linter:
     name: Ruff linter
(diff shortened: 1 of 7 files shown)


thowell:
@Kenny-Vilella @adenzler-nvidia is there a recommended approach for checking if any out of bounds reads occur on cpu? thanks!

Kenny-Vilella:
It will be great to have GPU runner to run compute-sanitizer.
Currently, I am running weekly compute-sanitizer run on a set of benchmark and a subset of UTs (not all due to memory limitation).
If we could automatize that as nightly run, it will probably be the most helpful.

To reply to your question, I guess using valgrind should work.
I tried to run on the UTs and got a lot of noise, but it does detect invalid write/read.

adenzler-nvidia:
I'm curious why debug_mode doesn't work with CPU for you? On my machine the tests pass, they are just a lot slower.

thowell:
@adenzler-nvidia

was noticing a case where `--debug_mode` would produce cuda error 710, but with `--cpu` no errors were raised.

to reproduce:
1. undo the fix here https://github.com/google-deepmind/mujoco_warp/pull/865/changes to induce out of bounds access
2. `pytest -k io_test --debug_mode` -> should raise cuda error 710
3. `pytest -k io_test --debug_mode --cpu` -> no errors (?)


however, this seems to work
```
import warp as wp

wp.set_device("cpu")
wp.config.mode = "debug"

@wp.kernel
def kernel(arr: wp.array(dtype=int)):
  arr[1] = 0 # intential out of bounds

arr = wp.zeros(1, dtype=int)
wp.launch(kernel, dim=1, inputs=[arr])

print(arr.numpy())
```

```
Module __main__ 7cb498a load on device 'cpu' took 1.15 ms  (cached)
Assertion failed: 'i >= -arr.shape[0] && i < arr.shape[0]'
```

adenzler-nvidia:
interesting, if you use -s here you can see the assert is being printed to stderr, so in theory it works. The problem seems to be connected to the warp assert handler and how assert(false) somehow isn't being picked up by the python side properly.

adenzler-nvidia:
I found the issue, made a PR to warp. The issue is that the CPU-side assert handler calls assert(false), but that code is part of the warp library instead of the kernel code so it doesn't respect NDEBUG.

thowell:
nice! thanks @adenzler-nvidia

thowell:
https://github.com/NVIDIA/warp/issues/1148

adenzler-nvidia:
@thowell the fix for warp has been merged and should appear in nightly builds soon.

.github/workflows/ci.yml
```diff
@@ -80,6 +80,9 @@ jobs:
       - name: Test with pytest
         run: |
           pytest -n 2
+      - name: Out-of-bounds check for batched model fields
+        run: |
+          pytest -k io_test --debug_mode
```
adenzler-nvidia, line 93: did you check runtimes, maybe we could enable debug mode for everything?

adenzler-nvidia (approved):
LGTM!

.github/workflows/ci.yml
thowell: when i run `pytest -n 8 --debug_mode` there are a many of errors, for example `RuntimeError: Failed to acquire primary context for device cuda:0`. will leave running all tests with debug_mode as future work.

PR #972, opened 2026-01-05
Optimization: solve_init_jaref

Kenny-Vilella:
I followed the optimization done for the `linesearch_jv_fused` kernel, since they are doing similar calculation.

The optimization is too minor to show up clearly on the benchmark result but it is clear on nsight:
With optimization
Screenshot from 2026-01-05 15-30-41
Without optimization
Screenshot from 2026-01-05 15-30-56

Results obtained running the following command:
```
python mujoco_warp/testspeed.py --function=step benchmark/humanoid/n_humanoid.xml --nworld=8192 --nstep=1000 --njmax=190 --nconmax=70 -o opt.ls_parallel=True
```

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -1002,21 +1002,6 @@ def solve_init_efc(
   efc_search_dot_out[worldid] = 0.0
 
 
-@wp.kernel
-def solve_zero_jaref(
-  # Data in:
-  nefc_in: wp.array(dtype=int),
-  # Data out:
-  efc_Jaref_out: wp.array2d(dtype=float),
-):
-  worldid, efcid = wp.tid()
-
-  if efcid >= nefc_in[worldid]:
-    return
-
-  efc_Jaref_out[worldid, efcid] = 0.0
-
-
 @cache_kernel
 def solve_init_jaref(nv: int, dofs_per_thread: int):
   @nested_kernel(module="unique", enable_backward=False)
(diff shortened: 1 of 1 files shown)


adenzler-nvidia (approved):
Looking good, thank you!

mujoco_warp/_src/solver.py
```diff
@@ -2043,10 +2072,29 @@ def create_context(
...
+
+  if m.nv > 50:
+    dofs_per_thread = 20
+  else:
+    dofs_per_thread = 50
+
+  threads_per_efc = ceil(m.nv / dofs_per_thread)
+  # we need to clear the jv array if we're doing atomic adds.
+  if threads_per_efc > 1:
+    wp.launch(
```
thowell: is the performance about the same if we call `d.efc.Jaref.zero_()` instead of utilizing a custom kernel? 

the former implementation is a bit simpler, but potentially zeros unnecessary elements.

thowell:
left a curiosity comment. otherwise lgtm. thanks @Kenny-Vilella!

mujoco_warp/_src/solver.py
Kenny-Vilella: I just followed the optimization in the `linesearch_jv_fused` kernel without putting too much thought about it, but you are right.
I am not sure how reliable nsight results are for this, but in any case the time spend on this is so negligible that it doesn't make sense to have an extra kernel.
Changed it.

Issue #942, opened 2025-12-22
set sliding friction is 0 cause Nan


PhiDCH:
```python
import mujoco
import mujoco_warp as mjw
import warp as wp
wp.config.quiet = True
import numpy as np
np.set_printoptions(precision=4, suppress=True)


scene = """
<mujoco model="MuJoCo Model">
  <option gravity="1.0 0 -1.0" timestep="0.01" integrator="RK4" iterations="100" tolerance="0" >
  </option>
  <worldbody>
    <geom name="ground" type="plane" size="2 1 1" pos="0 0 0" quat="1 0 0 0" condim="3" friction="0 0 0"/>
    <body name="box_1" pos="0.0 0.0 0.005">
      <freejoint />
      <inertial pos="0 0 0" mass="1" diaginertia="0.0008416666666666668 0.0008416666666666668 0.001666666666666667"/>
      <geom name="box_1_geom_0" size="0.05 0.05 0.005" type="box" condim="3" friction="0 0 0"/>
    </body>
  </worldbody>
</mujoco>
"""

nsteps = 10      # simulate in 10 steps


# sim in mujoco
mj_model = mujoco.MjModel.from_xml_string(scene)
mj_data = mujoco.MjData(mj_model)
for _ in range(nsteps):
    mujoco.mj_step(mj_model, mj_data)

mj_qpos = mj_data.qpos.copy()



# sim in mujoco warp
mj_model = mujoco.MjModel.from_xml_string(scene)
mj_data = mujoco.MjData(mj_model)

mjw_model = mjw.put_model(mj_model)
mjw_data = mjw.put_data(mj_model, mj_data, njmax=100, nworld=1)
with wp.ScopedCapture() as capture:
    mjw.step(mjw_model, mjw_data)

for _ in range(nsteps):
    wp.capture_launch(capture.graph)

mjw_qpos = mjw_data.qpos.numpy().copy()


print("in mujoco, qpos is:\n", mj_qpos)
print("in mujoco warp, qpos is:\n", mjw_qpos)
```

output is nan value 
```console
in mujoco, qpos is:
 [0.005 0.    0.005 1.    0.    0.    0.   ]
in mujoco warp, qpos is:
 [[nan nan nan  0.  0.  0.  1.]]
```

thowell:
[...]

PR #978, opened 2026-01-05
check friction in `io.put_model`

thowell:
instead of silently overriding, provide a warning if small friction values are provided

### DIFF
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -882,26 +882,27 @@ class CollisionTest(parameterized.TestCase):
     np.testing.assert_equal(d.nacon.numpy()[0], 4)
 
   def test_min_friction(self):
-    _, _, _, d = test_data.fixture(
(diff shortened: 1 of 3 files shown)


adenzler-nvidia:
I'm not sure about existing conventions, but it could be an option to override with a warning, and avoid the potential NaN in all cases.

But this looks good as well, I'm just not sure what is consistent behavior.

thowell:
the friction values are clamped [here](https://github.com/google-deepmind/mujoco_warp/blob/4ce1c3e9fd635f6589f8b91b2cfba1739cba01a3/mujoco_warp/_src/collision_primitive.py#L556) with `MJ_MINMU` (=1e-5).

the issue is that this minimum value was tuned for double instead of float and this value seems to be insufficient to prevent nans, see experiments with mujoco using floats [here](https://github.com/google-deepmind/mujoco_warp/issues/942#issuecomment-3714388647).

opened an issue to track updating relevant constants #1000.

PR #981, opened 2026-01-05
Address Warp Deprecation Warnings

christophercrouzet:
Following the [release of Warp v1.11.0](https://github.com/NVIDIA/warp/releases/tag/v1.11.0), deprecation warnings were added for symbols meant to be internal. This MR addresses most of these warnings by referencing the public API instead.

Additionally this:
- Fixes tests that couldn't be run in standalone mode.
- Bumps mujoco to `3.4.1.dev851315742` since the previously used version doesn't seem available anymore.

### DIFF
--- a/mujoco_warp/_src/broadphase_test.py
+++ b/mujoco_warp/_src/broadphase_test.py
@@ -25,7 +25,6 @@ from mujoco_warp import BroadphaseFilter
 from mujoco_warp import BroadphaseType
 from mujoco_warp import DisableBit
 from mujoco_warp import test_data
-
 from mujoco_warp._src import collision_driver
 
 
--- a/mujoco_warp/_src/collision_gjk_test.py
+++ b/mujoco_warp/_src/collision_gjk_test.py
@@ -21,7 +21,6 @@ from mujoco_warp import Data
 from mujoco_warp import GeomType
 from mujoco_warp import Model
 from mujoco_warp import test_data
-
 from mujoco_warp._src.collision_gjk import ccd
 from mujoco_warp._src.collision_gjk import multicontact
 from mujoco_warp._src.collision_primitive import Geom
--- a/mujoco_warp/_src/inverse_test.py
+++ b/mujoco_warp/_src/inverse_test.py
@@ -25,7 +25,6 @@ import mujoco_warp as mjw
 from mujoco_warp import DisableBit
 from mujoco_warp import IntegratorType
 from mujoco_warp import test_data
-
 from mujoco_warp._src import inverse
 
 
--- a/mujoco_warp/_src/ray_test.py
+++ b/mujoco_warp/_src/ray_test.py
@@ -21,7 +21,6 @@ from absl.testing import absltest
 
 import mujoco_warp as mjw
 from mujoco_warp import test_data
-
 from mujoco_warp._src.types import vec6
 
 # tolerance for difference between MuJoCo and MJX ray calculations - mostly
--- a/mujoco_warp/_src/solver_test.py
+++ b/mujoco_warp/_src/solver_test.py
@@ -25,7 +25,6 @@ import mujoco_warp as mjw
 from mujoco_warp import ConeType
 from mujoco_warp import SolverType
 from mujoco_warp import test_data
-
 from mujoco_warp._src import solver
 
 # tolerance for difference between MuJoCo and MJWarp solver calculations - mostly
--- a/mujoco_warp/_src/support_test.py
+++ b/mujoco_warp/_src/support_test.py
@@ -25,7 +25,6 @@ import mujoco_warp as mjwarp
 from mujoco_warp import ConeType
 from mujoco_warp import State
 from mujoco_warp import test_data
-
 from mujoco_warp._src.block_cholesky import create_blocked_cholesky_func
 from mujoco_warp._src.block_cholesky import create_blocked_cholesky_solve_func
 from mujoco_warp._src.warp_util import nested_kernel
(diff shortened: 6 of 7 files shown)


mujoco_warp/_src/broadphase_test.py
```diff
@@ -26,7 +26,7 @@
 from mujoco_warp import DisableBit
 from mujoco_warp import test_data
 
-from . import collision_driver
+from mujoco_warp._src import collision_driver
```
thowell, line 28: is this change necessary?

christophercrouzet: Only to solve running tests outside of `pytest`, otherwise the following error happen:

```
python ./mujoco_warp/_src/broadphase_test.py
Traceback (most recent call last):
  File "./mujoco_warp/_src/broadphase_test.py", line 29, in <module>
    from . import collision_driver
ImportError: attempted relative import with no known parent package
```

thowell: @erikfrey are we good to change the imports? we considered this before https://github.com/google-deepmind/mujoco_warp/pull/760 but decided against it at the time.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "etils[epath]",
     "mujoco>=3.3.7",
     "numpy",
-    "warp-lang>=1.9.1",
+    "warp-lang>=1.11.0",
```
erikfrey, line 32: Why do we need these changes?  we should only change this if mjwarp no longer works with `>=1.9.1` - is that the case?

pyproject.toml
```diff
@@ -56,7 +56,6 @@ dev = [
     "pygls>=1.0.0,<2.0.0",
     "lsprotocol>=2023.0.1,<2024.0.0",
     "mujoco>=3.3.7.dev0",
-    "warp-lang>=1.9.1.dev0",
```
erikfrey, line 58: We put this here because when we develop, we want to use a Warp nightly.  Please revert this change.

mujoco_warp/_src/broadphase_test.py
erikfrey: That's fair - I'm OK with this change.

erikfrey (changes requested):
Thanks for the cleanup!  I think we should revert the changes to `pyproject.toml` but otherwise LGTM

erikfrey (changes requested):
Oh please also clean up the linter errors.  I think you just need to run `ruff check --fix`.

pyproject.toml
christophercrouzet: That's correct. For example, the current MR replaces internal references like:

```
if wp._src.context.runtime.toolkit_version < (12, 4) or wp._src.context.runtime.driver_version < (12, 4):
```

with:

```
if not wp.is_conditional_graph_supported():
```

The latter was introduced as part of the public API overhaul commit [2d6d379a](https://gitlab-master.nvidia.com/omniverse/warp/-/commit/2d6d379acd3aab26035338d67268fc0502e759a7), and released in [v1.11.0](https://gitlab-master.nvidia.com/omniverse/warp/-/releases/v1.11.0).

This MR also uses `wp.type_is_composite()`, another API introduced in v1.11.0, to replace a reference to the internal symbol `wp.types.vector_types`.

pyproject.toml
christophercrouzet: Done! :heavy_check_mark:

mujoco_warp/_src/warp_util.py
```diff
@@ -220,8 +218,7 @@ def _hash_arg(a):
 
 
 def check_toolkit_driver():
-  if wp._src.context.runtime is None:
-    wp._src.context.init()
+  wp.init()
   if wp.get_device().is_cuda:
-    if wp._src.context.runtime.toolkit_version < (12, 4) or wp._src.context.runtime.driver_version < (12, 4):
+    if not wp.is_conditional_graph_supported():
```
thowell, line 223: we [note in the readme](https://github.com/google-deepmind/mujoco_warp?tab=readme-ov-file#installing-for-development) that the minimum cuda version is 12.4, so we still want to check that the toolkit and driver versions meet this requirement. 

not sure that the proposed check for conditional graph support is sufficient for these version checks?

christophercrouzet: `wp.is_conditional_graph_supported()` is currently implemented using the exact same checks against version 12.4 (see [source code](https://github.com/NVIDIA/warp/blob/a12c56c9693498aa8abb4767c41b17c8d535c991/warp/_src/context.py#L7644-L7653)).

As this is only used in `put_model()`, and the previous manual version checks aligned with conditional graph support semantics (see commit [300b5bd](https://github.com/google-deepmind/mujoco_warp/commit/300b5bde19712ea9496229e1a491b123f16c820f)), switching to this function seemed to make sense.

For a future Warp release, we could consider exposing `wp.get_cuda_toolkit_version()` and `wp.get_cuda_driver_version()` in the public API and revisiting this check if it is still deemed necessary?

pyproject.toml
erikfrey: Ah OK, understood.  SGTM.

christophercrouzet:
> Still seeing some ruff linter errors but otherwise LGTM. If ruff is misbehaving you, happy to hop on a quick chat to help you get it sorted. Just let us know!

Apologies, I totally forgot! These should now be addressed. I also rebased on `main` and resolved the merge conflicts.

thowell:
thank you for this contribution @christophercrouzet!

PR #984, opened 2026-01-06
github ci concurrency

thowell:
this should enable currently running tests to continue instead of being canceled when more tests are scheduled, for example when a new pr is created or a pr is updated


erikfrey:
@thowell what's the use case here?  I'm not sure what this PR is solving.

thowell:
`Error: The operation was canceled.` https://github.com/google-deepmind/mujoco_warp/actions/runs/20723476339/job/59493178675

PR #986, opened 2026-01-06
capsule<>capsule: add parallel axes case

thowell:
add parallel axes case to capsule<>capsule collider

mujoco reference: https://github.com/google-deepmind/mujoco/blob/fa044fe0e53d728a0a242bcb5d2a5c7c2c4f4551/src/engine/engine_collision_primitive.c#L436

### DIFF
--- a/mujoco_warp/_src/collision_primitive_core.py
+++ b/mujoco_warp/_src/collision_primitive_core.py
@@ -264,7 +264,7 @@ def capsule_capsule(
     x2 = wp.clamp((v - mb) / mc, -1.0, 1.0)
     vec2 = cap2_pos + axis2 * x2
     dist, pos, normal = sphere_sphere(vec1, cap1_radius, vec2, cap2_radius)
-    if dist <= margin and contact_count < 2:
+    if dist <= margin:
       contact_dist[contact_count] = dist
       contact_pos[contact_count] = pos
       contact_normal[contact_count] = normal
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_primitive.py
```diff
@@ -824,37 +824,38 @@ def capsule_capsule_wrapper(
...
-    contact_solreffriction_out,
-    contact_solimp_out,
-    contact_dim_out,
-    contact_geom_out,
-    contact_worldid_out,
-    contact_type_out,
-    contact_geomcollisionid_out,
-    nacon_out,
-  )
+  for i in range(2):
```
kbayes, line 828: Are we always producing two contacts? Only the parallel case needs them.

thowell: `capsule_capsule` returns fixed size objects and then `write_contact` below determines if a contact should be added based on `dist` and `margin`.

many of the wrapper functions follow this pattern. please see `plane_capsule_wrapper` below for an example.

mujoco_warp/_src/collision_primitive_core.py
```diff
@@ -202,26 +202,75 @@ def capsule_capsule(
...
+    contact_normal[0] = normal
+
+    vec1 = cap1_pos - axis1
+    x2 = wp.clamp((v + mb) / mc, -1.0, 1.0)
+    vec2 = cap2_pos + axis2 * x2
+    dist, pos, normal = sphere_sphere(vec1, cap1_radius, vec2, cap2_radius)
+    contact_dist[1] = dist
+    contact_pos[1] = pos
+    contact_normal[1] = normal
+
```
kbayes, line 306: What happened to the x2 = {-1, 1} cases from the MJC reference? I assume they're 1-1 with the four scenarios of the line segments overlapping.

thowell: those cases have been added

kbayes (approved):

PR #992, opened 2026-01-07
Reduce EPA memory usage

kbayes:
* Pack face data into one 32-bit integer instead of three ints.
* Remove EPA map logic

This gives a ~1/3 reduction of memory with ccd_iterations = 35. With simulating 100 box meshes, I'm seeing a 20% increase in ccd performance as well.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -1188,7 +1188,7 @@ def _polytope4(
 
 @wp.func
 def _get_face_verts(face: int) -> wp.vec3i:
-  """ "Return the three vertices of the face given by indices into the polytope vertex array."""
+  """Return the three vertices of the face given by indices into the polytope vertex array."""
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_gjk.py
```diff
@@ -1302,11 +1269,15 @@ def _epa(
 
     # compute horizon for w
     for i in range(pt.nface):
-      if pt.face_index[i] == -2:
+      if pt.face_norm2[i] == DELETED_FACE:
```
thowell: is there any scenario where this exact comparison of floats would be unsafe?
would it make sense to:
- store `face_norm` instead of `face_norm2`
- set invalid faces to `face_norm` < 0.0
- check for invalid faces with `face_norm` < 0

kbayes: I removed this hack with the bit packing.

thowell (approved):

erikfrey:
Faster, less memory, and simpler!  Triple whammy!

PR #1003, opened 2026-01-12
Added explicit check to raise ValueError on CPU devices

zer-art:
### Summary

This PR updates `mujoco_warp/testspeed.py` to prevent crashes when running on non-GPU devices (e.g., macOS CPU).

As discussed in [this comment](https://github.com/google-deepmind/mujoco_warp/pull/947#issuecomment-3729154091), `testspeed.py` calls CUDA functions unconditionally, which leads to errors on CPU environments. This change adds an explicit check to ensure the script is only run on GPU devices.

### Changes:

* Added a device check at the entry point of `testspeed.py`.
* Raises a clear `ValueError: testspeed available for gpu only` if the detected device is not a GPU, preventing obscure runtime errors.

### Verification
I verified this change by running the script in a CPU environment (macOS).

**Command:**
`conda run -n academic python -m mujoco_warp.testspeed mujoco_warp/test_data/humanoid/humanoid.xml --nstep 10 --nworld 10`

**Output (New Behavior):**
`ValueError: testspeed available for gpu only`


mujoco_warp/testspeed.py
```diff
@@ -161,6 +161,9 @@ def _main(argv: Sequence[str]):
   if _CLEAR_KERNEL_CACHE.value:
     wp.clear_kernel_cache()
 
+  if "cpu" in (_DEVICE.value or "", str(wp.get_device())):
```
thowell: does `"cpu" in (_DEVICE.value, wp.get_device())` work?

zer-art: **@thowell** Thanks for the review!

You were right about **str()**—I verified that `wp.get_device() == "cpu"` works directly, so the explicit cast isn't needed.

While testing the simpler in syntax, I noticed an edge case: if I am on a laptop (default CPU) but run with --`device=cuda:0, wp.get_device()` still returns cpu. The in check sees that and raises the error, even though I provided a valid GPU override.

**Would you mind if I switched to this logic instead?**
`if (_DEVICE.value or wp.get_device()) == "cpu":`

thowell: lgtm. thanks!

thowell:
@zer-art lets sync this pr branch with main and then we should be good to merge. thanks!

thowell:
@zer-art will probably need to sync with main again for the github ci. thanks!

zer-art:
Done! I've synced the branch with main again. Thanks for the heads-up!

thowell:
thank you for the contribution @zer-art!

PR #1004, opened 2026-01-12
Use CCD for box-box collisions

kbayes:
This PR switches out the box-box primitive collision method for CCD. For the supplied box100.xml with 100 boxes with 400 worlds we see a speedup from 17.60s to 13.52s. While the primitive method in isolation is faster, the contact pruning in CCD speeds up the solver performance.

### DIFF
--- a/mujoco_warp/_src/sensor_test.py
+++ b/mujoco_warp/_src/sensor_test.py
@@ -22,11 +22,11 @@ from absl.testing import absltest
 from absl.testing import parameterized
 
 import mujoco_warp as mjw
-from mujoco_warp._src.types import CollisionType
 from mujoco_warp import DisableBit
 from mujoco_warp import GeomType
 from mujoco_warp import test_data
 from mujoco_warp._src.collision_driver import MJ_COLLISION_TABLE
+from mujoco_warp._src.types import CollisionType
 
 # tolerance for difference between MuJoCo and MJWarp calculations - mostly
 # due to float precision


mujoco_warp/test_data/convex_collision/box100.xml
```diff
@@ -0,0 +1,37 @@
...
+        <replicate count="4" offset="0 0 .06">
+          <body pos="-.1 -.2 0.15">
+            <freejoint/>
+            <geom class="free"/>
+          </body>
+        </replicate>
+      </replicate>
+    </replicate>
+  </worldbody>
+</mujoco>
```
thowell: please add a new line

kbayes: Done.

erikfrey:
@kbayes exciting!  Can we delete `box_box` from collision_primitive_core and `box_box_wrapper` from `collision_primitive.py` ?

kbayes:
@erikfrey ideally I like to keep it for debugging purposes. When box contact issues pop up, it's nice to switch back and forth to rule out collision bugs.

erikfrey:
@kbayes OK that works for me, but can we make a TODO to remove it down the line?

mujoco_warp/_src/collision_driver.py
```diff
@@ -34,6 +35,44 @@
 
 wp.set_module_options({"enable_backward": False})
 
+# Corresponding table to MuJoCo's mjCOLLISIONFUNC table in engine_collision_driver.c
+# A "True" value corresponds to using a primitive collision method
+MJ_COLLISION_FUNC = {
```
thowell: since `MJ_COLLISION_FUNC` doesn't map keys to functions, would it make sense to rename to something like `MJ_COLLISION_TABLE`?

would it makes sense to introduce an enum with: primitive, ccd, and sdf and have each collider key map to one of these values?

thowell:
2 questions about `MJ_COLLISION_FUNC`, otherwise lgtm. thanks @kbayes!

mujoco_warp/_src/io.py
```diff
@@ -384,18 +384,8 @@ def geom_trid_index(i, j):
     )
   )
 
-  # compute nmaxpolygon and nmaxmeshdeg given the geom pairs for the model
-  nboxbox = m.geom_pair_type_count[geom_trid_index(types.GeomType.BOX, types.GeomType.BOX)]
```
erikfrey, line 388: since we're still supporting primitive box<>box if nativeccd is disabled, do we need to keep around the old way of calculating `nmaxpolygon` and `nmaxmeshdeg` ?

kbayes: Oh this is addressed in collision_convex.py already :)

 nboxbox = _pair_count(GeomType.BOX.value, GeomType.BOX.value)
  if (GeomType.BOX, GeomType.BOX) not in collision_table:
    nboxbox = 0

Issue #280, opened 2025-05-20
Sensors for geometric distance


erikfrey:
`geom_normal`, `geom_dist`, `geom_fromto`

thowell:
depends on #869 in order to match mujoco results for box<>box

PR #1006, opened 2026-01-12
changes for collision sensors

thowell:
primarily changes to enable collision sensors (distance, normal, fromto) to work with ccd box<>box collisions #280. 

`mjwarp-viewer mujoco/test/engine/testdata/collision_convex/perf/box.xml --nconmax=512 --njmax=1024`

https://github.com/user-attachments/assets/0a374a72-dff6-4286-a3c0-6f3b83bfcb01

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -796,10 +796,6 @@ def ccd_kernel_builder(
           geomtype2,
         )
 
-    # collision sensors always evaluate all collisions
-    if is_collision_sensor:
-      ncollision = 4
-
     for i in range(ncollision):
       points[i] = 0.5 * (witness1[i] + witness2[i])
     normal = witness1[0] - witness2[0]


mujoco_warp/_src/collision_gjk.py
```diff
@@ -1965,8 +1965,8 @@ def multicontact(
   geomtype1: int,
   geomtype2: int,
 ) -> Tuple[int, mat43, mat43]:
-  witness1 = mat43()
-  witness2 = mat43()
+  witness1 = mat43(wp.inf, wp.inf, wp.inf, wp.inf, wp.inf, wp.inf, wp.inf, wp.inf, wp.inf, wp.inf, wp.inf, wp.inf)
```
kbayes: Why are these initialized to inf?

thowell: reverted to `mat43()` initialization

mujoco_warp/_src/collision_convex.py
```diff
@@ -797,7 +796,11 @@ def eval_ccd_write_contact(
           geomtype2,
         )
 
-    for i in range(ncontact):
+    # collision sensors always evaluate all collisions
```
kbayes: If ncollision is 1 here, should we be duplicating the 1 witness pair  across the remaining 3?

thowell: removed the duplication for collision sensors

mujoco_warp/_src/collision_primitive.py
```diff
@@ -456,6 +456,8 @@ def write_contact(
     contact_solimp_out[cid] = solimp_in
     contact_type_out[cid] = contact_type
     contact_geomcollisionid_out[cid] = id_
+    return int(active)
```
kbayes, line 459: Just use 1?

thowell: `active` here means that the contact contributes to the physics. i think we want to return `active` instead of `1` since there is a case where we write a "contact" that is not active and only contributes to a contact sensor and would return `0`.

PR #1019, opened 2026-01-13
_flex_bending: fix edge lookup index

thowell:
Fix double vertex address offset when accessing `flex_vertbodyid`. The `v` vector already includes `flex_vertadr[f]`, so it should not be added again.


mujoco_warp/_src/passive.py
```diff
@@ -786,24 +786,24 @@ def passive(m: Model, d: Data):
...
-      m.flex_edgenum,
-      m.flex_vertbodyid,
-      m.flex_edge,
-      m.flex_edgeflap,
-      m.flex_bending,
-      d.flexvert_xpos,
-    ],
-    outputs=[d.qfrc_spring],
-  )
+    wp.launch(
```
quagla, line 789: This indentation looks weird

thowell: ah this has the changes to ~~#1019~~ #1018 included... once that pr is merged the diff here should just be the one line above for changing the indexing.

shi-eric:
Is the `Error: The operation was canceled.` message a new thing? I saw this when trying to get the pipelines for #981 to pass.

thowell:
@shi-eric this error has been occurring frequently. @erikfrey suggested it might be resulting from `pytest -n 4` in ci.yml and running out of resources.

mujoco_warp/_src/passive.py
thowell: merged so this should be resolved

PR #897, opened 2025-12-10
Add implementation of mjd_ellipsoidFluid derivative

Aaryan-549:
Fixes #750
Implements derivative of ellipsoid-based fluid forces with respect to velocities. Includes all component derivatives:
- Added mass forces
- Viscous torque and drag
- Kutta lift
- Magnus force

The implementation adds the fluid force derivatives to qDeriv, following the pattern from the reference C implementation.

PR #1022, opened 2026-01-14
Implement derivative of fluid forces

SUZ-tsinghua:
Hi, I implemented the derivative of fluid forces and also wrote tests for derivative and forward. It can pass all tests.
I also saw this pr #897 . My implementation might be helpful.

### DIFF
--- a/mujoco_warp/_src/forward_test.py
+++ b/mujoco_warp/_src/forward_test.py
@@ -257,25 +257,26 @@ class ForwardTest(parameterized.TestCase):
   @parameterized.product(
     jacobian=(mujoco.mjtJacobian.mjJAC_DENSE, mujoco.mjtJacobian.mjJAC_SPARSE),
     fluidshape=("none", "ellipsoid"),
+    integrator=("implicit", "implicitfast"),
   )
-  def test_implicit_fluid(self, jacobian, fluidshape):
+  def test_implicit_fluid(self, jacobian, fluidshape, integrator):
     mjm, mjd, m, d = test_data.fixture(
       xml=f"""
     <mujoco>
-      <option density="1.225" viscosity="1.8e-5" wind="0 0 0" integrator="implicitfast"/>
+      <option density="1.225" viscosity="1.8e-5" wind="0 0 0" integrator="{integrator}"/>
       <worldbody>
-        <body name="main_sphere" pos="0 0 1">
-          <freejoint name="root"/> <geom name="big_ball" type="sphere" size="0.2" rgba="0.8 0.2 0.2 1" mass="1" fluidshape="{fluidshape}"/>
-          <body name="small_sphere_1" pos="-0.3 0 0">
-            <geom name="ball_1" type="sphere" size="0.1" rgba="0.2 0.8 0.2 1" mass="0.2" fluidshape="{fluidshape}"/>
+        <body pos="0 0 1">
+          <freejoint/>
+          <body pos="0.025 0 0">
+            <geom type="box" size=".025 .01 0.0001" euler="20 0 0" mass="1.0" fluidshape="{fluidshape}"/>
           </body>
-          <body name="small_sphere_2" pos="0.4 0 0">
-            <geom name="ball_2" type="sphere" size="0.2" rgba="0.2 0.2 0.8 1" mass="0.2" fluidshape="{fluidshape}"/>
+          <body pos="-0.025 0 0">
+            <geom type="box" size=".025 .01 0.0001" euler="-19 0 0" mass="1.0" fluidshape="{fluidshape}"/>
           </body>
         </body>
       </worldbody>
       <keyframe>
-        <key qvel="100 -100 10 50 -40 100"/>
+        <key qvel="10 -10 1 5 -4 10"/>
       </keyframe>
     </mujoco>
     """,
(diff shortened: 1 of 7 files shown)


thowell:
@SUZ-tsinghua thank you for your contribution to mujoco warp!

please coordinate with @Aaryan-549 on this feature since it seems that there is overlap between this pr and #897. thanks!

Aaryan-549:
Hi @SUZ-tsinghua, thanks again. I reviewed your changes. Your test case test_smooth_vel_fluid is excellent and exactly what the maintainers requested for #897.

With your permission, I'm going to integrate your derivative_test.py and the io.py enablement into my PR so we can get the Ellipsoid feature merged. I'll stick to my kernel implementation for now to keep the scope focused on Ellipsoids (as you also included the Inertia Box fallback). I will add a Co-authored-by tag to my commit to ensure you get credit for the testing logic.

SUZ-tsinghua:
Hi @Aaryan-549, thank you. Another option is directly merging my pr, since I think my version is more complete. I can add you as a co-author for your credit.😊

Aaryan-549:
Thanks for the offer, but I don't think swapping PRs is the right move at this stage.

PR #897 has already undergone significant review iterations (30+ comments) with @thowell and is very close to merging. Switching to a new PR now would invalidate that review effort and force the maintainers to restart the process.

Also, I noticed your PR includes the Inertia Box implementation. While useful, that expands the scope beyond the original Ellipsoid feature request. It's generally safer to merge the Ellipsoid support first as a focused unit (which #897 does), and then add the Inertia Box as a separate, follow-up PR.

I’ve just updated #897 with the mj_forward parity tests (referencing your logic, thank you!) and the io.py fix. Let’s get the base Ellipsoid feature in through #897 first.

thowell:
@Aaryan-549 @SUZ-tsinghua
thank you for both contributing to the project!

probably makes sense to finish the review of #897 for adding the implementation of `mjd_ellipsoidFluid`. @SUZ-tsinghua it would great if you would add a review to that pr. 

after we land #897, we can coordinate the next features. sound like a plan? otherwise, please let us know if you have alternative suggestions. thanks!

SUZ-tsinghua:
Hi @thowell , thanks for your comment. I personally prefer this pr, because this pr is more complete and cleaner. I have implemented all necessary tests and derivatives for both ellipsoid and box fluid model. I think this pr requires less review and can alleviate further effort to implement a new feature (derivative for box fluid model).
It would be nice if you could spend a little time reviewing my changes. If after little reviewing and you still think finishing #897 is better, I would be happy to review #897 😊

mujoco_warp/_src/forward_test.py
```diff
@@ -254,6 +254,47 @@ def test_implicit_position(self):
...
+  )
+  def test_implicit_fluid(self, jacobian, fluidshape, integrator):
+    mjm, mjd, m, d = test_data.fixture(
+      xml=f"""
+    <mujoco>
+      <option density="1.225" viscosity="1.8e-5" wind="0 0 0" integrator="{integrator}"/>
+      <worldbody>
+        <body name="main_sphere" pos="0 0 1">
+          <freejoint/>
+          <geom type="sphere" size="0.2" rgba="0.8 0.2 0.2 1" mass="1" fluidshape="{fluidshape}"/>
```
yuvaltassa: I don't love that we are using a sphere here, prefer something with different dimensions. Perhaps we can just inline [this model](https://github.com/google-deepmind/mujoco/blob/main/test/engine/testdata/derivative/tumbling_thin_object.xml)? (with non zero qvel keyframe)

mujoco_warp/_src/forward_test.py
```diff
@@ -254,6 +254,47 @@ def test_implicit_position(self):
...
+    )
+
+    mujoco.mj_implicit(mjm, mjd)
+
+    mjw.solve(m, d)  # compute efc.Ma
+    mjw.implicit(m, d)
+
+    tol_pos = 2e-3 if integrator == "implicit" else _TOLERANCE * 10
+    tol_vel = 2e-2 if integrator == "implicit" else _TOLERANCE * 10
+    np.testing.assert_allclose(d.qpos.numpy()[0], mjd.qpos, atol=tol_pos, rtol=tol_pos, err_msg="qpos")
```
yuvaltassa: Could we reduce the scope here? I mean, okay we can *also* look at the next qvel (qpos is pointless btw), but we'll get a better read with potentially smaller tolerances if we compare qDeriv directly...

yuvaltassa (commented):
LGTM modulo comments

thowell:
some modifications to `_qderiv_box_fluid`:
- return early based on `madr < 0`
- instead of traversing chain use `body_isdofancestor` lookup
- guard atomic add

mujoco_warp/_src/forward_test.py
thowell: done

mujoco_warp/_src/forward_test.py
thowell: done

thowell (approved):

yuvaltassa (approved):
NICE

thowell:
thanks @SUZ-tsinghua!

PR #1023, opened 2026-01-14
`collision_gjk._add_edge`: update out of memory check

thowell:


kbayes:
@thowell Thanks! This reminded me that we can actually just pack horizon as we did with epa face. See #1025

PR #1024, opened 2026-01-14
_ray_capsule: vec -> lvec

thowell:
Fix capsule ray intersection bug: use local-frame direction (lvec) instead of world-frame direction (vec) when checking if intersection is within cylindrical section.


thowell:
this is already fixed in the render branch

PR #1028, opened 2026-01-14
`ccd_hfield_kernel`: change inner loop variable name: i -> j

thowell:
the inner loop is currently using the same variable as the outer loop.

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -439,11 +439,7 @@ def ccd_hfield_kernel_builder(
           geom1.hfprism = prism
 
           # prism center
-          x1 = geom1.pos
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_convex.py
```diff
@@ -441,8 +441,8 @@ def ccd_hfield_kernel(
           # prism center
           x1 = geom1.pos
           x1_ = wp.vec3(0.0, 0.0, 0.0)
```
kbayes: Can we just replace this all with a single line:

x1 = geom1.pos + geom1.rot @ ((prism[0] + prism[1] + prism[2] + prism[3] + prism[4] + prism[5]) / 6)

thowell: updated. additionally `/ 6.0` -> `* wp.static(1.0 / 6.0)`

PR #1029, opened 2026-01-14
Optimization: iterative linesearch using tile reductions

adenzler-nvidia:
The main optimization is using an entire thread-block to parallelize constraint row processing and use tile reductions to get the alphas. It also contains a lot of secondary optimizations, I inspected almost every line, shuffled a lot of code around and tried out a lot of things. There is an inline comment that explains some of the main optimization that works - it's a combination of fusing kernels, more parallelism, restructuring branches and removing redundant work, gating unnecessary memory accesses and playing with the placement of the loads.

Some performance numbers:

| Benchmark                 | FPS Main | FPS This | Parallel | Speedup vs Main | Speedup vs Parallel |
|---------------------------|---------:|---------:|---------:|----------------:|--------------------:|
| aloha_cloth               |      378 |      712 |      610 |       1.883597884 |         1.167213115 |
| aloha_pot                 |  2443856 |  2804299 |  2750685 |       1.147489459 |         1.019491145 |
| aloha_sdf                 |   507543 |   560905 |   552287 |       1.105137890 |         1.015604206 |
| apptronik_apollo_flat     |  2703462 |  3771940 |  3358311 |       1.395225825 |         1.123165782 |
| apptronik_apollo_hfield   |  1626559 |  2335867 |  2134352 |       1.436078863 |         1.094415073 |
| apptronik_apollo_terrain  |  1012693 |  1173106 |  1116483 |       1.158402398 |         1.050715506 |
| cloth                     |      569 |      665 |      621 |       1.168717047 |         1.070853462 |
| franka_emika_panda        | 26001116 | 24807900 | 25801338 |       0.954109047 |         0.961496648 |
| humanoid                  |  3647670 |  4893138 |  4380018 |       1.341442071 |         1.117150203 |
| three_humanoids           |   250020 |   425368 |   376680 |       1.701335893 |         1.129255602 |

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -59,7 +59,9 @@ def _eval_pt_direct_alpha_zero(jaref: float, jv: float, d: float) -> wp.vec3:
 
 
 @wp.func
-def _eval_pt_direct_3alphas(jaref: float, jv: float, d: float, lo_alpha: float, hi_alpha: float, mid_alpha: float) -> tuple[wp.vec3, wp.vec3, wp.vec3]:
+def _eval_pt_direct_3alphas(
+  jaref: float, jv: float, d: float, lo_alpha: float, hi_alpha: float, mid_alpha: float
+) -> tuple[wp.vec3, wp.vec3, wp.vec3]:
   """Eval quadratic constraint for 3 alphas."""
   x_lo = jaref + lo_alpha * jv
   x_hi = jaref + hi_alpha * jv
@@ -129,7 +131,9 @@ def _eval_frictionloss_pt_one(x: float, f: float, rf: float, half_d: float, jvD:
 
 
 @wp.func
-def _eval_frictionloss_pt_3alphas(x_lo: float, x_hi: float, x_mid: float, f: float, rf: float, jv: float, d: float) -> tuple[wp.vec3, wp.vec3, wp.vec3]:
+def _eval_frictionloss_pt_3alphas(
+  x_lo: float, x_hi: float, x_mid: float, f: float, rf: float, jv: float, d: float
+) -> tuple[wp.vec3, wp.vec3, wp.vec3]:
   """Eval frictionloss for 3 x values with shared precomputation."""
   jvD = jv * d
   half_d = 0.5 * d


adenzler-nvidia:
sharing to get early feedback. It still needs cleanup and double-checks:

* linesearch UTs are failing because the new version doesn't write alpha out anymore
* didn't look at kernel analyzer failures yet
* probably some experimental changes weren't reverted yet.

adenzler-nvidia:
It's kind of ready to review, but I forgot about the fact that syncthreads does not work on CPU. Need to find a solution for this.

adenzler-nvidia:
Ready for review now @thowell @erikfrey

mujoco_warp/_src/solver.py
```diff
@@ -44,20 +43,80 @@ def _in_bracket(x: wp.vec3, y: wp.vec3) -> bool:
   return (x[1] < y[1] and y[1] < 0.0) or (x[1] > y[1] and y[1] > 0.0)
 
 
+@wp.func
+def _compute_quad(jaref: float, jv: float, d: float) -> wp.vec3:
+  """Compute quad coefficients from Jaref, jv, and efc_D (for pyramidal cones)."""
```
thowell: should we expect an additional improvement by computing:
`jarefd = jaref * d` or `jvd = jv * d` and then reusing `jarefd` or `jvd`?

mujoco_warp/_src/solver.py
```diff
@@ -44,20 +43,80 @@ def _in_bracket(x: wp.vec3, y: wp.vec3) -> bool:
...
+
+  Equivalent to _eval_pt(_compute_quad(jaref, jv, d), alpha) but more efficient.
+  """
+  x = jaref + alpha * jv
+  jvD = jv * d
+  return wp.vec3(0.5 * d * x * x, jvD * x, jv * jvD)
+
+
+@wp.func
+def _eval_pt_direct_alpha_zero(jaref: float, jv: float, d: float) -> wp.vec3:
```
thowell, line 55: this function returns the same result as `_compute_quad` except a difference in the third element, not scaling by `0.5` here?

mujoco_warp/_src/solver.py
```diff
@@ -647,6 +358,42 @@ def linesearch_parallel_best_alpha(
 
 
 def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=float)):
+  """Parallel linesearch with setup and teardown kernels."""
+  dofs_per_thread = 20 if m.nv > 50 else 50
```
thowell, line 370: do we expect an optimal (or good) value for `dofs_per_thread` to remain the same across gpu architectures and cuda versions? would it makes sense to create a dataclass like `BlockDim` that contains settings like this?

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+# kernel_analyzer: off
+@wp.func
+def _compute_efc_eval_pt_pyramidal(
+  efcid: int,
+  alpha: float,
+  ne: int,
+  nf: int,
+  # Per-row data:
+  efc_D: float,
+  efc_frictionloss: wp.array(dtype=float),
```
thowell, line 464: why is this not a float?

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+
+
+@wp.func
+def _compute_efc_eval_pt_elliptic(
+  efcid: int,
+  alpha: float,
+  ne: int,
+  nf: int,
+  impratio_invsqrt: float,
+  # Per-row data (arrays for deferred load):
```
thowell, line 494: maybe this answers my question above about why not float?

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+    ls_iterations: Max linesearch iterations (compile-time constant for loop optimization).
+    cone_type: Friction cone type (PYRAMIDAL or ELLIPTIC) for compile-time optimization.
+    fuse_jv: Whether to compute jv = J @ search in-kernel (efficient for small nv).
+  """
+  LS_ITERATIONS = ls_iterations
+  IS_ELLIPTIC = cone_type == types.ConeType.ELLIPTIC
+  FUSE_JV = fuse_jv
+
+  # Native snippet for CUDA __syncthreads()
+  @wp.func_native(snippet="WP_TILE_SYNC();")
```
thowell, line 786: is this something that will make its way into warp?

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+      impratio_invsqrt = opt_impratio_invsqrt[worldid % opt_impratio_invsqrt.shape[0]]
+      nacon = nacon_in[0]
+
+      for efcid in range(tid, nefc, wp.block_dim()):
+        Jaref = efc_Jaref_in[worldid, efcid]
+        jv = efc_jv_inout[worldid, efcid]
+        efc_D = efc_D_in[worldid, efcid]
+
+        # scalar quadratic coefficients
+        quad = wp.vec3(0.5 * Jaref * Jaref * efc_D, jv * Jaref * efc_D, 0.5 * jv * jv * efc_D)
```
thowell: should we use `_compute_quad` here?

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+              vv = float(0.0)
+              for j in range(1, dim):
+                efcidj = contact_efc_address_in[conid, j]
+                if efcidj >= 0:
+                  jvj = efc_jv_inout[worldid, efcidj]
+                  jarefj = efc_Jaref_in[worldid, efcidj]
+                  dj = efc_D_in[worldid, efcidj]
+                  DJj = dj * jarefj
+
+                  quad += wp.vec3(0.5 * jarefj * DJj, jvj * DJj, 0.5 * jvj * dj * jvj)
```
thowell, line 899: `_compute_quad`?

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+    lo_less = lo_in[1] < p0[1]
+    lo = wp.where(lo_less, lo_in, p0)
+    lo_alpha = wp.where(lo_less, lo_alpha_in, 0.0)
+    hi = wp.where(lo_less, p0, lo_in)
+    hi_alpha = wp.where(lo_less, 0.0, lo_alpha_in)
+
+    # main iterative loop
+    alpha = float(0.0)
+
+    for _ in range(wp.static(LS_ITERATIONS)):
```
thowell: in general, is utilizing `wp.static` for loops like this more performant?

mujoco_warp/_src/solver.py
adenzler-nvidia: so far it's been pretty consistent in our testing, but might be that this changes with newer hardware coming in.

I'd say once we have some setup to tune block_dims and other threshold to a workload-GPU combination we should add that. I wouldn't go there yet though, and especially not in combination with this change.

mujoco_warp/_src/solver.py
adenzler-nvidia: Probably not, it's really just a bit of a hack. The main benefit of fusing this kernel in is to avoid the private caches being invalidated (which happens on kernel boundaries) so I don't expect this to be a more first-class warp feature at any point.

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+        local_mid[0],
+        local_lo[1],
+        local_hi[1],
+        local_mid[1],
+        local_lo[2],
+        local_hi[2],
+        local_mid[2],
+      )
+      combined_tile = wp.tile(local_combined, preserve_type=True)
+      combined_sum = wp.tile_reduce(wp.add, combined_tile)
```
thowell, line 1146: maybe add a comment explaining the `wp.tile_reduce` operation here?

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +428,827 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+
+    # Jaref update
+    for efcid in range(tid, nefc, wp.block_dim()):
+      efc_Jaref_out[worldid, efcid] += alpha * efc_jv_inout[worldid, efcid]
+
+  return kernel
+
+
+def _linesearch_iterative(m: types.Model, d: types.Data, fuse_jv: bool):
+  """Iterative linesearch with parallel reductions over efc rows.
```
thowell: there are also reductions over dofs (line 1199-1201)?

thowell:
@adenzler-nvidia very awesome pr! the implementation is excellent and i feel like i learned a bunch seeing these techniques applied to improve performance.

lgtm. deferring to @erikfrey for additional review and approval.

thanks!

adenzler-nvidia:
Thanks for the detailed feedback! I will go over it and see whether we can squeeze out some more on Monday.

mujoco_warp/_src/solver.py
adenzler-nvidia: I didn't see a noticeable difference, but it probably won't hurt. Done.

mujoco_warp/_src/solver.py
adenzler-nvidia: yes. I removed compute_quad now though because it made sense to inline into all call sites.

mujoco_warp/_src/solver.py
adenzler-nvidia: Not all branches of the function need frictionloss, so passing like this will make sure we only load the value if we really need it.

mujoco_warp/_src/solver.py
adenzler-nvidia: yes :)

mujoco_warp/_src/solver.py
adenzler-nvidia: we could, but it's gone now, I hope it's ok to just compute inline.

mujoco_warp/_src/solver.py
adenzler-nvidia: same comment as above.

mujoco_warp/_src/solver.py
adenzler-nvidia: It depends. I also tried it for the nv loops and it didn't help. It unrolls the loops and thus increases the optimization window, so it's always worth a try.

The static is not actually needed as warp will figure out that this is a constant. Let me remove it.

mujoco_warp/_src/solver.py
adenzler-nvidia: done

mujoco_warp/_src/solver.py
adenzler-nvidia: true, updated.

adenzler-nvidia:
next iteration online, not much more in terms of perf but a bit of cleanup and some comments.

mujoco_warp/_src/types.py
```diff
@@ -1506,8 +1506,8 @@ class Constraint:
     mv: qM @ search                                   (nworld, nv)
     jv: efc_J @ search                                (nworld, njmax)
     quad: quadratic cost coefficients                 (nworld, njmax, 3)
-    quad_gauss: quadratic cost Gauss coefficients     (nworld, 3)
-    alpha: line search step size                      (nworld,)
+    quad_gauss: quadratic cost Gauss coefficients     (nworld, 3)  # parallel linesearch only
```
erikfrey: nit: this implementation detail  (parallel linesearch only) is good for the code below but I'd leave it out of the docstring

mujoco_warp/_src/solver.py
```diff
@@ -44,39 +43,103 @@ def _in_bracket(x: wp.vec3, y: wp.vec3) -> bool:
...
+
+@wp.func
+def _eval_pt_direct_alpha_zero(jaref: float, jv: float, d: float) -> wp.vec3:
+  """Eval quadratic constraint at alpha=0."""
+  jvD = jv * d
+  return wp.vec3(0.5 * d * jaref * jaref, jvD * jaref, jv * jvD)
+
+
+@wp.func
+def _eval_pt_direct_3alphas(jaref: float, jv: float, d: float, lo_alpha: float, hi_alpha: float, mid_alpha: float):
```
erikfrey: nit: let's add return type to the function signature.  I've been trying to keep those in.

mujoco_warp/_src/solver.py
```diff
@@ -44,39 +43,103 @@ def _in_bracket(x: wp.vec3, y: wp.vec3) -> bool:
...
-    return wp.vec3(f * (-0.5 * rf - Jaref), -f * jv, 0.0)
-  # bound < x : linear positive
+    return wp.vec3(f * (-0.5 * rf - x), -f_jv, 0.0)
   else:
-    return wp.vec3(f * (-0.5 * rf + Jaref), f * jv, 0.0)
+    return wp.vec3(f * (-0.5 * rf + x), f_jv, 0.0)
+
+
+@wp.func
+def _eval_frictionloss_pt_3alphas(x_lo: float, x_hi: float, x_mid: float, f: float, rf: float, jv: float, d: float):
```
erikfrey: nit: same here, `-> tuple[wp.vec3, ...`

mujoco_warp/_src/solver.py
```diff
@@ -681,6 +432,813 @@ def _linesearch_parallel(m: types.Model, d: types.Data, cost: wp.array2d(dtype=f
...
+  FUSE_JV = fuse_jv
+
+  # Native snippet for CUDA __syncthreads()
+  @wp.func_native(snippet="WP_TILE_SYNC();")
+  def _syncthreads():
+    pass
+
+  # Select specialized helper functions based on cone type
+  if IS_ELLIPTIC:
+    _compute_efc_eval_pt = _compute_efc_eval_pt_elliptic
```
erikfrey, line 792: nice    :+1:

mujoco_warp/_src/solver.py
erikfrey: Although I have no problems checking this in, over time I would love it if MJWarp is using Warp in a "standard" way.  Things I would love to see eventually disappear from mjwarp codebase:

* this `WP_TILE_SYNC()` thingie
* `nested_kernel` / `cache_kernel`
* `block_cholesky`

mujoco_warp/_src/types.py
adenzler-nvidia: thanks, done

mujoco_warp/_src/solver.py
adenzler-nvidia: added return types for all

mujoco_warp/_src/solver.py
adenzler-nvidia: I will start those conversations. I'm not sure about block cholesky, but the others should be possible.

PR #1034, opened 2026-01-14
Fix sphere GJK support function returning center instead of surface point.

kevinzakka:
The sphere support was `pos + (0.5 * margin) * size[0] * dir`, which returns the sphere center when margin=0. Fixed to `pos + size[0] * dir` with margin handled at function end like other geom types.

**Before**

Screenshot From 2026-01-14 13-24-16

**After**

Screenshot From 2026-01-14 13-36-29

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -92,7 +92,7 @@ def support(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
   sp.cached_index = -1
   sp.vertex_index = -1
   if geomtype == GeomType.SPHERE:
-    sp.point = geom.pos + (0.5 * geom.margin) * geom.size[0] * dir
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_gjk.py
```diff
@@ -92,8 +92,7 @@ def support(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
   sp.cached_index = -1
   sp.vertex_index = -1
   if geomtype == GeomType.SPHERE:
-    sp.point = geom.pos + (0.5 * geom.margin) * geom.size[0] * dir
```
kbayes, line 95: I meant to write:

sp.point = geom.pos + ((0.5 * geom.margin) + geom.size[0]) * dir

kbayes:
Nice catch! See my comment. You need the early return or else you'll hit the extra logic.

thowell:
great fix, thank @kevinzakka!

PR #1037, opened 2026-01-15
Render Cleanup 2

StafaH:
This PR adds:
- Parsing of the new MuJoCo output field for rgb/depth types
- Use of warp's mesh query ray anyhit function for fast shadow calculation

### DIFF
--- a/mujoco_warp/_src/ray.py
+++ b/mujoco_warp/_src/ray.py
@@ -731,19 +731,14 @@ def ray_mesh_with_bvh_anyhit(
   pnt: wp.vec3,
   vec: wp.vec3,
   max_t: float,
-) -> float:
-  """Returns 1 if there is any hit for ray mesh intersections, otherwise -1.
+) -> bool:
+  """Returns True if there is any hit for ray mesh intersections.
 
   Requires wp.Mesh be constructed and their ids to be passed. This variant is useful
-  for shadow ray casts where the only goal if there is any ray hit.
+  for shadow ray casts where the only goal is if there is any ray hit.
   """
   lpnt, lvec = _ray_map(pos, mat, pnt, vec)
-  hit = wp.mesh_query_ray_anyhit(mesh_bvh_id[mesh_geom_id], lpnt, lvec, max_t)
-
-  if hit:
-    return 1.0
-
-  return -1.0
+  return wp.mesh_query_ray_anyhit(mesh_bvh_id[mesh_geom_id], lpnt, lvec, max_t)
 
 
 @wp.func
--- a/mujoco_warp/_src/render_context.py
+++ b/mujoco_warp/_src/render_context.py
@@ -223,23 +223,19 @@ class RenderContext:
 
     self.cam_res = wp.array(active_cam_res, dtype=wp.vec2i)
 
-    if render_rgb is not None:
-      if isinstance(render_rgb, bool):
-        render_rgb = [render_rgb] * ncam
-      assert len(render_rgb) == ncam, (
-        f"Render RGB must be provided for all active cameras (got {len(render_rgb)}, expected {ncam})"
-      )
-    else:
+    if render_rgb and isinstance(render_rgb, bool):
+      render_rgb = [render_rgb] * ncam
+    elif render_rgb is None:
       render_rgb = [mjm.cam_output[i] & mujoco.mjtCamOutBit.mjCAMOUT_RGB for i in active_cam_indices]
 
-    if render_depth is not None:
-      if isinstance(render_depth, bool):
-        render_depth = [render_depth] * ncam
-      assert len(render_depth) == ncam, (
-        f"Render depth must be provided for all active cameras (got {len(render_depth)}, expected {ncam})"
-      )
-    else:
+    if render_depth and isinstance(render_depth, bool):
+      render_depth = [render_depth] * ncam
+    elif render_depth is None:
       render_depth = [mjm.cam_output[i] & mujoco.mjtCamOutBit.mjCAMOUT_DEPTH for i in active_cam_indices]
+
+    assert len(render_rgb) == ncam and len(render_depth) == ncam, (
+      f"Render RGB and depth must be provided for all active cameras (got {len(render_rgb)}, {len(render_depth)}, expected {ncam})"
+    )
 
     rgb_adr = -1 * np.ones(ncam, dtype=int)
     depth_adr = -1 * np.ones(ncam, dtype=int)
--- a/mujoco_warp/_src/render_context_test.py
+++ b/mujoco_warp/_src/render_context_test.py
@@ -47,13 +47,13 @@ class RenderContextTest(parameterized.TestCase):
     self.assertIsNotNone(rc)
     self.assertEqual(rc.ncam, mjm.ncam)
 
-    self.assertEqual(rc.lower.shape, (nworld * rc.bvh_ngeom,), "render context lower shape")
-    self.assertEqual(rc.upper.shape, (nworld * rc.bvh_ngeom,), "render context upper shape")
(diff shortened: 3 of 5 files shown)


mujoco_warp/_src/ray.py
```diff
@@ -721,6 +721,31 @@ def ray_mesh_with_bvh(
...
+  pos: wp.vec3,
+  mat: wp.mat33,
+  pnt: wp.vec3,
+  vec: wp.vec3,
+  max_t: float,
+) -> float:
+  """Returns 1 if there is any hit for ray mesh intersections, otherwise -1.
+
+  Requires wp.Mesh be constructed and their ids to be passed. This variant is useful
+  for shadow ray casts where the only goal if there is any ray hit.
```
thowell: ... goal is if

mujoco_warp/_src/ray.py
```diff
@@ -721,6 +721,31 @@ def ray_mesh_with_bvh(
...
+  """
+  lpnt, lvec = _ray_map(pos, mat, pnt, vec)
+  hit = wp.mesh_query_ray_anyhit(mesh_bvh_id[mesh_geom_id], lpnt, lvec, max_t)
```
thowell: can this function return a `bool` instead of a `float` and then directly return the output from `wp.mesh_query_ray_anyhit`?

mujoco_warp/_src/render_context.py
```diff
@@ -211,29 +211,35 @@ def __init__(
...
+      active_cam_res = mjm.cam_resolution[active_cam_indices]
 
     self.cam_res = wp.array(active_cam_res, dtype=wp.vec2i)
 
-    if isinstance(render_rgb, bool):
-      render_rgb = [render_rgb] * ncam
-    assert len(render_rgb) == ncam, (
-      f"Render RGB must be provided for all active cameras (got {len(render_rgb)}, expected {ncam})"
-    )
+    if render_rgb is not None:
```
thowell: python `and` should short circuit, so can the check in this line and the check in the line below be combined into 1 line? something like `if render_rgb and isinstance(render_rgb, bool)`?

mujoco_warp/_src/render_context.py
```diff
@@ -211,29 +211,35 @@ def __init__(
...
+      )
+    else:
+      render_rgb = [mjm.cam_output[i] & mujoco.mjtCamOutBit.mjCAMOUT_RGB for i in active_cam_indices]
 
-    if isinstance(render_depth, bool):
-      render_depth = [render_depth] * ncam
-    assert len(render_depth) == ncam, (
-      f"Render depth must be provided for all active cameras (got {len(render_depth)}, expected {ncam})"
-    )
+    if render_depth is not None:
```
thowell: same comment as above

mujoco_warp/_src/render_context_test.py
```diff
@@ -74,18 +74,14 @@ def test_output_buffers(self):
     np.testing.assert_array_equal(rgb_adr, [0, width * height, 2 * width * height], err_msg="render context rgb adr values")
     np.testing.assert_array_equal(depth_adr, [0, width * height, 2 * width * height], err_msg="render context depth adr values")
 
-  def test_different_camera_resolutions(self):
-    """Tests render context with different resolutions per camera."""
+  def test_heterogeneous_camera(self):
+    """Tests render context with different resolutions and output ."""
```
thowell: please remove space after 'output'

mujoco_warp/_src/render_context_test.py
```diff
@@ -74,18 +74,14 @@ def test_output_buffers(self):
...
     cam_res = [(64, 64), (32, 32), (16, 16)]
     rc = mjw.create_render_context(mjm, m, d, cam_res=cam_res, render_rgb=True, render_depth=True)
 
     self.assertEqual(rc.ncam, 3, "render context ncam")
-
-    cam_res_np = rc.cam_res.numpy()
-    np.testing.assert_array_equal(cam_res_np[0], [64, 64], err_msg="render context cam res values")
-    np.testing.assert_array_equal(cam_res_np[1], [32, 32], err_msg="render context cam res values")
-    np.testing.assert_array_equal(cam_res_np[2], [16, 16], err_msg="render context cam res values")
+    np.testing.assert_array_equal(rc.cam_res.numpy(), cam_res, err_msg="render context cam res values")
```
thowell: we can simplify the error message to `cam_res`

mujoco_warp/_src/render_context_test.py
```diff
@@ -96,6 +92,15 @@ def test_different_camera_resolutions(self):
     np.testing.assert_array_equal(rgb_adr, [0, 64 * 64, 64 * 64 + 32 * 32], err_msg="render context rgb adr values")
     np.testing.assert_array_equal(depth_adr, [0, 64 * 64, 64 * 64 + 32 * 32], err_msg="render context depth adr values")
 
+    # Test that results are same when reading from xml values
```
thowell: what do we mean by 'reading from xml values'?

mujoco_warp/_src/render_context_test.py
```diff
@@ -123,23 +128,25 @@ def test_rgb_only_and_depth_only(self):
...
-    self.assertEqual(depth_adr[1], 0, "render context depth adr value")
-
-    self.assertEqual(rgb_adr[2], pixels, "render context rgb adr value")
-    self.assertEqual(depth_adr[2], pixels, "render context depth adr value")
-
-    render_rgb = rc.render_rgb.numpy()
-    render_depth = rc.render_depth.numpy()
-    np.testing.assert_array_equal(render_rgb, [True, False, True], err_msg="render context render rgb values")
-    np.testing.assert_array_equal(render_depth, [False, True, True], err_msg="render context render depth values")
+    np.testing.assert_array_equal(rc.rgb_adr.numpy(), [0, -1, pixels], "render context rgb adr value")
```
thowell: we can simplify the error messages for these tests to be the name of the field that is being tested.

mujoco_warp/_src/render_context_test.py
```diff
@@ -123,23 +128,25 @@ def test_rgb_only_and_depth_only(self):
...
+    np.testing.assert_array_equal(rc.depth_adr.numpy(), [-1, 0, pixels], "render context depth adr value")
+    np.testing.assert_array_equal(rc.render_rgb.numpy(), [True, False, True], err_msg="render context render rgb values")
+    np.testing.assert_array_equal(rc.render_depth.numpy(), [False, True, True], err_msg="render context render depth values")
+
+    # Test that results are same when reading from xml values
```
thowell: same question as above

thowell:
thanks @StafaH!

mujoco_warp/_src/ray.py
StafaH: Done. This changes the logic in the cast_ray_first_hit since we still need to convert the bool into a distance value, but we can just do that in one line after the check. (`d = 0.0 if hit else -1.0`)

mujoco_warp/_src/render_context.py
StafaH: Did a small refactor to make it cleaner for both, let me know if this is not what you intended

mujoco_warp/_src/render_context_test.py
StafaH: This is to make sure values read from the xml/mjmodel also create an equivalent result to when we specify them manually.

StafaH: Added small clarification in the comment

mujoco_warp/_src/render_context_test.py
StafaH: Fixed for all error messages across all tests for renderer

thowell:
lgtm. thanks for these improvements!

PR #1039, opened 2026-01-15
update uv lock file

adenzler-nvidia:
changes like #1035 should really also update uv.lock. One more reason to use uv in CI :)


thowell:
@adenzler-nvidia what is the status of this pr now that #1084 is merged?

adenzler-nvidia:
closing as it seems to run with the versions locked in uv.lock. I guess you added some backwards compatibility for older mujoco?

PR #1041, opened 2026-01-15
ci.yml: pytest update `-n 4` -> `-n 2`

thowell:
attempt to fix recurring `Error: The operation was canceled` issue.


erikfrey (approved):
I'm not sure this will fix the problems we are seeing but it certainly doesn't hurt to try!  So let's merge

PR #1043, opened 2026-01-15
QoL: avoid running menagerie tests from base directory

adenzler-nvidia:
With the move of the conftest file into mujoco_warp/, the menagerie tests are discovered again if you accidentally run pytest from the base directory. This change avoids going into benchmark/ and contrib/ for testing.


thowell (approved):
thank you!

Issue #1044, opened 2026-01-15
io_test fail


SUZ-tsinghua:
When I run io_test, I found it will fail for these 3 tests:

<img width="2816" height="182" alt="Image" src="https://github.com/user-attachments/assets/b91f19ab-7e30-4b53-8021-adbfd484b738" />

thowell:
thank you for reporting (and fixing! with #1045) this bug

SUZ-tsinghua:
Close this issue since it's solved

PR #1045, opened 2026-01-15
fix io_test error

SUZ-tsinghua:
Fix the io_test error raised in #1044

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -999,7 +999,8 @@ def get_data_into(
   result.cinert[:] = d.cinert.numpy()[world_id]
   result.flexvert_xpos[:] = d.flexvert_xpos.numpy()[world_id]
   flexedge_J = d.flexedge_J.numpy()[world_id]
-  mujoco.mju_dense2sparse(result.flexedge_J, flexedge_J, mjm.flexedge_J_rownnz, mjm.flexedge_J_rowadr, mjm.flexedge_J_colind)
+  if flexedge_J.size:
+    mujoco.mju_dense2sparse(result.flexedge_J, flexedge_J, mjm.flexedge_J_rownnz, mjm.flexedge_J_rowadr, mjm.flexedge_J_colind)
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/io.py
```diff
@@ -999,7 +999,8 @@ def get_data_into(
   result.cinert[:] = d.cinert.numpy()[world_id]
   result.flexvert_xpos[:] = d.flexvert_xpos.numpy()[world_id]
   flexedge_J = d.flexedge_J.numpy()[world_id]
-  mujoco.mju_dense2sparse(result.flexedge_J, flexedge_J, mjm.flexedge_J_rownnz, mjm.flexedge_J_rowadr, mjm.flexedge_J_colind)
+  if flexedge_J.size:
+    mujoco.mju_dense2sparse(result.flexedge_J, flexedge_J, mjm.flexedge_J_rownnz, mjm.flexedge_J_rowadr, mjm.flexedge_J_colind)
```
quagla, line 1003: Is this even needed now? flexedge_J is always sparse now in mujoco

erikfrey: Yep it's needed because this change to flexedge_J is not released yet, and in general we try to maintain backwards compatibility with previous MuJoCo releases.

thowell: @SUZ-tsinghua

the update to `io.get_data_into` should maintain backward compatibility with the release version of mujoco ([3.4.0](https://pypi.org/project/mujoco/3.4.0/)) and work with the nightly version so we need 2 cases here for now.

see [here](https://github.com/google-deepmind/mujoco_warp/pull/1036/files) for an example that checked the warp version for a feature.

in addition to the 2 cases, please added a todo `# TODO(team): ...` to replace the 2 cases with just the cases that currently works with the nightly build of mujoco once the next version of mujoco is available.

SUZ-tsinghua:
Close this pr since the issue has been solved.

mujoco_warp/_src/io.py
thowell: resolved with #1052

PR #1046, opened 2026-01-16
Update links in `README.md`

shi-eric:
- Added missing line breaks for better readability in Linux and Windows installation sections.
- Corrected the path for benchmark files from `benchmark/` to `benchmarks/`.
- Updated the link for differentiability documentation to the correct user guide page.


erikfrey:
Thanks so much @shi-eric ❤️

PR #1049, opened 2026-01-16
Fix potential OOB read

adenzler-nvidia:
wp.where evaluates both branches and then selects. This removes the branch and uses modulo indexing to avoid the potential overflow.


kbayes:
Awesome!!

PR #1051, opened 2026-01-16
Use MjMAXVAL to signal invalid/large values instead of wp.inf

adenzler-nvidia:
This change does not change behavior, but using wp.inf as a sentinel value is leading to a lot of false positives when using verify_fp = True in warp options.

I don't need this change 100% in, but I felt like I can make the PR because I had to make the change anyway for debugging.


kbayes:
@adenzler-nvidia  This is nice! Approval on collision code.

adenzler-nvidia:
actually, we might want to make the values something that is exactly representable in single precision. I just came across an issue in Newton (where I'm proposing the same change) where that was problematic.

Does C MuJoCo change that constant if compiled for single precision?

thowell:
@adenzler-nvidia 
float versus double differences [here](https://github.com/google-deepmind/mujoco/blob/77d1ac10feacd2c3b50088ea840426692fbc82e3/include/mujoco/mjtnum.h#L29) do not include `mjMAXVAL` (=[1e10](https://mujoco.readthedocs.io/en/latest/APIreference/APIglobals.html#glnumeric))

kbayes:
@adenzler-nvidia shouldn't mjMAXVAL be used only for comparisons? Any arithmetic operations with it would cause issues. What was the issue in Newton?

adenzler-nvidia:
I think it's actually a warp issue, but we could work around it by having a value that is exactly representable in float32.

Python will make 1e10 a float64 constant because otherwise the number is not representable. That led to issues where we compared the converted float32 value (which is smaller than 1e10) to the float64 constant, and then we failed some checks.

I think declaring MaxVal as `wp.constant(wp.float32(MjMaxVal))` in types.py should just also convert it to the right value, that might be the easier fix.

adenzler-nvidia:
ok that might have been unnecessary noise. The issue in newton is a rounding issue in warp.volume that seems unrelated to representation, I might just have to change 1 threshold there.

adenzler-nvidia:
confirming that. We are good with just using MjMaxVal. The Newton issues were caused by interpolation issues with wp.volumes.

thowell:
@adenzler-nvidia what is the status of this pr?

adenzler-nvidia:
Good to go from my side, will merge in latest main.

adenzler-nvidia:
I have no idea why the CI keeps failing though. I somehow got it to turn green by rebasing on other branches, but this one won't budge

mujoco_warp/_src/ray.py
```diff
@@ -922,7 +923,7 @@ def ray(
     d: The data object containing the current state and output arrays (device).
     pnt: Ray origin points.
     vec: Ray directions.
-    geomgroup: Group inclusion/exclusion mask. If all are wp.inf, ignore.
+    geomgroup: Group inclusion/exclusion mask.
```
thowell, line 926: nice catch!

PR #1053, opened 2026-01-16
check if mujoco is dev version with `_is_mujoco_dev`

thowell:
instead of checking for a specific dev version of mujoco, just check if dev.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -15,6 +15,7 @@
 
 import dataclasses
 import importlib.metadata
+import re
 import warnings
 from typing import Any, Optional, Sequence, Union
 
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/io.py
```diff
@@ -26,11 +27,23 @@
 from mujoco_warp._src import warp_util
 from mujoco_warp._src.warp_util import nested_kernel
 
-bleeding_edge_mujoco = tuple(map(int, mujoco.__version__.split(".")[:3])) >= (3, 4, 1)
-dist = importlib.metadata.distribution("mujoco")
-if "dev" in dist.version:
-  # breaking change: https://github.com/google-deepmind/mujoco/commit/a49576e469c8f42a496f5d71cd115f59b4762a06
-  bleeding_edge_mujoco = bleeding_edge_mujoco and int(dist.version.split("dev")[1]) >= 855718305
+
+def _is_mujoco_dev(package_name: str) -> bool:
```
btaba: since this is called `is_mujoco_dev` (which I think it should be for now), just hardcode the package_name?

mujoco_warp/_src/io.py
```diff
@@ -26,11 +27,23 @@
...
+
+  # fall back to metadata
+  dist_version = importlib.metadata.version(package_name)
+  if _DEV_VERSION_PATTERN.match(dist_version):
+    return True
+
+  return False
+
+
+bleeding_edge_mujoco = _is_mujoco_dev("mujoco")
```
btaba: style nit: globals are typically UPPER

mujoco_warp/_src/io.py
```diff
@@ -26,11 +27,23 @@
...
+
+  # fall back to metadata
+  dist_version = importlib.metadata.version("mujoco")
+  if _DEV_VERSION_PATTERN.match(dist_version):
+    return True
+
+  return False
+
+
+BLEEDING_EDGE_MUJOCO = _is_mujoco_dev()
```
btaba: super smol nit, "is_mujoco_dev" and "bleeding_edge" => better to pick one

mujoco_warp/_src/io.py
thowell: fixed

mujoco_warp/_src/io.py

PR #1057, opened 2026-01-19
Optimization: Minor shuffling in update_gradient_JTDAJ_dense_tiled

Kenny-Vilella:
Minor perf improvement:

On the humanoid:
Before:
main_humanoid

After:
optimized_humanoid

On the apollo:
Before:
main_apollo

After:
optimized_apollo


adenzler-nvidia (approved):
Thanks - nice catch! How did you come up with that one?

PR #1059, opened 2026-01-19
SolverContext

thowell:
## Overview

Refactored solver workspace memory management to allocate temporary arrays inline during solve rather than persisting them in `types.Data.efc`.

### 1. New `SolverContext` Class

```python
class SolverContext:
  def __init__(self, m: types.Model, d: types.Data, inverse: bool = False):
    # Allocates all temporary solver workspace arrays
    # inverse=True uses minimal allocation for inverse dynamics
```

**Encapsulated arrays:**
`Jaref`, `grad`, `grad_dot`, `Mgrad`, `search`, `search_dot`, `gauss`, `cost`, `prev_cost`, `mv`, `jv`, `quad`, `quad_gauss`, `alpha`, `prev_grad`, `prev_Mgrad`, `beta`, `done`, `h`, `hfactor`

---

### 2. Updated Solver Functions

All solver functions now accept `ctx: SolverContext`:

| Function | Old Signature | New Signature |
|----------|---------------|---------------|
| `create_context` | `(m, d, h, hfactor, grad)` | `(m, d, ctx, grad)` |
| `_solver_iteration` | `(m, d, h, hfactor, ...)` | `(m, d, ctx, ...)` |
| `_update_gradient` | `(m, d, h, hfactor)` | `(m, d, ctx)` |
| `_update_constraint` | `(m, d)` | `(m, d, ctx)` |
| `_linesearch` | `(m, d, cost)` | `(m, d, ctx, cost)` |

---

### 3. Removed Fields from `types.Constraint`

**Removed (17 fields):**
> `Jaref`, `grad`, `grad_dot`, `Mgrad`, `search`, `search_dot`, `gauss`, `cost`, `prev_cost`, `mv`, `jv`, `quad`, `quad_gauss`, `alpha`, `prev_grad`, `prev_Mgrad`, `beta`, `done`

**Kept (12 fields):**
> `type`, `id`, `J`, `pos`, `margin`, `D`, `vel`, `aref`, `frictionloss`, `force`, `state`, `Ma`

---

### 4. Updated Dependent Files

| File | Change |
|------|--------|
| `inverse.py` | Uses `SolverContext(m, d, inverse=True)` |
| `sensor.py` | Creates `mv` array inline in `energy_vel` |
| `support_test.py` | Uses inline arrays instead of `d.efc.*` |
| `solver_test.py` | Restored tests with `SolverContext` API |

---

## Benefits

1. **Memory efficiency**: Solver workspace freed after each solve call
2. **Cleaner API**: Temporary state encapsulated in `SolverContext`
3. **Reduced `types.Constraint` size**: 17 fewer persistent fields

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
   test-locked-deps:
     name: Test with locked uv dependencies
     # Tests against exact versions pinned in uv.lock
-    runs-on: ubuntu-22.04 # https://github.com/orgs/community/discussions/160680
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["3.10", "3.12"]
@@ -70,7 +70,7 @@ jobs:
     name: Test with stable PyPI dependencies
     # Tests against stable releases from pypi.org (not custom indexes like pypi.nvidia.com)
     # This ensures the library works for users installing dependencies from standard PyPI
-    runs-on: ubuntu-22.04 # https://github.com/orgs/community/discussions/160680
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["3.10", "3.12"]


mujoco_warp/_src/solver.py
```diff
@@ -34,6 +34,67 @@
 _BLOCK_CHOLESKY_DIM = 32
 
 
+class SolverContext:
```
erikfrey: I think I'd mildly prefer to make this a dataclass if we're going to do this.

The main reason is that dataclasses dissuade hanging logic off the class

mujoco_warp/_src/solver.py
```diff
@@ -34,6 +34,67 @@
...
+
+    # Arrays needed for both forward and inverse
+    self.Jaref = wp.zeros((nworld, njmax), dtype=float)
+    self.search_dot = wp.zeros((nworld,), dtype=float)
+    self.gauss = wp.zeros((nworld,), dtype=float)
+    self.cost = wp.zeros((nworld,), dtype=float)
+    self.prev_cost = wp.zeros((nworld,), dtype=float)
+    self.done = wp.zeros((nworld,), dtype=bool)
+
+    if inverse:
```
erikfrey: should we just make two context classes, one for inverse and one for everything else?

mujoco_warp/_src/solver.py
```diff
@@ -34,6 +34,67 @@
...
+      self.prev_Mgrad = wp.empty((nworld, 0), dtype=float)
+      self.beta = wp.empty((nworld,), dtype=float)
+      self.h = wp.empty((nworld, 0, 0), dtype=float)
+      self.hfactor = wp.empty((nworld, 0, 0), dtype=float)
+    else:
+      # Newton solver needs h; hfactor only needed if nv > _BLOCK_CHOLESKY_DIM
+      alloc_h = m.opt.solver == types.SolverType.NEWTON
+      alloc_hfactor = alloc_h and nv > _BLOCK_CHOLESKY_DIM
+
+      self.grad = wp.zeros((nworld, nv_pad), dtype=float)
```
erikfrey: can you double check that these all need to be zeros vs empty?  empty is cheaper.

mujoco_warp/_src/solver.py
erikfrey: some kernel analyzer stuff to clean up

erikfrey (approved):
Please report some benchmark deltas so we can confirm this isn't slowing stuff down.

Should we do this with the temp arrays in `collision_convex.py` too?  If we're going to do this context pattern I'd like to be consistent in the codebase.

Overall I think this is OK, just avoid the temptation to hang logic off these context classes if possible.

LGTM given the comments above and the code comments inline getting resolved.

mujoco_warp/_src/solver.py

mujoco_warp/_src/solver.py
thowell: context arrays are now created with `wp.empty`

mujoco_warp/_src/solver.py
thowell: there are now 2 dataclasses: `SolverContext` and `InverseContext`

mujoco_warp/_src/solver.py

thowell:
`mjwarp-testspeed benchmarks/humanoid/humanoid.xml --nworld=8192 --nconmax=24 --njmax=64`

this pr
```

Model
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: False
  ls_parallel: False
Data

Total JIT time: 0.29 s
Total simulation time: 2.07 s
Total steps per second: 3,950,532
Total realtime factor: 19,752.66 x
Total time per step: 253.13 ns
```

main 43ac254f99a909aecb3a423ed1c719895cbd5d47
```

Model
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: False
  ls_parallel: False
Data

Total JIT time: 6.88 s
Total simulation time: 2.12 s
Total steps per second: 3,862,773
Total realtime factor: 19,313.87 x
Total time per step: 258.88 ns
```

mujoco_warp/_src/solver.py
thowell: `h` and `hfactor` should be initialized as zeros to follow the existing pattern

PR #1064, opened 2026-01-19
Render Cleanup 4

StafaH:
More render code cleanup:
- Fixes failing tests due to new cam_fovy change in _sensor_pos kernel
- Moves bvh build for hfield into bvh.py
- Fixes ordering of types.py camera related fields to pass ordering test

### DIFF
--- a/mujoco_warp/_src/bvh.py
+++ b/mujoco_warp/_src/bvh.py
@@ -290,7 +290,7 @@ def build_mesh_bvh(
   constructor: str = "sah",
   leaf_size: int = 2,
 ) -> tuple[wp.Mesh, wp.vec3]:
-  """Create a Warp mesh BVH from mjcf mesh data."""
+  """Create a Warp mesh BVH from mesh data."""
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/bvh.py
```diff
@@ -310,6 +310,178 @@ def build_mesh_bvh(
...
+  return np.array(points_list, dtype=np.float32), np.array(indices_list, dtype=np.int32)
+
+
+def build_hfield_bvh(
+  mjm: mujoco.MjModel,
+  hfieldid: int,
+  constructor: str = "sah",
+  leaf_size: int = 2,
+) -> tuple[wp.Mesh, wp.vec3]:
+  """Create a Warp mesh BVH from mjcf heightfield data."""
```
thowell: we can remove 'mjcf'

StafaH: Removed for this and the mesh docstring

PR #1067, opened 2026-01-20
remove collision arrays from `Data` and utilize `CollisionContext` inline

thowell:


mujoco_warp/_src/broadphase_test.py
```diff
@@ -26,13 +26,18 @@
 from mujoco_warp import DisableBit
 from mujoco_warp import test_data
 from mujoco_warp._src import collision_driver
+from mujoco_warp._src.collision_driver import create_collision_context
```
erikfrey: style nit: since you already imported `collision_driver` right above, and you only reference `create_collision_context` in one place, I'd just use the module reference and delete this extra import.

mujoco_warp/_src/collision_primitive.py
erikfrey, line 52: not blocking, suggestion for a future PR:

might be time for a `collision_base.py` or `collision_core.py` where we move this stuff that really shouldn't be in `collision_primitive` like `Geom`, `contact_params`, `write_contact` etc.

Once we do that we could put `CollisionContext` there too

thowell: created an issue to track a refactor for collision https://github.com/google-deepmind/mujoco_warp/issues/1124

mujoco_warp/_src/broadphase_test.py
thowell: done

PR #1069, opened 2026-01-21
Replace face and edge tolerance with trig formula

kbayes:
Shouldn't change any behavior as the new values are:

FACE_TOL = 0.999998720000273
EDGE_TOL = 0.0015999993173334208

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -37,12 +37,12 @@ To run specific benchmarks, use the `-f` or `--filter` option with a regex patte
 ./run.sh -f cloth
 ```
 
-### Clearing the Kernel Cache
+### Clearing the Warp Cache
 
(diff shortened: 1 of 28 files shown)


thowell:
@adenzler-nvidia should we wrap these constants with `wp.constant` like the constants below? https://github.com/google-deepmind/mujoco_warp/blob/40ea9a78bf73c5333e21390c976b901eb9a47d25/mujoco_warp/_src/collision_gjk.py#L43

adenzler-nvidia:
If you don't see any compiler warnings, then it's fine to leave it as-is. Moving those other constants is really just a workaround for an issue in warp that arises as a combination of the python and c++ type systems for type promotions. For floats, leaving as-is should be fine.

PR #1074, opened 2026-01-22
Minor refactor of CCD

kbayes:


mujoco_warp/_src/collision_gjk.py
```diff
@@ -293,28 +293,25 @@ def _project_origin_plane(v1: wp.vec3, v2: wp.vec3, v3: wp.vec3) -> Tuple[wp.vec
...
-    return v, 0
+  if nv != 0.0 and nn > MINVAL:
+    return (nv / nn) * n, 0
 
   # n = (v2 - v1) x (v3 - v1)
   n = wp.cross(diff21, diff31)
   nv = wp.dot(n, v1)
   nn = wp.dot(n, n)
-  if nn == 0:
+  if nn == 0.0:
```
thowell, line 305: should we check for a small number here instead of exactly zero?

mujoco_warp/_src/collision_gjk.py
```diff
@@ -293,28 +293,25 @@ def _project_origin_plane(v1: wp.vec3, v2: wp.vec3, v3: wp.vec3) -> Tuple[wp.vec
...
+  if nv != 0.0 and nn > MINVAL:
+    return (nv / nn) * n, 0
 
   # n = (v1 - v3) x (v2 - v3)
   n = wp.cross(diff31, diff32)
   nv = wp.dot(n, v3)
   nn = wp.dot(n, n)
-  v = (nv / nn) * n
-  return v, 0
+  return (nv / nn) * n, 0
```
thowell, line 314: we don't check `nn != 0.0`, should we have a safe division?

thowell (approved):

mujoco_warp/_src/collision_gjk.py
kbayes: There's a different check down below

mujoco_warp/_src/collision_gjk.py
kbayes: I think if nn == 0.0 it'll be caught about. Leaving the behavior as-is for now.

PR #1077, opened 2026-01-23
Optimization: sparse mul_m

adenzler-nvidia:
This seems to help as well. We precompute a CSR layout and perform single-thread accumulation in mul_m.

numbers: (note that the flex benchmarks are heavily dominated by flex time, but the event_trace shows benefits).

| Benchmark | Step (main) | Step (this PR) | Step Δ | mul_m (main) | mul_m (this PR) | mul_m Δ |
|-----------|-------------|----------------|--------|--------------|-----------------|---------|
| aloha_cloth | 1301.98 µs | 1304.82 µs | +0.2% | 80.12 ns | 63.77 ns | **-20%** |
| cloth | 1378.18 µs | 1377.09 µs | -0.1% | 81.17 ns | 65.15 ns | **-20%** |
| three_humanoids | 1544.95 ns | 1527.84 ns | -1.1% | 6.74 ns | 3.75 ns | **-44%** |
| g1_hands | 1713.63 ns | 1673.05 ns | -2.4% | 6.01 ns | 3.54 ns | **-41%** |

We probably need better benchmark coverage in the 30-60dof range. I added numbers for a G1 with hands from the menagerie to cover that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-    types: [ opened, reopened, synchronize, edited ]
+    types: [opened, reopened, synchronize, edited]
 
 jobs:
   format:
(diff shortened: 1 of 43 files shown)


adenzler-nvidia:
I also readded module=unique for the dense kernel as well, because I'm not 100% sure about the interaction with cache_kernel. I think we should stay on the safe side while I figure this out separately.

PR #1079, opened 2026-01-23
Remove nested_kernel

adenzler-nvidia:
We already didn't really rely on it by passing module=unique, because it wasn't really working with kernel factories.

This change just removes everything and moves to plain old wp.kernel with module=unique.

The kernel analyzer now checks for module=unique in factory patterns.


adenzler-nvidia:
I had to fix a few issues in kernel analyzer to make the checks for nesting work. And then it turned out there were a bunch of kernels that weren't checked at all, so I fixed those as well. Sorry for all of this spiralling out of control.

@thowell - I realize this is conflicting a lot with your solverContext and CollisionContext changes. How do we want to sequence those?

thowell:
no problem, lets move forward with the changes in this pr without consideration for the solvercontext and collisioncontext prs

PR #1081, opened 2026-01-23
Render Cleanup - Ray

StafaH:
This PR cleans up the following:
- Allow for BVH accelerated `ray` and `rays` usage by passing in a `RenderContext`
- Cleanup of imports to use absolute imports, matching rest of repo
- Move BVH build into public API (fixes circular import issues, and makes it possible for downstream users to build BVH independently from the render context)
- Add bvh ray tests

### DIFF
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -29,12 +29,7 @@ from mujoco_warp._src.types import Model as Model
 from mujoco_warp._src.types import Data as Data
 # isort: on
 
-from mujoco_warp._src.bvh import build_scene_bvh as build_scene_bvh
-from mujoco_warp._src.bvh import refit_scene_bvh as refit_scene_bvh
-from mujoco_warp._src.bvh import build_mesh_bvh as build_mesh_bvh
-from mujoco_warp._src.bvh import build_hfield_bvh as build_hfield_bvh
-from mujoco_warp._src.bvh import build_flex_bvh as build_flex_bvh
-from mujoco_warp._src.bvh import refit_flex_bvh as refit_flex_bvh
+from mujoco_warp._src.bvh import refit_bvh as refit_bvh
 from mujoco_warp._src.collision_driver import collision as collision
 from mujoco_warp._src.collision_driver import nxn_broadphase as nxn_broadphase
 from mujoco_warp._src.collision_driver import sap_broadphase as sap_broadphase
--- a/mujoco_warp/_src/bvh.py
+++ b/mujoco_warp/_src/bvh.py
@@ -13,18 +13,31 @@
 # limitations under the License.
 # ==============================================================================
 
-from typing import Tuple
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Tuple
 
 import mujoco
 import numpy as np
 import warp as wp
 
-from mujoco_warp._src.render_context import RenderContext
 from mujoco_warp._src.types import Data
 from mujoco_warp._src.types import GeomType
 from mujoco_warp._src.types import Model
+from mujoco_warp._src.warp_util import event_scope
+
+if TYPE_CHECKING:
+  from mujoco_warp._src.render_context import RenderContext
 
 wp.set_module_options({"enable_backward": False})
+
+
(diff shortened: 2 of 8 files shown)


mujoco_warp/__init__.py
```diff
@@ -29,6 +29,12 @@
 from mujoco_warp._src.types import Data as Data
 # isort: on
 
+from mujoco_warp._src.bvh import build_scene_bvh as build_scene_bvh
```
thowell: @erikfrey @btaba what do we want to be part of the public api?

mujoco_warp/_src/ray.py
```diff
@@ -1050,41 +1208,74 @@ def rays(
...
-      geomgroup,
-      flg_static,
-      bodyexclude,
-      dist,
-      geomid,
-      normal,
-    ],
-    block_dim=m.block_dim.ray,
-  )
+  if rc is None:
```
thowell, line 1217: please add a todo to investigate creating a `RenderContext` if rc is None and then completely removing the current `_ray` code path

mujoco_warp/__init__.py
btaba: Why not expose `bvh`? Most of the other functions here seem to have an "equivalent in mujoco", so if we want to roughly keep that property, I propose `bvh`. Of course this is merely an opinion

kevinzakka: The bvh function is useful for downstream sensors that need to create or refit the BVH (e.g., [raycast sensor](https://github.com/mujocolab/mjlab/blob/bvh-raycast/src/mjlab/sensor/raycast_sensor.py#L697-L717) in mjlab).

mujoco_warp/_src/render_context.py
```diff
@@ -20,11 +20,11 @@
 import numpy as np
 import warp as wp
 
-from . import bvh
-from .types import Data
-from .types import GeomType
-from .types import Model
-from .types import ProjectionType
+import mujoco_warp as mjw
```
erikfrey: Let's use internal imports for internal code.  So this could be:

`from mujoco_warp._src import bvh`

mujoco_warp/_src/ray.py
erikfrey: Big +1, let's avoid keeping around old implementations wherever possible.

mujoco_warp/__init__.py
erikfrey: I don't have a great intuition for which parts of the render code should be offered up in the public API, and at what granularity.

For example, if we are to expose both a `refit_flex_bvh` and a `refit_scene_bvh` then presumably there's some reason the user wants to refit just flexes?  vs. a `refit_bvh` public function that just does it all?

Basically let's use making changes to the API in `__init__.py` as an invitation to discuss what the expected use cases are, maybe even with a few concrete user stories.  Can discuss this here or in chat or in a doc, whatever you prefer @StafaH 

In general we should be conservative about what we add to the API, the mistake of under-adding is much less painful to fix than the mistake of over-adding.

StafaH: Reduced to be just refit_bvh now

mujoco_warp/_src/bvh.py
```diff
@@ -13,20 +13,33 @@
...
-from .render_context import RenderContext
-from .types import Data
-from .types import GeomType
-from .types import Model
+from mujoco_warp._src.types import Data
+from mujoco_warp._src.types import GeomType
+from mujoco_warp._src.types import Model
+from mujoco_warp._src.warp_util import event_scope
+
+if TYPE_CHECKING:
```
thowell, line 29: is this check necessary? maybe add a brief comment

thowell:
left a question, otherwise lgtm. thanks @StafaH!

mujoco_warp/_src/bvh.py
StafaH: It stops a circular import between the two.

I think if we did a refactor to move render context into types and the init into the create method that already exists that would be cleaner and also resolve this

btaba (approved):

PR #1084, opened 2026-01-28
Refactor GitHub workflows to test stable, lockfile, and nightly dependencies

shi-eric:
- Add a `nightly.yml` workflow that tests against the latest nightly versions (using the `if-necessary-or-explicit` pre-release strategy of `uv lock -U`. This is for early warning detection of future dependency issues.
- Refactor `ci.yml` to use best-practices for `uv` usage and have two key testing jobs:
  - `test-locked-deps` runs the mjwarp tests against the dependencies pinned in the `uv.lock` fle
  - `test-stable-deps` runs the mjwarp tests against the stable versions of dependencies installed from pypi.org (no custom indices or pre-release versions published on pypi.org) 
- Update `pre-commit` versions of existing hooks and add the `uv-lock` hook, which ensures that the `uv.lock` file is up-to-date

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.2 # Use the latest Ruff version
+    rev: v0.14.14 # Use the latest Ruff version
     hooks:
       - id: ruff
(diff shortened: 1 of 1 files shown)


shi-eric:
@erikfrey: The branch protection rules will need to be updated to use the new names of the jobs

erikfrey (approved):
Really clean, thanks @shi-eric

PR #1089, opened 2026-01-29
Add kernel warning system for overflow detection

adenzler-nvidia:
## Summary

Adds a structured warning system for detecting and reporting overflow conditions in kernels. Previously, `wp.printf` calls went to stdout and cluttered output. Now warnings are captured via atomic flags and emitted to stderr with actionable messages.

### New Warning System

- `WarningType` enum for categorizing warnings: `NEFC_OVERFLOW`, `BROADPHASE_OVERFLOW`, `NARROWPHASE_OVERFLOW`, `CONTACT_MATCH_OVERFLOW`, `GJK_ITERATIONS`, `EPA_HORIZON`, `HFIELD_OVERFLOW`
- `check_warnings(d)` - reads flags from GPU and emits to stderr with suggested values
- `get_warnings(d)` / `clear_warnings(d)` - utility functions for programmatic access
- Warning arrays added to `Data`: `warning` (flags) and `warning_info` (suggested values)

### Kernel Changes

- All overflow `wp.printf` calls are now guarded by `wp.static(m.opt.warning_printf)` for compile-time removal
- Kernels set warning flags via `wp.atomic_max` with suggested parameter values
- Affected: `forward.py`, `sensor.py`, `collision_convex.py`, `collision_gjk.py`

### New Option

- `opt.warning_printf` (default: `True`) - controls whether printf warnings are compiled into kernels
- `testspeed` and `viewer` disable this and use `check_warnings()` instead

### Cleanup

- Removed dead `wp.printf` calls in `smooth.py` (unreachable due to `io.py` validation)

## API

Warnings are emitted per-step to stderr:

```
Warning: nefc overflow - increase njmax to 32
Warning: nefc overflow - increase njmax to 36
```

Programmatic access:

```python
m.opt.warning_printf = False  # disable printf, use check_warnings instead
mjw.step(m, d)
mjw.check_warnings(d)
```

## Testing

New `warning_test.py` with 8 tests covering initialization, detection, clearing, graph capture scenarios, and the printf option.


erikfrey:
Nice!

I wouldn't be against defaulting `warning_printf` to false if we want to discourage this for general use.  WDYT @adenzler-nvidia ?

adenzler-nvidia:
I'm torn on that one - I think before we had the printfs we had a lot of silent overflows. Any user not opting in to this warning system would be in that situation again.. So I think it's safer to default to True and give everyone using this warning handler or implementing it themselves the option to turn off the printfs.

thowell:
the overflow warnings are very helpful for debugging and my vote would be to keep them on by default. we could note in the  the [performance](https://mujoco.readthedocs.io/en/latest/mjwarp/index.html#performance-tuning) section of the documentation to set this option field to `False` for best performance?

thowell:
@adenzler-nvidia is it possible for us to utilize this (in combination with some sort of option) to error / stop simulation when a warning is detected?

adenzler-nvidia:
The error handler can definitely be expended to do something like this. I would say it's up to the application integrating MjWarp, I don't think this should be hardcoded into MjWarp.

I can see an app like MjLab implementing their own error handler to do this. We can do it for the renderer/testspeed as well.

This is also the reason I vote for keeping the printfs in by default - it's a good thing for anyone integrating MjWarp. Once you need more than that, you can turn off the prints and implement an error handler.

mujoco_warp/_src/warning.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2025 The Newton Developers
```
erikfrey: nit: 2026

mujoco_warp/_src/types.py
```diff
@@ -690,6 +715,9 @@ class Option:
       zeros out the contacts at each step)
     contact_sensor_maxmatch: max number of contacts considered by contact sensor matching criteria
                              contacts matched after this value is exceded will be ignored
+    warning_printf: if True, emit overflow warnings via printf to stdout in addition to setting
```
erikfrey: Hmm, but we're sort of leaving this up to the mjwarp user, right?

erikfrey (commented):
@adenzler-nvidia and I discussed this PR offline, main takeaways:

* this is an awesome feature that I would love to get in
* it feels more like this config belongs at the package level instead of as a model option (think sort of like `wp.config`)
* I'm wondering if we can use `wp.static` to somehow do bookkeeping *during compilation* so that we can avoid having to maintain an explicit WarningType enum in `types.py` - instead, a user can call `get_warnings` at any point on the host side and it will fetch all the formatted warnings that were emitted since the last time it was cleared

It's very possible that I'm asking for too much          :-)         but @adenzler-nvidia was intrigued by this and we agreed we'd play with these ideas before committing.

adenzler-nvidia:
I will probably not get to this in the next few days, but will try to take this to the warp module level, it seems like a good idea.

adenzler-nvidia:
Closing — we're pursuing a Warp-native approach to kernel logging instead (see warp MR !2184). The overflow detection functionality will be revisited once the Warp-side logging primitives land.

PR #1094, opened 2026-01-30
Update uv.lock to get warp-lang 1.12 pre-release.

erikfrey:
Ran the following:

```
UV_NO_CONFIG=1 uv lock -P warp-lang
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
   test-locked-deps:
     name: Test with locked uv dependencies
     # Tests against exact versions pinned in uv.lock
-    runs-on: ubuntu-22.04 # https://github.com/orgs/community/discussions/160680
+    runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


adenzler-nvidia:
This is good if you want to update all of these dependencies - otherwise a `uv lock warp-lang --prerelease allow` would suffice.

shi-eric:
I'm also second-guessing my suggestion of `--prerelease allow` since `uv` adds this to the `uv.lock` file in a wholesale manner, allowing any package to be upgraded to a prerelease version:

```
[options]
prerelease-mode = "allow"
```

Since MJWarp still has `"warp-lang>=1.11.0.dev0"` in their lockfile, a plain `uv lock -P warp-lang` should do the trick:

```
uv lock -P warp-lang
Resolved 61 packages in 1.57s
Updated warp-lang v1.11.0 -> v1.12.0.dev20260130
```

erikfrey:
OK that worked - @shi-eric  PTAL

PR #1095, opened 2026-01-30
Exclude flex/floppy.xml from nworld scaling test to reduce memory usage

erikfrey:
The test_put_data_nworld_array test was consuming ~14GB of memory when testing flex/floppy.xml with nworld=133. This was causing CI failures due to OOM on GitHub Actions runners (which have ~7GB RAM).

Reduced nworld to 7, which still tests batching but lowers memory usage for io_jax_test.py: 14GB -> 849MB

### DIFF
--- a/mujoco_warp/_src/io_jax_test.py
+++ b/mujoco_warp/_src/io_jax_test.py
@@ -39,14 +39,6 @@ _IO_TEST_MODELS = (
   "hfield/hfield.xml",
 )
 
-# Exclude flex/floppy.xml from nworld scaling test - it uses ~13GB with nworld=133
-_IO_TEST_MODELS_NWORLD = (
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/io_jax_test.py
```diff
@@ -39,6 +39,14 @@
   "hfield/hfield.xml",
 )
 
+# Exclude flex/floppy.xml from nworld scaling test - it uses ~13GB with nworld=133
+_IO_TEST_MODELS_NWORLD = (
```
thowell: maybe add a todo to add flex/floppy.xml to this list or utilize `_IO_TEST_MODELS` with `test_put_data_nworld_array` once we land sparsity #88?

PR #1097, opened 2026-01-30
SolverContext v2

thowell:
originally #1059 (seems the pr cannot be reopened?)

`mjwarp-testspeed benchmarks/humanoid/humanoid.xml --nconmax=24 --njmax=64`

this pr
```

Model
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: False
  ls_parallel: False
Data

Total JIT time: 7.01 s
Total simulation time: 2.07 s
Total steps per second: 3,961,963
Total realtime factor: 19,809.82 x
Total time per step: 252.40 ns
```

main 2e1ee3d767f860ddf9f37ba423859d28ce807d0f
```

Model
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: False
  ls_parallel: False
Data

Total JIT time: 0.33 s
Total simulation time: 2.11 s
Total steps per second: 3,875,279
Total realtime factor: 19,376.39 x
Total time per step: 258.05 ns
```


mujoco_warp/_src/solver.py
```diff
@@ -15,6 +15,7 @@
 
 from math import ceil
 from math import sqrt
+from typing import Union
```
erikfrey: nit: now that we're officially >= 3.10 I think it's okay to use pipe: " | " for type annotations

mujoco_warp/_src/solver.py
```diff
@@ -24,6 +25,8 @@
 from mujoco_warp._src import types
 from mujoco_warp._src.block_cholesky import create_blocked_cholesky_func
 from mujoco_warp._src.block_cholesky import create_blocked_cholesky_solve_func
+from mujoco_warp._src.types import InverseContext
```
erikfrey: nit: I could be easily convinced that these two classes should still be declared here since they are never used outside this file

mujoco_warp/_src/solver.py
```diff
@@ -2581,23 +2666,21 @@ def _solver_iteration(
...
     ],
-    outputs=[d.solver_niter, d.efc.done, nsolving],
+    outputs=[d.solver_niter, nsolving, ctx.done],
   )
 
 
-def create_context(
-  m: types.Model, d: types.Data, h: wp.array3d(dtype=float), hfactor: wp.array3d(dtype=float), grad: bool = True
-):
+def create_context(m: types.Model, d: types.Data, ctx: Union[SolverContext, InverseContext], grad: bool = True):
```
erikfrey: nit: we have some naming ambiguity here as we now have:

`solver.create_solver_context`

and

`solver.create_context`

and you are passing the context you create in the first to the context you create in the second.  so it reads like you need to create a context to create a context?

any way we can clean up the naming?

fwiw some of this logic is now called `mj_solPrimal` in C MuJoCo - not sure if we want to crib some naming from that.

mujoco_warp/_src/solver.py

mujoco_warp/_src/solver.py
thowell: the dataclasses are now declared in solver.py

mujoco_warp/_src/solver.py
thowell: the mujoco c functions for [creating a context](https://github.com/google-deepmind/mujoco/blob/7418d86eda3a58994230097c0d10351c7d56d4bb/src/engine/engine_solver.c#L1818) are: `PrimalPointers` and `PrimalAllocate` and there isn't a function that performs the same computation as (current) `solver.create_context`, the mujoco c computation is done directly in `mj_solPrimal`.

`solver.create_context` has been renamed to `solver.init_context` to hopefully avoid confusion with the functions `solver.create_solver_context` / `solver.create_inverse_context` that create a dataclass instance.

PR #1106, opened 2026-02-01
Add early CUDA availability check to mjwarp-viewer

ritikkumar1212:
mjwarp-viewer currently prints "CPU-only mode" and later crashes with
`RuntimeError: Must be a CUDA device`.

This PR adds an early CUDA availability check after wp.init(), so the viewer
fails fast with a clear, actionable error message on machines without NVIDIA GPUs.

This improves UX for CPU-only users and avoids a deep Warp stacktrace.

### DIFF
--- a/mujoco_warp/viewer.py
+++ b/mujoco_warp/viewer.py
@@ -135,6 +135,16 @@ def _main(argv: Sequence[str]) -> None:
   else:
     wp.config.quiet = flags.FLAGS["verbosity"].value < 1
     wp.init()
+    # ---- CUDA guard ----
+    if not wp.is_cuda_available():
+      raise RuntimeError(
+          "\n mjwarp-viewer requires an NVIDIA CUDA device.\n"
+          "No CUDA GPU was detected.\n\n"
+          "Possible fixes:\n"
+          "  • Install NVIDIA drivers + CUDA\n"
+          "  • Run on a GPU machine\n"
+          "  • Use standard MuJoCo viewer instead\n"
+      )
     if _CLEAR_WARP_CACHE.value:
       wp.clear_kernel_cache()
       wp.clear_lto_cache()


hartikainen:
I think it would be good to support non-CUDA devices. `warp-lang` [README](https://github.com/NVIDIA/warp) says:
> Warp can run on x86-64 and ARMv8 CPUs on Windows, Linux, and macOS.

I'm pretty close to getting this to work with the MJX frontend in https://github.com/google-deepmind/mujoco/pull/2948. I'm pretty sure that warp would work pretty much out-of-the box.

ritikkumar1212:
Thanks for the feedback!

My understanding was that wp.ScopedCapture() currently requires CUDA, which is why I added the early guard to avoid the late runtime crash.
I’m happy to adjust this to allow CPU execution if we can bypass graph capture on non-CUDA devices (or fall back to the C engine / non-captured Warp path).
Could you suggest the preferred behavior for CPU? For example:
skip ScopedCapture and call mjw.step directly?
or auto-switch to EngineOptions.C?

Happy to update the PR accordingly.

hartikainen:
I'm no expert on this, but the `warp-lang=1.11.0` [Scoped Devices documentation](https://nvidia.github.io/warp/user_guide/devices.html#scoped-devices) at least provides an example for `ScopedDevice("cpu")`:

> ```python
> with wp.ScopedDevice("cpu"):
>     # alloc and launch on "cpu"
>     a = wp.zeros(n)
>     wp.launch(kernel, dim=a.size, inputs=[a])
> ```

thowell:
@ritikkumar1212 thank you for contributing to mujoco warp!

#1107 should enable the viewer to be compatible with `device=cpu`

thowell:
with #1107 merged the viewer should now work on cpu with `--device=cpu`

ritikkumar1212:
That's great

PR #1107, opened 2026-02-02
viewer: `--device=cpu`

thowell:
make viewer compatible with `device=cpu`

`mjwarp-viewer benchmarks/humanoid/humanoid.xml --nconmax=48 --njmax=128 --device=cpu`

### DIFF
--- a/mujoco_warp/viewer.py
+++ b/mujoco_warp/viewer.py
@@ -191,11 +191,7 @@ def _main(argv: Sequence[str]) -> None:
         if mjm.opt != opt:
           opt = copy.copy(mjm.opt)
           m = mjw.put_model(mjm)
-          if graph:
-            with wp.ScopedCapture(_DEVICE.value) as capture:
-              mjw.step(m, d)
-            graph = capture.graph
-
+          graph = _compile_step(m, d) if wp.get_device().is_cuda else None
         if _VIEWER_GLOBAL_STATE["running"] or _VIEWER_GLOBAL_STATE["step_once"]:
           _VIEWER_GLOBAL_STATE["step_once"] = False
           if graph is None:


mujoco_warp/viewer.py
```diff
@@ -176,29 +182,32 @@ def _main(argv: Sequence[str]) -> None:
...
+          wp.copy(d.qvel, wp.array([mjd.qvel.astype(np.float32)]))
+          wp.copy(d.time, wp.array([mjd.time], dtype=wp.float32))
+
+          # if the user changed an option in the MuJoCo Simulate UI, go ahead and recompile the step
+          # TODO: update memory tied to option max iterations
+          if mjm.opt != opt:
+            opt = copy.copy(mjm.opt)
+            m = mjw.put_model(mjm)
+            if is_cuda:
+              with wp.ScopedCapture() as capture:
```
erikfrey: scopedcapture has a device param - could you do that instead of adding the ScopedDevice above?

mujoco_warp/viewer.py
```diff
@@ -145,6 +145,7 @@ def _main(argv: Sequence[str]) -> None:
         compute_cache.mkdir()
 
     with wp.ScopedDevice(_DEVICE.value):
+      is_cuda = wp.get_device().is_cuda
```
erikfrey: nit: we can use the graph itself as a sentinel, so this could be something like:

```py
graph = _compile_step(m, d) if wp.get_device().is_cuda else None
```

then further down

```py
if graph is None:
  print("Running Warp unoptimized on CPU.")
  mjw.step(m, d)  # warmup

...
if _VIEWER_GLOBAL_STATE["running"] or _VIEWER_GLOBAL_STATE["step_once"]:
  _VIEWER_GLOBAL_STATE["step_once"] = False
  if graph is None:
    mjw.step(m, d)
  else:
    wp.capture_launch(graph)
    wp.synchronize()
```

something like that?

thowell: done

mujoco_warp/viewer.py
thowell: refactored the code a bit to limit `ScopedDevice` usage.

instances like `mjw.put_model(mjm)` require `ScopedDevice` to put the `Model` arrays on the correct device. additionally, seems that `mjw.step` (no graph capture) requires `ScopedDevice` too, probably since arrays are allocated inline and they need to know the correct device?

thowell:
refactored the code a bit. `ScopedDevice` usage should now be more targeted.

mujoco_warp/viewer.py
```diff
@@ -148,21 +148,24 @@ def _main(argv: Sequence[str]) -> None:
...
-        f"  nbody: {m.nbody} nv: {m.nv} ngeom: {m.ngeom} nu: {m.nu} is_sparse: {m.is_sparse}\n"
-        f"  broadphase: {broadphase} broadphase_filter: {filter}\n"
-        f"  solver: {solver} cone: {cone} iterations: {iterations} {ls_str}\n"
-        f"  integrator: {integrator} graph_conditional: {m.opt.graph_conditional}"
-      )
       d = mjw.put_data(mjm, mjd, nconmax=_NCONMAX.value, njmax=_NJMAX.value)
-      print(f"Data\n  nworld: {d.nworld} nconmax: {int(d.naconmax / d.nworld)} njmax: {d.njmax}\n")
-      graph = _compile_step(m, d)
-      print(f"MuJoCo Warp simulating with dt = {m.opt.timestep.numpy()[0]:.3f}...")
+      graph = _compile_step(m, d) if wp.get_device().is_cuda else None
```
erikfrey: last nit, what if we do:

```py
wp.set_device(_DEVICE)
```

after `wp.init()` to save some of this extra `ScopedDevice` stuff?

thowell: done

PR #1111, opened 2026-02-03
`Option.is_sparse` -> `Model.is_sparse` and `Option.has_fluid` -> `Model.has_fluid`

thowell:
the `Option` fields `is_sparse` and `has_fluid` currently do not act as "options" since changing their values does not necessary modify the simulation as expected. these fields are "flags" determined by an `mjModel` instance and they indicate certain properties/code paths for simulation of the `Model` instance created by `io.put_model`. this pr moves these fields from `Option` to `Model`.

additionally, `io.override_model` is updated to accept `opt.jacobian`:

```
mjwarp-testspeed benchmarks/humanoid/humanoid.xml --nconmax=24 --njmax=64 -o "opt.jacobian=sparse"
```

```

Model
Option
  integrator: EULER
  cone: PYRAMIDAL
  is_sparse: True
  ls_parallel: False
Data
```

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -680,10 +680,8 @@ class Option:
 
   warp only fields:
     impratio_invsqrt: ratio of friction-to-normal contact impedance (stored as inverse square root)
-    is_sparse: whether to use sparse representations
     ls_parallel: evaluate engine solver step sizes in parallel
     ls_parallel_min_step: minimum step size for solver linesearch
-    has_fluid: True if wind, density, or viscosity are non-zero at put_model time
     broadphase: broadphase type (BroadphaseType)
     broadphase_filter: broadphase filter bitflag (BroadphaseFilter)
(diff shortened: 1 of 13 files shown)


mujoco_warp/_src/types.py
```diff
@@ -1115,6 +1111,8 @@ class Model:
     qM_mulm_rowadr: sparse matmul row pointers
     qM_mulm_col: sparse matmul column indices
     qM_mulm_madr: sparse matmul matrix addresses
+    is_sparse: whether to use sparse representations
```
erikfrey: nit: consider grouping these together with `has_sdf_geom` above

erikfrey (approved):
LGTM, just one nit

but heads up, once we cut our first pypi build for "release", such changes may become harder, as we'll try to keep non-major version changes backwards compatible

mujoco_warp/_src/types.py

PR #1113, opened 2026-02-03
mjwarp-render

StafaH:
# Summary                                                                                                        
                                                                             
This PR introduces a GPU-accelerated ray tracing renderer for MuJoCo Warp, enabling parallel rendering of RGB  
and depth images across thousands of simulation worlds.           

# Key Features

### Mesh Rendering with Textures

The renderer uses BVH accelerated mesh rendering with full texture support. Texture mapping includes CUDA texture support in provided through `warp==1.12` with bilinear filtering

539952308-fd61f463-3960-4fc9-82a0-a7275fac7335  render_preview_Franka Panda Visual_cam0

### Heightfield Rendering

Heighfield rendering from the MuJoCo heightfield data is possible. Rendering is optimized for high throughput through mesh optimizations during initialization.

render_preview_Apptronik Heightfield_cam0

### Flex Rendering

The new renderer includes an initial prototype for flex rendering. Currently only 2D and 3D flex objects are supported. Flex rendering is still a WIP and performance and features will continue to improve.

### Lighting and Shadows

Rendering supports dynamic lighting with configurable shadows, all of which can be domain randomized from the `Model` fields. Here is an example with two lights.

debug

### Additional features include:
- Multi-camera support with per-camera resolution, FOV, and intrinsics
- Batched rendering across parallel worlds with Domain Randomization for sim2real robotics.
- BVH accelerated ray/rays API for custom raycast sensors

# Public API

To work with rendering, a few additions have been made to the public API.

All BVH accelerated rendering or raycasting requires a RenderContext. RenderContext hold the BVH structures as well as rendering specific model fields, and output buffers.

```
rc = mjw.create_render_context(
    mjm, m, d,
    cam_res=(256, 256),           # Override camera resolution (or per-camera list)
    render_rgb=True,              # Enable RGB output (or per-camera list)
    render_depth=True,            # Enable depth output (or per-camera list)
    use_textures=True,            # Apply material textures
    use_shadows=False,            # Enable shadow casting (slower)
    enabled_geom_groups=[0, 1],   # Only render geoms in groups 0 and 1
    cam_active=[True, False],     # Selectively enable/disable cameras
    flex_render_smooth=True,      # Smooth shading for soft bodies
)
```

In the render context, you can customize what you want each camera to do. Each setting can be applied globally or per-camera. The render context also has the ability to read from the new MuJoCo spec that allows for camera customization:

```
<camera name="front_camera" pos="3 0 2" xyaxes="0 1 0 -0.6 0 0.8" resolution="64 64" output="rgb depth"/>
```

To render all cameras, users must first call refit to update the BVH trees, and then call render which will render all cameras that are meant to be rendered into the output buffers.

```
mjw.refit_bvh(m, d, rc)
mjw.render(m, d, rc)
```

The result can be accessed through output buffers. The output buffers are linear of shape `(nworld, pixels)`. We provide `rgb_adr` and `depth_adr` for users to correctly access camera data. RGB data is packed into a `uint32` and need to be unpacked for downstream use. Here is an example:

```
world, cam = 0, 0
width, height = rc.cam_res.numpy()[cam]
rgb_adr = rc.rgb_adr.numpy()[cam]
depth_adr = rc.depth_adr.numpy()[cam]

# Extract RGB (packed as uint32: 0xAARRGGBB)
rgb_packed = rc.rgb_data.numpy()[world, rgb_adr : rgb_adr + width * height]
rgb_packed = rgb_packed.reshape(height, width).astype(np.uint32)
rgb_image = np.dstack([
    ((rgb_packed >> 16) & 0xFF).astype(np.uint8),  # R
    ((rgb_packed >> 8) & 0xFF).astype(np.uint8),   # G
    (rgb_packed & 0xFF).astype(np.uint8),          # B
])

# Extract depth
depth = rc.depth_data.numpy()[world, depth_adr : depth_adr + width * height]
depth_image = depth.reshape(height, width)

# Display or save
Image.fromarray(rgb_image).save("rgb.png")
Image.fromarray((np.clip(depth_image / 5.0, 0, 1) * 255).astype(np.uint8)).save("depth.png")

```

# Benchmarks

Rendering can be benchmarked from the CLI using the mujoco-warp `testspeed` tool. Here is an example:

```
mjwarp-testspeed benchmarks/primitives.xml --function=render
```

Below are some benchmarks across a variety of scenes.

### Cartpole 

render_benchmark_total_fps_Cartpole

### Franka (Visual + Primitive)

render_benchmark_total_fps_Franka

### Primitives (100+ geometry)

render_benchmark_total_fps_Primitives Scene

### Height Field + Apptronik
render_benchmark_total_fps_Apptronik Heightfield

# Limitations

### 1. Visual Meshes vs Primitives
Where possible, users should use primitives over visual meshes. Mesh rendering is costly and scales with mesh complexity. For vision-based learning, especially for non-sim2real usage, it is recommended to use the primitives of roughly the same shape as the original mesh. Often the best practice here is to have both in your model, and label them in the MuJoCo xml/mjcf as `group=3` or `group=4` (collisions are often already marked as `group=3`). Rendering can be toggled between the visual and collision or primitive meshes by changing the `enabled_geom_groups` argument of the RenderContext

### 2. Flex Rendering
Flex rendering is currently WIP and will continue to evolve and improve.

### 3. Higher Resolution / More Cameras
The renderer currently scales linearly with resolution and number of cameras, as it scales with the total number of rays that need to be raycast. Higher resolutions or more cameras will lead to lower throughput. Improving high resolution rendering performance is WIP and on the roadmap.

# Happy Rendering!
The renderer will continue to improve as new features and performance optimizations are added. Please post a GitHub issue if there are any bugs or concerns or specific features that you need.


uv.lock
shi-eric, line 1: Can you check what version of `uv` you are running on your system? Odd that `revision` went from 3 to 1.

erikfrey:
@StafaH amazing!

Can you resolve the conflicts on the branch?  Then I'll run tests and we'll see where we are.

contrib/render.py
```diff
@@ -0,0 +1,362 @@
+# Copyright 2025 The Newton Developers
```
thowell: 2025 -> 2026

mujoco_warp/__init__.py
```diff
@@ -24,10 +24,12 @@
 
 # isort: off
 from mujoco_warp._src.forward import step as step
+from mujoco_warp._src.render import render as render
```
thowell: lets move render to the section below (the current order here is so `step`, `Model`, and `Data` are first in the [api documentation](https://mujoco.readthedocs.io/en/latest/mjwarp/api.html))

mujoco_warp/_src/benchmark.py
```diff
@@ -91,6 +92,7 @@ def benchmark(
   event_trace: bool = False,
   measure_alloc: bool = False,
   measure_solver_niter: bool = False,
+  render_context: Optional[RenderContext] = None,
```
thowell: `RenderContext | None`

lets replace `Optional` with `| None` everywhere

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,1013 @@
+# Copyright 2025 The Newton Developers
```
thowell: 2025 -> 2026

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,1013 @@
...
+
+
+@wp.func
+def _compute_box_bounds(
+  # In:
+  pos: wp.vec3,
+  rot: wp.mat33,
+  size: wp.vec3,
+) -> Tuple[wp.vec3, wp.vec3]:
+  min_bound = wp.vec3(wp.inf, wp.inf, wp.inf)
```
thowell: instead of `wp.inf` can we use `MJ_MAXVAL`? (see https://github.com/google-deepmind/mujoco_warp/pull/1051)

mujoco_warp/_src/bvh.py
```diff
@@ -0,0 +1,1013 @@
...
+@wp.func
+def _compute_capsule_bounds(
+  # In:
+  pos: wp.vec3,
+  rot: wp.mat33,
+  size: wp.vec3,
+) -> Tuple[wp.vec3, wp.vec3]:
+  radius = size[0]
+  half_length = size[1]
+  local_end1 = wp.vec3(0.0, 0.0, -half_length)
```
thowell: is it more efficient to create `z = wp.vec3(...` from the third column of `rot` and then use `(+/-)z * half_length` instead of `rot @ local_end(1/2)`?

mujoco_warp/_src/ray.py
```diff
@@ -687,6 +689,87 @@ def ray_mesh(
...
+  """Returns intersection information for ray mesh intersections.
+
+  Requires wp.Mesh be constructed and their ids to be passed.
+  """
+  t = float(-1.0)
+  u = float(0.0)
+  v = float(0.0)
+  sign = float(0.0)
+  n = wp.vec3(0.0, 0.0, 0.0)
+  f = int(-1)
```
thowell: is `int()` necessary here?

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,690 @@
...
+  if geom_type[geom_id] == GeomType.PLANE:
+    local = wp.transpose(rot) @ (hit_point - pos)
+    uv = wp.vec2(local[0], local[1])
+
+  if geom_type[geom_id] == GeomType.MESH:
+    if f < 0 or mesh_id < 0:
+      return wp.vec3(0.0, 0.0, 0.0)
+
+    face_adr = mesh_faceadr[mesh_id] + f
+    uv0 = mesh_texcoord[mesh_texcoord_offsets[mesh_id] + mesh_facetexcoord[face_adr].x]
```
thowell: is it possible to use `[0]`, `[1]`, `[2]` instead of `.x`, `.y`, `.z`?

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,690 @@
...
+      bary_u = u
+      bary_v = v
+      face_idx = f
+      geom_mesh_id = hit_mesh_id
+
+  return geom_id, dist, normal, bary_u, bary_v, face_idx, geom_mesh_id
+
+
+@wp.func
+def cast_ray_first_hit(
```
thowell, line 215: lets add a todo for trying to combine `cast_ray` and `cast_ray_first_hit` since it looks like there is a bunch of similar code in both functions

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,690 @@
...
+    )
+
+    if shadow_hit:
+      visible = 0.3
+
+  return ndotl * attenuation * visible
+
+
+@event_scope
+def render(m: Model, d: Data, rc: RenderContext):
```
thowell, line 409: @erikfrey should we update the kernel analyzer (probably in a separate pr) to check for something like `# RenderContext in:` and `# RenderContext out:`?

mujoco_warp/_src/render.py
```diff
@@ -0,0 +1,690 @@
...
+              v,
+              f,
+              mesh_id,
+            )
+            base_color = wp.cw_mul(base_color, tex_color)
+
+    len_n = wp.length(normal)
+    n = normal if len_n > 0.0 else wp.vec3(0.0, 0.0, 1.0)
+    n = wp.normalize(n)
+    hemispheric = 0.5 * (wp.dot(n, wp.vec3(0.0, 0.0, 1.0)) + 1.0)
```
thowell: this dot product could just be `n[2]`?

mujoco_warp/_src/sensor.py
```diff
@@ -138,8 +138,8 @@ def _cam_projection(
   refid: int,
 ) -> wp.vec2:
   sensorsize = cam_sensorsize[refid]
-  intrinsic = cam_intrinsic[refid]
-  fovy = cam_fovy[refid]
+  intrinsic = cam_intrinsic[worldid, refid]
```
thowell: should this use `worldid % cam_intrinsic.shape[0]` indexing? similar indexing update for `cam_fovy` below?

mujoco_warp/_src/types.py
```diff
@@ -870,10 +882,11 @@ class Model:
     cam_poscom0: global position rel. to sub-com in qpos0    (*, ncam, 3)
     cam_pos0: global position rel. to body in qpos0          (*, ncam, 3)
     cam_mat0: global orientation in qpos0                    (*, ncam, 3, 3)
-    cam_fovy: y field-of-view (ortho ? len : deg)            (ncam,)
+    cam_projection: projection type (ProjectionType)         (ncam,)
+    cam_fovy: y field-of-view (ortho ? len : deg)            (*, ncam,)
```
thowell: please remove `,` after `ncam`

mujoco_warp/_src/types.py
```diff
@@ -1590,8 +1604,7 @@ class Data:
     cdof: com-based motion axis of each dof (rot:lin)           (nworld, nv, 6)
     cinert: com-based body inertia and mass                     (nworld, nbody, 10)
     flexvert_xpos: cartesian flex vertex positions              (nworld, nflexvert, 3)
-    flexedge_J: edge length Jacobian                            (nworld, nflexedge, nv)
-    flexedge_length: flex edge lengths                          (nworld, nflexedge, 1)
+    flexedge_length: flex edge lengths                          (nworld, nflexedge)
```
thowell: are we missing `flexedge_J` now?

mujoco_warp/_src/types.py
```diff
@@ -1747,3 +1760,111 @@ class Data:
...
   collision_worldid: array("naconmax", int)
   ncollision: array(1, int)
+
+
+@dataclasses.dataclass
+class RenderContext:
+  """Context for rendering.
+
+  Attributes:
+    nrender: number of active cameras that are rendering
```
thowell: 'number of actively rendering cameras'?

thowell:
great work @StafaH!

left a few comments in the code. additionally, in order to merge this pr we need compatibility with the latest release warp (1.11.0). since the texture features depend on pre-release warp (1.12.0) we will need to add guards in the code for the warp version.

we should follow the pattern in `io.py` for mujoco that checks the version and introduces a global constant `BLEED_EDGE_MUJOCO` [see here](https://github.com/google-deepmind/mujoco_warp/blob/7f89cacecbf0baff92a631671d4a7a45c2b07e20/mujoco_warp/_src/io.py#L46). for warp we should check the version and introduce something like `BLEEDING_EDGE_WARP` to utilize for the guards in the code.

mujoco_warp/_src/render.py
StafaH: Left the comment. I actually tried this once but the result was lower perf, defaintely worth investigating again and also combining with the TODO for branch elimination

mujoco_warp/_src/sensor.py
StafaH: I modified cam_fovy and cam_intrinsic to be 2d arrays so that we can domain randomize then, the leading dim should be batched across worlds.

uv.lock
StafaH: I was using an older version. I've reverted the whole uv.lock, as this merge should not be changing dependencies.

mujoco_warp/_src/sensor.py
thowell: is the leading dimension always `nworld`? if not, we should utilize the modulo operator for indexing

StafaH: It should always be nworld just like the other uses of model DR across worlds

thowell: the leading dimension for batched `Model` fields is by default 1. we should follow the pattern used [here](https://github.com/google-deepmind/mujoco_warp/blob/de56a1c98ad75b1c3dfb1ecf92dd1be0c2d0740f/mujoco_warp/_src/smooth.py#L131) for `qpos0` to index any batched `Model` fields

StafaH: Understood. Fixed here and in other instances in rendering.

uv.lock
```diff
@@ -1271,4 +1271,4 @@ source = { registry = "https://pypi.org/simple" }
 sdist = { url = "https://files.pythonhosted.org/packages/e3/02/0f2892c661036d50ede074e376733dca2ae7c6eb617489437771209d4180/zipp-3.23.0.tar.gz", hash = "sha256:a07157588a12518c9d4034df3fbbee09c814741a33ff63c05fa29d26a2404166", size = 25547, upload-time = "2025-06-08T17:06:39.4Z" }
 wheels = [
     { url = "https://files.pythonhosted.org/packages/2e/54/647ade08bf0db230bfea292f893923872fd20be6ac6f53b2b936ba839d75/zipp-3.23.0-py3-none-any.whl", hash = "sha256:071652d6115ed432f5ce1d34c336c0adfd6a884660d1e9712a256d3d3bd4b14e", size = 10276, upload-time = "2025-06-08T17:06:38.034Z" },
-]
+]
```
thowell: can we fix this? thanks!

erikfrey (approved):
Nice clean code!

forrestthewoods:
@thowell this is very neat! I have several questions.

1. can you help me understand these benchmark numbers. what GPUs were they performed on?
2. how exactly do I run benchmarks other than primitive? how do I save out debug screenshots to see what I renderered? how do I run a render-only benchmark to test just rendering?
3. actually, how do I get render framerate for even primitives?

Running the command `uv run mjwarp-testspeed benchmarks/render/primitives.xml --function=render --nworld=512` on an RTX 4090 I got:

```
Summary for 512 parallel rollouts

Total JIT time: 0.02 s
Total simulation time: 0.98 s
Total steps per second: 522,101
Total realtime factor: 1,044.20 x
Total time per step: 1915.34 ns
Total converged worlds: 210 / 512
```
I don't really know how to interpret this. I don't know how to recreate the render FPS benchmark results you shared in this PR.

Thanks!

PR #1114, opened 2026-02-04
Add docstring to gjk_normalize function

arushsinghal:
Adds Google-style docstring to the gjk_normalize utility function explaining its bounds checking behavior for the GJK algorithm.


erikfrey:
Thanks for the contribution!  I do not think we expect to use `conftest.py` in the way you are proposing here.

The docstring contribution is appreciated, however we do not always add docstrings to small `wp.func` such as these.  It's a bit subjective and there's not a hard and fast rule.  At some point we may add a style guide for contributions.

PR #1115, opened 2026-02-04
height field: fix margin

thowell:
the current implementation for height field geoms adds `margin` to the top 3 prism vertices in the local z direction **and** sets `geom1.margin = margin`. as a result, the support function computation is incorrect when `margin > 0` since the returned support point includes an additional (incorrect in this case) `0.5 * margin * dir` term.

this pr fixes the issue by setting `geom1.margin = 0` and tests the height field support function.

thanks @kevinzakka for pointing out this issue!

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -363,8 +363,7 @@ def ccd_hfield_kernel_builder(
     min_pos = wp.vec3(MJ_MAXVAL, MJ_MAXVAL, MJ_MAXVAL)
     min_id = int(-1)
 
-    # TODO(team): height field margin?
-    geom1.margin = margin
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/collision_convex.py
```diff
@@ -363,8 +363,8 @@ def ccd_hfield_kernel(
     min_pos = wp.vec3(MJ_MAXVAL, MJ_MAXVAL, MJ_MAXVAL)
     min_id = int(-1)
 
-    # TODO(team): height field margin?
-    geom1.margin = margin
+    # margin added to hfield prism
+    geom1.margin = 0.0
```
kbayes: IIRC, this is set to zero initially and this line can be remove.  

The comment could be more clear by saying that the margin is added onto the z-axis of the prism later on.

PR #1122, opened 2026-02-06
Make stat.meaninertia multi-world compatible

adenzler-nvidia:
Changed stat.meaninertia from a scalar float to a per-world array to support different inertia values across worlds with varying configurations and mass properties.

The meaninertia value depends on both qpos0 and body masses/inertias, as it's computed from the diagonal of the joint-space inertia matrix qM. Since qpos0 and masses can vary per world, meaninertia must also be per-world to ensure correct solver convergence behavior.

Changes:
- types.py: Changed Statistic.meaninertia to array("*", float)
- io.py: Added _compute_meaninertia kernel to compute from qM diagonal in set_const_0, handling both dense and sparse qM storage formats
- solver.py: Updated kernels to index meaninertia array per-world
- io_test.py: Added parameterized test for dense/sparse jacobian modes verifying meaninertia matches MuJoCo after qpos0/mass modifications


erikfrey (approved):
Looks great, thanks.  Let's rebase and make sure tests are green before merging.

PR #1123, opened 2026-02-06
fix get_data_into

thowell:


adenzler-nvidia:
This should probably also include a lockfile upgrade for the mujoco version - such that the uv workflow targets bleeding edge mujoco?

PR #1126, opened 2026-02-06
add absolute links for kernel_analyzer in README

shi-eric:
## Summary
- Link `contrib/kernel_analyzer` directory and its README using absolute GitHub URLs so they render correctly on PyPI.


erikfrey (approved):
Good catch!

PR #1134, opened 2026-02-09
Bugfix: testspeed in headless mode when rc is None

Kenny-Vilella:


StafaH:
Might be cleaner to just do a one-liner:

`f"RenderContex ...." if rc else ""`

Kenny-Vilella:
If I make a one liner, the autoformatter split it weirdly.
Agree that the current format is not ideal, but not sure what is the cleanest way to do this.

StafaH:
I'm happy with removing the shadow + texture part of the print and making it short, then you can one-line it. Up to @erikfrey or @thowell for what the final output should look like.

mujoco_warp/testspeed.py
```diff
@@ -471,10 +471,13 @@ def trid_to_types(trid):
...
         collider_str = ""
 
+      bvh_ngeom = "None" if rc is None else rc.bvh_ngeom
+      ncam = "None" if rc is None else rc.nrender
+      cam_res = "None" if rc is None else rc.cam_res.numpy()
       print(
         f"Model\n{sizes_str}{opt_str}{collider_str}"
         f"Data\n  nworld: {d.nworld} naconmax: {d.naconmax} njmax: {d.njmax}\n"
-        f"RenderContext\n  shadows: {_USE_SHADOWS.value} textures: {_USE_TEXTURES.value} nlight: {m.nlight} bvh_ngeom: {rc.bvh_ngeom} ncam: {rc.nrender} cam_res: {rc.cam_res.numpy()}\n"
+        f"RenderContext\n  shadows: {_USE_SHADOWS.value} textures: {_USE_TEXTURES.value} nlight: {m.nlight} bvh_ngeom: {bvh_ngeom} ncam: {ncam} cam_res: {cam_res}\n"
```
thowell: can we make this entire line conditioned on something like `rc is not None`? then instead of the lines above for `bvh_geom`, `ncam`, and `cam_res` we can always print these fields with `RenderContext`

Kenny-Vilella: I didn't find any other way that are both cleaner than that and satisfy the autoformatter.
Do you have any suggestion?

thowell: suggestion here https://github.com/Kenny-Vilella/mujoco_warp/pull/3. thanks!

Kenny-Vilella: Thanks !

PR #1136, opened 2026-02-09
add `jac`: batched implementation of `mj_jac`

### LINKED ISSUE #1132: Expose jacobian functions in the public API
**TLDR**: No jacobian functions are in the public API. `support.jac` in support.py is a `@wp.func` device function and `_compute_body_jac_row` in io.py is internal.

**Request**: a public `mjwarp.jac(model, data, point, bodyid)` returning `(jacp, jacr)` with shape `(nworld, 3, nv)`. The warp equivalent of `mj_jac`.

The object type variants (site, body, bodycom, geom, subtreecom) all just call the generic jac with different (point, body_id), so we can build those on our side.

thowell:
- add `jac`: batched implementation of `mj_jac` ([documentation](https://mujoco.readthedocs.io/en/latest/APIreference/APIfunctions.html#mj-jac))
- rename `@wp.func` `jac` -> `jac_dof`

should resolve #1132

### DIFF
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -81,6 +81,7 @@ from mujoco_warp._src.smooth import transmission as transmission
 from mujoco_warp._src.solver import solve as solve
 from mujoco_warp._src.support import contact_force as contact_force
 from mujoco_warp._src.support import get_state as get_state
+from mujoco_warp._src.support import jac as jac
 from mujoco_warp._src.support import mul_m as mul_m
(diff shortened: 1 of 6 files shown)


kevinzakka:
FYI, take my review with a grain of salt, not a warp expert!

thowell:
1. refactored to utilize `@cache_kernel`
2. done
3. creating the array inline should be inexpensive so i think creating these empty array instances is ok

PR #1142, opened 2026-02-11
Fix implicitfast NaN with tendon actuators on serial chains

adenzler-nvidia:
- Fix NaN in implicitfast integrator when tendon actuators drive joints on a serial chain (dense mode)                                                                                                    
- The tiled actuator kernel writes the full tile block including sibling DOF pairs that are structurally zero in M. The subsequent _qderiv_actuator_passive kernel only iterates over ancestor pairs,
  leaving stale actuation values at sibling positions. This makes A = M - dt*qDeriv non-positive-definite, causing the tiled Cholesky to produce NaN.
- Fix by masking qDeriv at source: the tiled kernel now loads the qM tile and zeros entries where M is structurally zero, so stale values never reach the Cholesky. This matches MuJoCo CPU's implicitfast
   approximation (sibling cross-terms are dropped).
- Sparse mode was already unaffected (the sparse actuation kernel only writes to ancestor-pair positions).

### DIFF
--- a/mujoco_warp/testspeed.py
+++ b/mujoco_warp/testspeed.py
@@ -477,6 +477,7 @@ def _main(argv: Sequence[str]):
       print(
         f"Model\n{sizes_str}{opt_str}{collider_str}"
         f"Data\n  nworld: {d.nworld} naconmax: {d.naconmax} njmax: {d.njmax}\n"
+        f"RenderContext\n  shadows: {_USE_SHADOWS.value} textures: {_USE_TEXTURES.value} nlight: {m.nlight} bvh_ngeom: {rc.bvh_ngeom} ncam: {rc.nrender} cam_res: {rc.cam_res.numpy()}\n"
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/derivative_test.py
```diff
@@ -118,6 +118,94 @@ def test_smooth_vel(self, jacobian):
 
     _assert_eq(mjw_out, mj_out, "qM - dt * qDeriv")
 
+  _TENDON_SERIAL_CHAIN_XML = """
+    <mujoco>
+      <compiler angle="radian" autolimits="true"/>
+      <option integrator="implicitfast"/>
+      <default><general biastype="affine"/></default>
```
thowell: can we format default to be on multiple lines?

adenzler-nvidia: yes sure, fixed

PR #1146, opened 2026-02-11
Refactor and remove m/d from create_render_context

StafaH:
Remove the dependency of `create_render_context` on having model and data. Some side-effects of this change:

- `use_precomputed_rays` was added, users should explicitly toggle this. This is the easiest way for us to know that they intend to domain randomize and batch the camera intrinsic fields.
- `nworld` is the new required arg for making a RenderContext

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -2353,6 +2353,11 @@ def create_render_context(
 
   ray = wp.zeros(int(total), dtype=wp.vec3)
 
+  # TODO: remove after mjwarp depends on mujoco >= 3.4.1 in pyproject.toml
+  cam_projection = np.zeros(mjm.ncam, dtype=int)
+  if BLEEDING_EDGE_MUJOCO:
+    cam_projection = mjm.cam_projection
+
   offset = 0
   for idx, cam_id in enumerate(active_cam_indices):
     img_w = cam_res_np[idx][0]
@@ -2364,7 +2369,7 @@ def create_render_context(
         offset,
         img_w,
         img_h,
-        int(mjm.cam_projection[cam_id]),
+        int(cam_projection[cam_id]),
         float(mjm.cam_fovy[cam_id]),
         wp.vec2(mjm.cam_sensorsize[cam_id]),
         wp.vec4(mjm.cam_intrinsic[cam_id]),


mujoco_warp/_src/bvh.py
```diff
@@ -235,18 +234,23 @@ def compute_bvh_group_roots(
   group_root_out[tid] = root
 
 
-def build_scene_bvh(m: Model, d: Data, rc: RenderContext):
+def build_scene_bvh(mjm: mujoco.MjModel, mjd: mujoco.MjData, nworld: int, rc: RenderContext):
```
thowell: nit: maybe `nworld` last?

mujoco_warp/_src/bvh.py
```diff
@@ -865,23 +873,23 @@ def build_flex_bvh(
   nface = int(flex_faceadr[-1])
   flex_faceadr = flex_faceadr[:-1]
 
-  face_point = wp.zeros(nface * 3 * d.nworld, dtype=wp.vec3)
-  face_index = wp.zeros(nface * 3 * d.nworld, dtype=wp.int32)
-  group = wp.zeros(nface * d.nworld, dtype=int)
+  face_point = wp.zeros(nface * 3 * nworld, dtype=wp.vec3)
```
thowell: can these fields be initialized with `wp.empty`?

mujoco_warp/_src/io.py
```diff
@@ -2178,10 +2177,15 @@ def create_render_context(
     cam_active: List of booleans indicating which cameras to include in rendering.
                 If None, all cameras are included.
     flex_render_smooth: Whether to render flex meshes smoothly.
+    use_precomputed_rays: Use precomputed rays instead of computing during rendering.
+                          When using Domain Randomization for camera intrinsics, set this to False.
```
thowell: 'Domain Randomization' -> 'domain randomization'
'set this to False' -> 'set to False'

thowell (approved):

mujoco_warp/_src/bvh.py
hartikainen: Should this be `Sequence[mujoco.MjModel]`? This way one could domain randomize to e.g. different `mjm.geom_size`s? Or does it not matter if the initial scene bvh is computes for "wrong" model in the case with e.g. varying `geom_size`s passed to `render.render`?

StafaH: Correct, it does not matter. The first call to refit after creating the context would update to the correct size. Quality of the build shuoldn't change too much if using DR field vs assuming same size.

mujoco_warp/_src/bvh.py
StafaH: Most can, left `flexvert_norm` since its an accumulated value

mujoco_warp/_src/bvh.py
hartikainen: Makes sense!

PR #1149, opened 2026-02-12
update `_qderiv_actuator_passive_vel` with actuator force clamp case

thowell:
corresponding mujoco update https://github.com/google-deepmind/mujoco/commit/b46682948290a109654bdfa56f408eb66c2aabdb


thowell:
we should reduce the time this test takes to run

thowell:
modified the timesteps so that the total number of `mjw.step` calls is reduced from 100 to 20

PR #1155, opened 2026-02-13
Feat/split nonconvex meshes

NandiniDhanrale:
Summary  
This PR introduces a Python utility for preprocessing non-convex mesh assets in MuJoCo XML models by decomposing selected STL meshes into convex parts and rewriting all mesh references accordingly. [mujoco.readthedocs](https://mujoco.readthedocs.io/en/stable/modeling.html)

Motivation  
Issue #2985 requested a reusable helper for handling non-convex mesh geometry in MuJoCo models. This change provides a production-ready API that supports repeatable convex decomposition with disk caching and clean XML rewriting, so users can safely incorporate non-convex STL assets without manual editing or ad hoc scripts. 

Implementation  
- Added a new module: `mesh_utils.py`. 
- Introduced a public API:  
  `split_nonconvex_meshes(xml, mesh_names, coacd_kwargs=None, *, assets_dir=None, cache_dir=None, input_is_path=None, skip_if_already_convex=True, validate_output=True) -> str`.

- Key behavior:  
  - Accepts either an XML string or a file path as input.  
  - Resolves and validates the requested `.stl` mesh assets on disk.  
  - Invokes `coacd.run_coacd(...)` with configurable `coacd_kwargs` for approximate convex decomposition. 
  - Stores a deterministic cache on disk keyed by STL file bytes, CoACD version, and the effective kwargs to ensure repeatable results and avoid redundant work on large meshes. 
  - Replaces the original `<asset><mesh ...>` entries with generated `__cvx_###` mesh entries that correspond to the convex components.  
  - Rewrites or duplicates `<geom mesh="...">` references so that geoms point to the generated convex mesh names while preserving the original semantics of the model. 
  - Provides explicit error handling for missing mesh names, missing or unreadable STL files, invalid mesh assets, missing `coacd` installation, and malformed decomposition outputs. 
  - Optionally validates the transformed XML by compiling it with MuJoCo to catch structural issues early.

Tests  
- Added `mesh_utils_test.py` covering:  
  - XML rewrite behavior and cache-hit behavior.  
  - Error paths for missing meshes.  
  - The `skip_if_already_convex` behavior for meshes that are already convex.  
- Local test runs from the repo root:  
  - `python -m pytest mesh_utils_test.py -q` → 3 passed.  
  - `python -m pytest io_test.py mesh_utils_test.py -k "get_data_into or put_data_into or split_nonconvex" -q` → 11 passed.  

Documentation  
- Added a usage snippet and a synthetic before/after XML example to `README.md` to demonstrate typical usage and expected XML transformations. 
- Exported the new API at the top level in `__init__.py` so downstream users can call `split_nonconvex_meshes` directly.  

Feedback Requested  
- Whether `mesh_utils.py` is the right home and naming for this utility within the codebase.  
- API shape: parameter list, return type (XML string), and exception strategy for `split_nonconvex_meshes`.  
- Caching behavior: defaults, cache keying, and any expectations or concerns around performance and disk usage for large or numerous meshes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,3 +121,59 @@ Benchmark rendering with:
 ```bash
 mjwarp-testspeed benchmarks/primitives.xml --function=render
 ```
+
+# Python mesh utility
+
+`mujoco_warp` includes a Python helper for splitting non-convex STL meshes into
+cached convex parts using COACD:
+
+```python
+from pathlib import Path
+import mujoco
+import mujoco_warp as mjwarp
+
+xml_path = Path("models/scene.xml")
+
+xml_processed = mjwarp.split_nonconvex_meshes(
+		str(xml_path),
+		mesh_names=["tool", "housing"],
+		coacd_kwargs={"threshold": 0.05, "max_convex_hull": 32},
+)
+
+model = mujoco.MjModel.from_xml_string(xml_processed)
+data = mujoco.MjData(model)
+```
+
+Synthetic XML transformation example:
+
+Before:
+
+```xml
+<mujoco>
+	<asset>
+		<mesh name="tool" file="meshes/tool.stl"/>
+	</asset>
+	<worldbody>
+		<body>
+			<geom name="tool_geom" type="mesh" mesh="tool"/>
+		</body>
+	</worldbody>
+</mujoco>
+```
+
+After (illustrative):
+
+```xml
+<mujoco>
+	<asset>
+		<mesh name="tool__cvx_000" file=".mujoco_mesh_cache/tool/<hash>/tool__cvx_000.stl"/>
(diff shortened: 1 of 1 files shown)


erikfrey:
Hi @NandiniDhanrale - thank you, this is an interesting contribution!  Such a feature we would typically support directly in model compilation, which is shared functionality between MuJoCo, MJX, MJWarp, and other variants.  Model compilation lives here:

https://github.com/google-deepmind/mujoco/tree/main/src/user

I would recommend creating an issue here, proposing your idea in prose before moving to a PR: https://github.com/google-deepmind/mujoco/issues

The core MuJoCo maintainers may be able to give you some advice on how to proceed with this PR.

PR #1169, opened 2026-02-18
actuator and sensor delays

thowell:
[actuator and sensor delays](https://mujoco.readthedocs.io/en/latest/modeling.html#delays)

---

### API Signature Differences

All shape changes are driven by GPU batching over `nworld` parallel worlds.

| Parameter | MuJoCo C | MuJoCo Warp | Reason |
|-----------|----------|-------------|--------|
| `values` | 1D `(nsample,)` | 2D `(nworld, nsample)` | Per-world initial values |
| `phase` | scalar `float` | 1D `(nworld,)` | Per-world phase |
| `time` | scalar `float` | 1D `(nworld,)` | Per-world simulation time |
| `result` | scalar return / output param | `(nworld,)` or `(nworld, dim)` | One result per world |
| `interp` | default `-1` | no default | Explicit specification required |

#### MuJoCo C: Conditional Computation

Each sensor goes through `compute_or_read_sensor()`, which **skips computation** for delayed sensors:

- `delay > 0`: reads directly from history buffer, **never calls** `mj_computeSensor`
- `interval > 0, condition not met`: reads stale value from buffer
- `interval > 0, condition met`: computes fresh value

#### MuJoCo Warp: Always Compute, Then Overwrite

All sensors are computed unconditionally via a single kernel launch, then `apply_sensor_delay()` runs as a second pass:

1. Compute **all** sensor values (including delayed ones)
2. Insert fresh values into history buffers
3. Overwrite `sensordata` with delayed values from history

#### Justification

GPU kernels execute uniformly across all threads — per-sensor branching (compute vs skip) creates warp divergence and is inefficient. Computing everything then overwriting is the standard GPU tradeoff: redundant computation is cheaper than divergent control flow. The final result is identical.

### Public API Functions

| MuJoCo C | MuJoCo Warp | Description |
|----------|-------------|-------------|
| `mj_readCtrl(m, d, id, time, interp)` | `read_ctrl(m, d, ctrlid, time, interp, result)` | Read delayed ctrl for 1 actuator |
| `mj_readSensor(m, d, id, time, buf, interp)` | `read_sensor(m, d, sensorid, time, interp, result)` | Read delayed sensor for 1 sensor |
| `mj_initCtrlHistory(m, d, id, times, values)` | `init_ctrl_history(m, d, ctrlid, times, values)` | Initialize actuator history buffer |
| `mj_initSensorHistory(m, d, id, times, values, phase)` | `init_sensor_history(m, d, sensorid, times, values, phase)` | Initialize sensor history buffer |

### DIFF
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -47,6 +47,10 @@ from mujoco_warp._src.forward import implicit as implicit
 from mujoco_warp._src.forward import rungekutta4 as rungekutta4
 from mujoco_warp._src.forward import step1 as step1
 from mujoco_warp._src.forward import step2 as step2
+from mujoco_warp._src.history import init_ctrl_history as init_ctrl_history
+from mujoco_warp._src.history import init_sensor_history as init_sensor_history
+from mujoco_warp._src.history import read_ctrl as read_ctrl
+from mujoco_warp._src.history import read_sensor as read_sensor
 from mujoco_warp._src.inverse import inverse as inverse
 from mujoco_warp._src.io import create_render_context as create_render_context
 from mujoco_warp._src.io import get_data_into as get_data_into
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -20,6 +20,7 @@ import warp as wp
 from mujoco_warp._src import collision_driver
 from mujoco_warp._src import constraint
 from mujoco_warp._src import derivative
+from mujoco_warp._src import history
 from mujoco_warp._src import island
 from mujoco_warp._src import math
 from mujoco_warp._src import passive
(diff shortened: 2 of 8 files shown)


yuvaltassa (commented):
1. Can you add some tests for history-only mode (nhistory > 0, delay = 0)? Relatedly I think delay.py and delay_test.py should be renamed delay -> history, since delays are only one thing that is enabled by the history buffers.
2. Regarding the decision to always compute sensors. I'm sure this makes sense for many sensor types, but for the ones where you'll actually want interval/period it might not. Imagine a rangefinder array of depth sensor implemented using @StafaH 's raytracer. Do you really want to compute it every 2ms even if it's a 30hz camera? I mean, maybe, but I'd want to see a benchmark that proves that. We can decide that we revisit this decision in the future since the API is the same, if you like.

thowell:
@yuvaltassa the pr is updated

1. added a test for history + no delay `history_test.test_actuator_history_only` and renamed files with 'delay' to 'history'
2. added a todo to revisit always computing sensor values

yuvaltassa (approved):

PR #1171, opened 2026-02-19
Add util_pkg module for package version checking

erikfrey:
Add `check_version()` function that checks installed package versions using pyproject.toml-like semantics (e.g. `check_version('numpy>=1.20.0')`).

## Key features
- No dependency on `packaging` package - implemented manually
- Supports operators: `>=`, `<=`, `>`, `<`, `==`, `!=`
- Handles custom version strings like `3.5.0.dev869102767` and `3.5.0-google3`
- Both `.` and `-` are treated as separators
- Numeric components compared numerically, non-numeric lexicographically  
- Longer versions are greater (`3.5.0.dev > 3.5.0`)

## Files added
- `mujoco_warp/_src/util_pkg.py` - Main module
- `mujoco_warp/_src/util_pkg_test.py` - Tests (32 test cases)

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -14,7 +14,6 @@
 # ==============================================================================
 
 import dataclasses
-import importlib.metadata
 import warnings
 from typing import Any, Optional, Sequence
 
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/util_pkg_test.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2025 DeepMind Technologies Limited
```
thowell: 2025 -> 2026

mujoco_warp/_src/util_pkg.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2026 DeepMind Technologies Limited
```
thowell: this should be 'The Newton Developers'

mujoco_warp/_src/util_pkg_test.py
thowell: this should be 'The Newton Developers'

thowell:
should we remove [_is_mujoco_fresh](https://github.com/google-deepmind/mujoco_warp/blob/2710a5a6d781965ed09f5a5feada6efbd1b887ec/mujoco_warp/_src/io.py#L32) and update/remove (since it is not currently used) [BLEEDING_EDGE_MUJOCO](https://github.com/google-deepmind/mujoco_warp/blob/2710a5a6d781965ed09f5a5feada6efbd1b887ec/mujoco_warp/_src/io.py#L40C1-L40C21)?

PR #1172, opened 2026-02-19
Add incremental Hessian updates for Newton solver

adenzler-nvidia:
## Summary

- Skip the expensive J^T*D*J Hessian rebuild on subsequent solver iterations by tracking constraint state changes and incrementally updating H
- Only constraints that changed QUADRATIC state contribute a delta to the stored H, after which H is re-factorized via the existing Cholesky kernels

## Implementation

- Fuse state change detection into `update_constraint_efc` via `@cache_kernel(track_changes)` to avoid a separate kernel launch. Uses stream compaction (atomic add) to build a compact list of changed constraint IDs per world.
- Add `incremental_h_update` kernel: scalar per-thread kernel where each thread handles one (i,j) element of the lower triangle of H. Launches with `dim=(nworld, nv*(nv+1)/2)` using the standard triangular index formula. Skips zero J entries.
- On the first solver iteration per step, use the full JTDAJ path (unchanged). On subsequent iterations, use the incremental path: update H, then re-factorize and solve.
- Works for both small (nv <= 32, tile cholesky) and blocked (nv > 32, blocked cholesky) paths.

## Benchmarks

RTX PRO 6000 Blackwell, `graph_conditional=True` (default), `nworld=8192`.

All benchmarks below use Newton solver with pyramidal cones. Unaffected benchmarks (CG solver, elliptic cones) show no regression.

| Benchmark | nv | Main solve (us) | PR solve (us) | Solve | Main step (us) | PR step (us) | Step |
|---|---|---|---|---|---|---|---|
| humanoid | 27 | 91.1 | 83.6 | **-8.2%** | 184.0 | 177.3 | **-3.6%** |
| apollo_flat | 30 | 114.5 | 102.1 | **-10.8%** | 225.3 | 212.9 | **-5.5%** |
| apollo_terrain | 30 | 155.7 | 140.9 | **-9.5%** | 804.8 | 791.9 | -1.6% |
| unitree_g1 | 35 | 425.5 | 346.1 | **-18.7%** | 859.4 | 777.4 | **-9.5%** |
| unitree_g1_with_hands | 49 | 914.5 | 719.1 | **-21.4%** | 1723.1 | 1541.4 | **-10.5%** |
| three_humanoids | 81 | 1091.4 | 837.1 | **-23.3%** | 1553.4 | 1302.8 | **-16.1%** |

## Test plan

- [x] All 25 solver tests pass
- [x] Correctness test comparing incremental vs full Hessian paths (new test)
- [x] Verified state changes are actually detected during the test
- [x] No regressions on unaffected benchmarks (CG, elliptic)

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -43,6 +43,8 @@ class InverseContext:
   cost: wp.array(dtype=float)
   prev_cost: wp.array(dtype=float)
   done: wp.array(dtype=bool)
+  changed_efc_ids: wp.array2d(dtype=int)
+  changed_efc_count: wp.array(dtype=int)
 
 
 @dataclasses.dataclass
@@ -69,6 +71,9 @@ class SolverContext:
   beta: wp.array(dtype=float)
   h: wp.array3d(dtype=float)
   hfactor: wp.array3d(dtype=float)
+  # Incremental Hessian update (Newton only)
+  changed_efc_ids: wp.array2d(dtype=int)
+  changed_efc_count: wp.array(dtype=int)
 
 
 def create_inverse_context(m: types.Model, d: types.Data) -> InverseContext:
@@ -91,6 +96,8 @@ def create_inverse_context(m: types.Model, d: types.Data) -> InverseContext:
     cost=wp.empty((nworld,), dtype=float),
     prev_cost=wp.empty((nworld,), dtype=float),
     done=wp.empty((nworld,), dtype=bool),
+    changed_efc_ids=wp.empty((nworld, 0), dtype=int),
+    changed_efc_count=wp.empty((0,), dtype=int),
   )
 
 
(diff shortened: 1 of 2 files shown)


adenzler-nvidia:
Just for context - I did try actual rank-1 updates of the factorization. So far it's slower, I think because of 2 things:
1. storing out the factorization during the first factorize/solve is quite slow, something is off there. This is a problem especially for models that converge fast where we can't take advantage of reusing the factorization for more that 1-2 iterations
2. Doing Given's rotations only exposes parallelism in the colums which is a very bad access pattern in the warp tile API. I have a todo to try to store the factorization on the upper-left and see if that helps things.

Either way the main issue we're running into here is that this helps as soon as you need 5+ iterations for convergence, but for lower it's quite hard to be faster than what we already have.

Maybe it's a good idea to test-drive this change with a bit of RL and not only testspeed that doesn't do a lot of real stuff.

adenzler-nvidia:
needs a bit more cleanup, will let you know once it's ready

adenzler-nvidia:
Should be ready now. @thowell @erikfrey

adenzler-nvidia:
BTW are we supposed to maintain changelogs now that the first release is out?

thowell:
great feature, thanks @adenzler-nvidia!

erikfrey:
@adenzler-nvidia a changelog is a good idea.  let me circulate that with folks.

PR #1180, opened 2026-02-24
Fix mujoco nightly breakages: ten_J sparse format and MjOption field reorder

erikfrey:
## Summary

Two breaking changes in mujoco nightly require version-guarded fixes:

### 1. `ten_J` always sparse (since `3.5.1.dev872479828`)

`mjd.ten_J` is now stored in sparse format for all models, not just sparse ones. Previously, non-sparse models stored `ten_J` as a dense array that could be reshaped to `(ntendon, nv)`.

**Fix:** Updated `put_data` in `io.py` to always use `mju_sparse2dense` when mujoco version >= `3.5.1.dev872479828`, with version guard for backward compatibility. Updated corresponding test assertions in `smooth_test.py` and `forward_test.py`.

### 2. `MjOption` field reorder (since `3.5.1.dev869760513`)

`density` and `viscosity` fields moved from before `gravity` to after `magnetic` in `MjOption._all_fields`.

**Fix:** Reordered fields in `types.Option` dataclass to match the new order, with backward-compat version guard in `types_test.py`.

### Verification

- Breaking versions confirmed via bisection of 35 nightly builds from py.mujoco.org
- All 619 tests pass, 5 skipped

### Files changed

- `mujoco_warp/_src/io.py` — version-guarded `ten_J` sparse handling in `put_data`
- `mujoco_warp/_src/types.py` — reordered `density`/`viscosity` fields in `Option`
- `mujoco_warp/_src/types_test.py` — backward-compat guard for old field order
- `mujoco_warp/_src/smooth_test.py` — version-guarded `ten_J` test assertion
- `mujoco_warp/_src/forward_test.py` — version-guarded `ten_J` test assertion

### DIFF
--- a/mujoco_warp/_src/forward_test.py
+++ b/mujoco_warp/_src/forward_test.py
@@ -28,6 +28,7 @@ from mujoco_warp import EnableBit
 from mujoco_warp import GainType
 from mujoco_warp import IntegratorType
 from mujoco_warp import test_data
+from mujoco_warp._src.util_pkg import check_version
 
 # tolerance for difference between MuJoCo and mjwarp smooth calculations - mostly
 # due to float precision
@@ -426,10 +427,14 @@ class ForwardTest(parameterized.TestCase):
         actuator_moment = np.zeros((mjm.nu, mjm.nv))
         mujoco.mju_sparse2dense(actuator_moment, mjd.actuator_moment, mjd.moment_rownnz, mjd.moment_rowadr, mjd.moment_colind)
         mjd_arr = actuator_moment
-      elif arr == "ten_J" and mjm.ntendon:
-        ten_J = np.zeros((mjm.ntendon, mjm.nv))
-        mujoco.mju_sparse2dense(ten_J, mjd.ten_J, mjd.ten_J_rownnz, mjd.ten_J_rowadr, mjd.ten_J_colind)
-        mjd_arr = ten_J
+      elif arr == "ten_J":
+        if check_version("mujoco>=3.5.1.dev872479828"):
+          ten_J = np.zeros((mjm.ntendon, mjm.nv))
+          if mjm.ntendon:
+            mujoco.mju_sparse2dense(ten_J, mjd.ten_J, mjd.ten_J_rownnz, mjd.ten_J_rowadr, mjd.ten_J_colind)
+          mjd_arr = ten_J
+        else:
+          mjd_arr = mjd.ten_J.reshape((mjm.ntendon, mjm.nv))
       elif arr == "efc_J" or arr == "efc_id":
         # Already checked earlier
         continue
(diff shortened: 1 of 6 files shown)


mujoco_warp/_src/forward_test.py
```diff
@@ -426,10 +427,14 @@ def _getattr(arr):
         actuator_moment = np.zeros((mjm.nu, mjm.nv))
         mujoco.mju_sparse2dense(actuator_moment, mjd.actuator_moment, mjd.moment_rownnz, mjd.moment_rowadr, mjd.moment_colind)
         mjd_arr = actuator_moment
-      elif arr == "ten_J" and mjm.ntendon:
-        ten_J = np.zeros((mjm.ntendon, mjm.nv))
-        mujoco.mju_sparse2dense(ten_J, mjd.ten_J, mjd.ten_J_rownnz, mjd.ten_J_rowadr, mjd.ten_J_colind)
-        mjd_arr = ten_J
+      elif arr == "ten_J":
+        if check_version("mujoco>=3.5.1.dev872479828") or mujoco.mj_isSparse(mjm):
```
thowell: is `or mujoco.mj_isSparse(mjm)` necessary?

mujoco_warp/_src/io.py
```diff
@@ -983,13 +983,12 @@ def put_data(
 
   d.flexedge_J = wp.array(np.tile(mjd.flexedge_J.reshape(-1), (nworld, 1)).reshape((nworld, 1, -1)), dtype=float)
 
-  if mujoco.mj_isSparse(mjm):
-    ten_J = np.zeros((mjm.ntendon, mjm.nv))
+  ten_J = np.zeros((mjm.ntendon, mjm.nv))
+  if mujoco.mj_isSparse(mjm) or check_version("mujoco>=3.5.1.dev872479828"):
```
thowell: is `mujoco.mj_isSparse(mjm)` necessary?

mujoco_warp/_src/io.py
```diff
@@ -983,13 +983,12 @@ def put_data(
...
     mujoco.mju_sparse2dense(ten_J, mjd.ten_J.reshape(-1), mjd.ten_J_rownnz, mjd.ten_J_rowadr, mjd.ten_J_colind.reshape(-1))
```
thowell: should we add an `if mjm.ntendon` check here for the case where `ntendon == 0`?

mujoco_warp/_src/smooth_test.py
```diff
@@ -400,7 +401,12 @@ def test_tendon(self, xml):
     mjw.transmission(m, d)
 
     _assert_eq(d.ten_length.numpy()[0], mjd.ten_length, "ten_length")
-    _assert_eq(d.ten_J.numpy()[0], mjd.ten_J.reshape((mjm.ntendon, mjm.nv)), "ten_J")
+    if check_version("mujoco>=3.5.1.dev872479828") or mujoco.mj_isSparse(mjm):
```
thowell: is the `mujoco.mj_isSparse(mjm)` check necessary?

erikfrey:
Actually, I had to add `mj_isSparse` back to `io.py` (but kept it removed from the test files). In old mujoco, sparse models already stored `ten_J` sparsely — the version guard alone only handles the new behavior where *all* models store it sparsely. The test files don't need it because their model fixtures are non-sparse, but `put_data` handles arbitrary models.

PR #1193, opened 2026-02-27
kernel analyzer: check for incorrect batched `Model` field indexing

thowell:
updates the kernel analyzer to check for incorrect batched `Model` field indexing

### DIFF
--- a/mujoco_warp/_src/constraint.py
+++ b/mujoco_warp/_src/constraint.py
@@ -424,7 +424,7 @@ def _equality_joint(
     # Two joint constraint
     qposadr2 = jnt_qposadr[jntid_2]
     dofadr2 = jnt_dofadr[jntid_2]
-    dif = qpos_in[worldid, qposadr2] - qpos0[worldid, qposadr2]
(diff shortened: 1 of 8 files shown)


mujoco_warp/_src/constraint.py
```diff
@@ -424,7 +424,7 @@ def _equality_joint(
     # Two joint constraint
     qposadr2 = jnt_qposadr[jntid_2]
     dofadr2 = jnt_dofadr[jntid_2]
-    dif = qpos_in[worldid, qposadr2] - qpos0[worldid, qposadr2]
+    dif = qpos_in[worldid, qposadr2] - qpos0[qpos0_id, qposadr2]
```
adenzler-nvidia, line 427: Bug: this should be `qpos0[qpos0_id, qposadr2]` — `qpos0_id` is already computed at line 405 and used at lines 433/445, but was missed here. Since `qpos0` is `array("*", "nq", float)`, this is an OOB read when `qpos0.shape[0] == 1` and `worldid > 0`.

thowell: maybe this was incorrect/missed at some point, but indexing seems correct now

PR #1195, opened 2026-03-03
Fix depth rendering to output planar depth instead of Euclidean distance

kevinzakka:


thowell (approved):
lgtm, thanks @kevinzakka!

PR #1196, opened 2026-03-03
add flag to testspeed for setting warp kernel cache directory

thowell:


adenzler-nvidia:
note that this can also be done through an env variable, I usually have my agents do WARP_CACHE_PATH=/tmp/claude/warp-cache-$$ before any command to force separate cache and make sure there is no interference with multiple agents running at the same time.

erikfrey:
hmmm - it may be better to maintain one of way doing this to avoid confusing users - what do you think of adopting the env var approach @thowell ?

thowell:
setting the environment variable works. should we update `conftest.py` to remove the `--kernel_cache_dir` flag https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/conftest.py#L32?

PR #1198, opened 2026-03-03
updates for sparse constraint jacobian

thowell:


mujoco_warp/_src/solver.py
```diff
@@ -2084,6 +2098,59 @@ def update_gradient_h_incremental(
...
+      if Jj == 0.0:
+        continue
+      colindj = efc_J_colind_in[worldid, 0, sparseidj]
+      h = sign * Ji * Jj
+      # Ensure lower triangle: larger index first
+      if colindi >= colindj:
+        wp.atomic_add(ctx_h_out[worldid, colindi], colindj, h)
+      else:
+        wp.atomic_add(ctx_h_out[worldid, colindj], colindi, h)
+
```
adenzler-nvidia, line 2152: Thanks - this is a massive improvement, much faster than the tiled version I came up with. Great stuff.

PR #1203, opened 2026-03-05
Fix tactile sensor value duplication

quagla:
Change the parallelization of tactile sensors by removing the contact axis, which could results in accounting for the same geom multiple times when there are multiple contact points.

Also set the tangential components of the tactile measurement to zero if the mesh has undefined tangential direction in its vertex frames.

### DIFF
--- a/mujoco_warp/_src/sensor.py
+++ b/mujoco_warp/_src/sensor.py
@@ -2238,14 +2238,11 @@ def _sensor_tactile(
     pressure = depth / wp.max(kMaxDepth - depth, MJ_MINVAL)
     force = wp.mul(normal, pressure)
 
-    forceT = wp.vec3()
+    forceT = wp.vec3(0.0, 0.0, 0.0)
     forceT[0] = wp.dot(force, normal)
     if has_frame:
       forceT[1] = wp.abs(wp.dot(vel_rel, tang1))
       forceT[2] = wp.abs(wp.dot(vel_rel, tang2))
-    else:
-      forceT[1] = 0.0
-      forceT[2] = 0.0
 
     dim = sensor_dim[sensor_id] // 3
     wp.atomic_add(sensordata_out, worldid, sensor_adr[sensor_id] + 0 * dim + vertid, forceT[0])


mujoco_warp/_src/sensor.py
```diff
@@ -2151,58 +2176,80 @@ def _sensor_tactile(
...
-    plugin,
-    plugin_attr,
-    contact_type,
-    geom_size[worldid % geom_size.shape[0], geom],
-    plugin_id,
-    mesh_id,
-  )
+  has_frame = mesh_normalnum[mesh_id] == 3 * mesh_vertnum[mesh_id]
+  normal_stride = 3 if has_frame else 1
+  offset = mesh_normaladr[mesh_id] + normal_stride * vertid
```
thowell, line 2181: maybe read `mesh_quat[mesh_id]` once and the use a local variable?

mujoco_warp/_src/sensor_test.py
```diff
@@ -892,6 +892,52 @@ def my_sensor(m, d, stage):
...
+        <asset>
+          <mesh name="sensor_mesh" builtin="sphere" params="0"/>
+        </asset>
+        <worldbody>
+          <body name="sensor_body" pos="0 0 1">
+            <freejoint/>
+            <geom name="sensor_geom" type="mesh" mesh="sensor_mesh"/>
+          </body>
+          <body>
+            <geom name="contact_box" type="box" size=".7 .7 .3"/>
```
thowell: is this name necessary?

mujoco_warp/_src/sensor_test.py
```diff
@@ -892,6 +892,52 @@ def my_sensor(m, d, stage):
...
+          </body>
+        </worldbody>
+        <sensor>
+          <tactile geom="sensor_geom" mesh="sensor_mesh"/>
+        </sensor>
+      </mujoco>
+      """,
+    )
+
+    # Simulate until contact occurs (like the working C++ test)
```
thowell: is it possible to instead load a keyframe just a few timesteps before contact occurs so that the test requires fewer calls to `mjw.step`?

mujoco_warp/_src/sensor.py
```diff
@@ -2151,58 +2176,80 @@ def _sensor_tactile(
...
+    kMaxDepth = 0.05
+    pressure = depth / wp.max(kMaxDepth - depth, MJ_MINVAL)
+    force = wp.mul(normal, pressure)
 
-  # one row of mat^T * force
-  forceT = wp.vec3()
-  forceT[0] = wp.dot(force, normal)
-  forceT[1] = wp.abs(wp.dot(vel_rel, tang1))
-  forceT[2] = wp.abs(wp.dot(vel_rel, tang2))
+    forceT = wp.vec3()
```
thowell: this should initialize to all zeros so i think we can remove the else block below

PR #1215, opened 2026-03-11
_transmission sparsity

thowell:
this change to `nJmom` https://github.com/google-deepmind/mujoco/commit/53fadd9d639dae7f025555cae5076128373c2668 leads to out of bounds reads with mujoco warp since the current `_transmission` implementation assumes more memory has been allocated than is actually available (`actuator_moment` and `moment_colind` are allocated using `nJmom`).

this pr updates `_transmission` to construct site and slidercrank transmissions sparsely.

### DIFF
--- a/mujoco_warp/_src/smooth.py
+++ b/mujoco_warp/_src/smooth.py
@@ -2024,35 +2024,61 @@ def _transmission(
       dldv = axis
       dlda = vec
 
-    rowadr = wp.atomic_add(moment_nnz, worldid, nv)
-    moment_rowadr_out[worldid, actid] = rowadr
-    nnz = int(0)
+    # count dofs
+    b1 = body_weldid[site_bodyid[id]]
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/smooth.py
```diff
@@ -2024,35 +2024,61 @@ def _transmission(
       dldv = axis
       dlda = vec
 
-    rowadr = wp.atomic_add(moment_nnz, worldid, nv)
+    # count dofs
+    b1 = body_weldid[site_bodyid[id]]
+    b2 = body_weldid[site_bodyid[idslider]]
+    da1_init = int(-1)
```
kbayes, line 2030: Can this just be da1 and da2 directly?

kbayes: Resolved.

PR #1217, opened 2026-03-12
Rename gjk_iterations to ccd_iterations in EPA

kevinzakka:
When EPA exhausts its iteration limit, it prints a warning about needing more `ccd_iterations`. However, the warning was incorrectly using `gjk_iterations` instead of `epa_iterations`, resulting in wrong values reported to the user.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -1202,7 +1202,7 @@ def _is_invalid_face(face: int) -> bool:
 def _epa(
   # In:
   tolerance: float,
-  gjk_iterations: int,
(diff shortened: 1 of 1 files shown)


kbayes:
@kevinzakka This is actually not a typo. For memory reasons, epa_iterations <= opt.ccd_iterations while equality always holds for gjk_iterations. For instance box<->box uses epa_iterations = 16 which is the theoretical max needed for convergence.  While you have a opened PR, could you rename gjk_iterations to ccd_iterations in _epa? I think that'll clear up the confusion.

kevinzakka:
Thanks @kbayes, made the change.

PR #1218, opened 2026-03-12
updates for warp textures

thowell:


StafaH:
LGTM

thowell:
@erikfrey confirmed with @btaba that since `wp.Texture2D` is incorporated into `RenderContext` and not `Model` or `Data` these changes should be compatible with mjx.

PR #1226, opened 2026-03-14
Auto Differentiation Implementation 1/3

### LINKED ISSUE #500: Differentiability
https://nvidia.github.io/warp/modules/differentiability.html#differentiability

mar-yan24:
# Autodifferentiation Support 1/3

## Overview

So a bit ago I was sorta interested in implementing automatic differentiation into MJWarp cause I wanna do a project with diff contact geometry and I had some time on my hands so I decided to begin working on a personal implementation of AD support in MJWarp. I'll probably continue working on this over the month but I'm putting in a draft here to see if I can get some maintainer feedback and maybe discuss if there is still community desire for this as referenced in 'issue' #500.

Basically, these changes add reverse-mode AD support for the smooth dynamics pipeline of MuJoCo Warp. Most people should now be able to compute gradients of scalar loss functions with respect to `qpos`, `qvel`, `ctrl`, and other state variables by recording a `wp.Tape` over `kinematics -> fwd_velocity -> fwd_actuation -> euler`.

The implementation follows a **selective `enable_backward` strategy**: only the four modules that participate in the differentiable smooth-dynamics path have backward code generation enabled. All other modules (collision, constraint, solver, sensor, render, ray, etc.) remain at `enable_backward: False`. This should keep compilation time and binary size normal i think.

### Selective backward generation

| Module            | `enable_backward` | Notes                                     |
|-------------------|-------------------|-------------------------------------------|
| `smooth.py`       | `True`            | kinematics, crb, rne, com_vel, etc.       |
| `forward.py`      | `True`            | fwd_velocity, fwd_actuation, euler/rk4    |
| `passive.py`      | `True`            | passive forces (spring, damper, fluid)     |
| `derivative.py`   | `True`            | analytical derivatives (qDeriv)           |
| All others (13+)  | `False`           | collision, constraint, solver, sensor, ... |

Within the enabled modules, **tile kernels** (`wp.launch_tiled`) still have per-kernel `enable_backward=False` overrides (smooth.py lines 1053/2825/2903, forward.py line 309) because cuSolverDx LTO compilation does not support adjoint generation.

### Kernel compilation time

@Kenny-Vilella raised a good concern in discussion #993 which was about enabling backward **globally**. As I mentioned earlier, one of the issues we wanna avoid is hella long compile. This selective approach basically generates adjoint kernels for only ~30 smooth-dynamics kernels out of 100+ total. Warp caches compiled kernels, so the cost is one-time per kernel signature.

### New modules

- **`grad.py`** - coordination layer: `enable_grad()`, `disable_grad()`, `make_diff_data()`, `diff_step()`, `diff_forward()`, `SMOOTH_GRAD_FIELDS`.
- **`adjoint.py`** - centralizes `@wp.func_grad` registrations. Phase 1 provides a custom adjoint for `quat_integrate` (avoids gradient singularity at zero angular velocity).
- **`grad_test.py`** - AD test suite: kinematics, fwd_velocity, fwd_actuation, euler_step, quaternion integration, and utility tests.

## Summary

As a whole, my goal with this is to get some initial feedback and review from both the community and the maintainers on whether or not this project and implementation is feasible. Any words of advice and feedback is appreciated!

### DIFF
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -46,6 +46,12 @@ from mujoco_warp._src.forward import implicit as implicit
 from mujoco_warp._src.forward import rungekutta4 as rungekutta4
 from mujoco_warp._src.forward import step1 as step1
 from mujoco_warp._src.forward import step2 as step2
+from mujoco_warp._src.grad import SMOOTH_GRAD_FIELDS as SMOOTH_GRAD_FIELDS
+from mujoco_warp._src.grad import diff_forward as diff_forward
+from mujoco_warp._src.grad import diff_step as diff_step
+from mujoco_warp._src.grad import disable_grad as disable_grad
+from mujoco_warp._src.grad import enable_grad as enable_grad
+from mujoco_warp._src.grad import make_diff_data as make_diff_data
 from mujoco_warp._src.inverse import inverse as inverse
 from mujoco_warp._src.io import create_render_context as create_render_context
 from mujoco_warp._src.io import get_data_into as get_data_into
--- a/mujoco_warp/_src/adjoint.py
+++ b/mujoco_warp/_src/adjoint.py
@@ -1,0 +1,107 @@
+"""custom adjoint definitions for MuJoCo Warp autodifferentiation.
+
+This module centralizes all ``@wp.func_grad`` registrations. It must be
+imported before any tape recording so that custom gradients are registered
+with Warp's AD system.
+
+Import this module via ``grad.py`` dont import it directly
+"""
+
+import warp as wp
+
+from mujoco_warp._src import math
+
+
+@wp.func_grad(math.quat_integrate)
+def _quat_integrate_grad(q: wp.quat, v: wp.vec3, dt: float, adj_ret: wp.quat):
+  """Custom adjoint avoiding gradient singularity at |v|=0."""
+  EPS = float(1e-10)
+  norm_v = wp.length(v)
+  norm_v_sq = norm_v * norm_v
+  half_angle = dt * norm_v * 0.5
+
+  # sinc-safe rotation quaternion construction
+  if norm_v > EPS:
+    s_over_nv = wp.sin(half_angle) / norm_v  # sin(dt|v|/2) / |v|
+    c = wp.cos(half_angle)
+    # d(s_over_nv)/dv_j = ds_coeff * v_j
+    ds_coeff = (c * dt * 0.5 - s_over_nv) / norm_v_sq
+  else:
+    s_over_nv = dt * 0.5
+    c = 1.0
+    # Taylor limit: (c*dt/2 - s_over_nv) / |v|^2 -> -dt^3/24
+    ds_coeff = -dt * dt * dt / 24.0
+
+  q_rot = wp.quat(
+    c,
+    s_over_nv * v[0],
+    s_over_nv * v[1],
+    s_over_nv * v[2],
+  )
+
+  # recompute forward intermediates
+  q_len = wp.length(q)
+  q_inv_len = 1.0 / wp.max(q_len, EPS)
+  q_n = wp.quat(
+    q[0] * q_inv_len,
+    q[1] * q_inv_len,
+    q[2] * q_inv_len,
+    q[3] * q_inv_len,
+  )
+
+  q_res = math.mul_quat(q_n, q_rot)
+  res_len = wp.length(q_res)
+  res_inv = 1.0 / wp.max(res_len, EPS)
+
+  # result = normalize(q_res)
+  # adj_q_res_k = adj_ret_k / |q_res| - q_res_k * dot(adj_ret, q_res) / |q_res|^3
+  dot_ar = adj_ret[0] * q_res[0] + adj_ret[1] * q_res[1] + adj_ret[2] * q_res[2] + adj_ret[3] * q_res[3]
+  res_inv3 = res_inv * res_inv * res_inv
+  adj_qr = wp.quat(
+    adj_ret[0] * res_inv - q_res[0] * dot_ar * res_inv3,
+    adj_ret[1] * res_inv - q_res[1] * dot_ar * res_inv3,
+    adj_ret[2] * res_inv - q_res[2] * dot_ar * res_inv3,
+    adj_ret[3] * res_inv - q_res[3] * dot_ar * res_inv3,
+  )
+
+  # q_res = mul_quat(q_n, q_rot)
+  # adj_q_n  = mul_quat(adj_qr, conj(q_rot))
+  # adj_q_rot = mul_quat(conj(q_n), adj_qr)
+  q_rot_conj = wp.quat(q_rot[0], -q_rot[1], -q_rot[2], -q_rot[3])
+  adj_qn = math.mul_quat(adj_qr, q_rot_conj)
+
+  q_n_conj = wp.quat(q_n[0], -q_n[1], -q_n[2], -q_n[3])
+  adj_q_rot = math.mul_quat(q_n_conj, adj_qr)
+
+  # q_rot = (c, s_over_nv * v)
+  # d(c)/dv_j = -s_over_nv * dt/2 * v_j
+  # d(s_over_nv * v_i)/dv_j = ds_coeff * v_j * v_i + s_over_nv * delta_ij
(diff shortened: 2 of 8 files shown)


thowell:
@mar-yan24 thank you for this contribution!

thanks for scoping this to the smooth dynamics for now. have you identified any key blockers for adding differentiation support to other parts of the code like the collision pipeline or constraint solver?

thowell:
@erikfrey @adenzler-nvidia
how do we want to think about an api for differentiation?

@adenzler-nvidia @Kenny-Vilella
what are the performance implications for utilizing `wp.clone`? are these calls something we should consider guarding with `wp.static`?
what considerations should be made for tile operations and differentiability?

mar-yan24:
Thanks for the review @thowell! Yea I scoped to smooth dynamics after doing a rough survey of the collision and solver code. Within the collision pipeline and constraint solver, there are some blockers I need to look into and do some testing on but these are what I mainly found. 

For the collision pipeline, the fundamental issue is that collision detection is a discrete geometric query, with several integer configuration vars (`cltype`/`clface`/`clcorner`) controlling branching with no gradient, and atomic counters make the contact count data-dependent. There are also some algorithmic non-differentiabilities, which im pretty sure is not fixable just by enabling backward (`enable_backward`) on the existing code. Probably gotta bypass the discrete pipeline with smooth distance proxies and custom adjoins in `adjoint.py`.

For constraint solver I think I need to spend some more time looking into it, but from what ive seen, the biggest issues for `enable_backward` permission are `wp.capture_while` (runs until all worlds converse, so number of iterations varies per world -> tape needs fixed computation graph to replay backwards), constraint activation (constraint active depends on `pos < 0`, the discontinuity needs to be avoided, maybe add small perturbation to `qpos`?), and `wp.tile_cholesky` (does not support LTO adjoin generation, no flow for grad). 

I also don't mind taking a look at `wp.clone` and tile operations for performance optimization. I don't really know as much on performance optimization but im a student I have a decent amount of time on my hands lol.

If it would help, I can also write a MD file for the larger changes I need to implement and a high-level roadmap for the whole implementation.

Kenny-Vilella:
Some tile operations have some limitation on the adjoint calculation, this includes:
- `wp.tile_cholesky`
- `wp.tile_matmul`
- `wp.tile_*_solve`

But most of them should be OK.

For `wp.clone`, it's a memory operation so it may be expensive depending on the size of the array.
One thing to consider is whether it is faster to spin up a kernel, but it's not a major issue.

mar-yan24:
I think this should be ready for review, I did some testing in a personal suite and everything looks good. I already found out that multi-step tape AD is not supported properly in Warp so I can probably implement some workarounds for that. Also, you need `sm_70+` version to run on GPU cause Tile Cholesky needs `cuSolverDx.` 

I will probably request a PR draft soon for the constraint solver implementation which I have basically working just need to run some more tests. Let me know if anything looks sus and I'll take a look.

adenzler-nvidia:
Thanks for tackling this - I do have a comment about enable_backward in general. MuJoCo Warp is consumed as a library in most cases, so we need to make sure that enable_backwards is properly propagated from user code into the library, and can be toggled correctly. I realize this is currently broken, mostly due to the fact the differentiability does not work at all, so while enabling we should add proper testing to verify that any configuration coming from user code is respected.

nvtw:
We are working on a slightly approximate collision pipeline differentiability solution for Newton, see https://github.com/newton-physics/newton/pull/2164. Maybe some ideas can be applied to MjWarp as well.

mar-yan24:
@nvtw Thank you for the input, I'll take a look at the newton implementation today and see if there are improvements that can be made for the collision gradients on my end.

erikfrey:
hi @mar-yan24 - what a cool PR!  Thanks for taking this on.

Just as a heads up, we're still gathering requirements from teams / potential partners on what are the active use cases for differentiable simulations, which in turn feeds into thinking through ideal API design and needed code changes.

As a result, I think it's going to be a while yet before we're ready to give specific feedback on this PR and get it into a mergeable state.  So I want to offer a couple of options in the meantime:

* If you'd like to keep this PR open, we can come back to it from time to time as we collect new requirements or get feedback from the community
* If you have a specific research aim in mind and you're gung ho to complete this engineering, you could also consider maintaining a fork of mjwarp that supports autodiff, and we would be happy to refer folks to it who are interested in autodifferentiation in the meantime.

mar-yan24:
@erikfrey thanks for the input! Keeping this PR open might be a good call, but if anyone within the community wants to use a more polished version I do have my own fork which has a basic full pipeline working. I think having this PR open or another forum to keep a feedback loop might be nice especially if planned for future merging? I also don't know how much the fork might drift in the future from main but anyone is welcome to try out the AD implementation: [https://github.com/mar-yan24/mujoco_warp/tree/mark/autodifferentiation3](url).

If you or the rest of the team have any comments or insights feel free to just pop in and comment here!

johnnynunez:
hello guys,
how is it going this?

johnnynunez:
Continuing this differentiability work (newton → mjwarp → warp stack; determinism side tracked in #1281/#1300/NVIDIA/warp#1355).

**Rebased branch:** https://github.com/johnnynunez/mujoco_warp/tree/diff/mjwarp-autodiff — this PR rebased onto current main (122 commits forward). Full suite passes: 1072 passed, 23 skipped, including all 12 grad tests, on RTX PRO 6000 Blackwell / CUDA 13.3.

Two things worth review beyond the mechanical rebase:

1. **Conflict resolution in `smooth.py`:** main replaced the unfused sparse `_solve_LD_sparse` kernels with a fused single-kernel version (`enable_backward=False`, in-place updates — not differentiable). I kept both: the fused kernel remains the default, and `_solve_LD_sparse` dispatches to this PR's unfused differentiable kernels when `x.requires_grad or y.requires_grad`. Zero cost for non-grad users, gradients preserved for diff users.

2. **A bug fix to this PR** (also as a standalone commit on https://github.com/johnnynunez/mujoco_warp/tree/fix/autodiff-cdof-dot for direct cherry-pick): the unrolled `_process_joint_vel` introduced for warp AD compatibility dropped the explicit zeroing of `cdof_dot` rows 0-2 for FREE joints. Caught by `smooth_test.py::test_com_vel`, which prefills the array with inf — stale/garbage values leak into downstream RNE otherwise. Fix is 3 lines restoring the zero writes.

@mar-yan24 happy to open a PR into your branch with the rebase + fix, or the team can take the rebased branch directly — whichever keeps this moving.

johnnynunez:
The rebased branch is now also up as a reviewable PR into @mar-yan24's fork: https://github.com/mar-yan24/mujoco_warp/pull/7 (full rebase, 122 commits forward, 1072 tests incl. all 12 grad tests) — alongside https://github.com/mar-yan24/mujoco_warp/pull/5 (just the cdof_dot fix against the current branch). Whichever is easier to adopt.

johnnynunez:
Progress update: differentiability now extends **through the constraint solver** — https://github.com/mar-yan24/mujoco_warp/pull/9 (stacked on the rebase in mar-yan24/mujoco_warp#7).

Approach: implicit-function-theorem adjoint at the converged solver fixed point (active set frozen) instead of taping the iterative solve — one linear solve with the Newton Hessian `H = M + Jᴬᵀ Dᴬ Jᴬ` yields gradients w.r.t. `qfrc_smooth`, `efc_aref`, `efc_D`, `efc_J`, which chain into this branch's smooth-dynamics tape to reach `qpos/qvel/ctrl`. Validated against finite differences through the full GPU solver on persistent-contact scenes (sparse + dense), 1078 tests passing.

Current backward is research-grade (host NumPy solves per world); a GPU tile-Cholesky backward reusing the solver's factorization is the natural follow-up if there's interest in this direction for the main effort here.

johnnynunez:
Full autodiff chain is now ported onto current main and green: 1/3 smooth dynamics (mar-yan24/mujoco_warp#7) → 2/3 constraint solver implicit diff (mar-yan24/mujoco_warp#10) → 3/3 smooth contact AD (mar-yan24/mujoco_warp#11). 1093 tests passing on the tip, including the 29 gradient tests (solver KKT adjoint, euler-damp adjoint, smooth contact geometry gradients, friction surrogates). Together with the determinism stack (#1281/#1300), both chains from this effort are now reviewable against today's upstream.

thowell:
@mar-yan24 @johnnynunez what is the status of this pr?

thowell:
please see #1423

PR #1276, opened 2026-04-02
Fix TileSet dataclass behavior

tkelestemur:
## Summary

`TileSet` is a structural descriptor for tiled matrix layout metadata. Right now,  it relies on the default `dataclass` behavior, which means two independently constructed but semantically identical `TileSet`s do not compare equal, and `TileSet` is also unhashable.

That becomes a problem for downstream consumers that treat `TileSet` as structural metadata. Equivalent models built from separate compiles should compare equal when their tile layouts match.

A minimal repro is:

```python
import numpy as np
import warp as wp

from mujoco_warp._src.types import TileSet

wp.init()

for count in (1, 3):
    a = TileSet(wp.array(np.arange(count) * 6, dtype=int), 16)
    b = TileSet(wp.array(np.arange(count) * 6, dtype=int), 16)

    print(f"count={count}")
    print("equal:", a == b)
    print("hashes_equal:", hash(a) == hash(b))
```

## Fix
Define structural `TileSet.__eq__ `and `TileSet.__hash__` in terms of: `size` and `the value of adr`.
For hashing, this patch uses `(size, adr.dtype.str, adr.shape, adr.tobytes())`, so equal tile layouts hash equal.

### DIFF
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -17,6 +17,7 @@ import enum
 from typing import Callable
 
 import mujoco
+import numpy as np
 import warp as wp
 
 MJ_MINVAL = mujoco.mjMINVAL
@@ -795,6 +796,15 @@ class TileSet:
 
   adr: wp.array(dtype=int)
   size: int
+
+  def __eq__(self, other) -> bool:
+    if self.__class__ is not other.__class__:
+      return NotImplemented
+    return self.size == other.size and np.array_equal(np.asarray(self.adr.numpy()), np.asarray(other.adr.numpy()))
+
+  def __hash__(self) -> int:
+    adr = np.asarray(self.adr.numpy())
+    return hash((self.size, adr.dtype.str, adr.shape, adr.tobytes()))
 
 
 @dataclasses.dataclass
--- a/mujoco_warp/_src/types_test.py
+++ b/mujoco_warp/_src/types_test.py
@@ -18,6 +18,7 @@
 import dataclasses
 
 import mujoco
+import numpy as np
 import warp as wp
 from absl.testing import absltest
 from absl.testing import parameterized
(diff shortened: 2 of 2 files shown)


erikfrey:
hi @tkelestemur - can you tell me a bit more about how this fix impacts either mjwarp or mjwarp users?

For example, is this causing warp kernel cache misses?  What problem is this fix solving for you?

tkelestemur:
@erikfrey  The concrete problem we hit is downstream in the MJX WARP path, where `qM_tiles: tuple[TileSet, ...]` is treated as part of the model’s structural metadata. In our codebase we build two independently constructed but semantically identical WARP models for the same task (one for training, one for evaluation), and both later flow through the same jitted simulation step.

A reduced version of what that looks like on the MJX side is:

```python
import jax
from mujoco import mjx

step_fn = jax.jit(mjx.step)

model_a = mjx.put_model(mj_model_a, impl=mjx.Impl.WARP)
model_b = mjx.put_model(mj_model_b, impl=mjx.Impl.WARP)

data_a = mjx.make_data(mj_model_a, impl=mjx.Impl.WARP)
data_b = mjx.make_data(mj_model_b, impl=mjx.Impl.WARP)

step_fn(model_a, data_a)  # OK
step_fn(model_b, data_b)  # fails for our multi-object case before this patch
```

As I understand it mujoco_mjx syncs the mujoco_warp from here opened the PR here instead of mujoco_mjx.

Let me know if you need more context how this is affecting us.

erikfrey:
That's super helpful thanks - in that case adding @btaba to review.

btaba:
Thanks @tkelestemur , were any changes needed on the MJX side as well or just this? Can you share your full-fix branch so we can take a look as well? Otherwise LGTM

tkelestemur:
@btaba no separate MJX-side fix is needed needed assuming mjwarp is synced to mujoco side, but we do currently need a downstream monkeypatch until this change lands in the MJX copy we use.

By “full-fix branch,” I assume you mean the downstream branch in our application where we reproduced this and applied a temporary local workaround. I cannot share that since it's in our internal codebase but here's the monkey patch we applied to fix the mjx side:

```python
def _patch_warp_tileset_equality() -> None:
    """Makes WARP tile metadata comparisons safe for JAX cache keys.
    MJX treats only fields typed as `jax.Array` as pytree data. WARP's
    `ModelWarp.qM_tiles` therefore ends up in static pytree metadata, where JAX
    compares `TileSet` instances for cache-key equality. The upstream dataclass
    equality falls through to NumPy array equality on `TileSet.adr`, which
    raises for multi-element arrays and breaks repeated `jax.jit` calls across
    independently constructed multi-object environments.
    """
    tile_set_cls = mjx_warp_types.TileSet
    if getattr(tile_set_cls, "eka_tileset_equality_patched", False):
        return

    def _tileset_eq(
        self: mjx_warp_types.TileSet, other: object
    ) -> bool | NotImplementedType:
        if type(self) is not type(other):
            return NotImplemented
        return self.size == other.size and np.array_equal(self.adr, other.adr)

    def _tileset_hash(self: mjx_warp_types.TileSet) -> int:
        adr = np.asarray(self.adr)
        return hash((self.size, adr.dtype.str, adr.shape, adr.tobytes()))

    tile_set_cls.__eq__ = _tileset_eq
    tile_set_cls.__hash__ = _tileset_hash
    tile_set_cls.eka_tileset_equality_patched = True
```

btaba (approved):

PR #1278, opened 2026-04-07
Fix the box normal count for edge features in collision_gjk

camevor:
This PR fixes a typo in `_box_normals` that causes it to return a count of 2 when only 1 normal was written.

The caller (`_aligned_faces`) reads the unwritten slot from a `wp.empty()` workspace array, which may contain heap garbage that passes the dot-product threshold in `_aligned_faces`, producing corrupted contact positions and NaN in qpos.


kbayes:
Sweet! Nice catch! :)

kbayes (approved):

PR #1282, opened 2026-04-11
Fix cache_kernel hashing for numpy scalars

### LINKED ISSUE #1277: cache_kernel decorator does not work with numpy int
While doing some experimentation, I hit a limitation of the `cache_kernel` decorator that we may want to address at some point even though it is not needed right now.

Simple repro:
```
import numpy as np
from mujoco_warp._src.warp_util import cache_kernel


@cache_kernel
def make_kernel(n: int):
  """Returns n, simulating a kernel parameterized by n."""
  return n


# Python ints: works correctly
assert make_kernel(1) == 1
assert make_kernel(6) == 6
print("Python ints: OK")

# numpy scalars: BUG — second call returns cached result from first
a = make_kernel(np.int64(1))
b = make_kernel(np.int64(6))
print(f"np.int64(1) -> {a}")
print(f"np.int64(6) -> {b}")
assert b == 6, f"Expected 6, got {b} (reused cached result from np.int64(1))"
```

mar-yan24:
This PR fixes #1277. After doing some digging, I'm pretty sure the `hasattr(a, "size")` branch in `_hash_arg` used `TileSet.size` as a cache key. I guess luckily #1276 was merged recently cause now `TileSet` has a proper `__hash__`. Since this exists, I am pretty sure these two lines are redundant and causes numpy scalar collisions. Feel free to double check my analysis in case this is used somewhere else in pipeline where I can't find.

### DIFF
--- a/mujoco_warp/_src/warp_util.py
+++ b/mujoco_warp/_src/warp_util.py
@@ -127,8 +127,6 @@ def cache_kernel(func):
   @functools.wraps(func)
   def wrapper(*args):
     def _hash_arg(a):
-      if hasattr(a, "size"):
-        return a.size
       if isinstance(a, list):
         return hash(tuple(a))
       return hash(a)


thowell (approved):
lgtm, thanks @mar-yan24!

thowell:
@mar-yan24 if this pr's branch is synced with main i think the ci failures for the mujoco version should be resolved. thanks!

mar-yan24:
@thowell should be good to go. Thanks for the clarification on that check!

PR #1283, opened 2026-04-12
Semantic segmentation parity

tkelestemur:
## Summary

This PR adds first-class semantic segmentation to MJWarp's renderer.

- Adds `mjw.get_semantic_segmentation(...)`, returning per-pixel `(object_id, object_type)` pairs aligned with MuJoCo segmentation semantics.
- Preserves the existing `mjw.get_segmentation(...)` API and buffer for backward compatibility.
- Reports regular geom hits as `(geom_id, mjOBJ_GEOM)` and flex hits as `(flex_id, mjOBJ_FLEX)`, instead of reducing flex hits to the legacy `-2` sentinel only.
- Exposes the new API from the public package surface and adds test coverage for the new renderer/accessor behavior.
- Adds a dedicated semantic segmentation notebook demo, including flex examples, RGB/depth rendering, legacy segmentation, semantic object ids, semantic object types, and semantic flex ids.

PS: The new notebook imports `mujoco_warp` directly from the local checkout and disables IPython autoreload for the session, since Warp kernels require file-backed Python source in notebook environments.

Aside from the regular tests, I've also run `pytest contrib/kernel_analyzer/kernel_analyzer/ast_analyzer_test.py`.


thowell:
@tkelestemur thank you for contributing this feature to mujoco warp!

@StafaH @btaba would it make sense to combine this feature (in a potentially breaking way) with the existing segmentation api?

@tkelestemur instead of adding a separate notebook for this feature, can we add an example to the existing tutorial notebook?

StafaH:
Hi @tkelestemur, thanks for the PR!

In this case I believe this new feature should not be implemented in MJWarp, and can be implemented at the framework level that you use downstream. The existing segmentation output provides enough information to perform post processing downstream e.g. by implementing semantic segmentation as an extension in Mjlab with a kernel that converts the existing integer id to another value inplace.

tkelestemur:
@StafaH The existing segmentation buffer is sufficient to reconstruct semantic output for geom hits and background, but not for flex hits, because all flex hits are collapsed to -2 and lose the underlying `flex_id`.

StafaH:
I agree, we should fix existing segmentation output. I think what @thowell mentioned is correct, we should focus on having the existing segmentation output match MuJoCo exactly instead of making a seperate output, and we can update the tests to check that this segmentation output matches exactly the C MuJoCo output (similar to what was done for depth recently).

WDYT @thowell?

@tkelestemur this might also require opening an issue and a PR in mjlab to fix any changes that happen downstream there.

tkelestemur:
@StafaH Sounds good -- I'll update this PR to use the existing segmentation API and also open an issue and a PR in mjlab.

tkelestemur:
@thowell I've removed the extra notebook and added a section to the current notebook.

@StafaH I've reverted the new segmentation api and updated the current one. Soon, I'll open a PR on the mjlab side.

mujoco_warp/_src/io_test.py
```diff
@@ -2023,7 +2026,10 @@ def test_put_data_nefc_zero_dense(self):
     mujoco.mj_forward(mjm, mjd)
 
     self.assertFalse(mujoco.mj_isSparse(mjm))
-    self.assertEqual(mjd.nefc, 0)
+    if mjd.nefc != 0:
```
thowell: is this change necessary?

mujoco_warp/_src/io_test.py
```diff
@@ -1973,9 +1973,12 @@ def test_segmentation_from_camera_output(self):
     rc = mjwarp.create_render_context(mjm, nworld=1, cam_res=(32, 32))
     mjwarp.render(m, d, rc)
 
-    seg = rc.seg_data.numpy()
-    self.assertTrue(np.any(seg >= 0), "Expected geom hits from auto-detected seg")
-    self.assertGreater(np.unique(seg).shape[0], 1)
+    seg = wp.zeros((1, 32, 32, 2), dtype=int)
```
thowell: can we create `seg` from the render context?

notebooks/tutorial.ipynb
```diff
@@ -508,6 +508,82 @@
         "media.show_image(depth_grid, cmap='gray', vmin=0, vmax=1)"
       ]
     },
+    {
```
thowell: i think we can simplify this and make it similar to the depth render example (essentially a comment plus 2 lines of code and then an additional segmentation image to display.

thowell:
the updated pr looks great! added a few comments. please let us known if you have any questions. thanks!

mujoco_warp/_src/io_test.py
tkelestemur: Updated to size the segmentation output from rc.cam_res instead of hardcoding the resolution. Is that what you meant?

mujoco_warp/_src/io_test.py
tkelestemur: Removed this change.

notebooks/tutorial.ipynb
tkelestemur: Simplified the tutorial example to extend the existing rendering cell with a small segmentation extraction/display block. Lmk how it looks.

tkelestemur:
Thanks @thowell I've addressed your reviews. I also opened an downstream PR for mjlab here: https://github.com/mujocolab/mjlab/pull/911

mujoco_warp/_src/io_test.py
thowell: lets change this back to what you had before `wp.zeros((1, 32, 32, 2), dtype=int)`. thanks!

notebooks/tutorial.ipynb
```diff
@@ -505,7 +505,18 @@
...
+        "# Extract segmentation data. Channel 0 stores object ids and channel 1 stores\n",
+        "# the corresponding MuJoCo object types.\n",
+        "segmentation_data = wp.zeros((NWORLD, CAM_RES[1], CAM_RES[0], 2), dtype=int)\n",
+        "mjw.get_segmentation(rc, camera_index=0, seg_out=segmentation_data)\n",
+        "\n",
+        "# Display object ids in a grid.\n",
+        "segmentation_grid = segmentation_data.numpy()[..., 0].reshape(4, 4, CAM_RES[1], CAM_RES[0])\n",
+        "segmentation_grid = segmentation_grid.transpose(0, 2, 1, 3)\n",
+        "segmentation_grid = segmentation_grid.reshape(4 * CAM_RES[0], 4 * CAM_RES[1])\n",
+        "media.show_image(segmentation_grid, cmap='viridis', vmin=-1, vmax=max(1, int(segmentation_grid.max())))\n"
```
thowell: please confirm that this line runs as expected

```
ValueError: Type int32 is not a valid media data type (uint or float).
```

converting the array to floats produces

Image

mujoco_warp/_src/render_util.py
```diff
@@ -211,32 +211,35 @@ def get_depth(rc: RenderContext, camera_index: int, depth_scale: float, depth_ou
 @wp.kernel
 def _extract_seg_kernel(
   # In:
-  seg_data: wp.array2d[int],
+  seg_data: wp.array2d[wp.vec2i],
   seg_adr: wp.array[int],
   camera_index: int,
   # Out:
-  seg_out: wp.array3d[int],
+  seg_out: wp.array4d[int],
```
StafaH: Any reason we can't use wp.vec2i here as well and keep the array3d?

mujoco_warp/_src/render_util.py
```diff
@@ -211,32 +211,35 @@ def get_depth(rc: RenderContext, camera_index: int, depth_scale: float, depth_ou
...
-  """Extract per-pixel geom IDs from the render context buffers for a given camera index."""
+  """Extract per-pixel `(object_id, object_type)` pairs for a camera."""
   worldid, pixelid = wp.tid()
   xid = pixelid % seg_out.shape[2]
   yid = pixelid // seg_out.shape[2]
 
   seg_adr_offset = seg_adr[camera_index]
-  seg_out[worldid, yid, xid] = seg_data[worldid, seg_adr_offset + pixelid]
+  pair = seg_data[worldid, seg_adr_offset + pixelid]
+  seg_out[worldid, yid, xid, 0] = pair[0]
```
StafaH: Same as above, if we have vec2i array, we can store the value directly

mujoco_warp/_src/io_test.py
```diff
@@ -1973,9 +1973,13 @@ def test_segmentation_from_camera_output(self):
     rc = mjwarp.create_render_context(mjm, nworld=1, cam_res=(32, 32))
     mjwarp.render(m, d, rc)
 
-    seg = rc.seg_data.numpy()
-    self.assertTrue(np.any(seg >= 0), "Expected geom hits from auto-detected seg")
-    self.assertGreater(np.unique(seg).shape[0], 1)
+    cam_w, cam_h = map(int, rc.cam_res.numpy()[0])
+    seg = wp.zeros((1, cam_h, cam_w, 2), dtype=int)
+    mjwarp.get_segmentation(rc, 0, seg)
+    seg = seg.numpy()
```
StafaH: io tests should only check that the the initialization of render context was done correctly. This can be simplified to just check that the proper seg_out shape/size was made. See the above tests. Checking render output should happen in the render_test (which I see you already did).

mujoco_warp/_src/render_test.py
```diff
@@ -131,7 +131,35 @@ def test_render_rgb_and_segmentation(self):
...
+
+    rc = mjw.create_render_context(
+      mjm,
+      nworld=1,
+      cam_res=(cam_w, cam_h),
+      render_seg=[True],
+    )
+    mjw.render(m, d, rc)
+
+    warp_seg = wp.zeros((1, cam_h, cam_w, 2), dtype=int)
```
StafaH: This can all be avoided if you flatten the mujoco numpy output similar to how it is done for the depth test.

tkelestemur:
sorry forgot to enable `render_seg`. This is what I'm getting right now:

image

so should be good to go.

mujoco_warp/_src/render_test.py
```diff
@@ -131,7 +131,35 @@ def test_render_rgb_and_segmentation(self):
...
+    mjw.get_segmentation(rc, 0, warp_seg)
+    warp_seg_np = warp_seg.numpy()[0]
+
+    with mujoco.Renderer(mjm, height=cam_h, width=cam_w) as renderer:
+      renderer.update_scene(mjd, camera=0)
+      renderer.enable_segmentation_rendering()
+      mj_seg = renderer.render()
+
+    np.testing.assert_array_equal(warp_seg_np[..., 1], mj_seg[..., 1])
```
StafaH: Can we not test the full array matches? The goal of the recent changes was have the output match exactly the C MuJoCo output (other than shape, which would need to be flattened), so warp_seg == mj_seg (no slicing)

StafaH (commented):
This is shaping up nicely. Left some comments.

StafaH:
Also might be worth having the seg test xml include 1 flex object for completeness

mujoco_warp/_src/render_util.py
tkelestemur: Updated to use `wp.vec2i `and keep `seg_out` as a 3D array.

mujoco_warp/_src/render_util.py
tkelestemur: Done. The kernel now stores the `wp.vec2i` value directly.

mujoco_warp/_src/io_test.py
tkelestemur: Simplified this test to only verify render-context initialization `(seg_data, seg_adr, render_seg)`. Output validation stays in `render_test.py.`

mujoco_warp/_src/render_test.py
tkelestemur: Updated to flatten the MuJoCo segmentation output before comparison, matching the depth test pattern.

mujoco_warp/_src/render_test.py
tkelestemur: Updated to compare the full flattened segmentation arrays directly.

tkelestemur:
@StafaH thanks for the reviews, I think I addressed all of them. 

In https://github.com/google-deepmind/mujoco_warp/pull/1283/commits/f4bf916e02f5b18a238a4977131b4bee995a08e9, I also dded a small flex object to the synthetic segmentation XML for completeness, and the test now asserts the model includes one flex before checking the render-context segmentation setup.

Let me know if this is enough.

mujoco_warp/_src/render_test.py
```diff
@@ -131,7 +131,32 @@ def test_render_rgb_and_segmentation(self):
...
+
+    rc = mjw.create_render_context(
+      mjm,
+      nworld=1,
+      cam_res=(cam_w, cam_h),
+      render_seg=[True],
+    )
+    mjw.render(m, d, rc)
+
+    warp_seg = wp.zeros((1, cam_h, cam_w), dtype=wp.vec2i)
```
StafaH: Can we skip this and just compare directly to `rc.seg_data.numpy()` below?

StafaH (approved):
Thanks @tkelestemur, left one last comment, but otherwise LGTM

mujoco_warp/_src/render_test.py
tkelestemur: done in https://github.com/google-deepmind/mujoco_warp/pull/1283/commits/f77e74761819986824b7ef624233c750838bbcc2

thowell (approved):

thowell:
@tkelestemur some of the checks are not passing, please take a look. might just need to sync with main. thanks!

thowell:
there is a segmentation fault with one of the checks, please take a look. thanks! https://github.com/google-deepmind/mujoco_warp/actions/runs/24784925329/job/72532705830?pr=1283

tkelestemur:
@thowell on it!

tkelestemur:
@thowell I pushed a fix for the Linux CPU segfault in `98d6411`.

Root cause was that `build_flex_bvh()` creates a grouped `wp.Mesh`, but we were looking up roots with the BVH path. For flexes the correct lookup is `mesh_get_group_root`, and in this Warp version that is only exposed as a kernel builtin, not a normal Python host API. So the final change keeps the surface area small:

- add a private `_compute_mesh_group_roots` helper just to call the mesh-specific builtin
- switch flex BVH root lookup over to that helper
- fix `flex_group_root` layout in `create_render_context()` to match the renderer indexing (`[worldid, flexid]`)

I considered a smaller `nworld == 1 -> -1` workaround, but that would only paper over the single-world case instead of fixing the underlying mesh-vs-BVH mismatch. I also considered splitting flexes into one mesh per world, but that would be a much broader change.

I rechecked the targeted regressions locally on macOS and on Linux CPU (`l4`):
- `render_util_test.py::test_get_segmentation_preserves_flex_ids`
- `render_test.py -k segmentation_matches_mujoco` (skips on headless Linux as expected)
- `io_test.py -k segmentation_from_camera_output`

All of those pass with this change, so this should address the CI segfault without broadening the PR beyond the failing path.

mujoco_warp/_src/bvh.py
```diff
@@ -245,6 +245,19 @@ def compute_bvh_group_roots(
   group_root_out[tid] = root
 
 
+# Warp exposes mesh group-root lookup as a kernel builtin in this version.
+@wp.kernel
+def _compute_mesh_group_roots(
```
StafaH: nit: the _ is not necessary

thowell: @tkelestemur can we update this as @StafaH suggests? thanks!

tkelestemur: done in https://github.com/google-deepmind/mujoco_warp/pull/1283/commits/0fdb3e62d6a5e526b4ff40e1d97bf023456e3f5f

thowell:
@tkelestemur thank you for this contribution!

PR #1289, opened 2026-04-14
warp-lang 1.12.1

thowell:


adenzler-nvidia:
Is it required to also bump the min spec in pyproject.toml on your side? I would just bump the version in the lock file - pip will resolve to the latest released anyway, and 1.12.1 should be backwards compatible, so the min version bump just breaks workflows for users that can't upgrade for some other reason?

thowell:
@adenzler-nvidia wanted to checked that we could upgrade to the latest warp release version (1.12.1), but we don't need to upgrade so closing this pr

PR #1290, opened 2026-04-14
ci cache for warp with stable dependencies tests

thowell:


.github/workflows/ci.yml
```diff
@@ -62,56 +62,67 @@ jobs:
...
-  test-locked-deps:
-    name: Test with locked uv dependencies
-    # Tests against exact versions pinned in uv.lock
+  test:
+    name: Test with ${{ matrix.dep-type }} dependencies (${{ matrix.python-version }})
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["3.10", "3.13"]
+        dep-type: [locked, stable]
```
erikfrey: nit: can we s/locked/"locked uv" so we don't have to reconfigure the test names?

erikfrey (approved):
LGTM, just one nit

.github/workflows/ci.yml
thowell: done

PR #1291, opened 2026-04-14
CUDA support fix for cache_kernel numpy decorator

mar-yan24:
This follows up on #1277/#1282.

Hello again sorry for the last PR, I should've done a better test. The earlier patch fixed the numpy-scalar cache collision, but I missed that removing the old fallback caused dense CUDA graph-capture paths to hash `TileSet` directly. I think this follow-up fixes both issues. 

This patch should fix both sides:
  - numpy scalars are keyed by value
  - `TileSet` exposes an explicit capture-safe cache key for `cache_kernel`
  - `TileSet.__hash__` remains unchanged

Hope this works better and let me know if the regression tests show otherwise.


mar-yan24:
@thowell this should fix the issue from last time.

PR #1293, opened 2026-04-15
dc motor

thowell:
[mujoco dc motor](https://mujoco.readthedocs.io/en/latest/_static/dcmotor.pdf)

### DIFF
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -138,10 +138,13 @@ def _next_activation(
   actuator_actnum: wp.array[int],
   actuator_actlimited: wp.array[bool],
   actuator_dynprm: wp.array2d[vec10f],
+  actuator_gainprm: wp.array2d[vec10f],
+  actuator_biasprm: wp.array2d[vec10f],
   actuator_actrange: wp.array2d[wp.vec2],
   # Data in:
   act_in: wp.array2d[float],
   act_dot_in: wp.array2d[float],
+  actuator_velocity_in: wp.array2d[float],
   # In:
   act_dot_scale: float,
   limit: bool,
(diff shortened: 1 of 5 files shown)


mujoco_warp/_src/forward.py
```diff
@@ -670,6 +726,97 @@ def _actuator_force(
       dynprm = actuator_dynprm[worldid % actuator_dynprm.shape[0], uid]
       act = act_in[worldid, act_last]
       act_dot = util_misc.muscle_dynamics(ctrl, act, dynprm)
+    elif dyntype == DynType.DCMOTOR:
+      gainprm = actuator_gainprm[worldid % actuator_gainprm.shape[0], uid]
+      slots = util_misc.dcmotor_slots(dynprm, gainprm)
+      adr = act_first
+
+      # Initialize act_dot to 0.0 in case no slots are active (should not happen for DCMOTOR)
```
yuvaltassa: Why should not happen for DCMOTOR?
I expect most people will use the stateless version?
To be clear I don't think the code is wrong.

mujoco_warp/_src/util_misc.py
```diff
@@ -599,6 +600,61 @@ def muscle_dynamics(control: float, activation: float, prm: vec10) -> float:
   return dctrl / wp.max(MJ_MINVAL, tau)
 
 
+@wp.func
+def dcmotor_slots(dynprm: types.vec10, gainprm: types.vec10) -> types.vec6i:
+  s = types.vec6i(-1, -1, -1, -1, -1, 0)
```
yuvaltassa, line 622: Some comment here about the semantics of the slots? This is very unreadable as it is

yuvaltassa (approved):
Small nits but LGTM

mujoco_warp/_src/forward.py
thowell: removed this comment

mujoco_warp/_src/util_misc.py
thowell: added a docstring

PR #1295, opened 2026-04-16
Optimization: efc_contact kernels

Kenny-Vilella:
Optimized both `_contact_pyramidal` and `_contact_elliptic`:
- Separated the pipeline into sparse/dense instead of pyramidal/elliptic
- Break down the kernel into 3 kernels: Initialization, Jacobian calculation, constraint rows update
- Used a reinterpret to avoid unnecessary loads
- Used `dof_affects_body` to simplify the Jacobian calculation
- Dense pipeline now use tiles.
- Sparse pipeline is not as fast as it should be but I am out of idea

Some numbers on my RTX4080:
For `humanoid`, average of 10 runs:
Command:
```
python mujoco_warp/testspeed.py --function=step benchmarks/humanoid/humanoid.xml --nworld=8192 --nstep=1000 --njmax=60 --nconmax=25 -o opt.ls_parallel=true -o opt.solver=newton -o opt.cone=${CONE} -o opt.graph_conditional=true -o opt.integrator=euler
```
Case                       | main (RTF)  |  This PR (RTF) |  Delta
------------------------  | ----------------- |  ------------------- | -------
dense / pyramidal  |  12932          |  13798               |  -6.7
dense / elliptic        |  7173           |  7468                 |  -4.1

For `three_humanoids`, average of 10 runs:
Command:
```
python mujoco_warp/testspeed.py --function=step  benchmarks/humanoid/three_humanoids.xml --nworld=8192 --nstep=${NSTEP} --njmax=190 --nconmax=70  -o opt.ls_parallel=true -o opt.solver=newton -o opt.cone=${CONE} -o opt.graph_conditional=true -o opt.integrator=euler
```
Case                       | main (RTF)  |  This PR (RTF) |  Delta
------------------------  | ----------------- |  ------------------- | -------
sparse / pyramidal   |  2179            |  2244              |  -3.0
sparse / elliptic         |  396             |  407                 |  -2.8

And results on the benchmark, average of 5 runs. Excluded `aloha_cloth` because have overflow even on main.

Case                       | main (step)  |  This PR (step) |  Delta (step)  | main (make_constraint)  |  This PR (make_constraint) |  Delta (make_constraint)  | 
------------------------  | ----------------- |  ------------------- | -------  | ----------------- |  ------------------- | -------
`aloha_pot`                        |   573.09           |   580.54             |    1.3               |   14.39           |   14.21             |   -1.2
`aloha_sdf`                        |    3148.92          |   3163.37             |  0.5                  |   16.16           |   15.05             |  -6.9
`apptronik_apollo_flat`       |   416.91           |  412.85              |   -1.0                |  38.78            |    30.02            |   -22.5
`apptronik_apollo_hfield`   |     615.80         |   412.85             |    -0.15               |    37.59           |    31.10             |   -17.3
`apptronik_apollo_terrain`  |    1702.85          |   1710.17             |   0.4                |   43.38           |   37.41             |  -13.3
`cloth`                                 |   545598.72           |   552445.89             |  1.2                 |   3250.11           |   3408.51             |  4.9 
`franka_emika_panda`       |   66.52           |  67.97              |   2.1                |   1.12              |   1.61   |  43.7
`humanoid`                         |  330.23            |    313.10            |   -5.2                |   45.94              |  27.89  |  -39.0
`three_humanoids`             |   2034.70           |   2051.66             |   0.8                |   99.61            |   85.38             | -14.3  

So a fairly good speedup for dense model with large number of DoFs.
An OK speedup for sparse model.
A fairly large perf decrease for very simple system, the actual perf decrease on the franka should be something like 0.7%.


mujoco_warp/_src/io.py
```diff
@@ -266,6 +266,24 @@ def _check_friction(name: str, id_: int, condim: int, friction, checks):
   m.jnt_limited_ball_adr = np.nonzero(mjm.jnt_limited & (mjm.jnt_type == mujoco.mjtJoint.mjJNT_BALL))[0]
   m.dof_tri_row, m.dof_tri_col = np.tril_indices(mjm.nv)
 
+  # precompute dof_affects_body: which DOFs affect each body
```
thowell, line 269: @yuvaltassa is this information contained in an `mjModel` field?

Kenny-Vilella:
Just two more things:
- With this change, same contact constraint rows are contiguous which may bring additional optimization in the solver pipeline
- It will be easier in the future to build deterministic constraint rows by making the `init` kernel a serial kernel.

mujoco_warp/_src/io.py
yuvaltassa: `dof_bodyid`, obviously? Is this what you're asking?

yuvaltassa: Ah I see this is a map from each dof to all bodies

yuvaltassa: 1. I would call this `body_isdofancestor`. Definitely should start with `body_`.
2. `dof_bodyid` is never 0. The world has no dofs by definition.
3. You don't need to loop over dofs. Given a body `i`, you start with `body_dofadr[i]+body_dofnum[i]-1` and climb up the tree with `dof_parentid`. Or start with `body_dofadr[i]`, forward-mark `body_dofnum[i]` dofs and then climb up, it's the same thing.

yuvaltassa: Also, why are these ints? We could use bytes or even a bitmap, reducing memory 32x in the latter case.

Kenny-Vilella: I focused on the kernel part and did not spend enough time on the `put_model` part, sorry for that.
Before fixing everything, just want to confirm about the bitmap part.
This will have a non-measurable perf improvement, little memory reduction (as long as we assume homogeneous worlds), but more importantly the implementation will be `tile_size` dependent in the `_efc_contact_jac_dense` kernel.

mujoco_warp/_src/constraint.py
```diff
@@ -144,6 +144,7 @@ def _equality_connect(
   eq_solimp: wp.array2d[vec5],
   eq_data: wp.array2d[vec11],
   is_sparse: bool,
+  dof_affects_body: wp.array2d[int],
```
thowell: lets rename this as yuval suggests

mujoco_warp/_src/constraint.py
```diff
@@ -1665,97 +1675,183 @@ def _limit_tendon(
...
-  efc_margin_out: wp.array2d[float],
-  efc_D_out: wp.array2d[float],
-  efc_vel_out: wp.array2d[float],
-  efc_aref_out: wp.array2d[float],
-  efc_frictionloss_out: wp.array2d[float],
-  # Out:
-  efc_nnz_out: wp.array[int],
-):
-  conid, dimid = wp.tid()
+@cache_kernel
```
thowell: can this decorator be removed?

mujoco_warp/_src/constraint.py
```diff
@@ -1665,97 +1675,183 @@ def _limit_tendon(
...
+      for dim in range(ndim):
+        efcid = base_efcid + dim
+        if efcid < njmax_in:
+          efc_J_rowadr_out[worldid, efcid] = rowadr + dim * rownnz
+          efc_J_rownnz_out[worldid, efcid] = rownnz
+
+  return kernel
+
+
+@cache_kernel
```
thowell: can this decorator be removed?

mujoco_warp/_src/constraint.py
```diff
@@ -1858,173 +1919,275 @@ def _contact_pyramidal(
...
-        if is_sparse:
-          dofid = da
+        dofid = da
+
+    efc_Jqvel_out[worldid, efcid] = Jqvel
+
+  return kernel
+
+
+@cache_kernel
```
thowell: can this decorator be removed?

mujoco_warp/_src/constraint.py
```diff
@@ -1858,173 +1919,275 @@ def _contact_pyramidal(
...
+        affects1_tile = wp.tile_load(dof_affects_body[body1], shape=TILE_SIZE, offset=dof_start, bounds_check=True)
+        affects2_tile = wp.tile_load(dof_affects_body[body2], shape=TILE_SIZE, offset=dof_start, bounds_check=True)
+
+        jacp1_tile = wp.tile_map(support._compute_jacp, cdof_tile, offset1, affects1_tile)
+        jacp2_tile = wp.tile_map(support._compute_jacp, cdof_tile, offset2, affects2_tile)
+        jacp_dif_tile = wp.tile_map(wp.sub, jacp2_tile, jacp1_tile)
+
+        jacr1_tile = wp.tile_map(support._compute_jacr, cdof_tile, affects1_tile)
+        jacr2_tile = wp.tile_map(support._compute_jacr, cdof_tile, affects2_tile)
+        jacr_dif_tile = wp.tile_map(wp.sub, jacr2_tile, jacr1_tile)
```
thowell, line 2056: would it make sense to guard computing `jacr_dif_tile`?

mujoco_warp/_src/constraint.py
```diff
@@ -1858,173 +1919,275 @@ def _contact_pyramidal(
...
-  efc_aref_out: wp.array2d[float],
-  efc_frictionloss_out: wp.array2d[float],
-  # Out:
-  efc_nnz_out: wp.array[int],
-):
-  conid, dimid = wp.tid()
 
-  if conid >= nacon_in[0]:
-    return
+@cache_kernel
```
thowell: can this decorator be removed?

mujoco_warp/_src/io.py
thowell: lets add a todo to investigate an alternative approach like using a bitmap for now. thanks!

mujoco_warp/_src/types.py
```diff
@@ -1680,6 +1683,7 @@ class Constraint:
     state: constraint state                           (nworld, njmax_pad)
   warp only fields:
     Ma: M*qacc                                        (nworld, nv)
+    Jqvel: precomputed J*qvel per efc row             (nworld, njmax)
```
thowell: we can probably remove 'precomputed' from this description

thowell:
great work, thanks @Kenny-Vilella!

mujoco_warp/_src/constraint.py
Kenny-Vilella: Done

mujoco_warp/_src/constraint.py
Kenny-Vilella: Done

mujoco_warp/_src/constraint.py
Kenny-Vilella: Done

mujoco_warp/_src/constraint.py
Kenny-Vilella: Done

mujoco_warp/_src/constraint.py
Kenny-Vilella: Done

mujoco_warp/_src/types.py
Kenny-Vilella: Done

mujoco_warp/_src/io.py
Kenny-Vilella: All done

mujoco_warp/_src/constraint.py
Kenny-Vilella: We could do it two ways:
- Calculate in model creation whether jacr_dif_tile will ever be needed and add a static.
- Make the guard at the thread level.

What option you had in mind?
Personally, I am not convinced by the second option because you will still need to use the memory, so not so much gain. But I can still do some investigation.

thowell: not sure it is worth precomputing something in `io.put_model` unless the performance gain is significant and if we think the second option wont lead to significant improvement maybe we leave the code as is, maybe added a todo to investigate the first option in the future?

thowell (approved):

mujoco_warp/_src/constraint.py
thowell: i think we can leave this as is

PR #1296, opened 2026-04-16
Remove cache_kernel decorator

adenzler-nvidia:
## Summary

Remove the `cache_kernel` decorator, its `_KERNEL_CACHE` dict, and all usages. Warp's `module="unique"` kernel dedup handles these cases natively.

## Test plan

All tests on GPU (NVIDIA RTX PRO 6000 Blackwell, sm_120):

- [x] Full test suite: **727 passed, 0 failed, 7 skipped**
- [x] Full test suite `-n 2`: **727 passed, 0 failed, 7 skipped**
- [x] Humanoid benchmark (8192 worlds, 5 runs): no regression (+0.14%, within noise)
- [x] `pre-commit` passes

PR #1297, opened 2026-04-16
Skip Cholesky refactorization when no constraints changed

adenzler-nvidia:
## Summary

- In the incremental Newton solver (pyramidal cone), ~88% of solver iterations have zero constraint state changes. The blocked Cholesky path previously refactorized H every iteration even when H was unchanged.
- Since the blocked path already caches L in `hfactor`, add a per-world check on `changed_efc_count` and skip the factorization when it's zero, going straight to the forward/backward substitution solve.
- Only affects the blocked path (nv > 32). The tiled path (nv ≤ 32) is unchanged.

## Benchmarks (RTX PRO 6000 Blackwell, 8192 worlds)

| Model | nv | Solver kernel speedup | Step speedup |
|---|---|---|---|
| three_humanoids | 81 | 1.9x | ~2% |
| G1 on plane | 35 | 1.3x | ~2% |
| humanoid | 27 | unchanged (tiled path) | unchanged |

Kernel-level nsys breakdown (three_humanoids, 50 steps):
- `cholesky_blocked` (always factorize): 83.3ms (294 calls) → 23.8ms (50 init calls)
- `cholesky_blocked_skip_unchanged`: 43.2ms (243 solver-iteration calls)
- **Net Cholesky savings: -16.3ms**

## Test plan

- [x] `forward_test.py` — 60/60 passed
- [ ] Full CI

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -2785,6 +2785,37 @@ def update_gradient_cholesky_blocked(tile_size: int, matrix_size: int):
   return kernel
 
 
+@cache_kernel
+def update_gradient_cholesky_blocked_skip_unchanged(tile_size: int, matrix_size: int):
+  """Like update_gradient_cholesky_blocked but skips factorization per-world when no constraints changed."""
+
+  @wp.kernel(module="unique", enable_backward=False)
+  def kernel(
+    # In:
+    ctx_done_in: wp.array[bool],
+    ctx_grad_in: wp.array3d[float],
+    ctx_h_in: wp.array3d[float],
+    changed_count_in: wp.array[int],
+    ctx_hfactor: wp.array3d[float],
+    # Out:
+    ctx_Mgrad_out: wp.array3d[float],
+  ):
+    worldid = wp.tid()
+    TILE_SIZE = wp.static(tile_size)
+
+    if ctx_done_in[worldid]:
+      return
+
+    if changed_count_in[worldid] > 0:
+      wp.static(create_blocked_cholesky_func(TILE_SIZE))(ctx_h_in[worldid], matrix_size, ctx_hfactor[worldid])
+
+    wp.static(create_blocked_cholesky_solve_func(TILE_SIZE, matrix_size))(
+      ctx_hfactor[worldid], ctx_grad_in[worldid], matrix_size, ctx_Mgrad_out[worldid]
+    )
+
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -2785,6 +2785,37 @@ def kernel(
   return kernel
 
 
+@cache_kernel
```
thowell: we can remove this after #1296 is merged

mujoco_warp/_src/solver.py
```diff
@@ -2796,8 +2827,12 @@ def padding_h(nv: int, ctx_done_in: wp.array[bool], ctx_h_out: wp.array3d[float]
   ctx_h_out[worldid, dofid, dofid] = 1.0
 
 
-def _cholesky_factorize_solve(m: types.Model, d: types.Data, ctx: SolverContext):
-  """Cholesky factorize ctx.h and solve for Mgrad."""
+def _cholesky_factorize_solve(m: types.Model, d: types.Data, ctx: SolverContext, skip_unchanged: bool = False):
+  """Cholesky factorize ctx.h and solve for Mgrad.
+
+  If skip_unchanged is True (blocked path only), worlds where no constraints
+  changed reuse the cached L in hfactor instead of refactorizing.
```
thowell: instead of 'L in' can we say 'factorization'?

thowell (approved):

kevinzakka:
Hi @adenzler-nvidia, FYI this PR has produced a massive slowdown on CPU (M1 mac). I was previously able to simulate G1-related tasks in mjlab on my CPU at roughly 24-30 FPS and it's barely at 3-4 FPS after this change.

adenzler-nvidia:
Thanks for reporting - just got back from vacation, will investigate. This is very unintuitive. Could you point me to an example/snippet that best shows the slowdowns?

kevinzakka:
Hi @adenzler-nvidia, turns out there was a missing cache_kernel decorator (part of a larger issue that was reverted in mjwarp). https://github.com/google-deepmind/mujoco_warp/commit/88b55fc2696960b927bc12584994bb8412b36558 fixed my CPU slowdown issues.

adenzler-nvidia:
Oh, good to hear. I will look at the cache_kernel side as well now to figure out if we can improve this. So easy to miss these decorators..

PR #1298, opened 2026-04-17
polynomial stiffness and damping

thowell:
https://github.com/google-deepmind/mujoco/commit/efae9157a771957d08e9a4caa9523df4447268c1


mujoco_warp/_src/passive.py
```diff
@@ -86,22 +89,37 @@ def _spring_damper_dof_passive(
...
   dofid = jnt_dofadr[jntid]
+  jnttype = jnt_type[jntid]
   stiffness = jnt_stiffness[worldid % jnt_stiffness.shape[0], jntid]
+  spoly = jnt_stiffnesspoly[worldid % jnt_stiffnesspoly.shape[0], jntid]
   damping = dof_damping[worldid % dof_damping.shape[0], dofid]
+  dpoly = dof_dampingpoly[worldid % dof_dampingpoly.shape[0], dofid]
 
-  has_stiffness = stiffness != 0.0 and not (opt_disableflags & DisableBit.SPRING)
-  has_damping = damping != 0.0 and not (opt_disableflags & DisableBit.DAMPER)
+  has_stiffness = (stiffness != 0.0 or spoly[0] != 0.0 or spoly[1] != 0.0) and not (opt_disableflags & DisableBit.SPRING)
```
yuvaltassa, line 98: I like how you're inferring mjNPOLY in lots of places, but not here. One day mjNPOLY will increase and we'll have a hard-to-find bug... One effective way of preventing that is to have a check somewhere that mjNPOLY == 2 and error if not? ONTOH this would just miss a term, IDK

mujoco_warp/_src/passive.py
```diff
@@ -86,22 +89,37 @@ def _spring_damper_dof_passive(
...
+      for i in range(6):
+        qfrc_spring_out[worldid, dofid + i] = 0.0
+    elif jnttype == JointType.BALL:
+      for i in range(3):
+        qfrc_spring_out[worldid, dofid + i] = 0.0
+    else:
+      qfrc_spring_out[worldid, dofid] = 0.0
 
   if not has_damping:
-    qfrc_damper_out[worldid, dofid] = 0.0
```
yuvaltassa, line 99: Is this fixing an unrelated preexisting bug?

mujoco_warp/_src/util_misc.py
```diff
@@ -630,3 +630,21 @@ def inside_geom(pos: wp.vec3, mat: wp.mat33, size: wp.vec3, geomtype: int, point
     return plocal[2] < 0.0
 
   return False
+
+
+@wp.func
+def _poly_force(linear: float, poly: wp.vec2, x: float, flg_odd: int) -> float:
```
yuvaltassa, line 636: I find it weird that these functions are here but the potential function is elsewhere

yuvaltassa (commented):
Mostly LGTM except that I think we need to think about a future increase of mjNPOLY. Either we write code that is robust to that, or we error somewhere... I think?

mujoco_warp/_src/passive.py
thowell: added asserts to `io.put_model` for checking `mjNPOLY` and the second dimension size of `jnt_stiffnesspoly`, `tendon_stiffnesspoly`, `dof_dampingpoly`, and `tendon_dampingpoly`

mujoco_warp/_src/passive.py
thowell: there were existing bugs with `qfrc_spring_out` and `qfrc_damper_out` for joints with more than one dof where only the first dof was zeroed when there was no stiffness/damping. should be resolved now.

mujoco_warp/_src/util_misc.py
thowell: moved the potential function to util_misc.py

yuvaltassa (approved):

PR #1308, opened 2026-04-23
Add skybox rendering to the Batch Renderer

StafaH:
Adds skybox rendering using the skybox texture loaded from the XML

### DIFF
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -84,6 +84,72 @@ def sample_texture(
   return wp.vec3(tex_color[0], tex_color[1], tex_color[2])
 
 
+@wp.func
+def sample_skybox(
+  # In:
+  skybox_tex: wp.Texture2D,
+  face_width_inv: float,
+  ray_dir_world: wp.vec3,
+) -> wp.vec3:
+  # MuJoCo maps a world-space direction to cube-map space by rotating 90° about X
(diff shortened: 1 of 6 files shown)


mujoco_warp/_src/render.py
```diff
@@ -84,6 +84,72 @@ def sample_texture(
...
+    if rz > 0.0:
+      face = 4
+      sc = rx
+      tc = -ry
+    else:
+      face = 5
+      sc = -rx
+      tc = -ry
+
+  s = (sc / ma + 1.0) * 0.5
```
thowell: should we use safe division here and the line below?

mujoco_warp/_src/render.py
```diff
@@ -84,6 +84,72 @@ def sample_texture(
...
+  t = (tc / ma + 1.0) * 0.5
+
+  # Keep the linear filter from bleeding between adjacent faces in the vertical strip.
+  t_min = 0.5 * face_width_inv
+  t = wp.clamp(t, t_min, 1.0 - t_min)
+
+  v = (float(face) + t) / 6.0
```
thowell: `/ 6.0` -> `wp.static(1.0 / 6.0) *`

thowell:
this is great! thanks @StafaH

thowell (approved):

StafaH:
Done, thanks @thowell

PR #1309, opened 2026-04-23
update _flex_bending

thowell:


quagla:
What is the rationale?

thowell:
think there is a bug in line 720 with the indentation?

quagla (approved):

PR #1327, opened 2026-05-04
python 3.14

thowell:


erikfrey (approved):
Looks great, after we submit I'll update the CI settings to remove the deprecated actions and run the new actions

thowell:
@erikfrey i think the ci is waiting for 3.13 checks that have been removed, do we need to update the ci somewhere to remove these?

PR #1328, opened 2026-05-04
flex_bending update

thowell:
https://github.com/google-deepmind/mujoco/commit/dbd451138c673aba6cc38efecda5e37a510e5e1b

### DIFF
--- a/mujoco_warp/_src/passive.py
+++ b/mujoco_warp/_src/passive.py
@@ -702,10 +702,11 @@ def _flex_bending(
   flex_vertadr: wp.array[int],
   flex_edgeadr: wp.array[int],
   flex_edgenum: wp.array[int],
+  flex_bendingadr: wp.array[int],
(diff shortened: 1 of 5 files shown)


mujoco_warp/_src/passive.py
```diff
@@ -733,8 +735,12 @@ def _flex_bending(
     flex_vertadr[f] + flex_edgeflap[edgeid][1],
   )
 
+  adr = flex_bendingadr[f]
+  if not wp.static(check_version("mujoco>=3.8.1.dev909088123")):
```
quagla: To keep the engine clean, maybe you want to do an else branch in io.py and populate it?

quagla (approved):
Thank you!

mujoco_warp/_src/passive.py
thowell: removing this case since io.py already incorporates the appropriate stride

PR #1329, opened 2026-05-05
flex trilinear strain

thowell:
```
mjwarp-viewer ../mujoco/model/flex/trilinear.xml --nconmax=256 --njmax=1024 --engine=warp
```

https://github.com/user-attachments/assets/ecfc96ed-3389-4783-9bb5-290a8159a6d1

### DIFF
--- a/mujoco_warp/_src/collision_flex.py
+++ b/mujoco_warp/_src/collision_flex.py
@@ -353,6 +353,7 @@ def _flex_broadphase_plane(
 
   diff_center = flex_center - geom_pos
   dist_center = wp.dot(diff_center, plane_normal)
+
   if dist_center - proj_half > margin:
     return
 
@@ -415,6 +416,7 @@ def _write_flex_contact(
   contact_dim_out: wp.array[int],
   contact_geom_out: wp.array[wp.vec2i],
   contact_flex_out: wp.array[wp.vec2i],
+  contact_elem_out: wp.array[wp.vec2i],
   contact_vert_out: wp.array[wp.vec2i],
   contact_worldid_out: wp.array[int],
   contact_type_out: wp.array[int],
@@ -452,8 +454,9 @@ def _write_flex_contact(
     contact_solimp_out[c_idx] = solimp
     contact_dim_out[c_idx] = condim
     contact_geom_out[c_idx] = wp.vec2i(geomid, -1)
-    contact_flex_out[c_idx] = wp.vec2i(flexid, elemid)
-    contact_vert_out[c_idx] = wp.vec2i(vertid, -1)
+    contact_flex_out[c_idx] = wp.vec2i(-1, flexid)
+    contact_elem_out[c_idx] = wp.vec2i(-1, elemid)
+    contact_vert_out[c_idx] = wp.vec2i(-1, vertid)
     contact_worldid_out[c_idx] = worldid
     contact_type_out[c_idx] = ContactType.CONSTRAINT
     contact_geomcollisionid_out[c_idx] = collisionid
@@ -1004,6 +1007,7 @@ def _flex_plane_narrowphase(
   contact_dim_out: wp.array[int],
   contact_geom_out: wp.array[wp.vec2i],
   contact_flex_out: wp.array[wp.vec2i],
+  contact_elem_out: wp.array[wp.vec2i],
   contact_vert_out: wp.array[wp.vec2i],
   contact_worldid_out: wp.array[int],
   contact_type_out: wp.array[int],
@@ -1082,6 +1086,7 @@ def _flex_plane_narrowphase(
       contact_dim_out,
       contact_geom_out,
       contact_flex_out,
+      contact_elem_out,
       contact_vert_out,
       contact_worldid_out,
       contact_type_out,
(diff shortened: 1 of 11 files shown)


mujoco_warp/_src/io.py
```diff
@@ -236,6 +236,19 @@ def _check_friction(name: str, id_: int, condim: int, friction, checks):
   m.is_sparse = is_sparse(mjm)
   m.has_fluid = mjm.opt.wind.any() or mjm.opt.density > 0 or mjm.opt.viscosity > 0
 
+  # Compute total interpolated (non-strain) cells for passive interp kernel
```
quagla: Why non-strain?

mujoco_warp/_src/io.py
```diff
@@ -951,6 +974,19 @@ def make_data(
   sizes["naconmax"] = naconmax
   sizes["njmax"] = njmax
 
+  # Compute total interpolated (non-strain) cells for passive interp kernel
```
quagla: Haven't you already done this above?

mujoco_warp/_src/passive.py
```diff
@@ -758,6 +763,293 @@ def _flex_bending(
...
+    cz_fi = cellnum_fi[2]
+    ncells_fi = cx_fi * cy_fi * cz_fi
+    if cellid < cell_offset + ncells_fi:
+      f = fi
+      local_cell = cellid - cell_offset
+      ci = local_cell / (cy_fi * cz_fi)
+      cj = (local_cell - ci * cy_fi * cz_fi) / cz_fi
+      ck = local_cell - ci * cy_fi * cz_fi - cj * cz_fi
+      found = 1
+    cell_offset += ncells_fi
```
quagla: Would it be useful to do everything that happens up to here in io.py?

mujoco_warp/_src/passive.py
```diff
@@ -758,6 +763,293 @@ def _flex_bending(
...
+          grad_z = phi_x * phi_y * dphi_z
+
+          for r in range(3):
+            F[r, 0] += node_pos[r] * grad_x
+            F[r, 1] += node_pos[r] * grad_y
+            F[r, 2] += node_pos[r] * grad_z
+
+          idx += 1
+
+  # Extract rotation via Müller et al. (2016) quaternion iteration (mju_mat2Rot)
```
quagla: I guess this cannot go in support.py?

mujoco_warp/_src/io.py
quagla: Oh you mean cells for which the stretch force is passive and not a constraint.

mujoco_warp/_src/constraint.py
```diff
@@ -799,6 +805,418 @@ def kernel(
...
+  ):
+    worldid, eqstrainid = wp.tid()
+    eqid = eq_flexstrain_adr[eqstrainid]
+
+    if not eq_active_in[worldid, eqid]:
+      return
+
+    f = eq_obj1id[eqid]
+    order = flex_interp[f]
+    if order < 0:
```
quagla: This is dangerous because we now support the case flex_interp<0 and it has different sizes because each cell is a quadrilateral (surface element). I fear it could cause a segfault to just make the order positive and use flex_stiffness for the shell case.

thowell:
@quagla the pr is updated and should address the comments above

mujoco_warp/_src/constraint.py
```diff
@@ -2006,16 +2460,314 @@ def _get_contact_bodies_and_weights(
...
 
   else:
     return wp.vec4i(-1, -1, -1, -1), wp.vec4(0.0, 0.0, 0.0, 0.0)
 
 
+# TODO(flex): Add interpolated flex self-collision support. Currently, interpolated
+# flex contacts only handle flex-vs-geom contacts (e.g., flex on ground plane).
+# Self-collision requires distributing both sides of the contact across cell nodes.
+
+# TODO(flex): Add elem contact support for interpolated flex (vert < 0 case).
```
quagla: Is this true? I see that you are computing weights.

mujoco_warp/_src/collision_flex.py
```diff
@@ -1832,3 +1986,23 @@ def flex_narrowphase(m: Model, d: Data):
       d.nacon,
     ],
   )
+
+  # Filter flex contacts: limit contacts per (geom, flex) pair to
+  # MJ_MAXCONPAIR (matching filterFlexContacts in MuJoCo C)
+  if m.has_trilinear and m.eq_flexstrain_adr.size > 0:
```
quagla: It's weird to do it only if eq_flexstrain_adr.size>0?

quagla:
Does the video use passive forces or constraints? Would be nice to run it with both and just check that visually it's okay. But maybe you already tested other models.

thowell:
the video posted is with passive forces.

fixed a bug with the strain constraint implementation (and added a test) and confirmed that trilinear.xml with the following change
```
 <!-- <elasticity young="1e4" poisson="0.1" damping="0.01"/> -->
 <edge equality="strain"/>
```
looks to be correct now.

mujoco_warp/_src/collision_flex.py
thowell: removed this filter -- we now use the same filter for all flex contacts `_filter_flex_candidates`

mujoco_warp/_src/collision_flex.py
```diff
@@ -2105,6 +2110,162 @@ def _flex_narrowphase_unified(
     )
 
 
+@wp.kernel
+def _flex_narrowphase_dim3_detect(
```
quagla: The name of this function is very confusing and doesn't explain what it's actually doing.

mujoco_warp/_src/constraint.py
```diff
@@ -1406,6 +1412,337 @@ def kernel(
...
+
+    solref = eq_solref[worldid % eq_solref.shape[0], eqid]
+    solimp = eq_solimp[worldid % eq_solimp.shape[0], eqid]
+
+    # gather cell node positions and reference positions (order==1: npc=8)
+    # We compute the corotational quaternion from the deformation gradient
+    # at the cell center (0.5, 0.5, 0.5)
+
+    # Compute deformation gradient at center: F_ij = sum_n xpos_n_i * dN_n/dx_j
+    F = wp.mat33(0.0)
```
quagla: This is the same as the passive forces right? Would it be a performance regression to extract a common function in support?

mujoco_warp/_src/constraint.py
```diff
@@ -2115,16 +2519,312 @@ def _get_contact_bodies_and_weights(
...
+        W4 = -1.0
+      elif max_id == 5:
+        W5 = -1.0
+      elif max_id == 6:
+        W6 = -1.0
+      elif max_id == 7:
+        W7 = -1.0
+
+      # Pass 3
+      max_w = -1.0
```
quagla: Are all these IF's necessary or is there a more compact way to write it?

mujoco_warp/_src/collision_flex.py
thowell: renamed to `_flex_narrowphase_tet_detect` -- reason for 'detect' is that it does not write contacts yet but generates candidates that are filtered later

mujoco_warp/_src/constraint.py
thowell: updated to utilize a utility function `compute_interp_cell_quat` that is added to support.py

mujoco_warp/_src/constraint.py
thowell: refactored, please see the utility function `select_top4_weights` now added to support.py

quagla (approved):

PR #1332, opened 2026-05-05
Configurable ambient lighting

tkelestemur:
Adds a `use_ambient_lighting` option to MJWarp’s render context so callers can disable the renderer’s hardcoded hemispheric ambient term. This gives us the full control over how the RGB renderer uses the lights.

By default the option remains enabled, preserving existing render behavior. When set to `False`, RGB shading starts from black and only explicit model lights contribute to geometry color. MJWarp currently renders visible geometry even when the MJCF has no `<light>` elements and the MuJoCo headlight is disabled. This is because the renderer applies an unconditional ambient lighting term in `render.py`.

MuJoCo’s OpenGL renderer has a separate default camera-following headlight, but MJWarp does not currently implement that headlight path. Making the MJWarp ambient term configurable lets users produce dark RGB images when no explicit lights are present.

## Changes

- Added `use_ambient_lighting: bool = True` to `create_render_context`.
- Stored the option on `RenderContext`.
- Gated the hemispheric ambient color contribution in the render kernel.
- Added a regression test for a no-light scene:
  - default ambient lighting produces nonzero geometry pixels
  - `use_ambient_lighting=False` produces black geometry pixels


StafaH (commented):
Thanks @tkelestemur. This looks good to me.

In a future PR it might also be nice to make the ambient lighting configurable (mostly intensity) so user's can control what the look is like, and also try to make the default match more the MuJoCo ambient light

mujoco_warp/_src/types.py
```diff
@@ -1949,6 +1949,7 @@ class RenderContext:
     cam_id_map: camera id map
     use_textures: whether to use textures
     use_shadows: whether to use shadows
+    use_ambient_lighting: whether to use renderer ambient lighting
```
thowell: maybe just 'whether to use ambient lighting'?

tkelestemur: done in https://github.com/google-deepmind/mujoco_warp/pull/1332/commits/747716037f4e88fac201cb144909cadbfae0ab5d

tkelestemur:
Thanks @StafaH 

>In a future PR it might also be nice to make the ambient lighting configurable (mostly intensity) so user's can control what the look is like, and also try to make the default match more the MuJoCo ambient light

Yeah I thought about making the MuJoCo ambient lightning, it's a bit different in the sense the light follows the camera whereas here we are adding a constant bias to all pixels. 

I think it would be great if we have complete feature-parity with MuJoCo for all the renderer settings (to the degree we can).

thowell (approved):

thowell:
thanks @tkelestemur!

PR #1334, opened 2026-05-06
Fix _qderiv_actuator_passive_actuation_sparse: use qM_fullm_elemid lookup table

### LINKED ISSUE #1333: _qderiv_actuator_passive_actuation_sparse indexes qMj with wrong row offsets when free joint is in qvel
PR  (https://github.com/google-deepmind/mujoco_warp/pull/1243) — "Optimization: qderiv_actuator_passive_actuation", merged 2026-03-21, commit 80e146c.  
 

  Before this PR, the sparse path indexed qDeriv directly by element id and walked the moment-sparsity rows — no separate "row offset" lookup into qMj was needed, [Before PR Reference](https://github.com/google-deepmind/mujoco_warp/blob/7ffd2bc41b70429740bb6fd21eeefacec7667da1/mujoco_warp/_src/derivative.py#L119-L172). The PR rewrote the sparse kernel to search qMj for the right element id by iterating qMj[M_rowadr[row] + k] for k in [0, M_rownnz[row]). [After PR Reference](https://github.com/google-deepmind/mujoco_warp/blob/80e146c71709d09af817c0d85cb80134251b46ab/mujoco_warp/_src/derivative.py#L172-L222):
```                                                                                                                                                                                        
  @wp.kernel                                                        
  def _qderiv_actuator_passive_actuation_sparse(
      M_rownnz: wp.array(dtype=int),
      M_rowadr: wp.array(dtype=int),
      ...                                                                                                                                                                                   
      qMj: wp.array(dtype=int),
      ...                                                                                                                                                                                   
  ):                                                                
      ...
      row_startk = M_rowadr[row] - 1
      row_nnz   = M_rownnz[row]                                                                                                                                                             
```                                                                                                                                                                                    
  That assumes qMj and M_rownnz/M_rowadr share a layout. They don't — qMj is built chained (free-joint internal block expanded) while M_rownnz/M_rowadr is compact (free-joint block stored diagonal-only). For models with descendants whose dof_parentid chain would walk through a free-joint block, the offsets diverge.      For example:

a free-flying body + one hinge joint with a position/velocity actuator (kp or kv> 0).
                                                                                                                                                                                            
  ```
┌─────┬──────────────────────┬──────────────────┬────────────────────┐                                                                                                                    
  │ DoF │         Body         │ compact M_rowadr │ chained qMj offset │
  ├─────┼──────────────────────┼──────────────────┼────────────────────┤                                                                                                                    
  │ 0–5 │ free joint (chained) │   0,1,2,3,4,5    │   0,1,3,6,10,15    │
  ├─────┼──────────────────────┼──────────────────┼────────────────────┤
  │   6 │ actuated hinge       │        6         │         21         │                                                                                                                    
  └─────┴──────────────────────┴──────────────────┴────────────────────┘     
```                                                                                                               
                                                                                                                                                                                            
  m.nM=7 (compact), qMj.size=22 (chained).                                                                                                                                                  
                                                                    
  Kernel processes the hinge actuator's diagonal (6,6) contribution:                                                                                                                        
  - reads M_rowadr[6]=6, scans qMj[6] for col=6                     
  - but qMj[6] is in the free-joint block (row 3, col ≤ 3) — no match                                                                                                                       
  - wp.atomic_add skipped → damping dropped → implicit step diverges                                                    
                                                                                                                                                                                            
The new unit test added by the PR (test_smooth_vel_sparse_tendon_coupled (https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/_src/derivative_test.py#L217)) uses a 35-joint serial-chain fixture with no free joint, so compact and chained layouts coincide on every row. The bug stayed silent until a model with a free joint preceding actuated DoFs in qvel order.

### PR DESCRIPTION
**Pull request by ooctipus**
## Summary
Fixes #1333.

The sparse actuator implicit-damping kernel introduced in #1243 searches `qM_fullm` for the `(row, col)` element id using `m.M_rownnz` / `m.M_rowadr`. Those arrays describe MuJoCo's compact mass-matrix sparsity, where joints whose internal block is treated as diagonal-only by `mj_factorM` (e.g. free joints) contribute one entry per internal dof. `qM_fullm_i` / `qM_fullm_j`, however, are built in `io.py` by walking `dof_parentid` and include the full chained internal block — so the two layouts diverge whenever a joint with diagonal-only compact storage precedes any actuated dof in qvel order.

In that topology, the inner `qMj[row_startk] == col` check inside `_qderiv_actuator_passive_actuation_sparse` reads the wrong slice of `qMj`, never matches, and the actuator's contribution to `qDeriv` is silently dropped. Downstream `factor_solve_i` then sees `(M − dt·qDeriv)` with no actuator damping for the affected dof, and the implicit step diverges.

The bug only surfaces with this specific structure — free joint at qvel start + actuated dof after it — which is why the unit test added in #1243 (`test_smooth_vel_sparse_tendon_coupled`, a 35-joint serial chain with no free joint) does not catch it.

## Repro (also added as a regression test)

A single free body followed by an actuated hinge:

```xml
<worldbody>
  <body><joint type="free"/><geom type="sphere" size="0.05" mass="1"/></body>
  <body pos="1 0 0">
    <joint name="hinge0" type="hinge" axis="0 1 0"/>
    <geom type="sphere" size="0.05" mass="1"/>
  </body>
</worldbody>
<actuator><position joint="hinge0" kp="10" kv="1"/></actuator>
```

With the buggy indexing, `mjwarp`'s `deriv_smooth_vel` diagonal entry for the hinge dof is `qM` only (`0.001`); MuJoCo's reference is `qM + dt·(kp+kv)` (`0.003`). On real-world models (KUKA Iiwa7 + Allegro hand grasping a free-flying object), this manifested as ~94 % "abnormal robot velocity" terminations during RL training because the robot's implicit damping was fully disabled.

## Fix

Build a dense `(nv × nv)` `qM_fullm_elemid` lookup table in `io.py` mapping `(row, col) → elemid` in the chain-aware `qM_fullm_i / qM_fullm_j` layout (`-1` if `col` is not a chain ancestor of `row`). The sparse kernel uses an O(1) lookup instead of walking the row, which both fixes the layout-divergence bug and removes the inner search loop. (Approach proposed by @Kenny-Vilella in [this branch](https://github.com/google-deepmind/mujoco_warp/compare/main...Kenny-Vilella:mujoco_warp:dev/kvilella/fix_bug_actuator_derivative).)

Memory cost: `nv² × 4` bytes — 2.8 KiB at nv=27 (humanoid), 25.6 KiB at nv=81 (three_humanoids), 1 MiB at nv=500.

## Benchmarks (5090, CUDA-event-only timing, 1000 iters × 7 trials, medians)

For typical robot models (humanoid, three_humanoids) kernel time is within ±10% trial-to-trial noise of the previous chain-aware-search variant. For deep chains with high-`moment_rownnz` actuators (tendons spanning a long serial chain) the asymptotic improvement is visible:

| n_links | n_tendon | n_pos | nworld | search-variant (µs) | elemid (µs) | Δ |
|--:|--:|--:|--:|--:|--:|--:|
| 30  | 4 | 30  | 4096 | 535.52    | 164.33   | +69% |
| 60  | 4 | 60  | 4096 | 3,921.68  | 713.41   | +82% |
| 100 | 4 | 100 | 4096 | 20,180.72 | 2,558.77 | +87% |

## Test plan

- [x] New regression test `test_smooth_vel_sparse_free_joint_precedes_actuator` covers the minimum reproducer; verified it FAILS on `main` (slot `[6,6]` = `0.001` vs `0.003`) and PASSES with this fix.
- [x] `pytest mujoco_warp/_src/derivative_test.py` — 25/25 passed.
- [x] `pytest mujoco_warp/_src/{smooth,forward,io,passive}_test.py` — all passed.
- [x] `ruff check` + `ruff format --check` clean on all touched files.

### DIFF
--- a/mujoco_warp/_src/derivative_test.py
+++ b/mujoco_warp/_src/derivative_test.py
@@ -433,7 +433,7 @@ class DerivativeTest(parameterized.TestCase):
         <flag gravity="disable"/>
       </option>
       <worldbody>
-        <body name="floater">
+        <body>
           <joint type="free"/>
           <geom type="sphere" size="0.05" mass="1"/>
         </body>
@@ -453,7 +453,6 @@ class DerivativeTest(parameterized.TestCase):
       keyframe=0,
       overrides={"opt.jacobian": mujoco.mjtJacobian.mjJAC_SPARSE},
     )
-    self.assertTrue(m.is_sparse, "Model should use sparse path")
 
     mujoco.mj_step(mjm, mjd)
 
--- a/mujoco_warp/_src/derivative_test.py
+++ b/mujoco_warp/_src/derivative_test.py
@@ -415,6 +415,65 @@ class DerivativeTest(parameterized.TestCase):
     self.assertFalse(np.any(np.isnan(mjw_out)))
     _assert_eq(mjw_out, mj_out, "qM - dt * qDeriv (sparse tendon coupled)")
 
+  def test_smooth_vel_sparse_free_joint_precedes_actuator(self):
+    """Sparse qDeriv uses chain-aware row offsets when indexing qM_fullm.
+
+    A free joint's internal block is stored diagonal-only in the compact
+    (M_rownnz, M_rowadr) layout but fully chained (1+2+...+n entries per
+    internal dof) in qM_fullm_i / qM_fullm_j. For any actuated dof that
+    follows the free joint in qvel order, indexing qM_fullm with the compact
+    offsets lands in slots that belong to the free-joint block, the actuator
+    contribution to qDeriv is silently dropped, and the implicit step loses
+    damping for the actuated dof.
+    """
+    mjm, mjd, m, d = test_data.fixture(
+      xml="""
+    <mujoco>
+      <option integrator="implicitfast">
+        <flag gravity="disable"/>
+      </option>
+      <worldbody>
+        <body name="floater">
+          <joint type="free"/>
+          <geom type="sphere" size="0.05" mass="1"/>
+        </body>
+        <body pos="1 0 0">
+          <joint name="hinge0" type="hinge" axis="0 1 0"/>
+          <geom type="sphere" size="0.05" mass="1"/>
+        </body>
+      </worldbody>
+      <actuator>
+        <position joint="hinge0" kp="10" kv="1"/>
+      </actuator>
+      <keyframe>
+        <key qpos="0 0 0 1 0 0 0 0" qvel="0 0 0 0 0 0 1" ctrl="0.1"/>
+      </keyframe>
+    </mujoco>
+    """,
+      keyframe=0,
+      overrides={"opt.jacobian": mujoco.mjtJacobian.mjJAC_SPARSE},
+    )
+    self.assertTrue(m.is_sparse, "Model should use sparse path")
+
+    mujoco.mj_step(mjm, mjd)
+
+    out_smooth_vel = wp.zeros((1, 1, m.nM), dtype=float)
+    mjw.deriv_smooth_vel(m, d, out_smooth_vel)
+
+    mjw_out = np.zeros((m.nv, m.nv))
+    for elem, (i, j) in enumerate(zip(m.qM_fullm_i.numpy(), m.qM_fullm_j.numpy())):
+      mjw_out[i, j] = out_smooth_vel.numpy()[0, 0, elem]
+      mjw_out[j, i] = out_smooth_vel.numpy()[0, 0, elem]
+
+    mj_qDeriv = np.zeros((mjm.nv, mjm.nv))
+    mujoco.mju_sparse2dense(mj_qDeriv, mjd.qDeriv, mjm.D_rownnz, mjm.D_rowadr, mjm.D_colind)
+    mj_qM = np.zeros((m.nv, m.nv))
+    mujoco.mj_fullM(mjm, mj_qM, mjd.qM)
+    mj_out = mj_qM - mjm.opt.timestep * mj_qDeriv
+
+    self.assertFalse(np.any(np.isnan(mjw_out)))
+    _assert_eq(mjw_out, mj_out, "qM - dt * qDeriv (sparse, free joint precedes actuated dof)")
+
   def test_actearly_derivative(self):
     """Implicit derivatives should use next activation when actearly is set."""
     mjm, mjd, m, d = test_data.fixture(
--- a/mujoco_warp/_src/derivative.py
+++ b/mujoco_warp/_src/derivative.py
@@ -197,8 +197,8 @@ def _qderiv_actuator_passive_actuation_dense(
 @wp.kernel
 def _qderiv_actuator_passive_actuation_sparse(
   # Model:
-  M_rownnz: wp.array[int],
-  M_rowadr: wp.array[int],
+  qM_fullm_rownnz: wp.array[int],
+  qM_fullm_rowadr: wp.array[int],
   # Data in:
   moment_rownnz_in: wp.array2d[int],
   moment_rowadr_in: wp.array2d[int],
@@ -235,12 +235,16 @@ def _qderiv_actuator_passive_actuation_sparse(
 
       contrib = moment_i * moment_j * vel
 
-      # Search the corresponding elemid
+      # Search the corresponding elemid in qM_fullm. Use the chain-aware row
+      # offsets (qM_fullm_rownnz / qM_fullm_rowadr) that match qMj's layout
+      # rather than the compact (M_rownnz, M_rowadr); the two diverge whenever
+      # a joint with diagonal-only compact storage (e.g. free joint) precedes
+      # actuated dofs in qvel order.
       # TODO: This could be precalculated for improved performance
       row = dofi
       col = dofj
-      row_startk = M_rowadr[row] - 1
-      row_nnz = M_rownnz[row]
+      row_startk = qM_fullm_rowadr[row] - 1
+      row_nnz = qM_fullm_rownnz[row]
       for k in range(row_nnz):
         row_startk += 1
         if qMj[row_startk] == col:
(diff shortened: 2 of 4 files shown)


ooctipus:
@googlebot I signed it!

Kenny-Vilella:
I made another attempt to fix in there:
https://github.com/google-deepmind/mujoco_warp/compare/main...Kenny-Vilella:mujoco_warp:dev/kvilella/fix_bug_actuator_derivative

@thowell  You can choose the one you prefer ;)

mujoco_warp/_src/derivative_test.py
```diff
@@ -415,6 +415,65 @@ def test_smooth_vel_sparse_tendon_coupled(self):
...
+    damping for the actuated dof.
+    """
+    mjm, mjd, m, d = test_data.fixture(
+      xml="""
+    <mujoco>
+      <option integrator="implicitfast">
+        <flag gravity="disable"/>
+      </option>
+      <worldbody>
+        <body name="floater">
```
thowell: we can remove this name

mujoco_warp/_src/derivative_test.py
```diff
@@ -415,6 +415,65 @@ def test_smooth_vel_sparse_tendon_coupled(self):
...
+      </actuator>
+      <keyframe>
+        <key qpos="0 0 0 1 0 0 0 0" qvel="0 0 0 0 0 0 1" ctrl="0.1"/>
+      </keyframe>
+    </mujoco>
+    """,
+      keyframe=0,
+      overrides={"opt.jacobian": mujoco.mjtJacobian.mjJAC_SPARSE},
+    )
+    self.assertTrue(m.is_sparse, "Model should use sparse path")
```
thowell: we can remove this assert

thowell:
are the added fields equivalent to `mjModel`'s `D_rowadr` and `D_rownnz`? https://mujoco.readthedocs.io/en/latest/APIreference/APItypes.html#mjmodel

thowell:
@ooctipus can you please try again signing the cla? thanks!

ooctipus:
Thanks @thowell!

Removed both in 0cbd73f.

On the `D_rownnz` / `D_rowadr` question — I am not 100% familiar with mujoco conventions, but reading how they behave during debugging, it seems like they're not the same as the new fields. `D_*` describes the full-square qDeriv (both triangles, `nD` total), while `qM_fullm_i/j` only walks `dof_parentid` upward, so it's lower-triangle / chain-ancestors only (`nC` total). For a free joint, per-row counts are `[1,2,3,4,5,6]` in `qM_fullm_*` vs `[6,6,6,6,6,6]` in `D_*`, so `D_rowadr` would point into the wrong slots in `qM_fullm_j` — which is the same shape of bug this PR fixes.

Also, @Kenny-Vilella has an [alternative](https://github.com/google-deepmind/mujoco_warp/compare/main...Kenny-Vilella:mujoco_warp:dev/kvilella/fix_bug_actuator_derivative) that uses an `nv × nv` `qM_fullm_elemid` table to skip the inner search entirely. A bit more memory, O(1) lookup. Happy to switch to that if you'd rather go that way.

Re-signing the CLA now.

thowell:
@ooctipus can we benchmark the approach in this pr with kenny's approach?

ooctipus:
@thowell ran a proper stress benchmark this time (5090, CUDA-event-only timing, 1000 iters × 7 trials, medians). Previous attempt with humanoid / three_humanoids was within noise — the search is a tiny fraction of work on shallow trees, so I needed a workload where the inner search actually shows up.

**Stress model:** N-link serial hinge chain with 1–4 fixed tendons spanning all joints (so `moment_rownnz = N` per tendon actuator → inner double loop fires `N²/2` times per thread, each search walks `qM_fullm_rownnz[dofi]` ≈ chain depth at `dofi`). Total work per actuator thread: octi ~ depth³ / 6, kenny ~ depth² / 2.

| n_links | n_tendon | n_pos | nworld | octi (µs) | kenny (µs) | Δ |
|--:|--:|--:|--:|--:|--:|--:|
| 10  | 1 | 0   | 4096 | 29.72       | 20.83     | +29.9% |
| 10  | 4 | 10  | 4096 | 27.40       | 20.38     | +25.6% |
| 30  | 1 | 0   | 4096 | 355.47      | 136.92    | +61.5% |
| 30  | 4 | 30  | 4096 | 535.52      | 164.33    | +69.3% |
| 60  | 1 | 0   | 4096 | 2,478.52    | 719.24    | +71.0% |
| 60  | 4 | 60  | 4096 | 3,921.68    | 713.41    | +81.8% |
| 100 | 1 | 0   | 4096 | 10,822.11   | 2,490.40  | +77.0% |
| 100 | 4 | 100 | 4096 | 20,180.72   | 2,558.77  | +87.3% |

Trial-to-trial spread is tight (±0.5–1.5% on each row), so the deltas are well above noise.

**Realistic models from earlier bench (humanoid nv=27, three_humanoids nv=81): within ±10%, no consistent winner.** That's the regime most users hit, so the practical impact is small for typical robot configs.

**Memory cost** of Kenny's `qM_fullm_elemid` table: `nv² × 4` bytes — 2.8 KiB at nv=27, 40 KiB at nv=100, 1 MiB at nv=500. Trivial.

**Recommendation:** switch to Kenny's variant. Asymptotically better, big wins on deep chains with tendons (or any high-`moment_rownnz` actuator on a long kinematic tree), no measurable cost on shallow trees, memory cost negligible. Want me to fold it into this PR?

thowell:
lets move forward with kenny's approach (thanks kenny!)

ooctipus:
Switched the PR to the elemid-table approach in af4743f — credit @Kenny-Vilella. Title and description updated. All derivative / io / smooth / forward tests pass; pre-commit clean. @thowell I think it is ready :D. Thanks for all the review

thowell (approved):

PR #1340, opened 2026-05-07
testspeed eager mode and cpu

thowell:
add testspeed support for eager mode and cpu

## Benchmark Results: Humanoid

**Model:** `benchmarks/humanoid/humanoid.xml`
**Config:** `--nconmax=24 --njmax=64`
**GPU:** NVIDIA RTX 6000 Ada Generation (48 GB)

| Mode | nworld | Steps/sec | Realtime Factor | Time/step | Total Sim Time | JIT Time |
|------|--------|----------:|----------------:|----------:|---------------:|---------:|
| GPU (graph) | 8,192 | 4,734,044 | 23,670× | 211 ns | 1.73s | 0.32s |
| GPU (eager) | 8,192 | 359,861 | 1,799× | 2,779 ns | 22.76s | 0.32s |
| CPU | 1 | 49 | 0.24× | 20.5 ms | 20.52s | 0.26s |

### Key Observations

- **CUDA graph capture** provides a **13.2× speedup** over eager mode on GPU (211 ns vs 2,779 ns per step) by eliminating per-kernel launch overhead.
- **GPU eager mode** is still **~7,400× faster** than CPU single-world execution on a per-step basis, demonstrating massive parallelism benefits even without graph capture.
- **CPU mode** runs at 0.24× realtime for a single humanoid — useful for debugging and validation, not performance.

---

## Code Changes

Two files were modified to support `--eager` mode and `--device=cpu` execution.

#### 1. Added `--eager` flag

```python
EAGER = flags.DEFINE_bool("eager", False, "run without CUDA graph capture (eager mode)")
```

#### 2. Made `unroll()` device-aware

The original `unroll()` unconditionally used CUDA-only APIs (`ScopedCapture`, `capture_launch`, `ScopedStream`, `get_stream`). Changes:

- **Conditional graph capture:** Only uses `wp.ScopedCapture` / `wp.capture_launch` when `use_graph = is_cuda and not EAGER.value`. Otherwise calls `fn()` directly.
- **Removed `ScopedStream`:** The `wp.ScopedStream(wp.get_stream())` wrapper was removed entirely — it called `wp.get_stream()` which raises `RuntimeError` on CPU. The stream pinning wasn't needed for correctness.
- **Guarded `EventTracer`:** Added `and is_cuda` to the tracer enable condition since event tracing is CUDA-only.

---

#### 1. Removed GPU-only gate

```diff
-  if device == "cpu":
-    raise ValueError("testspeed available for gpu only")
+  is_cuda = device.is_cuda
```

#### 2. Guarded CUDA-only APIs

- `device.free_memory` → only accessed when `is_cuda`
- `--memory` flag → warns and skips on CPU
- `--event_trace` flag → warns and skips on CPU
- `total_memory` metric → conditionally included in JSON output

### DIFF
--- a/mujoco_warp/_src/cli.py
+++ b/mujoco_warp/_src/cli.py
@@ -40,6 +40,7 @@ NCCDMAX = flags.DEFINE_integer("nccdmax", None, "override maximum number of CCD 
 OVERRIDE = flags.DEFINE_multi_string("override", [], "Model overrides (notation: foo.bar = baz)", short_name="o")
 KEYFRAME = flags.DEFINE_integer("keyframe", 0, "keyframe to initialize simulation.")
 EVENT_TRACE = flags.DEFINE_bool("event_trace", False, "print an event trace report")
+EAGER = flags.DEFINE_bool("eager", False, "run without CUDA graph capture (eager mode)")
 NOISE_STD = flags.DEFINE_float("noise_std", 0.01, "add noise to ctrl signal (standard deviation)")
 NOISE_RATE = flags.DEFINE_float("noise_rate", 0.1, "add noise to ctrl signal (noise rate)")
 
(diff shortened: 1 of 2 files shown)


erikfrey (commented):
I think with the explicit cache compilation unit tests, in particular #1348, I'm OK with punting on this PR.

I DO want to guard against incorrect kernel caching behavior, so that regression test makes sense to me.  But I'm not prepared to scope creep beyond that, e.g. to advertise that we are optimizing either eager throughput or CPU throughput.  Not an explicit goal of the project right now (although it could change in the future)

PR #1344, opened 2026-05-08
kernel analyzer @cache_kernel check

thowell:
add kernel analyzer check that kernel factories have @cache_kernel


erikfrey (approved):
I'm OK with this, but I'm also OK punting because we are about to rip @cache_kernel out after the next warp release.

OK if you want to merge this, but JFYI we're almost immediately going to about-face and undo it.

PR #1347, opened 2026-05-08
update gh-pages: simulation and rendering tabs

thowell:
update website to have 'Simulation' and 'Rendering' tabs for benchmarks


erikfrey:
how many rendering benchmarks do we plan to have?  if it's just one or two, I think we can skip the tab and just add them as buttons.  if needed we can distinguish them with a little tag like this:

📷 render primitives

thowell:
probably add a few more based on benchmarks presented in #1113:
- mesh + texture
- flex
- height field

@StafaH are there additional rendering benchmarks you recommend adding?

StafaH:
I think these 5 cover the main scenes. It might be worth adding a collision rendering example from a robot scene, to highlight the performance difference if you were to render for example franka visual vs franka collision.

PR #1348, opened 2026-05-09
test kernel recompilation

thowell:


mujoco_warp/_src/io_test.py
```diff
@@ -2433,6 +2433,81 @@ def test_margin_pair_box_box(self):
...
+    def _tracking_init(self_kernel, *args, **kwargs):
+      original_init(self_kernel, *args, **kwargs)
+      created_kernels.append(self_kernel.key)
+
+    wp.Kernel.__init__ = _tracking_init
+    _uncached_factory(1)
+    _uncached_factory(1)  # second call re-creates the kernel
+    self.assertLen(created_kernels, 2, "kernel recompilation detection mechanism is broken")
+
+    # First step: compiles / loads all required modules.
```
erikfrey: @thowell can you make this test more targeted by doing something like:

```py
wp._src.context.user_modules['mujoco_warp._src.forward'].unload()
mjw.step(m, d) # check there's at least 1 kernel compilation
mjwp.step(m, d) # check there's zero kernel compilations
```

then we can probably leave out the first half of this test, no?

erikfrey (commented):
really lovely!  one question on how to keep the test minimal

mujoco_warp/_src/io_test.py
thowell: updated the test to remove parts unrelated to `step`

mujoco_warp/_src/io_test.py
```diff
@@ -2433,6 +2433,56 @@ def test_margin_pair_box_box(self):
       """)
       )
 
+  @parameterized.parameters(*_IO_TEST_MODELS)
+  def test_kernel_recompilation(self, xml):
+    """Test that subsequent steps do not trigger kernel recompilation."""
+    # Force unload all modules to clear memory cache.
+    for module in list(wp._src.context.user_modules.values()):
+      module.unload()
```
erikfrey: this is basically a full `wp.clear_kernel_cache()` no?

this could really impact unit test time depending on when this particular test gets run, and could invalidate the recent work we did to restore caches across github CI runs.

what is the goal of this first part of the test?  if it's just to "test the test", e.g. demonstrate that your kernel creation counting is actually working, is there a way to do that without invalidating the whole cache?

thowell: test is simplified to just check for any recompilation during the second call to `step`

erikfrey (approved):
🙏 🙏 🙏

PR #1351, opened 2026-05-12
add mesh render benchmark

thowell:
benchmark for rendering mesh + texture


benchmarks/render/mug.xml
```diff
@@ -0,0 +1,81 @@
...
+    <light pos="0 0 3" dir="0 0 -1" directional="true" />
+
+    <!-- A table to hold the mug -->
+    <geom name="table" pos="0.8 0.0 0.5" type="box" size="0.5 1.0 0.01" quat="1.0 0.0 0.0 0.0" contype="1" conaffinity="1"/>
+
+    <!-- A single mug above the table -->
+    <body name="mug" pos="0.8 0.0 0.6" euler="0 0 90">
+      <freejoint name="mug_joint"/>
+      <geom type="mesh" mesh="mug" material="mug" euler="90 0 0" group="1" contype="0" conaffinity="0"/>
+      <geom class="cup" pos="0.0418 0 0.038" euler="0 0 0"/>
```
StafaH: Do we need the collision geoms?

StafaH (approved):
LGTM, thanks Taylor! just have one comment,

benchmarks/render/mug.xml
thowell: removed

erikfrey:
same comment as #1353 - let's use the actual mjwarp render output for the webp

thowell:
mug.webp is updated to be the result from the batch renderer. do we think the resolution is too low?

erikfrey (approved):

PR #1352, opened 2026-05-12
add flex render benchmark

thowell:


benchmarks/cloth/README.md
```diff
@@ -20,3 +20,19 @@ A soft-body simulation benchmark featuring a cloth draped over the MuJoCo humano
...
+
+GPU ray-traced rendering of the draping soft-body flex cloth scene. This benchmark measures rendering performance of deformable meshes and soft-body grids on the GPU. The rollout video below is rendered from the viewpoint of the pre-defined `back` camera.
+
+| Property | Value |
+|----------|-------|
+| Bodies | 918 |
+| DoFs | 2706 |
+| Actuators | 0 |
+| Geoms | 21 |
+| Cameras | 3 |
```
StafaH: The default scene has 3 camers, maybe we should remove some from the xml, it would make the flex rendering look really slow on the benchmarks (in comparison to the others), but the others only have 1 camera

StafaH (approved):
LGTM

benchmarks/cloth/README.md
thowell: scene now has 1 camera

thowell:
updated video with new camera view

erikfrey (commented):
same comment as #1353 - let's use the actual mjwarp render output for the webp

thowell:
video is updated to use result from the batch renderer

erikfrey (approved):

PR #1353, opened 2026-05-12
add render benchmark for height field

thowell:
```
python3 mujoco_warp/testspeed.py /tmp/benchmark_assets/unitree_g1_hfield_render/scene_hfield.xml \
  --function render \
  --nworld 8192 \
  --nconmax 48 \
  --njmax 192 \
  --nstep 200 \
  --norender_depth \
  --replay /tmp/benchmark_assets/unitree_g1_hfield_render/shuffle_dance.npz
```

```
Loading model from: /tmp/benchmark_assets/unitree_g1_hfield_render/scene_hfield.xml...

Model
  nq: 36 nv: 35 nu: 29 nbody: 31 ngeom: 69
RenderContext
  shadows: False textures: True nlight: 2 bvh_ngeom: 36
  ncam: 1 cam_res: [(64, 64)]
Option
  integrator: IMPLICITFAST
  cone: PYRAMIDAL
  solver: NEWTON iterations: 10 ls_iterations: 20
  is_sparse: True
  ls_parallel: False
Data
  nworld: 8192 naconmax: 393216 njmax: 192
Rolling out 200 renders at dt = 0.005...

Total JIT time: 0.02 s
Total simulation time: 32.81 s
Total renders per second: 49,940
Total realtime factor: 249.70 x
Total time per render: 20023.83 ns
```


StafaH (approved):
LGTM!

StafaH:
We should probably mention if its RGBD or RGB

thowell:
@StafaH by default this is rendering rgb (no depth)?

StafaH:
Default behavior is RGBD if this uses the same path as testspeed, since default arg for render_rgb and render_rgb is True there

erikfrey (commented):
Benchmark code itself looks good!  A few questions about the .webp:

1) Can we make it the actual output of mjwarp render?
2) If so, we'll need to adjust [mjwarp-record](https://github.com/google-deepmind/mujoco_warp/blob/main/mujoco_warp/record.py) to support using the tiled renderer, fyi there's a TODO here:

https://github.com/google-deepmind/mujoco_warp/blob/092f3baecfc7317b28bd7b28d657aa52c7cc103c/mujoco_warp/record.py#L84

I guess if we want the movie to be be dynamic, you need to do an `mjwarp.step` after each benchmark measurement.

3) OOC, why is the `nconmax`, `njmax` higher in `unitree_g1_hfield_render` compared to `unitree_g1_hfield`.  For that matter, if up until now you were only benchmarking `render`, why did they need to be specified?

thowell:
@erikfrey
1. done
2. record.py is updated to use the batch renderer -- the todo remains since the update currently processes just the first world
3. updated to use the existing benchmark's setting

mujoco_warp/record.py
```diff
@@ -47,6 +48,7 @@
 _CAM_DISTANCE = flags.DEFINE_float("cam_distance", 1.5, "camera distance coefficient (multiplier for model extent)")
 _CAM_LOOKAT_Z = flags.DEFINE_float("cam_lookat_z", None, "camera lookat z value (absolute); default: mjm.stat.center[2]")
 _CAM_AZIMUTH_SPEED = flags.DEFINE_float("cam_azimuth_speed", 0.05, "camera azimuth orbit speed (degrees per step)")
+_RENDER_MODE = flags.DEFINE_enum("render_mode", "gpu", ["gpu", "cpu"], "rendering backend to use")
```
erikfrey: Well, it's not really cpu vs gpu.  Both renderers use GPU.

maybe we can call the modes `warp` and `python` or so?  or `mujoco` and `mujoco_warp`?  or `warp` and `opengl`? up to you.  i would default to python for now

mujoco_warp/record.py
```diff
@@ -62,37 +64,73 @@ def _main(argv: Sequence[str]):
   path = epath.Path(argv[1])
   print(f"Loading model from: {path}...\n")
   mjm = cli.load_model(path)
-  m, d, rc, ctrls = cli.init_structs(mjw.step, mjm)
 
-  frames = []
-  renderer = mujoco.Renderer(mjm, height=cli.RENDER_HEIGHT.value, width=cli.RENDER_WIDTH.value)
+  if _RENDER_MODE.value == "gpu":
+    m, d, rc, ctrls = cli.init_structs(mjw.render, mjm)
+    assert rc is not None, "GPU rendering requested, but RenderContext is None. Check if the model has cameras."
```
erikfrey: let's stick to the convention in this code, so we should `raise ValueError` or `raise app.UsageError` or so here

mujoco_warp/record.py
```diff
@@ -62,37 +64,73 @@ def _main(argv: Sequence[str]):
...
+      mujoco.mj_forward(mjm, mjd)
+      # symmetric orbit
+      cam.azimuth = 90 + (step - cli.NSTEP.value / 2) * _CAM_AZIMUTH_SPEED.value
+      renderer.update_scene(mjd, camera=cam)
+      rgb_frames.append(renderer.render())
 
   print(f"Recording {cli.NSTEP.value} steps...")
   cli.unroll(mjw.step, m, d, rc, callback, ctrls)
 
+  # Concatenate RGB and Depth frames sequentially (RGB first, then Depth)
```
erikfrey: let's pick one instead of concatenating.  the flashing back and forth between the two is a bit visually confusing.

may mean adding another flag to specify which render you want

mujoco_warp/record.py
```diff
@@ -62,37 +64,73 @@ def _main(argv: Sequence[str]):
...
-  cam.type = mujoco.mjtCamera.mjCAMERA_FREE
-  cam.lookat[:] = mjm.stat.center
-  if _CAM_LOOKAT_Z.value is not None:
-    cam.lookat[2] = _CAM_LOOKAT_Z.value
-  cam.distance = mjm.stat.extent * _CAM_DISTANCE.value
-  cam.elevation = -20
-  mjd = mujoco.MjData(mjm)
+  if _RENDER_MODE.value == "cpu":
+    renderer = mujoco.Renderer(mjm, height=cli.RENDER_HEIGHT.value, width=cli.RENDER_WIDTH.value)
+    cam = mujoco.MjvCamera()
```
erikfrey, line 78: how do we specify camera info for the warp renderer?  ideally params like _CAM_DISTANCE etc apply to both modes

mujoco_warp/_src/cli.py
```diff
@@ -194,7 +196,12 @@ def unroll(
     with warp_util.EventTracer(enabled=EVENT_TRACE.value) as tracer:
       jit_beg = time.perf_counter()
       with wp.ScopedCapture() as capture:
-        fn(*(m, d) if rc is None else (m, d, rc))
+        import inspect
```
erikfrey: instead of relying on inspect here, why not unroll `render` and then call `step` in the callback if it's needed?

erikfrey (commented):
Looks pretty good - main thing I'm not sure about is how the camera is getting set with mjwarp render.  Otherwise mostly nits

mujoco_warp/record.py
thowell: changed to "warp" and "python" as suggested

mujoco_warp/record.py
thowell: added `--channel` to specify the rendering mode: `rgb` or `depth`. this benchmark now just renders `rgb`.

mujoco_warp/record.py
thowell: done

mujoco_warp/record.py
thowell: the warp renderer requires camera info to be specified in xml/mjspec -- one exception is resolution that is specified when a `RenderContext` is constructed https://github.com/google-deepmind/mujoco_warp/blob/8d23f29fda16a30fa27a7b91af61f343b30115d3/mujoco_warp/_src/io.py#L2771

camera settings like camera distance are for the mujoco free camera and do not affect the warp renderer

mujoco_warp/_src/cli.py
thowell: removed inspect

thowell: updated testspeed for `--function='render'`
- create a callable with `refit_bvh` followed by `render`
- include `step` in the callback

mujoco_warp/_src/cli.py
```diff
@@ -194,7 +196,10 @@ def unroll(
     with warp_util.EventTracer(enabled=EVENT_TRACE.value) as tracer:
       jit_beg = time.perf_counter()
       with wp.ScopedCapture() as capture:
-        fn(*(m, d) if rc is None else (m, d, rc))
+        if rc is None:
```
erikfrey: nit: let's leave totally unchanged to preserve a nice change history

mujoco_warp/record.py
```diff
@@ -62,36 +58,73 @@ def _main(argv: Sequence[str]):
...
   print(f"Loading model from: {path}...\n")
   mjm = cli.load_model(path)
-  m, d, rc, ctrls = cli.init_structs(mjw.step, mjm)
 
-  frames = []
-  renderer = mujoco.Renderer(mjm, height=cli.RENDER_HEIGHT.value, width=cli.RENDER_WIDTH.value)
+  if _RENDER_MODE.value == "warp":
+    m, d, rc, ctrls = cli.init_structs(mjw.render, mjm)
+    if rc is None:
+      raise ValueError("Warp rendering requested, but RenderContext is None. Check if the model has cameras.")
```
erikfrey: nit: can we make this error more actionable?

"check if the model has cameras" - can we check if the model has cameras?  i briefly traced through the code and couldn't find why no cameras would result in a None rc?

"RenderContext is None" - this is asking a fair bit of the user to understand what this means, and to have read the code.  better would be something like:

"Could not initialize warp renderer on device {foo}: {exception bar}"

i would be fine with something further down the call stack raising an error if you ask to create a RenderContext with no cameras, then we just see a helpful error + stack trace if you do something invalid

mujoco_warp/record.py
erikfrey: can you add a TODO to have consistent camera behavior?  e.g. if we create a camera for the python renderer, we should be able to create a camera for the warp renderer too - just might have to do it in mjspec instead of on demand.

mujoco_warp/record.py
```diff
@@ -62,36 +58,73 @@ def _main(argv: Sequence[str]):
...
+        img_rgb = (rgb_dev.numpy() * 255.0).astype(np.uint8)
+        rgb_frames.append(img_rgb[0])
+
+      # 2. Extract and Normalize Depth via mjw.get_depth
+      elif _CHANNEL.value == "depth":
+        depth_dev = wp.zeros((d.nworld, res[1], res[0]), dtype=float)
+        mjw.get_depth(rc, 0, 5.0, depth_dev)
+        # Scale to uint8 grayscale and stack to match number of rgb channels
+        img_depth = ((1.0 - depth_dev.numpy()[0]) * 255.0).astype(np.uint8)
+        img_depth = np.stack([img_depth, img_depth, img_depth], axis=-1)
```
erikfrey: can you use `Image.fromarray(f, 'L')` for grayscale instead of repeating the channels like this?

mujoco_warp/record.py
```diff
@@ -62,36 +58,73 @@ def _main(argv: Sequence[str]):
...
+      cam.azimuth = 90 + (step - cli.NSTEP.value / 2) * _CAM_AZIMUTH_SPEED.value
+      renderer.update_scene(mjd, camera=cam)
+      rgb_frames.append(renderer.render())
 
   print(f"Recording {cli.NSTEP.value} steps...")
-  cli.unroll(mjw.step, m, d, rc, callback, ctrls)
+  cli.unroll(mjw.step, m, d, None, callback, ctrls)
+
+  # Select recorded channel frames
+  frames = rgb_frames if _CHANNEL.value == "rgb" else depth_frames
```
erikfrey: nit: i don't think you need `rgb_frames` and `depth_frames`, just define a single `frames` and change the code right above to only append to it

mujoco_warp/testspeed.py
```diff
@@ -260,7 +260,22 @@ def callback(step, step_trace, latency):
...
+  if _FUNCTION.value == "render":
+    with wp.ScopedCapture() as step_capture:
+      mjw.step(m, d)
+
+    def render_callback(step, step_trace, latency):
+      callback(step, step_trace, latency)
+      wp.capture_launch(step_capture.graph)
+      wp.synchronize()
+
+    def refit_and_render(m, d, rc):
```
erikfrey, line 274: nit: let's add a TODO that user can just specify more than one function and it will benchmark them together, e.g.

```mjwarp-testspeed --function=refit_bvh --function=render```

erikfrey (approved):
Nice, LGTM - just nits

mujoco_warp/record.py
thowell: added a todo

mujoco_warp/_src/cli.py
thowell: done

mujoco_warp/record.py
thowell: good catch -- dont think the render context could ever be `None` so this check is changed to checking `m.ncam == 0`. once we support rendering a free camera with the batch renderer we can remove the `ncam` check.

mujoco_warp/record.py
thowell: done

mujoco_warp/record.py
thowell: done

mujoco_warp/testspeed.py
thowell: added a todo

PR #1363, opened 2026-05-19
Render only active lights

tkelestemur:
This updates the MuJoCo Warp renderer to iterate only over active lights during shading.

Previously, the render kernel looped over all `m.nlight` entries and passed each light’s active flag into `compute_lighting`, which returned `0.0` for inactive lights. This meant inactive lights still paid the per-light loop overhead.

The active light indices are now computed when creating the `RenderContext`, stored as `rc.light_active_adr`, and used directly by the render kernel.
- Add `RenderContext.light_active_adr` for renderer-local active light indices.
- Update the render megakernel to loop over `rc.light_active_adr` instead of `m.nlight`.
- Remove the inactive-light early return from `compute_lighting`.
- Add a focused render test with one inactive light and one active light.

Tests

- `uv run --extra dev pytest mujoco_warp/_src/render_test.py`
- `uv run --extra dev pytest mujoco_warp/_src/io_test.py -k model_batched_fields`
- `uv run --extra dev pytest mujoco_warp/_src/types_test.py`
- `uv run --extra dev ruff check mujoco_warp/_src/render.py mujoco_warp/_src/io.py mujoco_warp/_src/types.py mujoco_warp/_src/render_test.py`
- `uv run --extra dev ruff format --check mujoco_warp/_src/render.py mujoco_warp/_src/io.py mujoco_warp/_src/types.py mujoco_warp/_src/render_test.py`


thowell:
@tkelestemur thanks for addressing this todo!

do we have any benchmark results for before and after the change?

tkelestemur:
Here are the benchmarks I run on L4 machine:

```bash
 COMMON="--function=render --nworld=2048 --nstep=100 --render_width=128 --render_height=128 --render_rgb=true --render_depth=false --render_textures=false --render_shadows=false --format=short --event_trace=true --clear_warp_cache=true"
```

Primitives
```
 uv run mjwarp-testspeed mujoco_warp/test_data/primitives.xml $COMMON
```
**Current main:** (19d4a45ee1f2b3a5aeb27a4607b62c8335420651)
```
jit_duration: 6.093866650946438
run_time: 1.1235463940538466
steps_per_second: 182279.96732833167
converged_worlds: 2048
model_memory: 5348741
data_memory: 343220232
total_memory: 658505728
ncon_mean: 0.0
ncon_p95: 0.0
nefc_mean: 0.0
nefc_p95: 0.0
solver_niter_mean: 0.0
solver_niter_p95: 0.0
render: 5469.287051819265
````

**My branch: sync'd with (19d4a45ee1f2b3a5aeb27a4607b62c8335420651)**

```
jit_duration: 6.015553653705865
run_time: 1.104588994756341
steps_per_second: 185408.3292267242
converged_worlds: 2048
model_memory: 5348741
data_memory: 343220232
total_memory: 658505728
ncon_mean: 0.0
ncon_p95: 0.0
nefc_mean: 0.0
nefc_p95: 0.0
solver_niter_mean: 0.0
solver_niter_p95: 0.0
render: 5376.186557114124
```

erikfrey:
@tkelestemur interesting, thanks!  I think this PR is deleting two features:

* ability to toggle lights on/off dynamically (without destroying and recreating rendercontext)
* ability to specify light active per world

Maybe these aren't important features, e.g. we aren't simulating discotheques, but let's make sure we're all OK with what we're losing here.  cc @StafaH

erikfrey:
@tkelestemur thank you - I see you posted a followup commit to preserve dynamic active lights.  With this in place, is there still a performance benefit?  Was there a reason to pursue this change besides performance benefit?

tkelestemur:
Thanks @erikfrey . I reran the render benchmark on an L4 after the follow-up commit that preserves dynamic/per-world `light_active`.

Using:
`--function=render --nworld=2048 --nstep=100 --render_width=128 --render_height=128 --render_rgb=true --render_depth=false --render_textures=false --render_shadows=false --format=short --event_trace=true --clear_warp_cache=true`

On `primitives.xml`:
- base `19d4a45`: `render: 5466.37`
- PR `30be72c`: `render: 5357.46`

So the dynamic-safe version is still about 2.0% faster on this benchmark.

I also ran a `primitives.xml` variant with 8 declared lights but only 1 active:
- base: `render: 5498.42`, `jit_duration: 57.10s`, `total_memory: 901.8MB`
- PR: `render: 5361.29`, `jit_duration: 6.38s`, `total_memory: 664.8MB`

So the main reason to keep the change is still performance: it avoids doing inactive-light work in the pixel lighting loop, and it also avoids the compile/code-size cost of statically unrolling the render light loop over every declared light. There is not a separate correctness reason beyond making the loop match the active-light semantics; the follow-up commit keeps dynamic toggles and per-world `light_active` intact by rebuilding the active-light list each render.

tkelestemur:
Merged latest `main` into this branch and resolved the renderer conflict with #1392’s render-kernel cleanup. The current PR keeps the new specialized ray-cast / lighting factory structure from `main`, while preserving this PR’s behavior of compacting active lights per world each render and iterating only over `nactive_light[worldid]`.

I reran the L4 benchmarks against the updated base:

Command:
`uv run mjwarp-testspeed ... --function=render --nworld=2048 --nstep=100 --render_width=128 --render_height=128 --render_rgb=true --render_depth=false --render_textures=false --render_shadows=false --format=short --event_trace=true --clear_warp_cache=true`

Results:

| Benchmark | Base `193404a` | PR `c414200` | Delta |
|---|---:|---:|---:|
| `primitives.xml`, 1 active / 1 total light | `render: 13181.80` | `render: 13747.16` | ~4.3% slower |
| `primitives.xml` variant, 1 active / 8 total lights | `render: 16574.72` | `render: 13814.11` | ~16.7% faster |

So after the newer render perf changes on `main`, this is no longer a universal render win. For the common all-lights-active / single-light case, the extra per-render active-light compaction is overhead. For the case this PR is targeting, declared lights with only a subset active, it is still a clear win.

Given that, I think the reason to keep this change is specifically to optimize models that keep inactive lights around while preserving dynamic/per-world `light_active`. If that workload is not important enough to justify the small all-active overhead, I’m fine dropping the PR.

erikfrey:
@tkelestemur I think that's the right call.  Let's close this for now, but if we see folks wanting to simulate raves we can revisit:

disco rave

PR #1365, opened 2026-05-19
types.py update

thowell:


mujoco_warp/_src/types.py
erikfrey, line 44: Don't you think it's safe to delete this?  have we ever changed these in practice?

No action required, just thinking out loud      :-)

erikfrey (approved):
Nice!

mujoco_warp/_src/types.py
thowell: good question -- think the values for these block_dim fields have basically been stable since they were introduced?

PR #1368, opened 2026-05-20
update height field rendering benchmark: set render height and width

thowell:


benchmarks/unitree_g1/__init__.py
```diff
@@ -32,6 +32,8 @@
     "nconmax": 48,
     "njmax": 192,
     "nstep": 200,
+    "render_width": 64,
```
thowell, line 35: note that these resolution (height and width) settings are the defaults (see [cli.py](https://github.com/google-deepmind/mujoco_warp/blob/a28e20b187a73facb2d9eddb6d623995d67559d3/mujoco_warp/_src/cli.py#L49)) but i think we probably want them defined here

benchmarks/unitree_g1/rollout_hfield_render.webp
thowell, line 1: the current rendered image is not correct and doesn't correspond to the render resolution (64x64)

erikfrey (approved):
Makes sense, and yes I like that the benchmark image resolution matches the render output!

PR #1303, opened 2026-04-20
island solver

thowell:
add island solver
- jacobian: dense, sparse
- solver: newton, cg

additionally, fixes a bug with island discovery where the adjacency matrix `tree_tree` was not properly constructed

---

**humanoid: monolithic, newton, dense**

```
mjwarp-testspeed benchmarks/humanoid/humanoid.xml --nworld=8192 --nconmax=24 --njmax=64 -o "opt.use_islands=False" -o "opt.solver="newton"" -o "opt.jacobian="dense"" --event_trace --memory --info --measure_alloc --measure_solver
```

```

Summary for 8192 parallel rollouts

Total JIT time: 0.37 s
Total simulation time: 1.99 s
Total steps per second: 4,107,666
Total realtime factor: 20,538.33 x
Total time per step: 243.45 ns
Total converged worlds: 8192 / 8192

Event trace:

step: 241.46
  
    solve: 122.45
      mul_m: 5.97
    sensor_acc: 3.18

nacon alloc:

mean    std      min     max   
-------------------------------
125330  42559.5   49152  188823
184071    12864  151170  193299
190897  1166.86  189094  192776
191825  703.674  189959  192596
190507  688.169  189219  191474
189070  288.936  188687  189631
189313  184.752  188949  189676
188363  218.536  188108  188994
188766   495.19  188152  189892
191547  934.322  189962  193035

nefc alloc:

mean   std       min  max
-------------------------
36.32   12.1679   11   49
 48.7   1.23693   47   51
51.65   1.26787   50   55
   55         0   55   55
   55         0   55   55
   55         0   55   55
   55         0   55   55
55.31  0.462493   55   56
55.86  0.632772   55   57
56.41  0.491833   56   57

solver niter:

mean     std       min  max
---------------------------
 1.9597  0.991898    1    8
1.81859  0.863628    1    7
1.27772  0.552244    1    8
1.27851   0.61241    1    8
1.34154  0.623869    1    8
1.32148  0.576217    1    7
1.38368  0.641771    1    8
1.39315  0.645936    1    8
1.35529  0.615481    1    8
1.35027  0.610405    1    8

Model memory 0.03 MiB (0.01% of used memory):
 (no field >= 1% of used memory)
Data memory 247.91 MiB (67.37% of used memory):

Other memory: 120.06 MiB (32.62% of used memory)
Total memory: 368.00 MiB (0.76% of total device memory)
```

**humanoid: island, newton, dense**

```
[...]

PR #1369, opened 2026-05-20
sleeping

### LINKED ISSUE #887: sleeping islands
implement [sleeping islands](https://mujoco.readthedocs.io/en/latest/computation/index.html#sleeping-islands)

thowell:
initial implementation of sleeping for constraint islands and the collision driver

based on #1303

---

```
mjwarp-viewer ../mujoco/model/sleep/dominos.xml \
--nconmax=512 \
--njmax=2048 \
--islands_experimental_override \
-o "opt.enableflags=sleep" \
-o "opt.disableflags=nativeccd" 
```
*note: needed to use the primitive box<>box collider otherwise the blocks don't achieve the sleep tolerance

https://github.com/user-attachments/assets/8a687fba-700a-42b0-a6cc-1e90eb28cb1d

### DIFF
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -25,6 +25,7 @@ from mujoco_warp._src import island
 from mujoco_warp._src import math
 from mujoco_warp._src import passive
 from mujoco_warp._src import sensor
+from mujoco_warp._src import sleep
 from mujoco_warp._src import smooth
 from mujoco_warp._src import solver
 from mujoco_warp._src import types
@@ -328,6 +329,11 @@ def _advance(m: Model, d: Data, qacc: wp.array, qvel: Optional[wp.array] = None)
   )
 
   wp.copy(d.qacc_warmstart, d.qacc)
+
+  if not (m.opt.disableflags & DisableBit.ISLAND) and (m.opt.enableflags & EnableBit.SLEEP):
+    sleep.sleep(m, d)
+    fwd_velocity(m, d)
+    sleep.update_sleep(m, d)
 
 
 @wp.kernel
@@ -658,13 +664,37 @@ def fwd_position(m: Model, d: Data, factorize: bool = True):
   smooth.camlight(m, d)
   smooth.flex(m, d)
   smooth.tendon(m, d)
+
+  sleep_enabled = not (m.opt.disableflags & DisableBit.ISLAND) and (m.opt.enableflags & EnableBit.SLEEP)
+
+  if sleep_enabled and m.ntendon > 0:
+    sleep.wake_tendon(m, d)
+    sleep.update_sleep_trees(m, d)
+
   smooth.crb(m, d)
   smooth.tendon_armature(m, d)
   if factorize:
     smooth.factor_m(m, d)
   if m.opt.run_collision_detection:
-    collision_driver.collision(m, d)
+    if sleep_enabled:
+      # pass 1
+      collision_driver.collision(m, d)
+      # check for newly awake
+      skip = wp.zeros(1, dtype=int)
+      sleep.wake_collision(m, d, skip)
+      sleep.update_sleep(m, d)
+      # pass 2: broadphase kernels early-return if skip[0] is 0
+      collision_driver.collision(m, d, skip)
+    else:
+      collision_driver.collision(m, d)
+
   constraint.make_constraint(m, d)
+
+  if sleep_enabled:
+    if m.neq > 0:
+      sleep.wake_equality(m, d)
+    sleep.update_sleep(m, d)
+
   if m.ntree > 1 and not (m.opt.disableflags & types.DisableBit.ISLAND):
     island.island(m, d)
   smooth.transmission(m, d)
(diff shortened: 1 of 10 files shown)


mujoco_warp/_src/sleep.py
```diff
@@ -0,0 +1,1040 @@
...
+  qvel_in: wp.array2d[float],
+  qfrc_applied_in: wp.array2d[float],
+  xfrc_applied_in: wp.array2d[wp.spatial_vector],
+  # In:
+  worldid: int,
+  treeid: int,
+  sleep_tolerance: float,
+) -> bool:
+  policy = tree_sleep_policy[treeid]
+  if policy == SleepPolicy.AUTO_NEVER:
```
yuvaltassa, line 290: What happens if a user loads a model with `mjSLEEP_NEVER`? I realize this is not yet supported but will they get an error?

Maybe add `or policy == SleepPolicy.NEVER` as defensive coding?

mujoco_warp/_src/sleep.py
```diff
@@ -0,0 +1,1040 @@
...
+  # Data in:
+  tree_awake_in: wp.array2d[int],
+  contact_geom_in: wp.array[wp.vec2i],
+  nacon_in: wp.array[int],
+  # Out:
+  tree_asleep_out: wp.array2d[int],  # kernel_analyzer: ignore
+  nwoke_out: wp.array[int],  # kernel_analyzer: ignore
+):
+  worldid, conid = wp.tid()
+  if conid >= nacon_in[0]:
```
yuvaltassa, line 383: Should this be `if conid >= nacon_in[worldid]:` ?

If yes and this is a bug, consider adding a failing test before fixing.

mujoco_warp/_src/sleep.py
```diff
@@ -0,0 +1,1040 @@
...
+  # Data out:
+  nbody_awake_out: wp.array[int],
+  body_awake_out: wp.array2d[int],
+  body_awake_ind_out: wp.array2d[int],
+):
+  worldid, bodyid = wp.tid()
+
+  tree = body_treeid[bodyid]
+  root = body_rootid[bodyid]
+  mocap = body_mocapid[root] if root >= 0 else -1
```
yuvaltassa: `if root >= 0` is always true? not a bug per se but maybe implies a bug elsewhere?

mujoco_warp/_src/io.py
```diff
@@ -1882,6 +1906,47 @@ def reset_contact(
...
+    if elemid < ntree:
+      tree_asleep_out[worldid, elemid] = -(1 + mj_minawake)
+      tree_awake_out[worldid, elemid] = 1
+
+    if elemid < nbody:
+      if body_treeid[elemid] < 0:
+        if body_mocapid[elemid] >= 0:
+          body_awake_out[worldid, elemid] = 1
+        else:
+          body_awake_out[worldid, elemid] = -1
```
yuvaltassa: Why `-1`? I think you are using the convention `SleepState.STATIC = 2`, while `mjS_STATIC = -1` is the MJC convention. 

See also collision_driver.py L525.

Also, why are we changing the enum values? Seems like a potential source of confusion?

mujoco_warp/_src/forward.py
```diff
@@ -658,13 +664,30 @@ def fwd_position(m: Model, d: Data, factorize: bool = True):
...
+
   smooth.crb(m, d)
   smooth.tendon_armature(m, d)
   if factorize:
     smooth.factor_m(m, d)
   if m.opt.run_collision_detection:
     collision_driver.collision(m, d)
+    if sleep_enabled:
+      sleep.wake_collision(m, d)
+      sleep.update_sleep_trees(m, d)
```
yuvaltassa: Wait what?! This seems like a serious bug but nothing would work so I'm very confused.

How is there no second call to `collision_driver.collision(m, d)`?

Please compare to C code. If we detected a contact with a sleeping tree and woke its island then we have to rerun the entire collision pipeline (well technically, just for this island) because now we now that, almost certainly, ther are missing contacts (eg between this island and the world).

If this is the bug that I think it is, please write a failing test first.

mujoco_warp/_src/sleep.py
```diff
@@ -0,0 +1,1040 @@
...
+
+  # Sleep unconstrained trees
+  for t in range(ntree):
+    island_id = tree_island_in[worldid, t]
+    if island_id < 0 or island_id >= num_islands:
+      if tree_asleep_out[worldid, t] == -1:
+        tree_asleep_out[worldid, t] = t  # self-cycle
+        wp.atomic_add(nslept_out, worldid, 1)
+
+      # Ensure sleeping tree dof velocity and acceleration remain exactly zero
```
yuvaltassa, line 967: I don't quite understand this, but I'm also not sure where it's called. At first glance it looks like you are forcing the velocities of all sleeping trees to zero at every step (and not just when they are put to sleep)?

Maybe this is intentional, but please add a test that manually touching a velocity before a step wakes up an already sleeping island.

yuvaltassa (commented):
Consider adding a test which is equivalent to `SleepingUnaffectedByWaking` which I expect would have caught the one major bug I seem to have detected (if it indeed is a bug, maybe I misunderstood)

Also consider adding a multi-world sleeping test that has nworld > 1 with different sleep states in each world.

mujoco_warp/_src/sleep.py
thowell: added not implemented errors for unsupported policies to `io.put_model`

mujoco_warp/_src/sleep.py
thowell: always true, so the check has been removed

mujoco_warp/_src/forward.py
thowell: updated to have a second call to `collision`. the collision pipeline has been refactored to include early returns.

mujoco_warp/_src/sleep.py
thowell: not a bug. `nacon` is the total number of contacts for all worlds and has size 1 https://github.com/google-deepmind/mujoco_warp/blob/452712f4e31113bb1dc5e83b74885b5d88ffe2fd/mujoco_warp/_src/types.py#L2094

mujoco_warp/_src/io.py
thowell: updated to use the mujoco enum `mjtSleepState`

thowell:
hadn't pushed all of the changes -- files should be updated now

mujoco_warp/_src/forward.py
thowell: added a test to verify the second `collision` call: `test_sleep_rerun_collision`

mujoco_warp/_src/sleep.py
thowell: `test_manual_velocity_wake` checks that setting a non-zero velocity will wake a sleeping island.

added a todo regarding always setting the velocity and acceleration to zero for sleeping island dof velocities and accelerations. don't think it should be necessary to zero velocities and accelerations for already asleep islands and this is something we could remove in the future and might result in a performance improvement?

yuvaltassa (approved):
LGTM, feel free to ack and merge, but two tests that I think might be valuable:

1. No `SleepPolicy.AUTO_NEVER` test: The code handles this but there's no test verifying a tree with this policy (eg a tree with actuators) never sleeps.
2. No multi-world test: Given the global contact indexing, a multi-world sleep test might be valuable.

thowell:
- added a test with an actuator to check `AUTO_NEVER` prevents a tree with an actuator from sleeping
- parameterized tests in sleep_test.py to test nworld=1 and nworld=2

PR #1374, opened 2026-05-22
Fix line search cost precision

adenzler-nvidia:
This fixes a numerical issue in line-search convergence when absolute solver costs are large but the accepted improvement is below float32 resolution.

The old path computed `previous_cost - current_cost`, which can round to zero when the two costs are large and nearly equal. This PR evaluates line-search candidates as deltas from the current point instead:

```text
delta_cost(alpha) = cost(alpha) - cost(0)
improvement = -min(delta_cost)
```

For quadratic terms this removes the constant analytically as `alpha * grad0 + 0.5 * alpha^2 * hessian`. For friction, limits, and elliptic contacts, it evaluates the same piecewise cost at the trial point and shifts it by the `alpha = 0` piecewise cost.

The commit stack is intentionally split into the numerical fix, removal of unused monolithic `cost`/`prev_cost` and `gauss` tracking, inactive-constraint handling, current `main` rebase fixes, and a final formatting-only commit. The island solver still uses absolute cost bookkeeping; applying the same shifted-improvement treatment there should be a follow-up.

8k envs, mean of 3 alternating `mjwarp-testspeed` runs versus `upstream/main`:

| model | main | this PR | delta | niter |
| --- | ---: | ---: | ---: | ---: |
| humanoid | 1.145113s | 1.073829s | -6.23% | 1.450097 -> 1.447349 |
| three_humanoids | 1.425615s | 1.391913s | -2.36% | 2.698686 -> 2.690980 |
| g1_flat | 0.725750s | 0.695636s | -4.15% | 4.282551 -> 4.281379 |

CUDA 12.9 `nsys`, `linesearch_iterative` kernel:

| model | main | this PR | delta |
| --- | ---: | ---: | ---: |
| humanoid | 81.492 ms | 79.811 ms | -2.06% |
| three_humanoids | 28.983 ms | 27.640 ms | -4.64% |
| g1_flat | 33.576 ms | 32.954 ms | -1.85% |

Verified locally with:

- `uv run --extra dev pytest mujoco_warp/_src/solver_test.py --tb=short`
- `uv run ruff check mujoco_warp/_src/solver.py mujoco_warp/_src/solver_test.py mujoco_warp/_src/types.py`
- `uv run pre-commit run --all-files`
- `git diff --check upstream/main...HEAD`

I also ran the repo-required `uv run pytest -n 8`. In this environment it fails in two render comparisons, `RenderTest.test_depth_matches_mujoco` and `RenderTest.test_segmentation_matches_mujoco`; both reproduce on current `upstream/main` with the same assertions.


mujoco_warp/_src/solver.py
```diff
@@ -406,50 +476,37 @@ def linesearch_parallel_fused(
...
-          pass
-        # mu * N + T <= 0 : bottom zone
-        elif mu * N + T <= 0.0:
-          out += _eval_cost(ctx_quad_in[worldid, efcid], alpha)
-        # otherwise middle zone
-        else:
-          out += 0.5 * dm * (N - mu * T) * (N - mu * T)
+
+      out += _eval_elliptic_cost(
+        opt_impratio_invsqrt[worldid % opt_impratio_invsqrt.shape[0]],
```
thowell: should we compute the index just once?

mujoco_warp/_src/solver.py
```diff
@@ -406,50 +476,37 @@ def linesearch_parallel_fused(
...
+        contact_friction_in[conid],
```
thowell: should we consider reading values from contact_friction_in and ctx_quad_in once and then passing the values to the cost functions?

mujoco_warp/_src/solver_test.py
```diff
@@ -73,16 +73,15 @@ def test_constraint_update(self, cone, solver_, jacobian):
         overrides={"opt.solver": solver_, "opt.cone": cone, "opt.jacobian": jacobian, "opt.iterations": 0},
       )
 
-      def cost(qacc):
+      def update_mujoco_constraints(qacc):
```
thowell: maybe something like 'update_constraints'? (not sure we need 'mujoco')

mujoco_warp/_src/solver_test.py
```diff
@@ -287,6 +283,76 @@ def test_parallel_linesearch(self, cone, jacobian):
     _assert_eq(Ma_iterative, Ma_parallel, name="Ma")
     _assert_eq(Jaref_iterative, Jaref_parallel, name="Jaref")
 
+  @parameterized.parameters(False, True)
+  def test_linesearch_accepts_sub_ulp_improvement(self, ls_parallel):
```
thowell: what does 'ulp' stand for here?

thowell (approved):
NICE!

adenzler-nvidia:
@thowell @erikfrey should I wait with this one for upstream MuJoCo changes or can this be merged?

adenzler-nvidia:
either way, will look at the comments now.

thowell:
the mujoco changes have landed https://github.com/google-deepmind/mujoco/commit/cd6db9ebe24d448e47018d357d7d131b42715221

mujoco_warp/_src/solver.py
adenzler-nvidia: Done in 0b1f441d — hoisted `impratio_invsqrt` into a local that both elliptic-cost calls share.

mujoco_warp/_src/solver.py
adenzler-nvidia: Done in 0b1f441d — `friction` and the three `quad` values are now read once and passed to both `_eval_elliptic_cost` calls, matching the iterative kernel.

mujoco_warp/_src/solver_test.py
adenzler-nvidia: Renamed to `update_constraints`.

mujoco_warp/_src/solver_test.py
adenzler-nvidia: "ulp" = unit in the last place (the float32 resolution limit). Renamed the test to `test_linesearch_accepts_sub_float32_improvement` to make that explicit.

adenzler-nvidia:
@thowell addressed all four comments in 0b1f441d — hoisted the per-contact reads in the parallel line-search kernel, and renamed `update_constraints` and `test_linesearch_accepts_sub_float32_improvement`. Also merged in the latest upstream/main (clean). Mind taking another look?

The island line-search still uses absolute cost bookkeeping; we'll apply the same shifted-improvement treatment there as a follow-up once this lands.

thowell (approved):

PR #1376, opened 2026-05-22
update cloth benchmark video

thowell:
```
mjwarp-record benchmarks/cloth/scene.xml --output benchmarks/cloth/rollout.webp --nworld 32 --nconmax 4000 --njmax 32000
```


erikfrey (approved):
Oh wonderful!  Did we fix something in flex or did I just record it wrong the first time?

thowell:
there were some changes to flex in the last ~1 month but not really sure what would have affected this benchmark. @quagla any ideas?

PR #1378, opened 2026-05-23
two-pass GJK+EPA to reduce naccdmax memory

erikfrey:
Split ccd() in collision_gjk.py into a GJK phase and an EPA phase, with a thin wrapper preserving the original signature. In the non-hfield kernel, the EPA scratch slot is now allocated only when GJK has confirmed penetration, so naccdmax (= nworld * nccdmax) tracks the count of actually-penetrating pairs rather than all broadphase pairs.

The hfield kernel keeps the original single-pass behavior (multiple prisms per thread make a two-pass split materially more complex; the wins there are smaller).

Lower nccdmax in the aloha benchmark configs accordingly: pot 12 -> 1, cloth 12 -> 1, clutter 320 -> 64.

Benchmarks on a single device:

| benchmark     | step before  | step after   | mem before  | mem after  |
|---------------|--------------|--------------|-------------|------------|
| aloha_pot     |    1281 ns   |    1291 ns   |   2032 MiB  |   592 MiB  |
| aloha_cloth   |   17.27 ms   |   17.26 ms   |    412 MiB  |   412 MiB  |
| aloha_clutter |     329 us   |     307 us   |   2820 MiB  |   772 MiB  |

Cloth memory is unchanged because nworld=32 makes the EPA scratch trivially small either way.


kbayes (approved):

kbayes:
Yeah, I don't see any issues.  👍

mujoco_warp/_src/collision_gjk.py
```diff
@@ -2258,10 +2249,31 @@ def ccd(
...
+def epa_phase(
+  # In:
+  tolerance: float,
+  epa_iterations: int,
+  result: GJKResult,
+  geom1: Geom,
+  geom2: Geom,
+  geomtype1: int,
+  geomtype2: int,
+  is_discrete: bool,
```
kbayes: is_discrete is an easy compute and doesn't need to be passed in.

mujoco_warp/_src/collision_gjk.py
```diff
@@ -2237,19 +2230,17 @@ def ccd(
     geom2.margin = 0.0
     geom2.size = wp.vec3(0.0, geom2.size[1], geom2.size[2])
 
-  # special handling for sphere and capsule (shrink to point and line respectively)
```
kbayes, line 2240: I think this comment and the shallow penetration comment shouldn't be deleted. The code doesn't have enough context without them.

erikfrey: oh yeah agreed, this is a draft by the agent, if the rough sketch looks good to you i'll clean it up and make it a non-draft PR soon

mujoco_warp/_src/collision_gjk.py
erikfrey: Agreed! I've removed is_discrete from the parameters of epa_phase and return values of gjk_phase, and instead computed it internally inside epa_phase.

mujoco_warp/_src/collision_gjk.py
erikfrey: Added all the comments back in the latest commit. Thank you!

erikfrey:
I've addressed both review comments:
1. **Removed `is_discrete` parameter passing**: We now compute `is_discrete` locally within `epa_phase`, which simplifies the signatures for `gjk_phase` and `epa_phase` and their callers in `collision_convex.py`.
2. **Restored deleted comments**: All descriptive comments that had been draft-deleted in `gjk_phase` have been fully restored.

@kbayes would you mind giving this one final review?

kbayes (approved):

PR #1385, opened 2026-05-26
Fix per-step CollisionContext allocation causing OOM in long runs

gholmes829:
`collision()` runs once per physics step and appears to allocate three broadphase scratch arrays (`collision_pair`, `collision_pairid`, `collision_worldid`, each sized `naconmax`) with `wp.empty()` every call via the `create_collision_context()`.

Over longer training runs this alloc/free churn seems to fragment Warp's memory pool to the point where a small allocation fails despite ample free memory (e.g. ~190KB failing on a 24GB GPU after 100M+ steps). Inspecting the history shows this behavior introduced by #1067.

This PR aims to fix this behavior by allocating `CollisionContext` once in `make_data`/`put_data`, storing it as `d.collision_ctx`, and reusing it each step; `ncollision` is reset and the broadphase output rewritten every step. Also moving `CollisionContext` to `types.py` and where it becomes built via the standard `_create_array` machinery like `Contact`.

Lastly, adds new test `test_collision_ctx_persistent`, which fails on the old per-step-allocation behavior and passes with this change.

### DIFF
--- a/mujoco_warp/_src/broadphase_test.py
+++ b/mujoco_warp/_src/broadphase_test.py
@@ -30,7 +30,7 @@ from mujoco_warp._src import collision_driver
 
 def broadphase_caller(m, d):
   """Run broadphase and return the CollisionContext with results."""
-  ctx = collision_driver.create_collision_context(d.naconmax)
+  ctx = d.collision_ctx
   d.ncollision.zero_()
   if m.opt.broadphase == BroadphaseType.NXN:
     collision_driver.nxn_broadphase(m, d, ctx)
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -17,7 +17,6 @@ from typing import Tuple
 
 import warp as wp
 
-from mujoco_warp._src.collision_core import CollisionContext
 from mujoco_warp._src.collision_core import Geom
 from mujoco_warp._src.collision_core import contact_params
 from mujoco_warp._src.collision_core import geom_collision_pair
@@ -36,6 +35,7 @@ from mujoco_warp._src.types import MJ_MAX_EPAHORIZON
 from mujoco_warp._src.types import MJ_MAXCONPAIR
 from mujoco_warp._src.types import MJ_MAXVAL
 from mujoco_warp._src.types import NEW_GAP_SEMANTICS
+from mujoco_warp._src.types import CollisionContext
 from mujoco_warp._src.types import Data
 from mujoco_warp._src.types import DisableBit
 from mujoco_warp._src.types import GeomType
--- a/mujoco_warp/_src/collision_core.py
+++ b/mujoco_warp/_src/collision_core.py
@@ -15,7 +15,6 @@
 
 """Core collision types and utilities shared across collision modules."""
 
-import dataclasses
 from typing import Tuple
 
 import warp as wp
@@ -344,27 +343,3 @@ def contact_params(
   )
 
   return geoms, margin, gap, condim, friction, solref, solreffriction, solimp
-
-
-@dataclasses.dataclass
-class CollisionContext:
-  """Collision driver intermediate arrays.
-
-  Attributes:
-    collision_pair: collision pairs from broadphase             (naconmax, 2)
-    collision_pairid: ids from broadphase                       (naconmax, 2)
-    collision_worldid: collision world ids from broadphase      (naconmax,)
-  """
-
-  collision_pair: wp.array
-  collision_pairid: wp.array
-  collision_worldid: wp.array
-
-
-def create_collision_context(naconmax: int) -> CollisionContext:
-  """Create a CollisionContext with allocated arrays."""
-  return CollisionContext(
-    collision_pair=wp.empty(naconmax, dtype=wp.vec2i),
-    collision_pairid=wp.empty(naconmax, dtype=wp.vec2i),
-    collision_worldid=wp.empty(naconmax, dtype=int),
-  )
--- a/mujoco_warp/_src/collision_driver.py
+++ b/mujoco_warp/_src/collision_driver.py
@@ -18,8 +18,6 @@ from typing import Any
 import warp as wp
 
 from mujoco_warp._src.collision_convex import convex_narrowphase
-from mujoco_warp._src.collision_core import CollisionContext
-from mujoco_warp._src.collision_core import create_collision_context
 from mujoco_warp._src.collision_flex import flex_narrowphase
 from mujoco_warp._src.collision_primitive import primitive_narrowphase
 from mujoco_warp._src.collision_sdf import sdf_narrowphase
@@ -27,6 +25,7 @@ from mujoco_warp._src.math import upper_tri_index
 from mujoco_warp._src.types import MJ_MAXVAL
 from mujoco_warp._src.types import BroadphaseFilter
 from mujoco_warp._src.types import BroadphaseType
+from mujoco_warp._src.types import CollisionContext
 from mujoco_warp._src.types import CollisionType
 from mujoco_warp._src.types import Data
 from mujoco_warp._src.types import DisableBit
@@ -789,7 +788,7 @@ def collision(m: Model, d: Data):
     d.nacon.zero_()
     return
 
-  ctx = create_collision_context(d.naconmax)
+  ctx = d.collision_ctx
 
   # zero counters
   wp.launch(_zero_nacon_ncollision, dim=1, outputs=[d.nacon, d.ncollision])
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -895,6 +895,39 @@ class CollisionTest(parameterized.TestCase):
 
     self.assertEqual(d.nacon.numpy()[0], mjd.ncon)
 
+  def test_collision_ctx_persistent(self):
+    """Tests that collision() reuses the persistent d.collision_ctx buffers across calls."""
+    _, _, m, d = test_data.fixture("humanoid/humanoid.xml", keyframe=0)
+
+    ctx0 = d.collision_ctx
+    self.assertIsInstance(ctx0, types.CollisionContext)
+    ptrs0 = (ctx0.collision_pair.ptr, ctx0.collision_pairid.ptr, ctx0.collision_worldid.ptr)
+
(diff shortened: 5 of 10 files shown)


erikfrey:
@gholmes829 wow super interesting, thank you for raising this issue.  Let me tag in some Warp folks because we rely on inline allocation throughout mjwarp.

Can you share a bit more about your training setup?  Are you using pytorch on top of this to train or something else?

cc @adenzler-nvidia @Kenny-Vilella

adenzler-nvidia:
Thanks for reporting - What version of warp was this tested on? I remember there being some issues with not properly freeing graphed allocations, but of course the fragmentation problem could be a new thing as well.

gholmes829:
Hey @erikfrey and @adenzler-nvidia, thanks for the responses! 

I'm working to train locomotion on a couple of rigid body robots -- quads, bipeds, some hybrids. I've been keeping `num_envs` in the 1-4k range depending on morphology complexity and such. 

Running Brax PPO (0.14.1) on MuJoCo Playground `MjxEnv` with the `warp` backend running JAX (not pytorch), `warp-lang` 1.12.0, `mujoco-warp` 3.5.0.2, `mujoco`/`mujoco-mjx` 3.5.0, `jax` 0.9.1.

I first noticed the problem when some of my overnight runs would unexpectedly die despite my monitors showing available vram to be healthy. Adding the contents of this PR as patch fixed it.

It looks like cuda graph capture runs by default, I don't do anything to explicitly configure it though. I'd need to study up bit more on the graph capture, but I'm pretty sure my proposed changes move the allocation out of the captured step?

adenzler-nvidia:
This unreleased fix looks slightly relevant: https://github.com/NVIDIA/warp/issues/1429 but looking at your report, I don't think it will fix your problem given that you state that the memory consumption is actually staying low. Still might be worth a try.

gholmes829:
> This unreleased fix looks slightly relevant: [NVIDIA/warp#1429](https://github.com/NVIDIA/warp/issues/1429) but looking at your report, I don't think it will fix your problem given that you state that the memory consumption is actually staying low. Still might be worth a try.

Ah thanks for pointing that one out; and yeah, I think in my case it's more an issue of repeated unnecessary allocations resulting in prohibitive fragmentation (despite having enough memory) rather than growing memory from over-retained graphs.

nvlukasz:
> > This unreleased fix looks slightly relevant: [NVIDIA/warp#1429](https://github.com/NVIDIA/warp/issues/1429) but looking at your report, I don't think it will fix your problem given that you state that the memory consumption is actually staying low. Still might be worth a try.
> 
> Ah thanks for pointing that one out; and yeah, I think in my case it's more an issue of repeated unnecessary allocations resulting in prohibitive fragmentation (despite having enough memory) rather than growing memory from over-retained graphs.

There is another improvement for graph allocations that is slated for the Warp 1.14 release: https://github.com/NVIDIA/warp/issues/1418

That one improves repeated allocation/deallocation behaviour when capturing graphs with multiple CUDA streams. I don't think it will have much impact on single-stream captures, but I'm not sure what you use.

In general, if memory is allocated and deallocated in the same graph, the CUDA mempool allocators should be able to reuse the memory. Not sure if there are some knobs we can tweak to influence this behaviour, but using persistent buffers is an effective way to get around the problem.

gholmes829:
> > > This unreleased fix looks slightly relevant: [NVIDIA/warp#1429](https://github.com/NVIDIA/warp/issues/1429) but looking at your report, I don't think it will fix your problem given that you state that the memory consumption is actually staying low. Still might be worth a try.
> > 
> > 
> > Ah thanks for pointing that one out; and yeah, I think in my case it's more an issue of repeated unnecessary allocations resulting in prohibitive fragmentation (despite having enough memory) rather than growing memory from over-retained graphs.
> 
> There is another improvement for graph allocations that is slated for the Warp 1.14 release: [NVIDIA/warp#1418](https://github.com/NVIDIA/warp/issues/1418)
> 
> That one improves repeated allocation/deallocation behaviour when capturing graphs with multiple CUDA streams. I don't think it will have much impact on single-stream captures, but I'm not sure what you use.
> 
> In general, if memory is allocated and deallocated in the same graph, the CUDA mempool allocators should be able to reuse the memory. Not sure if there are some knobs we can tweak to influence this behaviour, but using persistent buffers is an effective way to get around the problem.

Hrmm yeah I'm using single-stream via MJX's default `GraphMode.WARP`, so unfortunately it doesn't seem that improvement will apply to this case.

The persistent buffer approach is what I landed on too after messing around a bit, thus is the approach I propose in this PR.

nvlukasz:
> Hrmm yeah I'm using single-stream via MJX's default `GraphMode.WARP`, so unfortunately it doesn't seem that improvement will apply to this case.
> 
> The persistent buffer approach is what I landed on too after messing around a bit, thus is the approach I propose in this PR.

Sounds like a good fix in this case, since the allocations are always of the same size. Nonetheless, this is something that we should want to take a look at in Warp, just to make sure we're not doing anything funky. Ideally, we would like graph allocations to be supported seamlessly, with no funny stuff and no need for such workarounds. If you have a straightforward way for me to repro, please let me know.

gholmes829:
> > Hrmm yeah I'm using single-stream via MJX's default `GraphMode.WARP`, so unfortunately it doesn't seem that improvement will apply to this case.
> > The persistent buffer approach is what I landed on too after messing around a bit, thus is the approach I propose in this PR.
> 
> Sounds like a good fix in this case, since the allocations are always of the same size. Nonetheless, this is something that we should want to take a look at in Warp, just to make sure we're not doing anything funky. Ideally, we would like graph allocations to be supported seamlessly, with no funny stuff and no need for such workarounds. If you have a straightforward way for me to repro, please let me know.

I don't have a clean standalone repro handy -- it presents during a full MJX/Brax run after several hours.

One potentially repro-relevant detail is that I run `platform` allocator without preallocation... maybe this could be inhibiting the ability to cleanly recycle the same-size graph allocs due to JAX and Warp interleaving on the same device?

I can try to find some time to poke around at it some more, but in the meantime the persistent buffer in this PR fully mitigates this problem.

erikfrey:
@gholmes829 I'd like to see if we can find a fix in Warp if possible.  We've been moving towards relying more on intermediate allocation.  MJWarp uses far less total on device memory if different parts of the physics step can reuse memory, rather than allocating a bunch of arrays up front that are each only used for a portion of the physics step.

The less memory MJWarp uses, the less it needs to compete with JAX wanting to always grab 75% of the device memory up front          :-)

@gholmes829 the fact that you're using `XLA_PYTHON_CLIENT_ALLOCATOR=platform` is super relevant.  Is there a way to show that the problem goes away when removing that flag?  If so, we have a useful signal to debug.

gholmes829:
Thanks @erikfrey. I don't recall hitting the OOM issue on the few runs when using flags which pre-allocate the memory -- I can try this again on an upcoming run to confirm. The pre-allocation approach is really not ideal for my setup which is why I prefer `platform`, but that's beside the point anyways since you're asking for debug signal.

I've also spent some time looking through Warp's source. A few things stood out that feel worth mentioning:

NVIDIA/warp#1391 (pre-allocated functors for `warp.optim.linear`, shipped in v1.14.0) looks like the same architectural pattern. From the issue body: *"Solve function always re-allocate scratch buffers, which prevent them to be used within a CUDA subgraph."* The `CudaDefaultAllocator` error message at `context.py:3583` recommends *"Try pre-allocating memory before capture begins"*, and the GraphMode.WARP capture comment at `ffi.py:911` notes the design intent to *"reuse it with same buffers."*

These feel like Warp endorsing the persistent-buffer pattern this PR uses, and I was not able to find any other compelling issues or findings which point to upstream fix. I may very well be missing something though.

I'm also curious whether the per-step alloc/free still matters independent of allocator, since it runs as captured graph nodes on every replay regardless.

btaba:
I'm wondering if you can try a different [GraphMode](https://github.com/google-deepmind/mujoco/blob/6957966c7dab0386806b06ae09bab3121ee7f3c5/mjx/mujoco/mjx/_src/io.py#L524) such as `WARP_STAGED` and if that is related to this issue at all.

gholmes829:
@btaba I gave that a try, and `WARP_STAGED` mode does seem to prevent the OOM from manifesting!

Looking into it a bit, this mode seems to be performance motivated, but perhaps the stable graphs prevent the churn which caused the fragmentation and OOM? Just a theory, I'll leave it to the experts to confirm or refute.

Where might this land us for how we want to treat the default mode? It seems the issue can be resolved both by this PR (default mode) and by switching to other mode. On one hand I'd advocate that ideally the default mode wouldn't OOM, but on the other I can understand how this change could be seen as deviating from the intended pattern as described by @erikfrey.

btaba:
@gholmes829 is there a way to provide a repro script? I'd hope this can be solved on the Warp side (as indicated by the PR fix), and a repro would go a long way. Otherwise we'd just add a note to MJX-Warp docs to try different graph modes if facing OOMs (without a better remedy).

gholmes829:
@btaba sure I should be able to put something together to share in next few days, will ping once it's ready

gholmes829:
@btaba hey sorry for the delay, got super busy and swept up in some things.

Good news though, I have prepared a minimal repro to share. Please see https://github.com/gholmes829/mjx-warp-callable-leak.

Running `python repro.py` (~1 min for me) shows Warp's FFI callable registry growing by +2 per identical `jax.jit` re-trace, without bound.

With `python repro.py --oom` (~4–7 min for me on 24 GB 3090), this pattern is driven further via a fresh jit per "eval" and crashes with `Failed to allocate 131072 bytes` despite under 2 GB used / ~22 GB free.

Building the jit once (--reuse-jit) or WARP_STAGED runs this workload without trouble.

Interestingly, the OOM still occurs even with this PR's change applied. My interpretation of this is that there is indeed a deeper issue in the FFI/graph-capture path, and in real use cases the per-step collision allocation is what aggravated it. Meaning the PR here does solve the issue in a pragmatic way, but admittedly does deviate from the intermediate allocation paradigm that @erikfrey mentioned.

I'm wondering if perhaps some fix to memoization in `mjx/mujoco/mjx/warp/ffi.py` would be better alternative to this PR?

thowell:
@gholmes829 thank you for providing a repro!

thowell:
@gholmes829 did any recent changes to mujoco warp or warp resolve the reported issue with training?

PR #1392, opened 2026-05-31
Render kernel perf improvement and cleanup

StafaH:
## Render kernel performance improvements

Optimizes the render kernels to reduce register pressure and improve throughput, plus a couple of small robustness/cleanup fixes.

### Changes
- **Configurable render block dim.** Added `BlockDim.render` (default `64`) and pass `block_dim=m.block_dim.render` to the render kernel launch.
- **Static branch elimination for ray casting.** `create_render_context` now computes `geom_ray_types` — the set of `GeomType`s actually present among enabled geoms (plus `FLEX` when flex primitives exist). The per-type intersection branches in the ray-cast kernels are gated behind `wp.static(...)`, so intersection code for geom types not in the scene is eliminated at compile time, avoiding the register cost of unreachable paths.
- **Unified `cast_ray` / `cast_ray_first_hit`.** Both are now produced by a single `_make_cast_ray(geom_ray_types, first_hit=...)` factory, with the closest-hit vs. any-hit (shadow ray) variant resolved at compile time via `wp.static`. Removes the duplicated kernel body that the old `TODO` flagged. `compute_lighting` is built via `_make_compute_lighting` bound to the specialized first-hit caster.
- **Skybox fix.** `render_skybox=True` on a model with no skybox texture no longer asserts; it now degrades gracefully by disabling skybox rendering.
- **Naming cleanup.** `cam_idx` → `camid` for consistency.

### Performance

| Scene | FPS (main) | FPS (this PR) | Speedup |
|-------|-----------|---------------|---------|
|  primitives.xml | 1,049,688 | 2,272,441 | 2.16× |
| ray.xml | 1,254,526 | 1,355,787 | 1.08×|
| franka | 670,820 | 863,465 | 1.29× |

Important note: Almost all the performance improvement comes from defaulting the block_dim to 64. The remaining changes have small improvements, but are mostly refactoring changes that clean up the code without impacting perf.

### DIFF
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -934,4 +934,5 @@ def render(m: Model, d: Data, rc: RenderContext):
       rc.depth_data,
       rc.seg_data,
     ],
+    block_dim=m.block_dim.render,
   )
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -676,31 +676,31 @@ def render(m: Model, d: Data, rc: RenderContext):
   ):
     worldid, rayid = wp.tid()
 
-    # Map global rayid -> (cam_idx, rayid_local) using cumulative sizes
-    cam_idx = int(-1)
+    # Map global rayid -> (camid, rayid_local) using cumulative sizes
+    camid = int(-1)
     rayid_local = int(-1)
     accum = int(0)
     for i in range(nrender):
       num_i = cam_res[i][0] * cam_res[i][1]
       if rayid < accum + num_i:
-        cam_idx = i
+        camid = i
         rayid_local = rayid - accum
         break
       accum += num_i
-    if cam_idx == -1 or rayid_local < 0:
+    if camid == -1 or rayid_local < 0:
       return
 
-    if not render_rgb[cam_idx] and not render_depth[cam_idx] and not render_seg[cam_idx]:
+    if not render_rgb[camid] and not render_depth[camid] and not render_seg[camid]:
       return
 
     # Map active camera index to MuJoCo camera ID
-    mujoco_cam_id = cam_id_map[cam_idx]
+    mujoco_cam_id = cam_id_map[camid]
 
     if wp.static(rc.use_precomputed_rays):
       ray_dir_local_cam = ray[rayid]
     else:
-      img_w = cam_res[cam_idx][0]
-      img_h = cam_res[cam_idx][1]
+      img_w = cam_res[camid][0]
+      img_h = cam_res[camid][1]
       px = rayid_local % img_w
       py = rayid_local // img_w
       ray_dir_local_cam = compute_ray(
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/render.py
```diff
@@ -151,449 +151,347 @@ def sample_skybox(
...
-        vec = v1 - v0
-
-        length = wp.length(vec)
-        edgeq = math.quat_z2vec(vec)
-        mat = math.quat_to_mat(edgeq)
-        size = wp.vec3(flex_radius[flexid], 0.5 * length, 0.0)
-
-        d, n = ray_capsule(pos, mat, size, ray_origin_world, ray_dir_world)
-        hit_mesh_id = flexid
+def _make_cast_ray(geom_ray_types, first_hit=False):
```
thowell: can we type the arguments?

mujoco_warp/_src/render.py
```diff
@@ -151,449 +151,347 @@ def sample_skybox(
...
-      geom_id = hit_geom_id
-      bary_u = u
-      bary_v = v
-      face_idx = f
-      geom_mesh_id = hit_mesh_id
+  return cast_ray
 
-  return geom_id, dist, normal, bary_u, bary_v, face_idx, geom_mesh_id
 
+def _make_compute_lighting(cast_ray_first_hit):
```
thowell: can we add type information to the argument?

mujoco_warp/_src/render.py
```diff
@@ -151,449 +151,347 @@ def sample_skybox(
...
+  """Build compute_lighting bound to a specialized cast_ray_first_hit."""
```
thowell: we could probably simplify this to something like 'Build specialized compute_lighting'

mujoco_warp/_src/render.py
```diff
@@ -151,449 +151,347 @@ def sample_skybox(
...
-        ray_dir_world,
-      )
-    if gtype == GeomType.HFIELD:
-      d, n, u, v, f, geom_hfield_id = ray_mesh_with_bvh(
+      L, dist_to_light = math.normalize_with_norm(lightpos - hitpoint)
+      attenuation = 1.0 / (1.0 + 0.02 * dist_to_light * dist_to_light)
+      if lighttype == 0:  # spot light
+        spot_dir = wp.normalize(lightdir)
+        cos_theta = wp.dot(-L, spot_dir)
+        spot_factor = wp.min(1.0, wp.max(0.0, (cos_theta - 0.85) / (0.95 - 0.85)))
```
thowell: `/ (0.95 - 0.85)` -> `* 10.0`?

mujoco_warp/_src/render.py
```diff
@@ -151,449 +151,347 @@ def sample_skybox(
...
+    ndotl = wp.max(0.0, wp.dot(normal, L))
+    if ndotl == 0.0:
+      return light_contribution
+
+    visible = float(1.0)
+
+    if use_shadows and lightcastshadow:
+      # Nudge the origin slightly along the surface normal to avoid
+      # self-intersection when casting shadow rays
+      eps = 1.0e-4
```
thowell: since `eps` is only used once (?), maybe we just inline the value?

mujoco_warp/_src/render.py
```diff
@@ -151,449 +151,347 @@ def sample_skybox(
...
+      shadow_origin = hitpoint + normal * eps
+      # Distance-limited shadows: cap by dist_to_light (for non-directional)
+      max_t = float(dist_to_light - 1.0e-3)
```
thowell: is `float()` necessary?

mujoco_warp/_src/render.py
```diff
@@ -151,449 +151,347 @@ def sample_skybox(
...
+      if lighttype == 1:  # directional light
+        max_t = float(1.0e8)
```
thowell: is `float()` necessary?

thowell (approved):
great stuff! thanks @StafaH

StafaH:
Thanks @thowell, all fixed!

PR #1396, opened 2026-06-01
Make active-DOF compaction the default sleeping solver

erikfrey:
Active-DOF compaction packs the awake degrees of freedom of the active kinematic trees into a fixed `nvmax`-sized block each step and runs a dense Newton solve on just that block, freezing everything that is asleep. This promotes it to the default constraint solver for sleeping Newton models, in place of the island solver. It keeps static, user-set sizes (no per-island data structures, CUDA-graph friendly) and runs one cooperative dense solve per world over the contiguous active set, rather than the island solver's thread-per-island factorization with per-island indirection.

Compaction is selected by a model-derived flag `m.is_compact`, which is true when the solver is Newton and sleeping is enabled, unless `enable_islands` is set to keep the old island solver. So enabling sleeping (`mjENBL_SLEEP`) on a Newton model now gives you compaction. `nvmax` sizes the compacted block and should be set to the expected peak number of awake DOFs; left at its default of `nv` the block is full width and the solve degenerates into a dense full solve with no benefit. Requesting sleeping with a solver that cannot act on it (CG without `enable_islands`) is now rejected at `put_model` instead of being silently ignored.

On the aloha_clutter pick replay the trajectory has two regimes. Early on the grippers move in free space and nearly everything is asleep (about 16 of 136 DOFs awake); late in the trajectory they manipulate the clutter and more wakes up (about 52 of 136). Per-world-step on an RTX 6000 Ada at nworld=2048 (lower is better):

| phase | none | island | compact |
|---|---|---|---|
| free-flight (~16 awake) | 23357 | 29763 | 2539 |
| full replay (~52 awake) | 24792 | 33827 | 7241 |

That is compaction at 3.4x over the no-islands baseline and 4.7x over the island solver on the full replay, and 9.2x / 11.7x in free-flight (138k and 394k steps/s, against about 40k for the baseline). The constraint solve alone is roughly 5x faster than the full solve and about 10x faster than the island solver's. Worth calling out: on this workload the island solver is actually slower than running no islands at all, because its per-island factorization overhead is not recovered by sleeping, so compaction is the first sleeping path here that beats the plain solver.

Making it the default surfaced two latent bugs in the compaction path, fixed here. The elliptic-cone JTCJ Hessian assembly fell back to a dense O(nvmax_pad^2) scan whenever `nvmax` was a multiple of the tile size, instead of the sparse-aware kernel that maps contacts to only the DOFs they couple. And `cdof_dof` was sized `nvmax` but indexed to `nvmax_pad`, an out-of-bounds read on non-tile-multiple sizes. `solve()` now rebuilds the active-DOF map itself so it works when called directly, and CG models fall through to the existing solver.

The island solver is left in place behind `enable_islands` and is intended for removal in a follow-up.


thowell:
one thing we could change is that currently `nvmax < nv` enables the compact solver + initial sleeping + islands + sleep/wake. we could consider enabling these features when `nvmax` is provided but `nvmax == nv` but that could be future work.

thowell (approved):

thowell:
was able to run the following with #1448 (needed to fix `io.get_data_into`)

```
/tmp/benchmark_assets/aloha_clutter/scene_clutter.xml --nconmax=512 --nccdmax=64 --njmax=1024 --nvmax=64 -o "opt.enableflags=sleep" --replay=/tmp/benchmark_assets/aloha_clutter/pick_clutter.npz
```

and confirm visually that things worked as expected.

thowell:
lgtm

thowell (approved):

PR #1398, opened 2026-06-02
Parallelize CG beta calculation using atomic accumulations.

quagla:
Split the `solve_beta` and `solve_beta_island` kernels into accumulate and finalize stages. The accumulate kernels now run over individual DOFs and use atomic additions to sum up the numerator and denominator of the Polak-Ribière beta formula. The finalize kernels then compute the final beta values per world or per island. This improves parallelism by avoiding per-world/per-island loops over DOFs within a single kernel thread.

Benchmark Name | Parallel Worlds (nworld) | Baseline (steps/sec) | Optimized (steps/sec) | Speedup (%)
-- | -- | -- | -- | --
cloth | 32 | 319.1 | 391.8 | +22.78%
cloth | 1 | 153 | 207 | +35.3%

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -3131,6 +3131,7 @@ def solve_prev_grad_Mgrad(
 
 @wp.kernel
 def solve_beta_zero(
+  # Out:
   ctx_beta_num_out: wp.array[float],
   ctx_beta_den_out: wp.array[float],
 ):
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -3271,10 +3286,18 @@ def _solver_iteration(
 
   # polak-ribiere
   if m.opt.solver == types.SolverType.CG:
+    ctx.beta.zero_()
```
thowell: lets write a kernel that zeros both `beta` and `beta_den`

mujoco_warp/_src/solver.py
```diff
@@ -5448,19 +5488,26 @@ def _solver_iteration_island(
 
   # Polak-Ribière beta (CG only)
   if is_cg:
+    ctx.beta.zero_()
```
thowell: same thing here and we can use the same kernel to zero these arrays

thowell (approved):

PR #1399, opened 2026-06-02
Implement tiled kernels for CG solver

quagla:
Benchmark Name | Parallel Worlds | Baseline | Optimized | Speedup (%)
-- | -- | -- | -- | --
cloth | 32 | 1,077.51 steps/sec | 1,613.63 steps/sec | +49.76%
humanoid | 8,192 | 1,506,732.57 steps/sec | 1,512,146.94 steps/sec | +0.36%

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -3148,7 +3148,6 @@ def _update_gradient_incremental(m: types.Model, d: types.Data, ctx: SolverConte
   _cholesky_factorize_solve(m, d, ctx, skip_unchanged=True)
 
 
-
 @wp.kernel
 def _solve_beta_zero(
   # Out:
@@ -3223,7 +3222,6 @@ def _solve_beta_accumulate(
   den = ctx_prev_grad_in[worldid, dofid] * prev_Mgrad
   wp.atomic_add(ctx_beta_num_out, worldid, num)
   wp.atomic_add(ctx_beta_den_out, worldid, den)
-
 
 
 @wp.kernel
@@ -3351,7 +3349,6 @@ def _solve_cg_finalize(
   if done or solver_niter_out[worldid] == opt_iterations:
     ctx_done_out[worldid] = True
     wp.atomic_add(nsolving_out, 0, -1)
-
 
 
 @wp.kernel
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -3527,7 +3527,7 @@ def _solver_iteration(
 
 
 def init_context(m: types.Model, d: types.Data, ctx: SolverContext | InverseContext, grad: bool = True):
-  if grad and m.opt.solver == types.SolverType.CG and isinstance(ctx, SolverContext):
+  if grad and m.opt.solver == types.SolverType.CG and isinstance(ctx, (SolverContext, IslandSolverContext)):
     ctx.grad_dot.zero_()
 
   # initialize some efc arrays
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -1963,6 +1963,8 @@ def solve_init_search_cg_tiled(
 
   if tid == 0:
     ctx_search_dot_out[worldid] = search_dot_sum[0]
+
+
 @cache_kernel
 def _update_constraint_efc(track_changes: bool):
   TRACK_CHANGES = track_changes
@@ -2973,9 +2975,9 @@ def _update_gradient(m: types.Model, d: types.Data, ctx: SolverContext):
       _update_gradient_grad,
       dim=(d.nworld, m.nv),
       inputs=[d.qfrc_smooth, d.qfrc_constraint, d.efc.Ma, ctx.done],
-    outputs=[ctx.grad, ctx.grad_dot],
-    block_dim=m.block_dim.update_gradient_grad,
-  )
+      outputs=[ctx.grad, ctx.grad_dot],
+      block_dim=m.block_dim.update_gradient_grad,
+    )
 
   if m.opt.solver == types.SolverType.CG:
     smooth.solve_m(m, d, ctx.Mgrad, ctx.grad)
@@ -3437,7 +3439,6 @@ def _solver_iteration(
   nsolving: wp.array[int],
 ):
   _linesearch(m, d, ctx, step_size_cost)
-
 
   # Incremental H is only valid for non-elliptic cones. The elliptic cone
   # path in _update_constraint_efc has early returns that skip state change
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -3148,24 +3148,6 @@ def _update_gradient_incremental(m: types.Model, d: types.Data, ctx: SolverConte
   _cholesky_factorize_solve(m, d, ctx, skip_unchanged=True)
 
 
-@wp.kernel
-def _solve_prev_grad_Mgrad(
-  # In:
-  ctx_grad_in: wp.array2d[float],
-  ctx_Mgrad_in: wp.array2d[float],
-  ctx_done_in: wp.array[bool],
-  # Out:
-  ctx_prev_grad_out: wp.array2d[float],
-  ctx_prev_Mgrad_out: wp.array2d[float],
-):
-  worldid, dofid = wp.tid()
-
-  if ctx_done_in[worldid]:
-    return
-
-  ctx_prev_grad_out[worldid, dofid] = ctx_grad_in[worldid, dofid]
-  ctx_prev_Mgrad_out[worldid, dofid] = ctx_Mgrad_in[worldid, dofid]
-
 
 @wp.kernel
 def _solve_beta_zero(
@@ -3242,22 +3224,6 @@ def _solve_beta_accumulate(
   wp.atomic_add(ctx_beta_num_out, worldid, num)
   wp.atomic_add(ctx_beta_den_out, worldid, den)
 
-
-@wp.kernel
-def _solve_beta_finalize(
-  # In:
-  ctx_beta_num_in: wp.array[float],
-  ctx_beta_den_in: wp.array[float],
-  ctx_done_in: wp.array[bool],
-  # Out:
-  ctx_beta_out: wp.array[float],
-):
-  worldid = wp.tid()
-
-  if ctx_done_in[worldid]:
-    return
-
-  ctx_beta_out[worldid] = wp.max(0.0, ctx_beta_num_in[worldid] / wp.max(types.MJ_MINVAL, ctx_beta_den_in[worldid]))
 
 
 @wp.kernel
@@ -3345,7 +3311,7 @@ def _solve_search_update_cg_tiled(
 
 
 @wp.kernel
-def solve_cg_finalize(
+def _solve_cg_finalize(
   # Model:
   nv: int,
   opt_tolerance: wp.array[float],
@@ -3356,12 +3322,11 @@ def solve_cg_finalize(
   ctx_beta_den_in: wp.array[float],
   ctx_improvement_in: wp.array[float],
   ctx_done_in: wp.array[bool],
+  ctx_grad_dot_in: wp.array[float],
   # Data out:
   solver_niter_out: wp.array[int],
   # Out:
-  ctx_grad_dot_out: wp.array[float],
   ctx_beta_out: wp.array[float],
-  ctx_search_dot_out: wp.array[float],
   nsolving_out: wp.array[int],
   ctx_done_out: wp.array[bool],
 ):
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/types.py
```diff
@@ -90,7 +90,8 @@ class BlockDim:
   update_gradient_cholesky_blocked: int = 32
   update_gradient_JTDAJ_sparse: int = 64
   update_gradient_JTDAJ_dense: int = 96
-  linesearch_iterative: int = 32
+  linesearch_iterative: int = 256
```
thowell: lets revert this change unless this is key to the performance improvements and we confirm that this doesnt introduce performance regressions

mujoco_warp/_src/solver.py
```diff
@@ -3231,46 +3367,75 @@ def _solver_iteration(
...
+        ctx.cost,
+        ctx.prev_cost,
+        ctx.done,
+      ],
+      outputs=[d.solver_niter, nsolving, ctx.done],
+    )
 
 
 def init_context(m: types.Model, d: types.Data, ctx: SolverContext | InverseContext, grad: bool = True):
+  if grad and m.opt.solver == types.SolverType.CG and hasattr(ctx, "grad_dot"):
```
thowell: can we remove the `hasattr` check?

mujoco_warp/_src/solver.py
```diff
@@ -2859,13 +2927,15 @@ def _JTDAJ_sparse(
...
-  wp.launch(
-    update_gradient_grad,
-    dim=(d.nworld, m.nv),
-    inputs=[d.qfrc_smooth, d.qfrc_constraint, d.efc.Ma, ctx.done],
+  wp.launch_tiled(
+    update_gradient_grad_tiled,
+    dim=d.nworld,
+    inputs=[m.nv, d.qfrc_smooth, d.qfrc_constraint, d.efc.Ma, ctx.done],
     outputs=[ctx.grad, ctx.grad_dot],
+    block_dim=m.block_dim.cg_helpers,
```
thowell: lets add a `BlockDim` field, something like `update_gradient_grad`?

mujoco_warp/_src/solver.py
```diff
@@ -2859,13 +2927,15 @@ def _JTDAJ_sparse(
...
   # grad = Ma - qfrc_smooth - qfrc_constraint
-  wp.launch(update_gradient_zero_grad_dot, dim=(d.nworld), inputs=[ctx.done], outputs=[ctx.grad_dot])
+  if m.opt.solver != types.SolverType.CG:
+    wp.launch(update_gradient_zero_grad_dot, dim=(d.nworld), inputs=[ctx.done], outputs=[ctx.grad_dot])
 
-  wp.launch(
-    update_gradient_grad,
-    dim=(d.nworld, m.nv),
-    inputs=[d.qfrc_smooth, d.qfrc_constraint, d.efc.Ma, ctx.done],
+  wp.launch_tiled(
```
thowell: since this change affects all of the benchmarks we should check for regressions

mujoco_warp/_src/solver.py
```diff
@@ -3231,46 +3367,75 @@ def _solver_iteration(
...
-    wp.launch(
-      solve_beta_accumulate,
-      dim=(d.nworld, m.nv),
-      inputs=[ctx.grad, ctx.Mgrad, ctx.prev_grad, ctx.prev_Mgrad, ctx.done],
+    wp.launch_tiled(
+      solve_beta_accumulate_tiled,
+      dim=d.nworld,
+      inputs=[m.nv, ctx.grad, ctx.Mgrad, ctx.prev_grad, ctx.prev_Mgrad, ctx.done],
       outputs=[ctx.beta, ctx.beta_den],
+      block_dim=m.block_dim.cg_helpers,
```
thowell: lets add a `BlockDim` field, something like 'solve_beta_accumulate`?

mujoco_warp/_src/solver.py
```diff
@@ -3231,46 +3367,75 @@ def _solver_iteration(
...
+        nsolving,
+        ctx.done,
+      ],
+    )
+    wp.launch_tiled(
+      solve_search_update_cg_tiled,
+      dim=d.nworld,
+      inputs=[m.nv, ctx.grad, ctx.Mgrad, ctx.search, ctx.beta, ctx.done],
+      outputs=[ctx.search, ctx.search_dot, ctx.prev_grad, ctx.prev_Mgrad],
+      block_dim=m.block_dim.cg_helpers,
```
thowell: lets add a BlockDim field for this setting

mujoco_warp/_src/solver.py
```diff
@@ -3339,12 +3504,22 @@ def _solve(m: types.Model, d: types.Data, ctx: SolverContext):
...
-    inputs=[ctx.Mgrad],
-    outputs=[ctx.search, ctx.search_dot],
-  )
+  if m.opt.solver == types.SolverType.CG:
+    wp.launch_tiled(
+      solve_init_search_cg_tiled,
+      dim=d.nworld,
+      inputs=[m.nv, ctx.grad, ctx.Mgrad],
+      outputs=[ctx.search, ctx.search_dot, ctx.prev_grad, ctx.prev_Mgrad],
+      block_dim=m.block_dim.cg_helpers,
```
thowell: lets add a BlockDim field for this setting

mujoco_warp/_src/types.py
quagla: As it turns out, most of the performance comes from this change..

quagla: Okay I have done further benchmarks. For nworld=1, this is where the improvement comes from, but for nworld=32, the tiles do make a very large difference.

mujoco_warp/_src/solver.py
quagla: I added humanoid to the table

mujoco_warp/_src/solver.py
```diff
@@ -3262,45 +3464,73 @@ def _solver_iteration(
...
+        ctx.grad_dot,
+        ctx.improvement,
+        ctx.done,
+      ],
+      outputs=[d.solver_niter, nsolving, ctx.done],
+    )
 
 
 def init_context(m: types.Model, d: types.Data, ctx: SolverContext | InverseContext, grad: bool = True):
+  if grad and m.opt.solver == types.SolverType.CG and isinstance(ctx, (SolverContext, IslandSolverContext)):
```
thowell: lets remove the `isinstance` check

mujoco_warp/_src/solver.py
```diff
@@ -1931,6 +1931,40 @@ def _solve_init_search(
   wp.atomic_add(ctx_search_dot_out, worldid, search * search)
 
 
+@wp.kernel
+def solve_init_search_cg_tiled(
```
thowell: lets add an `_` prefix

mujoco_warp/_src/solver.py
```diff
@@ -2285,6 +2319,40 @@ def _update_gradient_grad(
   wp.atomic_add(ctx_grad_dot_out, worldid, grad * grad)
 
 
+@wp.kernel
+def update_gradient_grad_tiled(
```
thowell: lets add an `_` prefix

mujoco_warp/_src/solver.py
```diff
@@ -3101,6 +3178,47 @@ def _solve_beta_zero(
   ctx_beta_den_out[worldid] = 0.0
 
 
+@wp.kernel
+def solve_beta_accumulate_tiled(
```
thowell: lets add an `_` prefix

mujoco_warp/_src/solver.py
```diff
@@ -3184,6 +3302,98 @@ def _solve_search_update(
   wp.atomic_add(ctx_search_dot_out, worldid, search * search)
 
 
+@wp.kernel
+def solve_search_update_cg_tiled(
```
thowell: lets add an `_` prefix

mujoco_warp/_src/solver.py
```diff
@@ -3184,6 +3302,98 @@ def _solve_search_update(
...
+
+  search_dot_tile = wp.tile(local_search_dot, preserve_type=True)
+  search_dot_sum = wp.tile_reduce(wp.add, search_dot_tile)
+
+  if tid == 0:
+    ctx_search_dot_out[worldid] = search_dot_sum[0]
+
+
+@wp.kernel
+def solve_cg_finalize(
```
thowell: lets add an `_` prefix

mujoco_warp/_src/solver.py
thowell: @adenzler-nvidia can you please take a look and confirm that the pattern for this kernel launched with `wp.launch_tiled` is structured correctly and should be performant? thanks!

mujoco_warp/_src/solver.py
```diff
@@ -2893,14 +2961,23 @@ def _JTDAJ_sparse(
...
+    wp.launch_tiled(
+      update_gradient_grad_tiled,
+      dim=d.nworld,
+      inputs=[m.nv, d.qfrc_smooth, d.qfrc_constraint, d.efc.Ma, ctx.done],
+      outputs=[ctx.grad, ctx.grad_dot],
+      block_dim=m.block_dim.update_gradient_grad,
+    )
+  else:
+    wp.launch(_update_gradient_zero_grad_dot, dim=d.nworld, inputs=[ctx.done], outputs=[ctx.grad_dot])
+    wp.launch(
```
thowell, line 2974: should this be a `wp.launch_tiled` call?

quagla: I think we want to keep the change limited to CG for now?

mujoco_warp/_src/solver.py
```diff
@@ -3184,6 +3302,98 @@ def _solve_search_update(
...
+
+  search_dot_tile = wp.tile(local_search_dot, preserve_type=True)
+  search_dot_sum = wp.tile_reduce(wp.add, search_dot_tile)
+
+  if tid == 0:
+    ctx_search_dot_out[worldid] = search_dot_sum[0]
+
+
+@wp.kernel
+def solve_cg_finalize(
```
adenzler-nvidia: Every sibling solver kernel is module-private (`_solve_*`). Rename to `_solve_cg_finalize` for consistency.

adenzler-nvidia (commented):
Nice result on cloth, and I traced the CG path as functionally correct — the prev_grad/prev_Mgrad relocation, the fused solve_cg_finalize, and the reordered done-check are all equivalent to the kernels they replace.

A few cleanups I'd like before merge; inline comments cover the rest:

- Dead kernel `_solve_prev_grad_Mgrad` (solver.py:3152) is no longer launched anywhere — the per-iteration prev_grad/prev_Mgrad snapshot now happens inside `_solve_init_search_cg_tiled` and `_solve_search_update_cg_tiled`. No remaining references repo-wide (only the `_island` variant is still used). Please remove.
- Dead kernel `_solve_beta_finalize` (solver.py:3247) — its logic is now inlined into solve_cg_finalize (step 1) and it is no longer launched. Please remove.

mujoco_warp/_src/solver.py
```diff
@@ -3184,6 +3302,98 @@ def _solve_search_update(
...
+  grad_dot = ctx_grad_dot_out[worldid]
+
+  improvement = _rescale(nv, meaninertia, ctx_improvement_in[worldid])
+  gradient = _rescale(nv, meaninertia, wp.sqrt(grad_dot))
+  done = (improvement < tolerance) or (gradient < tolerance)
+  if done or solver_niter_out[worldid] == opt_iterations:
+    ctx_done_out[worldid] = True
+    wp.atomic_add(nsolving_out, 0, -1)
+
+  # 4. Zero grad_dot for the next iteration's update_gradient_grad
```
adenzler-nvidia: These zeroings are unnecessary now that the tiled kernels overwrite their reductions instead of accumulating. `_update_gradient_grad_tiled` writes grad_dot with `=` (not atomic_add), so the next iteration overwrites it before it is read — this step-4 zero, the step-2 search_dot zero (line 3376), and init_context's `ctx.grad_dot.zero_()` (line 3532) are all redundant, left over from the old atomic-add kernels that needed pre-zeroing. The comment here is also misleading: it implies the next update_gradient_grad accumulates, which it no longer does. Please remove all three plus the comment.

mujoco_warp/_src/types.py
```diff
@@ -92,6 +96,10 @@ class BlockDim:
   update_gradient_JTDAJ_sparse: int = 64
   update_gradient_JTDAJ_dense: int = 128
   linesearch_iterative: int = 32
+  update_gradient_grad: int = 256
```
adenzler-nvidia, line 99: These four defaults (lines 99-102) are hardcoded at 256, which is tuned for high-nv CG (cloth) but ~2x too large for low-nv models. A robot under CG at high world counts (humanoid nv=27 at 8192 worlds) runs ~14% faster end-to-end at block_dim=32 than at 256; cloth (nv=2706) conversely wants >=256. No single constant is correct — the optimum is a function of nv and nworld (and the GPU's SM count). Please derive these from nv at model-build time in io.py where BlockDim() is populated, e.g. clamp(round_up_to_32(nv), 32, 256), optionally lowering the cap as nworld grows.

mujoco_warp/_src/solver.py
thowell: lets remove the `block_dim` argument to this function

quagla: Done

thowell (approved):

PR #1410, opened 2026-06-05
remove parallel linesearch

thowell:
remove parallel linesearch

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -257,9 +257,6 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   opt.tolerance = max(opt.tolerance, 1e-6)
 
   # warp only fields
-  ls_parallel_id = mujoco.mj_name2id(mjm, mujoco.mjtObj.mjOBJ_NUMERIC, "ls_parallel")
-  opt.ls_parallel = (ls_parallel_id > -1) and (mjm.numeric_data[mjm.numeric_adr[ls_parallel_id]] == 1)
-  opt.ls_parallel_min_step = 1.0e-6  # TODO(team): determine good default setting
   opt.broadphase = types.BroadphaseType.NXN
   opt.broadphase_filter = types.BroadphaseFilter.PLANE | types.BroadphaseFilter.SPHERE | types.BroadphaseFilter.OBB
   opt.graph_conditional = True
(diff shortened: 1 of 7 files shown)


mujoco_warp/_src/io.py
```diff
@@ -2841,7 +2838,6 @@ def override_model(model: types.Model | mujoco.MjModel, overrides: dict[str, Any
 
   Overrides are of the format:
     opt.iterations = 1
-    opt.ls_parallel = True
```
erikfrey, line 2850: let's specifically check for `opt.ls_parallel` and throw a more helpful message calling out that this was removed, so something like:

`ls_parallel was removed in MuJoCo Warp 3.9.1.  The default iterative linesearch is now faster for all use cases."

Something like that.

erikfrey (approved):
Looks good, let's add one helpful error for future users.

mujoco_warp/_src/io.py
erikfrey: Actually, I have a better idea, resolving this.

mujoco_warp/_src/types.py
```diff
@@ -834,8 +832,6 @@ class Option:
   sdf_iterations: int
   # warp only fields:
   impratio_invsqrt: array("*", float)
-  ls_parallel: bool
```
erikfrey, line 845: let's add a custom helper like this for a few versions:

```py
def __post_init__(self):
  if self.ls_parallel:
    raise TypeError("ls_parallel was removed in bla bla")

@retries.setter
def ls_parallel(self, value):
    if value:
      raise AttributeError("ls_parallel was removed in bla bla")```

**Review by erikfrey — APPROVED** [review]
OK here's a better idea i think

File: mujoco_warp/_src/types.py [hunk 3]
```diff
@@ -842,14 +832,29 @@ class Option:
...
   broadphase: BroadphaseType
   broadphase_filter: BroadphaseFilter
   graph_conditional: bool
   run_collision_detection: bool
   contact_sensor_maxmatch: int
 
+  # TODO(team): remove in future version
+  @property
+  def ls_parallel(self) -> bool:
+    raise AttributeError("ls_parallel was removed in MuJoCo Warp 3.9.1.")
```
erikfrey, line 844: I think we could warn and return False here, and only error if you try to pass True in the setter

mujoco_warp/_src/types.py
```diff
@@ -842,14 +832,29 @@ class Option:
...
+
+  @ls_parallel.setter
+  def ls_parallel(self, value: bool):
+    raise AttributeError("ls_parallel was removed in MuJoCo Warp 3.9.1.")
+
+  @property
+  def ls_parallel_min_step(self) -> float:
```
erikfrey, line 851: similar idea here, let's just warn if read, and error if set to non-default

erikfrey (approved):

PR #1411, opened 2026-06-05
Iterate elliptic-cone JTCJ over contact support pairs

adenzler-nvidia:
`update_gradient_JTCJ_sparse` assembled the elliptic-cone Hessian output-stationary: one thread per (contact, dense dof-pair), scanning the contact's `colind` to locate each pair. With `nefc >> nv` almost every dof-pair is absent from a given contact's support, so ~99% of the work was scan-and-skip.

This makes it input-stationary — one thread per (contact, support-pair), decoding the pair index directly into its two dofs, register-summing the cone block, and accumulating into H with a single atomic add (which also fixes a latent write race in the old `+=`). The pair count is bounded by a new `jtcj_max_pairs`, computed once in `put_model` from MuJoCo's mass-matrix sparsity. The math is unchanged; results are bit-identical to the old kernel (differences ~1e-4, from float atomic ordering).

End-to-end, elliptic cone, vs main:

| model | step time | speedup |
|---|---|---|
| aloha clutter (~60k contacts) | 82.1 → 29.0 µs | 2.8× |
| three_humanoids | 4.6 → 1.8 µs | 2.5× |
| g1 + hands | 3.6 → 2.6 µs | 1.4× |

The JTCJ kernel alone drops from ~6.2 ms to ~140 µs (~44×) on aloha; gains scale with contact count. The dense (small-`nv`) and island paths are untouched. Existing elliptic solver/forward tests pass.

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -2435,8 +2435,6 @@ def update_gradient_JTDAJ_dense_tiled(nv_pad: int, tile_size: int, njmax: int):
 def update_gradient_JTCJ_sparse(
   # Model:
   opt_impratio_invsqrt: wp.array[float],
-  dof_tri_row: wp.array[int],
-  dof_tri_col: wp.array[int],
   # Data in:
   contact_dist_in: wp.array[float],
   contact_includemargin_in: wp.array[float],
@@ -2460,10 +2458,7 @@ def update_gradient_JTCJ_sparse(
   # Out:
   ctx_h_out: wp.array3d[float],
 ):
-  conid_start, elementid = wp.tid()
-
-  dof1id = dof_tri_row[elementid]
-  dof2id = dof_tri_col[elementid]
+  conid_start, pairid = wp.tid()
 
   for i in range(nblocks_perblock):
     conid = conid_start + i * dim_block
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -2521,22 +2483,24 @@ def _update_gradient_JTCJ_sparse(
...
-      if pos1 >= 0 and pos2 >= 0:
-        break
-    if pos1 < 0 or pos2 < 0:
+    npairs = rownnz * (rownnz + 1) // 2
+    if pairid >= npairs:
       continue
+    rowadr0 = efc_J_rowadr_in[worldid, efcid0]
+    pos1 = int(0)
+    rem = pairid
+    while rem >= rownnz - pos1:
```
thowell, line 2496: do we save some computations with something like the following?

```
dif = rownnz - pos1
while rem >= dif
  rem -= dif
  ...
  dif = rownnz - pos1
```

thowell (approved):
this is AWESOME 

thanks @adenzler-nvidia!

mujoco_warp/_src/solver.py
erikfrey: fyi i tested this and didn't see a speedup on clutter, so leaving as is

erikfrey:
#1413 is merged - thanks again @adenzler-nvidia

PR #1417, opened 2026-06-08
Block Jacobi preconditioner for CG

quagla:
On cloth:

Before:
1,639.66 steps/sec

After:
2,004.96 steps/sec

The new preconditioner solves a 3×3 block (M + J^T D J)⁻¹ analytically for each body. This is beneficial for flexes but loses the full mass coupling for articulated robots. For large kinematic trees (> 6 DOFs), the preconditioner additionally re-factorizes a copy of M augmented with J^T D J at M's existing non-zero positions, preserving the parent-child mass coupling at O(n) cost.

### DIFF
--- a/benchmarks/aloha/__init__.py
+++ b/benchmarks/aloha/__init__.py
@@ -40,6 +40,7 @@ BENCHMARKS = [
     "nccdmax": 1,
     "njmax": 8192,
     "nstep": 100,
+    "override": "opt.jac_preconditioner=true",
     "assets": [(ASSETS[0], "aloha")],
   },
   {
--- a/benchmarks/cloth/__init__.py
+++ b/benchmarks/cloth/__init__.py
@@ -5,6 +5,7 @@ BENCHMARKS = [
     "nworld": 32,
     "nconmax": 2200,
     "njmax": 8000,
+    "override": "opt.jac_preconditioner=true",
   },
   {
     "name": "cloth_render",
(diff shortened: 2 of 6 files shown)


erikfrey:
@quagla  can you run the benchmarks `python benchmarks/run.py` and report any meaningful changes in performance compared to `main` for benchmarks besides the cloth one?

quagla:
> @quagla can you run the benchmarks `python benchmarks/run.py` and report any meaningful changes in performance compared to `main` for benchmarks besides the cloth one?

The problem is that they all use Newton. Only these two use CG

cloth: 1,584 → 1,904 steps/sec (+20.2%), 26.25 → 17.58 iters (-33%)
aloha_cloth: 80 → 129 steps/sec (+60.1%), 99.99 → 48.49 iters (-51.5%)

thowell:
have you benchmarked the performance of the approach where each block in the preconditioner is the size of the corresponding tree `i` and is `(M + J'DJ)_i^-1`? think it would make sense to have this instead of the 3x3 block preconditioner since in general scenes will have tree sizes not equal to 3. what do you think?

quagla:
> have you benchmarked the performance of the approach where each block in the preconditioner is the size of the corresponding tree `i` and is `(M + J'DJ)_i^-1`? think it would make sense to have this instead of the 3x3 block preconditioner since in general scenes will have tree sizes not equal to 3. what do you think?

Yes I agree, the performance would be 1,822.80 steps/s.

quagla:
Switched to `(M + J'DJ)_i^-1`.

quagla:
I made a new version that combines the analytic Jacobi preconditioner for flex vertices and an approximation of (M + J'DJ)_i^-1 for articulated bodies. I updated the PR description.

quagla:
Will continue after #1424 is submitted

PR #1418, opened 2026-06-09
Add per-world material texture IDs

tkelestemur:
This adds support for per-world material texture assignment in MuJoCo Warp rendering. `mat_texid` can now be expanded from the default shared material texture table to one table per world, allowing downstream users to randomize which preloaded texture a material uses without recompiling the model or rebuilding the render context.

The render path already indexes material fields by world, so the change keeps the shared-model fast path by default and only allocates per-world texture ID storage when `expand_mat_texid` is called. Tests cover in-place texture ID updates and verify that two worlds can render the same material with different textures, then swap them without recompilation.

This is how it can be used:

```python
import mujoco
import mujoco_warp as mjw

mjm = mujoco.MjModel.from_xml_path("scene.xml")
mjd = mujoco.MjData(mjm)

nworld = 4
m = mjw.put_model(mjm)
d = mjw.put_data(mjm, mjd, nworld=nworld)

mjw.expand_mat_texid(m, nworld)

mat_id = mujoco.mj_name2id(mjm, mujoco.mjtObj.mjOBJ_MATERIAL, "table_mat")
tex_ids = [
    mujoco.mj_name2id(mjm, mujoco.mjtObj.mjOBJ_TEXTURE, name)
    for name in ("wood", "marble", "metal", "checker")
]
rgb_role = int(mujoco.mjtTextureRole.mjTEXROLE_RGB)

mat_texid = m.mat_texid.numpy()
mat_texid[:, mat_id, rgb_role] = tex_ids
m.mat_texid.assign(mat_texid)

rc = mjw.create_render_context(mjm, nworld=nworld, use_textures=True)
mjw.render(m, d, rc)
```

And here's an example of 4 worlds table texture randomization with panda:

https://github.com/user-attachments/assets/e4eb8b82-62b6-4c19-be72-0d6b23609be9

This should not affect the rendering performance but I can run the benchmarks if desired.

### DIFF
--- a/mujoco_warp/__init__.py
+++ b/mujoco_warp/__init__.py
@@ -53,6 +53,7 @@ from mujoco_warp._src.history import read_ctrl as read_ctrl
 from mujoco_warp._src.history import read_sensor as read_sensor
 from mujoco_warp._src.inverse import inverse as inverse
 from mujoco_warp._src.io import create_render_context as create_render_context
+from mujoco_warp._src.io import expand_mat_texid as expand_mat_texid
 from mujoco_warp._src.io import get_data_into as get_data_into
 from mujoco_warp._src.io import make_data as make_data
 from mujoco_warp._src.io import put_data as put_data
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -819,14 +819,18 @@ def expand_mat_texid(m: types.Model, nworld: int) -> wp.array:
   if nworld < 1:
     raise ValueError(f"nworld must be positive, got {nworld}.")
 
-  expected_shape = (nworld, m.nmat, int(mujoco.mjtTextureRole.mjNTEXROLE))
   current_shape = tuple(m.mat_texid.shape)
+  if len(current_shape) != 3:
+    raise ValueError(f"m.mat_texid has shape {current_shape}; expected (1, {m.nmat}, nrole) or ({nworld}, {m.nmat}, nrole).")
+
+  expected_shape = (nworld, m.nmat, current_shape[-1])
   if current_shape == expected_shape:
     return m.mat_texid
 
   if current_shape[0] != 1:
     raise ValueError(
-      f"m.mat_texid already has {current_shape[0]} world tables; expected 1 or {nworld}. "
+      f"m.mat_texid already has {current_shape[0]} world tables; expected 1 "
+      f"or {nworld}. "
       "Assign into m.mat_texid to update existing per-world storage."
     )
 
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -798,6 +798,44 @@ def put_model(mjm: mujoco.MjModel) -> types.Model:
   return m
 
 
+def expand_mat_texid(m: types.Model, nworld: int) -> wp.array:
+  """Expands material texture ids to allow per-world texture assignment.
+
+  `put_model` initializes `mat_texid` with one shared material texture table.
+  Call this once before CUDA graph capture when different worlds need different
+  material texture ids. Reset-time changes should then assign into
+  `m.mat_texid` directly.
+
+  Texture ids must refer to textures already present in the compiled
+  `mujoco.MjModel` used to create the render context.
+
+  Args:
+    m: The model containing kinematic and dynamic information (device).
+    nworld: The number of per-world material texture tables to allocate.
+
+  Returns:
+    The expanded `m.mat_texid` array.
+  """
+  if nworld < 1:
+    raise ValueError(f"nworld must be positive, got {nworld}.")
+
+  expected_shape = (nworld, m.nmat, int(mujoco.mjtTextureRole.mjNTEXROLE))
+  current_shape = tuple(m.mat_texid.shape)
+  if current_shape == expected_shape:
+    return m.mat_texid
+
+  if current_shape[0] != 1:
+    raise ValueError(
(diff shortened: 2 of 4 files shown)


thowell:
@tkelestemur thanks for contributing this feature! 

similar to per-world meshes where mujoco warp implements array indexing but does not provide utilities for constructing all of the necessary arrays, per-world textures might be a feature that we leave to environment frameworks. @erikfrey what do you think?

tkelestemur:
Thanks, that framing makes sense. I agree we should not put environment-level texture randomization policy in MJWarp: choosing texture datasets, sampling distributions, reset schedules, and asset-management conventions should stay in downstream frameworks.

The piece I think belongs in MJWarp is narrower: the renderer needs a backend-level way to index material texture IDs per world. Downstream apps can construct all texture assets and choose IDs themselves, but without a per-world `mat_texid` table they cannot apply different texture IDs across worlds inside one batched model/render context. The alternatives are rebuilding models/render contexts per reset, running separate single-world render contexts, or encoding randomization through duplicated geoms/materials, which breaks the usual batched-reset path and is awkward for CUDA graph capture.

So I see this PR as analogous to the low-level per-world mesh support: MJWarp provides the storage/indexing primitive, while environment frameworks remain responsible for constructing and sampling the actual randomized assets. If the concern is API surface, I’m happy to keep it minimal: just `expand_mat_texid(m, nworld)` plus direct assignment into `m.mat_texid`, without adding any high-level sampling utilities.

Without this feature, A downstream app cannot assign different texture IDs per world in one batched render context, because `m.mat_texid` is shared:

```python
# No per-world mat_texid support:
# m.mat_texid has shape (1, nmat, mjNTEXROLE), so this is global.
mat_texid = m.mat_texid.numpy()
mat_texid[0, table_mat_id, rgb_role] = sampled_texture_id
m.mat_texid.assign(mat_texid)

# All worlds render with the same table texture.
mjw.render(m, d, rc)
```

erikfrey:
Hi folks!

At first glance, `expand_mat_texid` doesn't seem like it has any logic particularly tied to `mat_texid` and could be repurposed to expand any batch model field, e.g. any field in `Model` with a leading `*` in its shape annotation might be expandable by the user, no?  So if we have `expand_mat_texid` we should also have `expand_mat_texrepeat` and `expand_pair_solimp` and `expand_eq_solref` and so on and so forth.

I'm looking at the pattern here:

https://github.com/mujocolab/mjlab/blob/0cdc56246999409b83622764f5b38edb660cf16e/src/mjlab/entity/variants.py#L1317

Where it looks like the environment logic creates a default `Model` and then overwrites some batch fields after the model is created.  Perhaps the one downside of letting users do it this way is a bit of wasted device memory - `put_model` allocates an array which is then immediately discarded and replaced by the user.

So perhaps one compromise we could consider that maximizes the generality/utility of changing the API surface would be a new param to `put_model` called, say, `batch_sizes` or so, it might look like:

```py
m = io.put_model(model, batch_sizes={'mat_texid': 8192})
```

Which at least allocates the array up front with the correct size.  It's still up to the user to copy their data into the array after it's been allocated.  Let me know what you think.

cc @kevinzakka would adding this to `put_model` allow the env code to be a bit tidier?

kevinzakka:
Agreed that we shouldn't treat `mat_texid` in any special way. In mjlab we have a helper that expands batched model fields to (nworld, ...), and we figure out which fields to expand by looking at which domain randomization functions the user has enabled (@erikfrey and I have gone back and forth on this a fair bit). A `put_model(batch_sizes=...)` would be cleaner for us for two reasons. First, today we let put_model allocate the (1, ...) array and then swap in our own per-world array afterward. Beyond the wasted allocation, swapping the array out invalidates our captured CUDA graph and forces a re-capture. Doing it up front lets us drop both the throwaway allocation and the rebuild. Second, it's more generic and less fiddly to maintain going forward (and I bet users who write their own libraries on top of mujoco_warp might shoot themselves in the foot regarding the graph invalidation business!).

tkelestemur:
100% agree with @kevinzakka. Being able to dynamically change the fields by having batch fields would make the downstream apps much cleaner. And we don't need to do any breaking changes for this.

I also agree with the sentiment that this shouldn't be `mat_texid` specific and more general.

Just for my understanding, the reason some fields are not batched is to keep the memory low, right? 

@erikfrey let me know if you want me to turn this PR into a more general solution as mentioned above, or open a new PR? I'm happy to implement the suggest generic version. 

Edit: one downside is that this not a great change for the MJX side because the environments now need to know the `nworld` and if I have `vmapped` step or reset functions I cannot call it w/ arbitrary batch sizes but I think that's comprise I'm willing to make. The render context already has that problem. 

cc @hartikainen

thowell:
@tkelestemur lets move forward with a new pr with the general solution following the api @erikfrey suggested. please let us know if you have any questions. thanks!

erikfrey:
@tkelestemur the main reason some fields are not batched is they alter the control flow / operational graph being executed on device.

It’s an oversimplification, but you can think of two categories of params: continuous params (usually floats) that act as scalars in the physics algorithms and discrete (e.g. ints, bools) that cause branching / thread divergence.  The latter we have to be careful with, if we made them per-world it would tank performance.

tkelestemur:
Thanks for explaining @erikfrey, makes sense.

btaba:
would prefer `m = io.put_model(model, nworld=8192, batch_fields={'mat_texid',})`

PR #1420, opened 2026-06-10
Prune edge multicontacts to a max of two contacts

kbayes:
See related commit from MuJoCo: https://github.com/google-deepmind/mujoco/commit/3188801

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -1895,6 +1895,26 @@ def _polygon_clip(
   if npolygon < 1:
     return 0, witness1, witness2
 
+  # if the face is an edge, remove potential duplicates
+  if nface2 == 2 and npolygon > 2:
+    best1 = int(0)
+    best2 = int(1)
+    max_d = float(0.0)
+    for i in range(npolygon):
+      for j in range(i + 1, npolygon):
+        diff = polygon_out[j] - polygon_out[i]
+        d2 = wp.dot(diff, diff)
+        if d2 > max_d:
+          max_d = d2
+          best1 = i
+          best2 = j
+
+    witness2[0] = polygon_out[best1]
+    witness1[0] = witness2[0] - dir
+    witness2[1] = polygon_out[best2]
+    witness1[1] = witness2[1] - dir
+    return 2, witness1, witness2
+
   if npolygon > 4:
     quad = _polygon_quad(polygon_out, npolygon)
     for i in range(4):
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_gjk.py
```diff
@@ -1895,6 +1895,26 @@ def _polygon_clip(
...
     return 0, witness1, witness2
 
+  # if the face is an edge, remove potential duplicates
+  if nface2 == 2 and npolygon > 2:
+    best1 = int(0)
+    best2 = int(1)
+    max_d = float(0.0)
+    for i in range(npolygon):
+      for j in range(i + 1, npolygon):
+        diff = polygon_out[j] - polygon_out[i]
```
thowell: can we read `polygon_out[i]` once above the `j` loop?

mujoco_warp/_src/collision_gjk.py
```diff
@@ -1895,6 +1895,26 @@ def _polygon_clip(
...
+        d2 = wp.dot(diff, diff)
+        if d2 > max_d:
+          max_d = d2
+          best1 = i
+          best2 = j
+
+    witness2[0] = polygon_out[best1]
```
thowell, line 1913: might be more efficient to track something like `polygon_best1` and `polygon_best2` instead of `best1` and `best2` since we can avoid 2 extra global reads to `polygon_out`, here and below.

mujoco_warp/_src/collision_gjk.py
kbayes: Done.

mujoco_warp/_src/collision_gjk.py
kbayes: Is there an extreme penalty hit reading from global memory, because if I understand correctly we would have to do a copy on each update of best1 and best2, instead of tracking the indices?

erikfrey:
@kbayes can you run the benchmarks `python benchmarks/run.py` and report any meaningful changes in performance compared to `main`?

kbayes:
@erikfrey @thowell This is kinda of a rare case so I have an assumption that this will have no impact on performance. I just ran `uv run python3 benchmarks/run.py` with this pr and main and did a quick glance to verify this.

mujoco_warp/_src/collision_gjk.py
thowell: something to benchmark i think -- not blocking

thowell (approved):

PR #1421, opened 2026-06-11
update flex collisions

thowell:
update flex collisions to better follow the mujoco implementation

fixes aloha cloth benchmark scene (nworld=1)

https://github.com/user-attachments/assets/68a0d3c9-6afd-429f-be5f-643b469212a2

```
mjwarp-testspeed /tmp/benchmark_assets/aloha_cloth/scene_cloth.xml --nworld=32 --nconmax=4096 --nccdmax=1 --njmax=40_000
```

```
Total JIT time: 0.49 s
Total simulation time: 113.24 s
Total steps per second: 283
Total realtime factor: 0.57 x
Total time per step: 3538754.84 ns
Total converged worlds: 32 / 32
```

---
these changes should not affect benchmarks w/o flex

humanoid

```
mjwarp-testspeed benchmarks/humanoid/humanoid.xml --nworld=8192 --nconmax=24 --njmax=64
```

this pr
```
Total JIT time: 0.45 s
Total simulation time: 1.59 s
Total steps per second: 5,157,683
Total realtime factor: 25,788.41 x
Total time per step: 193.89 ns
Total converged worlds: 8192 / 8192
```

083e5efd730c57fbcda50e6796c9f803f093bb04
```
Total JIT time: 0.89 s
Total simulation time: 1.58 s
Total steps per second: 5,189,397
Total realtime factor: 25,946.99 x
Total time per step: 192.70 ns
Total converged worlds: 8192 / 8192
```

three humanoids

```
mjwarp-testspeed benchmarks/humanoid/three_humanoids.xml --nworld=8192 --nconmax=100 --njmax=192
```

this pr
```
Total JIT time: 0.48 s
Total simulation time: 10.11 s
Total steps per second: 810,624
Total realtime factor: 4,053.12 x
Total time per step: 1233.62 ns
Total converged worlds: 8192 / 8192
```

083e5efd730c57fbcda50e6796c9f803f093bb04
```
Total JIT time: 0.49 s
Total simulation time: 10.06 s
Total steps per second: 814,538
Total realtime factor: 4,072.69 x
Total time per step: 1227.69 ns
Total converged worlds: 8192 / 8192
```

PR #1416, opened 2026-06-08
update aloha cloth benchmark

thowell:
updates for aloha cloth benchmark

- filter duplicate flex contacts
- reduce benchmark nconmax and njmax
- reduce cloth-box condim from 6 to 3
- update rollout video

```
mjwarp-testspeed /tmp/benchmark_assets/aloha_cloth/scene_cloth.xml --nworld=32 --nconmax=1024 --nccdmax=1 --njmax=8192 --nstep=100
```

```
Loading model from: /tmp/benchmark_assets/aloha_cloth/scene_cloth.xml...

Model
  nq: 2716 nv: 2716 nu: 14 nbody: 921 ngeom: 95
Option
  integrator: EULER
  cone: PYRAMIDAL
  solver: CG iterations: 100 ls_iterations: 50
  is_sparse: True
  broadphase: NXN broadphase_filter: PLANE|SPHERE|OBB
Data
  nworld: 32 naconmax: 32768 njmax: 8192
Rolling out 100 steps at dt = 0.002...

Summary for 32 parallel rollouts

Total JIT time: 0.44 s
Total simulation time: 1.53 s
Total steps per second: 2,091
Total realtime factor: 4.18 x
Total time per step: 478164.05 ns
Total converged worlds: 32 / 32
```

```
mjwarp-viewer /tmp/benchmark_assets/aloha_cloth/scene_cloth.xml --nconmax=1024 --njmax=8192
```

https://github.com/user-attachments/assets/1bc1fdf1-dc30-4bf5-a23e-6948c7eb4bc2

PR #1422, opened 2026-06-11
Deterministic simulation: stable contact ordering, constraint row allocation, and solver reductions

johnnynunez:
MJWarp simulation results currently vary run-to-run because contact ordering from the collision drivers and atomic reductions in the constraint solver are non-deterministic. This makes debugging, regression testing, and reproducible RL training difficult.

This PR makes stepping bitwise deterministic, opt-in via opt.deterministic, in three pieces. First, contacts are sorted into a stable permutation after narrowphase, so downstream kernels always see the same contact order regardless of kernel scheduling. Second, constraint row allocation is restructured from atomic slot grabbing into a count-scan-emit pattern so each constraint always lands in the same row. Third, the sparse solver's atomic accumulations (qfrc_constraint and JTDAJ) are replaced with deterministic gathers using a per-solve column index built once per step.

The goal of this PR is correctness without regressing the default path: with all three changes, qacc/qpos/qvel are bitwise identical across repeated runs over 1000 graph-captured steps in the included determinism tests, and with the flag off the kernels compile to the original atomic code via wp.static dispatch, measuring within run-to-run noise of main from 1 to 8192 worlds (humanoid and three_humanoids, RTX PRO 6000). Deterministic mode itself costs roughly 30-80% of step throughput depending on world count; the detailed characterization is in the comments. Reducing that overhead (tile-parallel scans, a precomputed pair index for the JTDAJ gather) is deferred to a follow-up PR, per review discussion.

Authored with Mark Yang (@mar-yan24); developed and reviewed in mar-yan24/mujoco_warp#4, #6, and #8 before being combined here.

credits to @mar-yan24


johnnynunez:
Test and benchmark results on this branch as pushed (RTX PRO 6000, CUDA 13.3).

Full test suite: `uv run pytest -n 4` -> 1108 passed, 23 skipped.

Determinism: the included determinism_test.py verifies qacc/qpos/qvel are bitwise identical across repeated runs over 1000 graph-captured steps.

Forward-step throughput with `opt.deterministic` left **off** (the default), `mjwarp-testspeed benchmarks/humanoid/humanoid.xml --nstep 500` (steps/second, this branch vs current main):

| nworld | main | this PR (det off) | delta |
|---|---|---|---|
| 1 | 4230 | 4205 | -0.6% |
| 256 | 621692 | 618133 | -0.6% |
| 1024 | 2231204 | 2223241 | -0.4% |
| 4096 | 6178023 | 6177181 | -0.0% |

This is the goal of this PR: bitwise-deterministic stepping as an opt-in flag, with the default path unchanged within run-to-run noise (the `wp.static(deterministic)` dispatch compiles the off path to the original atomic kernels). The cost of deterministic=on itself, measured on three_humanoids, is characterized in https://github.com/google-deepmind/mujoco_warp/pull/1422#issuecomment-4692575678; optimizing that overhead (tile-parallel scans, JTDAJ pair index) is deferred to a follow-up PR.

*(Edited: an earlier version of this comment claimed the deterministic gathers were faster than the atomic reductions; that comparison was against the naive deterministic kernels, not the atomics — see the linked comment for the corrected numbers.)*

thowell:
awesome work!

johnnynunez:
> awesome work!

I already have differentiability in the other PR. I’d also like to add these two features, since I think they’re both important for the research community.

Let me know if you have any comments or if you spot anything wrong. I tried to make the implementation as clean as possible.

adenzler-nvidia:
This is awesome stuff indeed, I really like it.

Question about the performance numbers though - I though we might as well do the precomputation for JTDAJ always based on your claims, but I cannot reproduce that it is faster. I used three_humanoids to test that, what model did you measure this on?

adenzler-nvidia:
same question about your testspeed numbers - what are you actually comparing there? Is this just to show that deterministic = off did not get any slower? Would love to have a good set of numbers to characterize the overhead of deterministic = on.

johnnynunez:
Good catch, and apologies, it was my own test environment with the previous PRs.... Let me explain:

 The perf claim in the original description was wrong, and you were right not to be able to reproduce it. The "-35% / -71%" figures came from the optimization history in mar-yan24/mujoco_warp#8: we compared the segmented-gather version of the deterministic reductions against the first naive O(nefc)-scan version of the same deterministic kernels, on a local 20-body contact pile scene. So we describe how much the det-mode kernels improved during development, not a win over the atomic baseline. The deterministic gathers are slower than the atomics everywhere, as you'd expect... there is no case where you'd want the precomputation with deterministic off. I've rewritten the PR description accordingly.

Here are fresh numbers on three_humanoids so we're on the same model (mjwarp-testspeed, graph captured, nstep=1000, njmax=160, RTX PRO 6000 Blackwell sm_120, CUDA 13.3):

`--function step` (worlds·steps/s):

| nworld | upstream main | det=off | det=on | det on-vs-off |
|---|---|---|---|---|
| 1 | 1588 | 1510 | 1042 | −31% |
| 256 | 201k | 205k | 100k | −51% |
| 1024 | 664k | 662k | 199k | −70% |
| 8192 | 1420k | 1388k | 259k | −81% |

`--function solve` (solver only):

| nworld | upstream main | det=off | det=on | det on-vs-off |
|---|---|---|---|---|
| 1 | 8051 | 8049 | 6089 | −24% |
| 256 | 1693k | 1704k | 1200k | −30% |
| 1024 | 5088k | 5089k | 3346k | −34% |
| 8192 | 7247k | 7449k | 4486k | −40% |

On your second question: yes, the testspeed numbers in the earlier comments were exactly that, checking that deterministic=off doesn't regress. The off path compiles to the original atomic kernels via wp.static dispatch, and the table bears that out: det=off matches upstream main within run-to-run noise at every world count (I re-ran the nworld=1 pair 3x to confirm the small deltas are noise).

For det=on, the step-level overhead is larger than the solve-level overhead because the count-scan-emit row allocation and the contact sort also run outside the solver, and the per-world exclusive scans serialize at high world counts. If the high-nworld overhead matters for the RL use case, the clearest headroom is making those scans tile-parallel and giving the JTDAJ gather a precomputed (row, col) pair index.... happy to pursue that in this PR or a follow-up, whichever you prefer.

adenzler-nvidia:
Thanks for the update, makes a lot of sense.

I'd say the JTDAJ follow-up is definitely material for a follow-up PR, and this PR should focus on correctness while not regressing the "off" path, which you already showed.

johnnynunez:
> Thanks for the update, makes a lot of sense.
> 
> I'd say the JTDAJ follow-up is definitely material for a follow-up PR, and this PR should focus on correctness while not regressing the "off" path, which you already showed.

thanks! Let me do that... Really appreciate it the feedback..
Last push the CPU failures are fixed (the host-side overflow check called wp.get_stream() on CPU devices, now guarded behind is_cuda) and the kernel analyzer is clean, pre-commit and the full determinism suite pass locally on both CPU and GPU. The workflows just need approval to run on this PR.

johnnynunez:
Follow-up PR with the first round of deterministic-mode overhead reductions is up as a draft: #1425 (JTDAJ merge-intersection gather + parallel contact world boundaries; det=on overhead at 8192 worlds goes from -81% to -66%, bitwise determinism unchanged). It's stacked on this branch and I'll rebase it once this lands.

thowell:
@johnnynunez @mar-yan24

couple of questions:

- do we have comparisons with warp's built-in determinism https://github.com/NVIDIA/warp/pull/1355?
- what features are not currently support with determinism in this pr (eg, rendering, island solver, flex)?
- how do these changes affect memory and compile time?

thanks!

johnnynunez:
> @johnnynunez @mar-yan24
> 
> couple of questions:
> 
> * do we have comparisons with warp's built-in determinism [Warp determinism NVIDIA/warp#1355](https://github.com/NVIDIA/warp/pull/1355)?
> * what features are not currently support with determinism in this pr (eg, rendering, island solver, flex)?
> * how do these changes affect memory and compile time?
> 
> thanks!

@eric-heiden @adenzler-nvidia are the expert here… i think that in that thread @eric-heiden answered me that probably we get better performance if we do it in in subset of kernels.

related to the other questions, let me check in detail, anyways i haver ready the other PR getting better performance than here.

johnnynunez:
Following up on the first question properly — I went and actually ran NVIDIA/warp#1355 on mujoco_warp rather than leaving it at "not merged." I built mmacklin/warp@warp-deterministic locally (reports as 1.15.0.dev0), ran mjwarp `main` (the unmodified atomic solver, none of this PR's changes) against it, and just flipped Warp's global `wp.config.deterministic = RUN_TO_RUN` before importing mjwarp so every module compiles with the deterministic atomic rewrite. Same GPU as before (RTX PRO 6000 Blackwell sm_120), humanoid.xml, sparse Jacobian, Newton.

The headline: NVIDIA/warp#1355 does make unmodified mjwarp bitwise-reproducible — so the approach genuinely works end to end. But applied wholesale to the solver it isn't practical on this model/hardware yet, for two concrete reasons that both surfaced as hard errors.

Matched run at 256 worlds (bit_stable = sha1(qacc,qpos,qvel) identical across two in-process runs):

| config | bit-stable | graph capture | µs/step | worlds·steps/s | used_MB |
|---|---|---|---|---|---|
| main, atomics (det off) | no | yes | 368 | 6.96e5 | 1038 |
| main + warp#1355 RUN_TO_RUN | yes | no (eager fallback) | 13765 | 1.86e4 | 45132 |
| this PR, det=off | no | yes | 377 | 6.80e5 | 1038 |
| this PR, det=on | yes | yes | 655 | 3.91e5 | 1038 |

So at 256 worlds this PR is ~1.8x over baseline with graph capture intact and no extra memory, whereas NVIDIA/warp#1355 is ~37x with graph capture lost and +44 GB of scratch.

Why NVIDIA/warp#1355 struggles specifically on the solver:

1. Scatter-buffer sizing on many-atomics-per-thread kernels. The generic scatter-sort-reduce sizes a temp buffer of roughly `num_threads × deterministic_max_records` per atomic target. mjwarp's `_M_sparse` (CRB) and especially `_JTDAJ_sparse` issue many `atomic_add`s per thread into the Hessian, so the default bound overflows right away: `Deterministic scatter buffer overflow in kernel '_JTDAJ_sparse'`. Bumping `deterministic_max_records` to 64 restores correctness but the buffer becomes ~44 GB at just 256 worlds, and at 8192 worlds the record count overflows int32 (2,733,721,343 > 2^31) so the scatter array can't be allocated at all.

2. Graph-capture incompatibility with the solver loop. mjwarp runs Newton inside a CUDA graph conditional-while (`capture_while`). NVIDIA/warp#1355 allocates its deterministic scratch inside the launch, and conditional graph bodies disallow allocation: `Conditional body graph contains an unsupported operation (memory allocation)`. So even where correctness holds, the step falls back to eager — that's the 37x above.

This PR sidesteps both by exploiting that J is fixed for the duration of a solve: it builds one compact per-dof column index (~87 MB at 8192 worlds) once per step and gathers in a fixed order, which stays capture-safe and costs ~1.8x at 256 worlds / ~3x at 8192 rather than 37x.

My read is unchanged but now backed by numbers: the two are complementary — NVIDIA/warp#1355 is the zero-rewrite general tool, this PR is the domain-specific fast path for the solver. The genuinely useful next step is feeding these `_JTDAJ_sparse` / `_M_sparse` cases back to the Warp team as a stress test for scatter-buffer sizing and for allocation inside conditional graphs; if those two are addressed, swapping individual hand-written gathers for the generic path becomes worth re-measuring. Happy to share the harness if useful.

johnnynunez:
Rebased onto current main to clear the merge conflict, which pulled in mujoco v3.9.0 (#1428), the flex collision changes (#1421), the aloha cloth benchmark update (#1416) and per-world skybox (#1429).

The only textual conflict was in constraint.py and resolved cleanly: main added the has_flex branch around contact initialization, so the deterministic count-scan-emit row allocation is threaded through both the flex and non-flex contact init kernels.

Merging those changes surfaced a real crash in deterministic mode that I also fixed here.
The flex collision work added a new contact.elem field and made contact.flex, contact.elem and contact.vert lazily allocated, so they are zero-length null arrays whenever a model has no flex contacts.
The deterministic contact permutation kernel _permute_contacts dereferenced src_flex_in and src_vert_in unconditionally, which read from a null pointer and raised CUDA 700 illegal memory access for any model without flex (for example humanoid) whenever opt.deterministic was set.
compute-sanitizer pinned it precisely: Invalid __global__ read of size 4 bytes at _permute_contacts, Access to 0x0 is out of bounds.

The fix guards the flex, elem and vert gathers behind a runtime shape check so the kernel never dereferences an empty array, and it also carries the new elem field through the sort so flex contact element ids stay consistent after reordering.
The determinism test helper had the same latent assumption on the empty arrays and is updated to skip empty contact fields and to track elem.

Verified with compute-sanitizer memcheck (0 errors) and the full determinism, constraint and collision_driver suites on both GPU and CPU: determinism 48 passed, constraint plus collision_driver 119 passed.

johnnynunez:
Rebased onto current main via merge (the conflicts were against the new flexstrain constraints, the compact sleep solve, and the stable-state fast path; the deterministic gathers cover the standard solve and the compact sleep path keeps upstream's kernels) and pushed e4d4a0a, which closes coverage gaps found while stress-testing the deterministic machinery:

- equality_flexstrain rows allocate through count-scan-emit instead of raw atomics; row placement raced once more than a handful of strain constraints were active per world
- the deterministic flex-contact count kernel now shares its rownnz computation with `_efc_contact_init_flex` through a common `wp.func`; the two could disagree and overlap adjacent J storage
- tendon-limit rows no longer double-reserve efc_nnz, which could raise spurious deterministic-mode overflow errors on tightly sized buffers
- contact sort keys are int64 with worldid as the leading radix; SDF contacts get unique geomcollisionids from the initpoint index and flex contacts get content-stable keys built from element and vertex ids, so key ties no longer fall back to racy emission order
- deterministic mode raises on collision-side buffer overflows (nacon, broadphase pairs, CCD, flex candidates) the same way it already does for nefc
- Newton termination uses an order-fixed reduction, matching the CG path

Determinism tests grew from 48 to 55; the full suite passes (1250 passed, 18 skipped).

PR #1424, opened 2026-06-12
Store M as CSR and factor it per-block (dense / sparse / simple)

adenzler-nvidia:
Apologies for the size of this one. It is a single coherent change — making M's factorization per-block — but that representation threads through every consumer of M: how it is stored (io), the factor and solve kernels (smooth), the Newton Hessian assembly and the island solver (solver, support), plus the type definitions and tests. Most of the diff is the mechanical consequence of M always being CSR rather than many independent changes, so it is best read as one idea propagated outward. Happy to split it if that helps review.

This makes the inertia matrix M's factorization a per-block decision instead of a single global dense-or-sparse choice. Previously `is_sparse` picked one factorization for the whole matrix: a dense tile-Cholesky for small models, or MuJoCo's level-structured sparse LDL for everything else. A large model therefore paid the sparse LDL solve even for dofs whose diagonal block is trivial, and a model with one big tree plus many tiny joints could not use a dense factor for the tree without forcing it on everything.

M is now always stored in compact CSR form, and `is_sparse` only selects the constraint Jacobian/Hessian layout — it no longer governs how M is stored or factored. Each diagonal block of M (a connected kinematic sub-tree, a contiguous dof range) is classified independently into one of three categories: a decoupled diagonal block (a "simple" dof such as a point mass on orthogonal slides, or a free cloth particle) needs no factorization at all and uses D = 1/diag; a small coupled block (≤ 64 dofs) factors as a packed dense tile-Cholesky; a larger coupled block uses the sparse LDL factor. The factor is packed as qLD = [dense region | LDL region], the three passes write disjoint dofs, and a mixed model runs whichever passes it needs (e.g. one large tree plus many small free joints). Dense blocks are densified from the CSR layout with a `tile_load_indexed` gather, which is why this requires warp-lang 1.14 (the lockfile is pinned to the 1.14.0 release), and the block-Cholesky thread count scales with block size. The island solver's M-solve mirrors the same three-way split.

The win lands wherever a coupled block fits the dense tile, or where decoupled "simple" dofs dominate; a single tree larger than the dense cutoff stays on the sparse LDL path and is essentially unchanged. End-to-end throughput (steps/s, best of 8 on a single-process GPU) and the M factor+solve GPU time per step (the kernel this PR changes), branch vs upstream/main on the same GPU and the same warp 1.14.0:

| model | nv | M path | baseline steps/s | this PR steps/s | speedup | M factor+solve baseline → PR |
|---|---:|---|---:|---:|---:|---|
| humanoid | 27 | dense block | 7,640,133 | 8,037,783 | 1.05× | 88.6 → 52.6 µs (−41%) |
| g1 + hands (512 worlds) | 49 | dense block | 181,060 | 189,100 | 1.04× | 194 → 46 µs (−76%) |
| g1 + hands (8192 worlds) | 49 | dense block | 962,968 | 1,019,926 | 1.06× | 941 → 636 µs (−32%) |
| three_humanoids | 81 | 3 dense blocks | 1,323,305 | 1,470,055 | 1.11× | 888 → 290 µs (−67%) |
| flybody | 108 | sparse LDL | 93,783 | 91,890 | 0.98× | 319 → 328 µs (+3%) |
| aloha_clutter | 136 | small blocks | 41,868 | 41,567 | 0.99× | 58 → 23 µs (−60%) |
| cloth | ~2,706 | simple + dense | 7,033 | 7,811 | 1.11× | 17.1 → 3.5 µs/CG-iter (−80%) |

Per-kernel figures are median per-launch GPU time from Nsight Systems over a real stepping rollout; unchanged control kernels (the Newton Hessian factor, `_JTDAJ_sparse`) match within 1% between baseline and branch, confirming the measurement. cloth's row is the M⁻¹ preconditioner applied every CG iteration (a free-joint dense solve plus a 1/diag solve for the 2700 particle dofs, replacing one sparse LDL solve over all dofs).

Two honest caveats. flybody (a single 108-dof tree) stays on the sparse LDL path, so it sees no win and a small overhead (the per-dof simple-skip check, with no payoff when there are no simple or dense blocks) — that is the expected cost of the unified path on a pure-sparse model. And `mul_m` regresses on dense-M models (humanoid: 2.9 → 5.2 µs/launch) because always-CSR M replaces the dense matmul with a CSR gather; this only affects models that were previously `is_sparse=False`, and the factor win outweighs it. Everything that was already sparse keeps its CSR `mul_m` unchanged.

A few non-obvious points for review. `is_sparse` changes meaning: it no longer describes M, only the constraint system, so any code that read it as "M is sparse" must use the new `qLD_has_dense` / `qLD_has_sparse` / `qLD_has_simple` flags. The `get_data_into` round-trip recomputes MuJoCo's LDL factor via `mj_factorM` whenever any block is dense or simple, since those factors do not live in the qLD LDL region — a pure-simple model has no LDL region at all. The Newton solver's Hessian assembly now reconstructs M's contribution from CSR (a scatter into the dense H tile, or an `M_elemid` lookup for the island Hessian) rather than a dense load, since M is no longer stored dense; the Hessian factorization itself is unchanged. And warp-lang 1.14 is now required (and locked to the 1.14.0 release) for `tile_load_indexed`.


erikfrey:
@adenzler-nvidia I notice the test against stable deps is failing with NaNs, but that you updated `pyproject.toml` so it's not immediately obvious to me why this would be.  Let me know if you'd like us to review it even with the failing tests, or if you still have a bit more cleanup to do.

adenzler-nvidia:
will put it to draft and investigate - sorry for the noise.

adenzler-nvidia:
Ready for review now - sorry this got a bit big, but it's not trivial to separate this into individual steps.

mujoco_warp/_src/types.py
```diff
@@ -1984,11 +2005,10 @@ class Data:
     moment_colind: column indices in sparse actuator_moment     (nworld, nJmom)
     actuator_moment: actuator moments                           (nworld, nJmom)
     crb: com-based composite inertia and mass                   (nworld, nbody, 10)
-    M: total inertia                                            (nworld, nv_pad, nv_pad) if dense
-                                                                (nworld, 1, nC) if sparse
-    qLD: upper Cholesky factorization                           (nworld, nv, nv) if dense
-         L'*D*L factorization of M                              (nworld, 1, nC) if sparse
-    qLDiagInv: 1/diag(D)                                        (nworld, nv)
+    M: total inertia, CSR                                       (nworld, 1, nC)
```
erikfrey: can we drop the `1` now that we no longer have a dense option?

e.g. `(nworld, nC)`

mujoco_warp/_src/types.py
```diff
@@ -1984,11 +2005,10 @@ class Data:
...
+    qLD: per-block factor: packed dense region, then the nC     (nworld, 1, qLD_block_total + nC)
```
erikfrey: same here

mujoco_warp/_src/smooth.py
```diff
@@ -823,7 +823,7 @@ def _crb_accumulate(
 
 
 @wp.kernel
-def _M_sparse(
+def _M(
```
erikfrey, line 826: nice.   :-).  only one M!

btw i think you can clean up `_tendon_armature` a bit, it is still checking an `is_sparse` which is no longer needed.

mujoco_warp/_src/solver.py
```diff
@@ -5047,7 +5069,7 @@ def _update_gradient_incremental_island(m: types.Model, d: types.Data, ctx: Isla
     outputs=[ctx.h],
   )
 
-  # Add mass matrix
+  # Add mass matrix. Both island variants read M via M_elemid; is_sparse selects the H layout.
   if m.is_sparse:
     wp.launch(
       _update_gradient_set_h_M_sparse_island,
```
erikfrey: not totally following, but do we need both of these kernels still?

since H is always dense and M is always sparse, can we delete `_update_gradient_set_h_M_dense_island`

erikfrey (changes requested):
This is super cool, and brings us one step closer to strong MJC : MJW parity which is great.

There might be a few opportunities to simplify / clean up a bit more as a result of these changes, some suggestions below

mujoco_warp/_src/io.py
```diff
@@ -726,14 +805,33 @@ def _check_margin(name, t1, t2, margin):
...
   m.M_elemid = M_elemid
+  m.M_hinit_i = M_hinit_i
+
+  # Precompute per-block gather indices for the dense-block densify (tile_load_indexed). For each
+  # dense block and flat slot (row, col), store the CSR address of M[max(i,j), min(i,j)], or nC
+  # (out of bounds -> read as 0) for structurally absent pairs. Laid out [block, slot] so the kernel
+  # reads slice (block_size^2,) at offset blk * block_size^2.
+  for tile in m.M_tiles:
+    sz = tile.size
+    starts = tile.adr.numpy()
```
thowell: can we refactor this a bit so that we dont need this device-to-host transfer?

mujoco_warp/_src/io.py
```diff
@@ -131,6 +131,72 @@ def is_sparse(mjm: mujoco.MjModel) -> bool:
...
+  return [(corners[k], (corners[k + 1] if k + 1 < len(corners) else nv) - corners[k]) for k in range(len(corners))]
+
+
+def _m_allow_dense(mjm: mujoco.MjModel) -> bool:
+  """Whether any block may use the packed dense layout (tendon armature forces all-sparse)."""
+  # tendon armature accumulates into M in CSR layout, which the packed block layout cannot represent
+  return not (mjm.ntendon and np.any(mjm.tendon_armature))
+
+
+def m_block_layout(mjm: mujoco.MjModel) -> dict:
```
thowell, line 162: can we utilize `mjModel` fields like `tree_dofadr`, `tree_dofnum`, and `dof_simplenum` to construct the layout?

mujoco_warp/_src/passive.py
```diff
@@ -630,12 +630,6 @@ def _flex_elasticity(
       f = i
       break
 
-  stiffness_adr_base = flex_stiffnessadr[f]
```
thowell, line 629: think we need this address check?

mujoco_warp/_src/passive.py
```diff
@@ -727,10 +721,6 @@ def _flex_bending(
       f = i
       break
 
-  bendingadr = flex_bendingadr[f]
-  if bendingadr < 0:
```
thowell, line 727: similarly, think we need this check

mujoco_warp/_src/types.py
```diff
@@ -1273,7 +1275,16 @@ class Model:
...
     nmaxpolygon: maximum number of verts per polygon
     nmaxmeshdeg: maximum number of polygons per vert
-    is_sparse: whether to use sparse representations
+    is_sparse: constraint Jacobian/Hessian layout (sparse vs dense). Does not affect M, whose
+      factorization is a per-block decision -- see qLD_* and m_block_layout
+    qLD_has_dense: any M block factors as a packed dense block
+    qLD_has_simple: any M block is simple (diagonal -> 1/diag, no factorization)
+    qLD_has_sparse: any M block factors via sparse LDL (oversized block / tendon armature)
+    qLD_block_total: packed length of the dense region per world (also the offset of the LDL region)
+    qLD_block_adr: packed offset of each dof's diagonal block in the dense region; 0 if sparse (nv,)
```
thowell: nit: lets align the sizes on this and the following rows

mujoco_warp/_src/types.py
adenzler-nvidia: Done — `M` is now `(nworld, nC)`. Flattened the whole CSR matrix family (`M`, `qLD`, `qLU`, `qDeriv`) together, since they flow through the shared `factor_solve_i`/`solve_m`/`mul_m` kernels.

mujoco_warp/_src/types.py
adenzler-nvidia: Done — `qLD` is now `(nworld, qLD_block_total + nC)`.

mujoco_warp/_src/smooth.py
adenzler-nvidia: Done — dropped the `is_sparse` branch in `_tendon_armature`. The dense `M_out[i, j]` path can't exist with CSR-only M (it would be out of bounds), so it's the CSR atomic path now.

mujoco_warp/_src/solver.py
adenzler-nvidia: Done — deleted `_update_gradient_set_h_M_dense_island`. The element-parallel sparse-island variant builds the dense island Hessian for every layout, so a single pass suffices (also dropped its now-unused `nidof` input).

mujoco_warp/_src/io.py
adenzler-nvidia: Switched the block decomposition to `tree_dofadr`/`tree_dofnum` (the kinematic trees are exactly M's diagonal blocks; verified identical to the old corner derivation across the test models). I kept the `M_rownnz` coupling check instead of `dof_simplenum`, though: `dof_simplenum` is a contiguous-suffix run-length, so it only flags simple dofs that form a trailing run and misses interspersed decoupled dofs (e.g. pendula's isolated single-dof bodies). It's a suffix fast-path rather than a complete decoupled classifier, so it would silently reclassify those dofs.

mujoco_warp/_src/io.py
adenzler-nvidia: Done — the block starts now come straight from the host `tiles` dict (`np.array(tiles[sz])`) instead of `tile.adr.numpy()`, so there's no device-to-host round-trip in `put_model`.

mujoco_warp/_src/passive.py
adenzler-nvidia: Good catch — yes, restored. `flex_stiffnessadr` is `-1` for flexes with no elasticity, and without the guard `_flex_elasticity` indexes `flex_stiffness` at a `-1` base (out of bounds). It had been dropped inadvertently in a rebase; it now matches `main`.

mujoco_warp/_src/passive.py
adenzler-nvidia: Same — restored the `flex_bendingadr < 0` guard, for the same out-of-bounds reason.

mujoco_warp/_src/types.py
adenzler-nvidia: Done — aligned.

erikfrey (approved):
So good!

PR #1425, opened 2026-06-12
Reduce deterministic-mode overhead: JTDAJ merge-intersection and parallel contact world boundaries

johnnynunez:
Stacked on #1422 (includes its commits; will rebase to contain only the optimization commits once that lands). Per the review discussion in #1422, this follow-up reduces the overhead of opt.deterministic=True without touching the default path or the determinism guarantees.

Three commits, driven by nsys node-level profiling of the captured graphs. First, the deterministic JTDAJ gather finds common efc rows with a two-pointer merge-intersection over the two per-dof entry lists (both already emitted in ascending efc order) instead of re-scanning the row entries, and the contact world-boundary scan becomes a one-thread-per-contact neighbor comparison instead of O(nworld * nacon). Second — the biggest win — the incremental Newton path is made deterministic instead of disabled: change tracking uses per-efcid flags rather than a nondeterministically-ordered atomic-append list, and the H delta is applied by an ordered merge-intersection gather restricted to flagged rows, so deterministic mode no longer pays a full JTDAJ rebuild plus full Cholesky refactorization on every solver iteration. Third, structural early-outs: disjoint efcid-range tests that skip dof pairs from independent kinematic trees, a per-dof changed flag that skips untouched elements in the incremental gather, and a measured per-row [min, max] colind range that lets the index emit skip rows from other trees in O(1). None of these change the float accumulation order (ascending efcid), so results remain bitwise identical.

Measured on three_humanoids (mjwarp-testspeed, graph captured, nstep=1000, njmax=160, RTX PRO 6000 Blackwell), deterministic=on step throughput in steps/s, deterministic=off unchanged within noise:

| nworld | det=off | det=on at #1422 | det=on this PR | overhead |
|---|---|---|---|---|
| 1 | 1519 | 1042 | 1307 | -31% -> -14% |
| 256 | 205k | 100k | 134k | -51% -> -34% |
| 1024 | 664k | 199k | 426k | -70% -> -36% |
| 8192 | 1395k | 259k | 895k | -81% -> -36% |

The overhead is now roughly flat in world count. Bitwise reproducibility verified beyond the unit suite with a 1000-step graph-captured contact-pile rollout, bitwise identical in qpos/qvel/qacc at every 100-step checkpoint across fresh-data runs. Full suite passes (1108 passed, 23 skipped), pre-commit and kernel analyzer clean.

viz @eric-heiden @adenzler-nvidia


johnnynunez:
Pushed two more rounds of optimization after profiling with nsys (--cuda-graph-trace=node to see inside the captured graphs).

The profile showed the real bottleneck wasn't the gathers themselves: deterministic mode had the incremental Newton path disabled entirely, so it ran the full JTDAJ rebuild plus a full Cholesky refactorization on every solver iteration (~5.5x per step), while the off path rebuilds once and then applies incremental H updates with the skip-unchanged factorization cache. The second commit makes the incremental path deterministic instead of disabled: the change tracker in _update_constraint_efc marks per-efcid flags instead of atomic-appending to a compacted list (the append order was the nondeterminism; the count remains an order-invariant integer sum used only as an any-changes test), and a new _update_gradient_h_incremental_sparse_det applies the H delta with the same merge-intersection gather as the full rebuild, restricted to flagged rows. This was the single biggest win.

The third commit adds early-outs that exploit structure without touching the float accumulation order: (a) both gathers exit in O(1) when the two dofs' efcid ranges are disjoint — dofs in independent kinematic trees never share constraint rows, which kills ~2/3 of the upper-triangle elements on three_humanoids; (b) the incremental gather also exits when neither dof is touched by a changed row, via a per-dof flag built from the changed rows each iteration; (c) the column-index emit skips rows from other trees using a measured per-row [min, max] colind range (measured, not assumed: contact rows have descending colind but tendon and two-joint equality rows are unsorted, so a sortedness-based break would corrupt the index on tendon models).

Cumulative deterministic=on step overhead versus deterministic=off on three_humanoids (graph captured, nstep=1000, njmax=160, RTX PRO 6000):

| nworld | #1422 baseline | first commit | + det incremental | + early outs |
|---|---|---|---|---|
| 1 | -31% | -22% | -16% | -14% |
| 256 | -51% | -49% | -36% | -34% |
| 1024 | -70% | -50% | -39% | -36% |
| 8192 | -81% | -66% | -44% | -36% |

At 8192 worlds that's 259k -> 895k steps/s (3.5x), and the overhead curve is now flat in world count. Every commit was validated for bitwise reproducibility beyond the unit suite: a 1000-step graph-captured rollout on a 12-body contact pile, compared at every 100-step checkpoint across fresh-data runs, stays bitwise identical in qpos/qvel/qacc. Full suite 1108 passed / 23 skipped, kernel analyzer clean.

Remaining cost is now dominated by the once-per-step full JTDAJ rebuild and kernels shared with the off path (qLD_acc, blocked Cholesky), so I'd expect ~-25-30% to be the floor without deeper solver changes.

johnnynunez:
@adenzler-nvidia took your suggestion — forgot about contacts entirely and sorted the constraint rows directly (2978b06). You were right that it simplifies things a lot.

**What changed:**
- Contact slot allocation reverts to the **same fast atomic path as the default mode** — row totals are order-invariant, only slot order varied, so no special-casing there anymore.
- Each contact constraint row gets a canonical key (`conid * nmaxdim + dim`); a parallel rank within the world's row range (rows are already per-world, as you said) gives each row its canonical position, accumulating the nnz prefix in the same pass.
- `contact.efc_address` is remapped to the canonical slots *before* the Jacobian/update kernels, so all constraint data is written directly to its final location — no floats are ever moved.

**Deleted:** the contact count kernel, world-boundaries kernels, the serial per-world scan, and all O(naconmax) scratch. Net **-59 lines**, and zero loops over naconmax remain.

**Numbers** (three_humanoids, nstep=1000, njmax=160, RTX PRO 6000, 3 alternating A/B cycles):

| | row-sort | boundaries (prev HEAD) |
|---|---|---|
| det=on, 256 worlds | 130-138k | 135k |
| det=on, 1024 worlds | 386-424k | 423k |
| det=on, 8192 worlds | 806-884k | 895-930k |
| det=off, 8192 worlds | ~1428k | ~1421k |

Honest tradeoff: the O(n²) per-world rank scan costs ~43us vs ~5us for the old serial scan, which shows up as **5-12% lower det-mode throughput at 8192 worlds** (noise-level at 256, default mode untouched). I also tried CUB segmented sort first — rejected, the sort kernels alone cost ~315us/step at 8192 worlds.

Given this is an opt-in mode and the structural simplification is exactly what you asked for, I think the review-surface win is worth it, but happy to chase the rank-scan cost (bitonic in shared memory) if you'd rather keep the throughput.

Tests: determinism suite 48/48 on GPU and CPU, full suite 1108 passed / 23 skipped.

johnnynunez:
Follow-up (ee7e5a5): recovered the 5-12% det-mode regression from the rank scan, keeping the no-contact-bookkeeping structure.

The O(n²) rank wasn't needed at all: since contacts are already sorted world-major in deterministic mode, each contact's canonical row slot follows directly from an **exclusive scan of per-contact row counts**. No sort, no rank, no remap — rows are written straight into canonical slots and nothing ever moves:

1. count rows (and sparse nnz) per contact — same activity/ndim logic as `_efc_contact_init`
2. `wp.utils.array_scan` (CUB exclusive scan) over naconmax
3. record the prefix at each world's first contact to localize the scan (contacts are contiguous per world)
4. bump per-world nefc/nnz totals at each world's last contact
5. deterministic twin of `_efc_contact_init` writes rows at `row_base[world] + prefix[conid] − world_start[world]` instead of `atomic_add`

**Numbers** (three_humanoids, nstep=1000, njmax=160, RTX PRO 6000, 3 cycles each):

| det=on | scan (new) | rank (prev) | boundaries (original) |
|---|---|---|---|
| 256 worlds | 130-143k | 130-138k | 135k |
| 1024 worlds | 426-431k | 386-424k | 423k |
| 8192 worlds | 871-936k | 806-884k | 895-930k |

det=off unchanged (~1425k @8192) in all three. So: parity with (or slightly above) the boundaries approach, with the structural simplification intact — zero contact-boundary machinery, allocation logic mirrors the default path 1:1.

Tests: determinism suite 48/48 GPU and CPU, full suite 1108 passed / 23 skipped.

johnnynunez:
viz @thowell continuity from https://github.com/google-deepmind/mujoco_warp/pull/1422

johnnynunez:
Rebased onto the updated feature/determinism so this stays stacked on the conflict-free base, which now carries the main merge (mujoco v3.9.0, flex collisions) plus the deterministic-mode crash fix from #1422.

The conflict was a genuine design overlap rather than a textual one: main restructured contact initialization around a has_flex branch with a dedicated _efc_contact_init_flex kernel, while this branch replaced the deterministic contact machinery twice over (first the canonical constraint-row sort, then the exclusive-scan row allocation).
I reconciled them so both survive.

The exclusive-scan fast path (_contact_row_counts, array_scan, _contact_world_starts, _contact_world_totals, _efc_contact_init_det) mirrors the non-flex body resolution only, so it is gated to det and not has_flex.
Flex models in deterministic mode fall back to main's atomic _efc_contact_init_flex, which is the same path they already took before this PR, so no flex behavior regresses.
The deterministic param threading that the earlier strategy needed is gone from both init kernels since the scan approach writes rows directly into canonical slots with no atomic offset.

Verified on both GPU and CPU: determinism 48 passed, constraint plus collision_driver 119 passed, solver 184 passed, and compute-sanitizer memcheck reports 0 errors on the deterministic scan path.

johnnynunez:
Merged the updated #1422 (this branch still includes its commits) and pushed 7a2c07d:

- the deterministic contact-row scan is now flex-aware: models containing any flex no longer silently drop all contact rows, rigid pairs included, to the atomic allocation path
- the deterministic contact emit writes the same JTDAJ block descriptors as the non-deterministic kernels; the sleep compact solve consumes that list, and the missing blocks produced NaN accelerations when deterministic mode ran with SLEEP enabled
- the limit-ball nnz reservation from #1422 that was lost in the rework is restored

Full suite passes (1251 passed, 18 skipped).

PR #1429, opened 2026-06-15
per-world skybox

thowell:


mujoco_warp/_src/render.py
```diff
@@ -709,17 +711,19 @@ def _render_megakernel(
...
-          ray_dir_world,
-        )
-        rgb_out[worldid, rgb_adr[camid] + rayid_local] = pack_rgba_to_uint32(
-          skybox_color[0] * 255.0,
-          skybox_color[1] * 255.0,
-          skybox_color[2] * 255.0,
-          255.0,
-        )
+        skybox_id = skybox_tex_id[worldid % skybox_tex_id.shape[0]]
+        if skybox_id >= 0:
```
StafaH: This if should not be necessary since < 0 should only happen when skybox = False, and our wp.static compiles that out.

StafaH (approved):
LGTM

mujoco_warp/_src/render.py
thowell: fixed

Issue #1430, opened 2026-06-15
Native CCD box-box collider misses/over-penetrates contacts on large box geoms


bd-pmorais:
## Summary

mujoco_warp's default box-box collider (the native CCD / GJK-EPA convex path) progressively over-penetrates and then fails to detect contacts as the box geoms grow large, while the MuJoCo C engine and mujoco_warp's own analytic `box_box` collider (reached by disabling NATIVECCD) remain correct and size-invariant. 

## Minimal repro
```python
import mujoco
import mujoco_warp as mjwarp
import numpy as np
import warp as wp

FREE_HALF = 0.1
FIXED_HALF_Z = 0.25
DT = 0.002
N_STEPS = 1500
SIZES = [1.0, 10.0, 25.0, 50.0, 100.0]


def model_xml(fixed_half_xy: float) -> str:
    top = 2 * FIXED_HALF_Z
    rest_z = top + FREE_HALF
    return f"""
<mujoco>
  <option timestep="{DT}" gravity="0 0 -9.81"/>
  <worldbody>
    <geom name="fixed" type="box" pos="0 0 {FIXED_HALF_Z}"
          size="{fixed_half_xy} {fixed_half_xy} {FIXED_HALF_Z}"/>
    <body name="free" pos="0 0 {rest_z}">
      <freejoint/>
      <geom name="free" type="box" size="{FREE_HALF} {FREE_HALF} {FREE_HALF}" mass="10"/>
    </body>
  </worldbody>
</mujoco>
"""


def settle_mujoco(mjm) -> tuple[int, float]:
    mjd = mujoco.MjData(mjm)
    for _ in range(N_STEPS):
        mujoco.mj_step(mjm, mjd)
    mujoco.mj_forward(mjm, mjd)
    dist = np.array([mjd.contact[i].dist for i in range(mjd.ncon)])
    return mjd.ncon, pen_mm(dist)


def settle_mjwarp(mjm, device) -> tuple[int, float]:
    mjd = mujoco.MjData(mjm)
    mujoco.mj_forward(mjm, mjd)
    with wp.ScopedDevice(device):
        m = mjwarp.put_model(mjm)
        d = mjwarp.put_data(mjm, mjd, nworld=1, nconmax=64, njmax=256)
        for _ in range(N_STEPS):
            mjwarp.step(m, d)
        mjwarp.forward(m, d)
        wp.synchronize()
        ncon = int(d.nacon.numpy()[0])
[...]

PR #1433, opened 2026-06-16
create a copy of collision_table before overriding box-box collider

thowell:
should fix collision table override issue reported in #1430


kbayes (approved):

kbayes:
Thanks for the fix!

PR #1439, opened 2026-06-17
remove `ngravcomp`

### LINKED ISSUE #893: mj_setConst
implement mj_setConst
https://github.com/google-deepmind/mujoco/blob/97eb307ce95d67b3a35c8d08bdc7ec0e559a90bf/src/engine/engine_setconst.c#L818

this routine is useful for model editing and should enable safe domain randomization involving certain `Model` fields on device.
https://mujoco.readthedocs.io/en/latest/programming/simulation.html#mjmodel-changes

- [ ] `setFixed` #905 
- [ ] `set0` #905 
- [ ] `setStat`
- [ ] `setSpring`

thowell:


kevinzakka (approved):

mujoco_warp/_src/passive.py
```diff
@@ -872,10 +873,9 @@ def passive(m: Model, d: Data):
       outputs=[d.qfrc_spring],
     )
 
-  gravcomp = m.ngravcomp and not (m.opt.disableflags & DisableBit.GRAVITY)
-
-  if gravcomp:
-    d.qfrc_gravcomp.zero_()
+  gravity_enabled = not (m.opt.disableflags & DisableBit.GRAVITY)
+  d.qfrc_gravcomp.zero_()
```
erikfrey, line 876: is it safe to move this under the conditional?

```py
if gravity_enabled:
  d.qfrc_gravcomp.zero_()
  wp.launch(
    ...
```

erikfrey (commented):
What purpose is `nvgravcomp` serving?  Is it just a way of short circuiting here for slightly faster performance:

```
if gravity_enabled and (ngravcomp[worldid % ngravcomp.shape[0]] > 0) and not jnt_actgravcomp[dof_jntid[dofid]]:
    qfrc_passive += qfrc_gravcomp_in[worldid, dofid]
```

if that's all it's doing, maybe we can just delete the field entirely?  i can't imagine it's that meaningful for perf?

mujoco_warp/_src/passive.py
thowell: think we want to always zero `qfrc_gravcomp` since it could include non-zeros and then if the disableflags option is changed to disable gravity `qfrc_gravcomp` would continue with persistent incorrect non-zero values

thowell:
removed `ngravcomp`

erikfrey (approved):

PR #1440, opened 2026-06-17
set_const_spring

thowell:


kevinzakka (approved):

mujoco_warp/_src/io.py
```diff
@@ -2777,6 +2790,34 @@ def set_const_0(m: types.Model, d: types.Data):
...
+  smooth.transmission(m, d)
+
+  wp.launch(
+    _resolve_tendon_lengthspring,
+    dim=(d.nworld, m.ntendon),
+    inputs=[d.ten_length],
+    outputs=[m.tendon_lengthspring],
+  )
+
+  wp.copy(d.qpos, qpos_saved)
```
erikfrey, line 3249: hmmm, don't you have to restore the calculated values (e.g. `xpos`) after restoring `qpos_saved` here?

naively, i think this would be (at the bottom):

```py
wp.copy(d.qpos, qpos_saved)
smooth.kinematics(m, d)
smooth.com_pos(m, d)
smooth.tendon(m, d)
smooth.transmission(m, d)
```

although that seems pretty expensive.

erikfrey (approved):
Looks good, just one question

mujoco_warp/_src/io.py
thowell: the set const functions are updated to include an argument `restore` that defaults to `True` that will recompute `Data` fields. additionally a parameterized test has been added to check that `Data` fields have been restored after calling set const functions

mujoco_warp/_src/io.py
```diff
@@ -3068,8 +3082,53 @@ def set_const_0(m: types.Model, d: types.Data):
...
+
+  if restore:
+    smooth.kinematics(m, d)
+    smooth.com_pos(m, d)
+    smooth.tendon(m, d)
+    smooth.transmission(m, d)
+
 
-def set_const(m: types.Model, d: types.Data):
+def set_const(m: types.Model, d: types.Data, restore: bool = True):
```
erikfrey, line 3258: an alternative is to make `d` optional, and call `make_data` if `d` is not provided.

then you would only restore if `d` is provided

would that be a cleaner API?

erikfrey (commented):
One question

mujoco_warp/_src/io.py
thowell: with #1439 `set_const` should be compatible with graph capture. if we include a call to `make_data` in this function then it is no longer compatible with graph capture if a `Data` is not provided.

erikfrey: ah got it, that's a really good reason.  okay let's go with this and we can always revisit if we get reports that it is too slow.

PR #1450, opened 2026-06-23
Make the sleeping collision second pass incremental

erikfrey:
Speeds up the sleeping collision pipeline on the aloha clutter benchmark by making the second collision pass incremental. The two columns after the baseline show the contribution of each part of the change:

| metric | baseline | + incremental filter | + graph conditional |
|---|---|---|---|
| **steps_per_second** | 69,189 | 75,115 | **79,276** |
| run_time | 81.43 | 75.01 | 71.07 |
| collision pass 2 | 2,384 | 1,234 | 581 |
| &nbsp;&nbsp;└ nxn_broadphase | 1,058 | 950 | 294 |
| &nbsp;&nbsp;└ convex_narrowphase | 1,288 | 249 | 252 |
| nefc_mean | 114.80 | 115.05 | 115.05 |
| converged_worlds | 2048 | 2048 | 2048 |

About a 15% throughput improvement, with identical nefc and convergence. All numbers are from the standard benchmark with event tracing on (time units are from the event trace); with tracing off the final build runs around 81k steps/sec.

When sleeping is enabled the collision pipeline runs twice per step: pass 1 collides the awake bodies, wake_collision wakes any sleeping tree touched by an awake one, and pass 2 picks up collisions for the newly-awakened bodies. Previously pass 2 re-ran the entire collision pipeline from scratch (zeroing nacon and recomputing every contact) whenever anything woke anywhere across the batch, which in a large batched scene is essentially every step.

This makes pass 2 incremental. The broadphase keeps the pass-1 contact buffer and only emits pairs that pass 1 skipped because both bodies were asleep (or one asleep and one static) but are collidable now that a body has woken; those contacts are appended via the existing atomic nacon counter. The resulting contact set is identical to the old full recompute, since pass-1 pairs are always a subset of the post-wake pairs (waking never re-sleeps a body) and the incremental pass adds exactly the difference. This cuts the pass-2 narrowphase down to the handful of newly-awakened pairs (the "incremental filter" column above).

On top of that, the NXN broadphase wraps its launch in a CUDA graph conditional so the whole pass is skipped on steps where nothing woke, rather than launching kernels that early-return per thread (the "graph conditional" column). The wake condition is derived locally by comparing the pre-wake awake snapshot against the post-wake body_awake, which also makes it robust to wakes from sources other than collision (e.g. tendon limits) that update_sleep folds in before pass 2. SAP relies on the per-pair incremental filter only, since its sort/scan utilities allocate internally and cannot run inside a conditional body.

Flex collision is intentionally left out of the incremental pass: it is not sleeping-aware, so pass 1 already emits every flex contact regardless of awake state and the incremental pass has nothing to add.

This also removes the now-dead skip/nwoke/nslept wake counters that were write-only outputs threaded through the sleeping kernels, and drops the bespoke nacon/ncollision zeroing kernel in favor of plain zero_() calls.

uv run pytest -n 8 passes (1110 passed, 6 skipped).

### DIFF
--- a/mujoco_warp/_src/collision_driver.py
+++ b/mujoco_warp/_src/collision_driver.py
@@ -93,26 +93,6 @@ MJ_FLEX_COLLISION_TABLE = {
 }
 
 
-@cache_kernel
-def _zero_nacon_ncollision(enable_sleep: bool = False):
-  @wp.kernel(module="unique", enable_backward=False)
-  def zero_nacon_ncollision(
-    # In:
-    skip_in: wp.array[int],
-    # Data out:
-    nacon_out: wp.array[int],
-    ncollision_out: wp.array[int],
-  ):
-    ncollision_out[0] = 0
-    if wp.static(enable_sleep):
-      if skip_in[0] != 0:
-        nacon_out[0] = 0
-    else:
-      nacon_out[0] = 0
-
-  return zero_nacon_ncollision
-
-
 @wp.func
 def _plane_filter(
   size1: float, size2: float, margin1: float, margin2: float, xpos1: wp.vec3, xpos2: wp.vec3, xmat1: wp.mat33, xmat2: wp.mat33
@@ -402,7 +382,7 @@ def _binary_search(values: wp.array[Any], value: Any, lower: int, upper: int) ->
 
 
 @cache_kernel
-def _sap_project(opt_broadphase: int, enable_sleep: bool = False):
+def _sap_project(opt_broadphase: int):
   @wp.kernel(module="unique", enable_backward=False)
   def sap_project(
     # Model:
@@ -415,7 +395,6 @@ def _sap_project(opt_broadphase: int, enable_sleep: bool = False):
     nworld_in: int,
     # In:
     direction_in: wp.vec3,
-    skip_in: wp.array[int],
     # Out:
     projection_lower_out: wp.array2d[float],
     projection_upper_out: wp.array2d[float],
@@ -423,10 +402,6 @@ def _sap_project(opt_broadphase: int, enable_sleep: bool = False):
     segmented_index_out: wp.array[int],
   ):
     worldid, geomid = wp.tid()
-
-    if wp.static(enable_sleep):
-      if skip_in[0] == 0:
-        return
 
     xpos = geom_xpos_in[worldid, geomid]
     rbound = geom_rbound[worldid % geom_rbound.shape[0], geomid]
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/collision_driver.py
```diff
@@ -455,44 +430,40 @@ def sap_project(
...
-    worldid, geomid = wp.tid()
-
-    if wp.static(enable_sleep):
-      if skip_in[0] == 0:
-        range_out[worldid, geomid] = 0
-        return
-
-    # current bounding geom
-    idx = sort_index_in[worldid, geomid]
+@wp.kernel(enable_backward=False)
```
thowell: we dont need to include `enable_backward=False` since this setting is defined at the module level

mujoco_warp/_src/collision_driver.py
```diff
@@ -797,8 +805,28 @@ def kernel(
   return kernel
 
 
+@wp.kernel(enable_backward=False)
```
thowell: we dont need to include `enable_backward=False` since this setting is defined at the module level

mujoco_warp/_src/collision_driver.py
```diff
@@ -882,27 +937,42 @@ def collision(m: Model, d: Data, skip: Optional[wp.array] = None):
...
+
+  # The incremental sleeping pass appends newly-awakened contacts to the pass-1 buffer, so nacon is
+  # preserved and only ncollision (the broadphase pair counter) is reset. nxn_broadphase also skips
+  # its launch wholesale via a graph conditional when nothing woke; pre-zeroing ncollision here
+  # keeps the narrowphase below a no-op in that case. SAP cannot use a conditional (its sort/scan
+  # allocate internally), so it relies on the per-pair incremental filter instead.
+  if incremental:
+    d.ncollision.zero_()
+  else:
+    d.nacon.zero_()
```
thowell, line 961: we could combine these initializations in a fused kernel

mujoco_warp/_src/forward.py
```diff
@@ -628,12 +628,16 @@ def fwd_position(m: Model, d: Data, factorize: bool = True):
...
+      sleep.wake_collision(m, d)
+      # snapshot the awake state pass 1 used, before update_sleep overwrites it. a body is "newly
+      # awakened" if it was asleep here but awake after update_sleep below.
+      awake_prev = wp.clone(d.body_awake)
       sleep.update_sleep(m, d)
-      # pass 2: broadphase kernels early-return if skip[0] is 0
-      collision_driver.collision(m, d, skip)
+      # pass 2: passing awake_prev runs the incremental pass, emitting only pairs involving a
+      # newly-awakened body and appending them to the pass-1 buffer. collision() skips the
+      # broadphase entirely via a CUDA graph conditional on steps where nothing woke.
```
thowell: should we update this comment to note that sap broadphase is not skipped entirely?

thowell (approved):
nice!

thowell (approved):

PR #1458, opened 2026-06-24
Optimization: sparse JTDAJ

adenzler-nvidia:
## Summary

The sparse Newton Hessian assembly `H += JᵀDJ` (D diagonal) previously launched one thread per efc row, each looping its row's nonzeros and doing a `wp.atomic_add` per upper-triangular entry. Constraints whose efc rows share a dof support — a contact's `condim` rows, the rows of a connect or weld — redundantly re-read the same column indices and contend on the same `H` entries.

make_constraint already lays out each constraint's efc rows contiguously with a shared `colind`, so those rows form one dense |S|×|S| block (S = the dof support). This change has make_constraint append each constraint as one entry to a per-world dense list — its first efc row and contiguous row count (`efc.jtdaj_group_{head,size,count}`) — and assembles each block with one warp: the 32 threads stride the block's upper-triangular entries, read the shared column indices once, sum `JᵀDJ` over the member rows, and issue a single `wp.atomic_add` per `H` entry.

The assembly kernel grid-strides the dense list, so threads only ever touch real group heads — no thread is spent on the non-head rows that make up ~3/4 of `njmax`. The launch width targets one warp per group, capped so the grid stays within a few device waves (sized from the kernel's queried occupancy), which keeps each warp's serial chain short and load-balances the variable group sizes.

The flat upper-triangular entry index is decoded to `(block_row, block_col)` with the triangular-number inverse. Column boundaries land on perfect squares (`8*entry+1 = (2*block_col+1)²`), where float32 `sqrt` is exact, so the decode needs no integer fix-up.

## Performance

RTX PRO 6000 Blackwell, nworld=8192, sparse Jacobian, Newton solver (nsys, CUDA 12.9).

JTDAJ assembly kernel, old per-row vs new, controlled same-state (interleaved capture, min instance):

| scene | old | new | speedup |
|---|---|---|---|
| three_humanoids | 1077 µs | 196 µs | 5.5× |
| unitree_g1 (flat) | 555 µs | 133 µs | 4.2× |
| aloha (clutter) | 2076 µs | 623 µs | 3.3× |

End-to-end, per solver step (total GPU time, and JTDAJ's share of it):

| scene | total GPU / step | JTDAJ / step | JTDAJ % of step |
|---|---|---|---|
| three_humanoids | 4.09 → 3.23 ms (1.27×) | 1129 → 255 µs | 27.6% → 7.9% |
| unitree_g1 (flat) | 2.97 → 2.42 ms (1.23×) | 726 → 172 µs | 24.4% → 7.1% |
| aloha (clutter) | 163 → 152 ms (1.08×) | 17.0 → 5.6 ms | 10.4% → 3.7% |

JTDAJ's share of GPU runtime drops ~3.5×: 27.6%→7.9% of the three_humanoids step, 24.4%→7.1% for g1, and 10.4%→3.7% for aloha. (Step times use a per-scene njmax/nconmax sized for the steady-state contact count; the unchanged non-JTDAJ work is held constant across old/new to isolate the assembly change.)

The only addition to make_constraint is the dense-list build: a per-world `jtdaj_group_count` reset before allocation, and one atomic append per constraint — its first efc row and block size into `jtdaj_group_{head,size}` — alongside the `nefc`/`efc_id` atomics each allocator already does.


mujoco_warp/_src/constraint.py
```diff
@@ -3534,6 +3620,8 @@ def make_constraint(m: types.Model, d: types.Data):
     inputs=[d.ne, d.nf, d.nl, d.nefc, efc_nnz],
   )
 
+  d.efc.jtdaj_group_count.zero_()
```
thowell: would it make sense to incorporate this initialization in `_zero_constraint_counts` above?

mujoco_warp/_src/types.py
```diff
@@ -1913,6 +1913,9 @@ class Constraint:
   Attributes:
     type: constraint type (ConstraintType)            (nworld, njmax)
     id: id of object of specific type                 (nworld, njmax)
+    jtdaj_group_head: first efc row of each JTDAJ block   (nworld, njmax)
```
thowell: what if we drop `_group` from these names be more concise?

mujoco_warp/_src/types.py
```diff
@@ -1913,6 +1913,9 @@ class Constraint:
...
+    jtdaj_group_size: efc rows per JTDAJ block            (nworld, njmax)
```
thowell: what about `_nrow`?

mujoco_warp/_src/types.py
```diff
@@ -1913,6 +1913,9 @@ class Constraint:
...
+    jtdaj_group_count: number of JTDAJ blocks             (nworld,)
```
thowell: what about `_nblock`?

mujoco_warp/_src/constraint.py
```diff
@@ -195,6 +198,10 @@ def _equality_connect(
   if efcid >= njmax_in - 3:
     return
 
+  jgid = wp.atomic_add(efc_jtdaj_group_count_out, worldid, 1)
```
thowell: since these fields are only used by the newton solver can we guard these operations?

mujoco_warp/_src/types.py
thowell: would `_adr` be more similar to existing field names?

thowell (approved):

adenzler-nvidia:
@thowell thanks for the review — addressed in the latest commits:

- **Renamed** the block-list fields to `jtdaj_adr` / `jtdaj_nrow` / `jtdaj_nblock`.
- **Guarded the block-list assembly**: the constraint kernels that append to it are now `cache_kernel` factories gated on `wp.static(is_sparse and newton)`, so dense / CG / non-Newton builds skip it entirely. The block-count reset now folds into `_zero_constraint_counts` alongside the other per-world counts.
- **Scoped the allocation** to sparse Newton (its only consumer); the arrays are plain workspace otherwise.
- **`put_data` now builds the block list** from the loaded `efc`. `make_constraint` builds it in-kernel, but `put_data` doesn't run that — so solving a `put_data` state previously assembled an empty `JᵀDJ` (it still converged via the gradient, which is why the solver tests passed). MuJoCo lays each constraint's rows out contiguously, so grouping maximal runs sharing `(efc_type, efc_id)` reproduces the in-kernel grouping; added an io test covering it.

All solver / constraint / io tests pass and kernel-analyzer is clean. PTAL.

thowell (approved):

PR #1460, opened 2026-06-25
Remove deprecated parallel island solver code path

erikfrey:
Remove the deprecated sleep-aware parallel island solver code path.

This change completely removes the parallel island solver (`_solve_island`) and its related constraint-gathering/scattering logic, migrating the codebase to use the active-DOF compaction solver (`is_compact` Newton path) as the sole sleep-aware solver.

Specifically, this refactor:
- Removes the `io.ENABLE_ISLANDS` constant and its corresponding `enable_islands` CLI flag.
- Removes `IslandSolverContext` and all island-local solver kernels/helper functions in `solver.py`, `island.py`, and `support.py` (e.g. `gather_island_inputs`, `scatter_island_results`, `mul_m_island`, `solve_m_island`).
- Simplifies `island.compute_island_mapping` to run without a solver context.
- Updates the active-DOF compaction allocations in `io.py` to be determined solely by Newton solver and sleep flags (or when `nvmax` is explicitly requested).
- Cleans up and simplifies tests in `solver_test.py`, `island_test.py`, `sleep_test.py`, and `support_test.py` by removing obsolete island solver cases and adapting the test configurations to default to active-DOF compaction.
- Adjusts `test_data.fixture` to allocate the compact workspace by default for tests.

All tests are passing.


mujoco_warp/_src/cli.py
```diff
@@ -43,11 +42,7 @@
 EVENT_TRACE = flags.DEFINE_bool("event_trace", False, "print an event trace report")
 NOISE_STD = flags.DEFINE_float("noise_std", 0.01, "add noise to ctrl signal (standard deviation)")
 NOISE_RATE = flags.DEFINE_float("noise_rate", 0.1, "add noise to ctrl signal (noise rate)")
-ENABLE_ISLANDS = flags.DEFINE_bool(
-  "enable_islands",
-  False,
-  "Enable constraint islands solver",
-)
+
```
thowell: nit: we can remove this empty line

mujoco_warp/_src/forward.py
```diff
@@ -647,7 +646,7 @@ def fwd_position(m: Model, d: Data, factorize: bool = True):
       sleep.wake_equality(m, d)
     sleep.update_sleep(m, d)
 
-  if m.is_compact or (m.ntree > 1 and not (m.opt.disableflags & types.DisableBit.ISLAND)):
+  if m.opt.enableflags & EnableBit.SLEEP:
```
thowell: maybe we add a todo here to eventually add the `disableflags` logic?

mujoco_warp/_src/io.py
```diff
@@ -1297,16 +1262,16 @@ def _allocate_island_arrays(
   d: types.Data,
   nworld: int,
   njmax: int,
-  island_enabled: bool,
+  sleep_enabled: bool,
```
thowell: not sure we want to make the island array allocation always dependent on sleeping since we might want to be able to enable islands independent from sleeping?

thowell: maybe we just call this variable `enabled`? and then we can pass a `sleep_enabled` variable to this function?

thowell: looks like below `compact_alloc` is passed to this function 👍

mujoco_warp/_src/io_test.py
```diff
@@ -458,7 +458,10 @@ def _assert_eq(a, b, name):
 class IOTest(parameterized.TestCase):
   def test_make_put_data(self):
     """Tests that make_data and put_data are producing the same shapes for all arrays."""
-    mjm, _, _, d = test_data.fixture("pendula.xml")
+    path = test_data.epath.resource_path("mujoco_warp") / "test_data" / "pendula.xml"
```
thowell: revert?

mujoco_warp/_src/types.py
```diff
@@ -1931,13 +1929,7 @@ class Constraint:
...
     state: constraint state                           (nworld, njmax_pad)
     island: island ID per constraint                  (nworld, njmax)
-    itype: island constraint type                     (nworld, njmax)
-    iid: island constraint id                         (nworld, njmax)
-    iD: island constraint mass                        (nworld, njmax_pad)
-    iaref: island aref                                (nworld, njmax)
-    ifrictionloss: island frictionloss                (nworld, njmax)
-    iforce: island force                              (nworld, njmax)
-    istate: island state                              (nworld, njmax_pad)
+
```
thowell: nit: we could remove this empty line

mujoco_warp/_src/types.py
```diff
@@ -2086,10 +2070,7 @@ class Data:
     map_iefc2efc: island-local EFC -> global EFC                (nworld, njmax)
     dof_islandid: island ID per island-DOF                      (nworld, nv)
     efc_islandid: island ID per island-EFC                      (nworld, njmax)
-    iqacc: island-local qacc                                    (nworld, nv)
-    iqacc_smooth: island-local qacc_smooth                      (nworld, nv)
-    iqfrc_smooth: island-local qfrc_smooth                      (nworld, nv)
-    iqfrc_constraint: island-local qfrc_constraint              (nworld, nv)
+
```
thowell: nit: we could remove this empty line

mujoco_warp/_src/types.py
```diff
@@ -2232,10 +2213,7 @@ class Data:
   map_iefc2efc: array("nworld", "njmax", int)
   dof_islandid: array("nworld", "nv", int)
   efc_islandid: array("nworld", "njmax", int)
-  iqacc: wp.array2d[float]
-  iqacc_smooth: wp.array2d[float]
-  iqfrc_smooth: wp.array2d[float]
-  iqfrc_constraint: wp.array2d[float]
+
```
thowell: nit: we could remove this empty line

thowell (approved):
thanks @erikfrey!

mujoco_warp/_src/cli.py
erikfrey: Done.

mujoco_warp/_src/forward.py
erikfrey: We decided to go ahead and fully replicate the behavior instead of leaving a TODO. Now, if islands are disabled in `disableflags`, sleeping is completely bypassed/deactivated, matching native MuJoCo.

mujoco_warp/_src/io.py
erikfrey: Done. Renamed to `enabled` and also fixed the ordering bug where `compact_alloc` was computed after `nvmax` was resolved, which was causing compaction arrays to be unconditionally allocated.

mujoco_warp/_src/io_test.py
erikfrey: Reverted back to using the fixture. I changed the fixture's default `nvmax` to `-1` (acting as a sentinel for resolving to `mjm.nv`), which lets `test_make_put_data` explicitly pass `nvmax=None` to verify default non-compaction layout matching.

mujoco_warp/_src/types.py
erikfrey: Done.

mujoco_warp/_src/types.py
erikfrey: Done.

mujoco_warp/_src/types.py
erikfrey: Done.

erikfrey:
All reviewer comments have been addressed:
- Fixed the empty line nits in cli.py and types.py.
- Replaced the TODO in forward.py with full support for DisableBit.ISLAND, bypassing the sleep and island logic if islands are deactivated (fully matching native MuJoCo behavior).
- Renamed sleep_enabled to enabled in _allocate_island_arrays, and fixed the ordering bug in io.py where compact_alloc was evaluated too late (which was causing compaction arrays to be unconditionally allocated).
- Reverted the verbose XML loading in test_make_put_data back to using test_data.fixture, updating the fixture default to nvmax=-1 to act as a sentinel for resolving to mjm.nv by default while supporting nvmax=None for default non-compaction shape testing.

erikfrey:
Additionally decoupled island array allocation and discovery from the sleep solver:
- Allocated the minor island and mapping arrays by default (as long as they are not explicitly disabled in options.disableflags). This ensures that custom debug tools or test fixtures can explicitly call island discovery and mapping without causing CUDA crashes.
- Bypassed the allocation of the heavy compaction solver matrices (like cM, cqLD) when sleeping is disabled to maintain memory efficiency.
- Refined forward.py to run the island discovery pass if there are multiple trees (m.ntree > 1) and islands are enabled, ensuring island topology is tracked independent of sleep status.

erikfrey:
Completed a full benchmark run comparison between our branch and the gh-pages nightly baselines:
- Resolved a performance regression in the cloth benchmarks (aloha_cloth went from -61% to +0.2%, and cloth went from -88% to +4.4% compared to the baseline).
- **The regression reason**: A cloth model consists of a grid of independent masses (e.g., 901 separate trees). Running island discovery automatically on every step was allocating a large `(nworld, 901, 901)` temporary matrix (~6.6 GB) inside the step loop on the GPU, causing severe synchronization overhead.
- **The fix**: Restrained the automatic execution of `island.island` inside `forward.py` to only run when sleep is enabled. Since the island and mapping arrays remain allocated on startup (by default), debug tools, unit tests, or the visualizer can still explicitly call island discovery and mapping on-demand without any performance impact on the hot simulation loop.
- **Other Benchmarks**: All other benchmarks run successfully with 0% to +7% performance gains (humanoid: +5.0%, three_humanoids: +7.6%) compared to their baselines.

PR #1461, opened 2026-06-25
Implement programmatic overflow flagging system using per-world bits and warn_overflow option

### LINKED ISSUE #692: `wp.printf` alternative for reporting overflow
Investigate alternatives to `wp.printf` for reporting overflow.

erikfrey:
This pull request implements a programmatic overflow flagging system in MuJoCo Warp, allowing users to query and handle simulation overflows. It introduces an OverflowType bitmask enum representing standard and collision-specific overflow bits (NEFC, NJMAX_NNZ, BROADPHASE, NARROWPHASE, CCD, HFIELD, CONTACT_MATCH, NVMAX), alongside a per-world overflow array on the Data struct.

A new warn_overflow configuration option (defaulting to True) is added to Option to control the legacy stdout print warnings, which are now checked dynamically inside the physics and collision kernels. All updated kernels have their parameters ordered, prefixed (opt_warn_overflow), and documented to satisfy strict kernel analyzer constraints.

Benchmarks on the Franka Emika Panda environment verify that the dynamic checks do not introduce any noticeable performance regression (~0.7% change, well within GPU execution noise).

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -41,6 +41,7 @@ from mujoco_warp._src.types import Data
 from mujoco_warp._src.types import DisableBit
 from mujoco_warp._src.types import GeomType
 from mujoco_warp._src.types import Model
+from mujoco_warp._src.types import OverflowType
 from mujoco_warp._src.types import mat43
 from mujoco_warp._src.types import mat63
 from mujoco_warp._src.types import vec5
@@ -171,6 +172,7 @@ def ccd_hfield_kernel_builder(
   def ccd_hfield_kernel(
     # Model:
     opt_ccd_tolerance: wp.array[float],
+    opt_warn_overflow: bool,
     geom_type: wp.array[int],
     geom_condim: wp.array[int],
     geom_dataid: wp.array2d[int],
@@ -242,6 +244,8 @@ def ccd_hfield_kernel_builder(
     contact_type_out: wp.array[int],
     contact_geomcollisionid_out: wp.array[int],
     nacon_out: wp.array[int],
+    # Data out:
+    overflow_out: wp.array[int],
   ):
     collisionid = wp.tid()
     if collisionid >= ncollision_in[0]:
@@ -280,7 +284,9 @@ def ccd_hfield_kernel_builder(
 
     ccdid = wp.atomic_add(nccd_in, wp.static(geomgeomid), 1)
     if ccdid >= naccdmax_in:
-      wp.printf("CCD overflow - please increase naccdmax to %u\n", ccdid)
+      if opt_warn_overflow:
+        wp.printf("CCD overflow - please increase naccdmax to %u\n", ccdid)
+      wp.atomic_or(overflow_out, worldid, wp.static(OverflowType.CCD))
       return
 
     _, margin, gap, condim, friction, solref, solreffriction, solimp = contact_params(
@@ -414,11 +420,13 @@ def ccd_hfield_kernel_builder(
       for c in range(cmin + 1, cmax + 1):
         # add both triangles from this cell
         for i in range(2):
-          if count >= MJ_MAXCONPAIR:
-            wp.printf(
-              "height field collision overflow, number of collisions >= %u - please adjust resolution: \n decrease the number of hfield rows/cols or modify size of colliding geom\n",
-              MJ_MAXCONPAIR,
-            )
+          if ncollision_in[0] + count > MJ_MAXCONPAIR:
+            if worldid == 0 and opt_warn_overflow:
+              wp.printf(
+                "height field collision overflow, number of collisions >= %u - please adjust resolution: \n decrease the number of hfield rows/cols or modify size of colliding geom\n",
+                MJ_MAXCONPAIR,
+              )
+            wp.atomic_or(overflow_out, worldid, wp.static(OverflowType.HFIELD))
             continue
 
           # add vert
@@ -714,6 +722,7 @@ def ccd_kernel_builder(
   def eval_ccd_write_contact(
     # Model:
     opt_ccd_tolerance: wp.array[float],
+    opt_warn_overflow: bool,
     # Data in:
     naconmax_in: int,
     naccdmax_in: int,
@@ -766,6 +775,8 @@ def ccd_kernel_builder(
     contact_type_out: wp.array[int],
     contact_geomcollisionid_out: wp.array[int],
     nacon_out: wp.array[int],
+    # Data out:
+    overflow_out: wp.array[int],
   ) -> int:
     points = mat43()
     witness1 = mat43()
@@ -795,7 +806,9 @@ def ccd_kernel_builder(
     if needs_epa:
       ccdid = wp.atomic_add(nccd_in, geomgeomid, 1)
       if ccdid >= naccdmax_in:
-        wp.printf("CCD overflow - please increase naccdmax to %u\n", ccdid)
+        if opt_warn_overflow:
+          wp.printf("CCD overflow - please increase naccdmax to %u\n", ccdid)
+        wp.atomic_or(overflow_out, worldid, wp.static(OverflowType.CCD))
         return 0
       dist, ncollision, w1, w2, multiccd_idx = epa_phase(
         opt_ccd_tolerance[worldid % opt_ccd_tolerance.shape[0]],
@@ -914,6 +927,7 @@ def ccd_kernel_builder(
   def ccd_kernel(
     # Model:
     opt_ccd_tolerance: wp.array[float],
+    opt_warn_overflow: bool,
     geom_type: wp.array[int],
     geom_condim: wp.array[int],
     geom_dataid: wp.array2d[int],
@@ -990,6 +1004,8 @@ def ccd_kernel_builder(
     contact_type_out: wp.array[int],
     contact_geomcollisionid_out: wp.array[int],
     nacon_out: wp.array[int],
+    # Data out:
+    overflow_out: wp.array[int],
   ):
     collisionid = wp.tid()
     if collisionid >= ncollision_in[0]:
@@ -1052,6 +1068,7 @@ def ccd_kernel_builder(
 
     eval_ccd_write_contact(
       opt_ccd_tolerance,
+      opt_warn_overflow,
       naconmax_in,
       naccdmax_in,
       epa_vert_in,
@@ -1101,6 +1118,7 @@ def ccd_kernel_builder(
       contact_type_out,
       contact_geomcollisionid_out,
       nacon_out,
+      overflow_out,
     )
 
   return ccd_kernel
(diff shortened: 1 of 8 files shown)


mujoco_warp/_src/collision_convex.py
```diff
@@ -280,7 +284,9 @@ def ccd_hfield_kernel(
 
     ccdid = wp.atomic_add(nccd_in, wp.static(geomgeomid), 1)
     if ccdid >= naccdmax_in:
-      wp.printf("CCD overflow - please increase naccdmax to %u\n", ccdid)
+      if opt_warn_overflow:
+        wp.printf("CCD overflow - please increase naccdmax to %u\n", ccdid)
+      wp.atomic_or(overflow_out, worldid, wp.static(OverflowType.CCD))
```
adenzler-nvidia, line 289: ideally we would make the if a wp.static, because printf generally uses a lot of registers just by being compiled into the kernel.

adenzler-nvidia:
also putting this one here for reference: https://github.com/google-deepmind/mujoco_warp/pull/1089/changes

mujoco_warp/_src/collision_convex.py
```diff
@@ -414,11 +420,13 @@ def ccd_hfield_kernel(
       for c in range(cmin + 1, cmax + 1):
         # add both triangles from this cell
         for i in range(2):
-          if count >= MJ_MAXCONPAIR:
-            wp.printf(
-              "height field collision overflow, number of collisions >= %u - please adjust resolution: \n decrease the number of hfield rows/cols or modify size of colliding geom\n",
-              MJ_MAXCONPAIR,
-            )
+          if ncollision_in[0] + count > MJ_MAXCONPAIR:
```
thowell: is this correct?

for height field collisions we inline allocate memory for `MJ_MAXCONPAIR` (50) potential collisions so the height field "broadphase"/"midphase" collision overflow should be a check on `count > MJ_MAXCONPAIR`?

the potential collisions are then filtered and a maximum of 4 contacts get written. we could have "narrowphase" overflow checks on `nacon_out[0] > naconmax_in` here after each `write_contact` call below or at the end of the entire height field routine?

mujoco_warp/_src/collision_convex.py
```diff
@@ -414,11 +420,13 @@ def ccd_hfield_kernel(
...
+            if worldid == 0 and opt_warn_overflow:
```
thowell: are there cases where guarding on `worldid == 0` does not work?

this kernel is launched over contacts `naconmax` so isn't there the possibility that overflow occurs when `worldid != 0` and then there is not a print warning?

mujoco_warp/_src/forward.py
```diff
@@ -236,28 +237,35 @@ def _next_time(
...
-    if ncollision > naconmax_in:
-      nconmax = int(wp.ceil(float(ncollision) / float(nworld_in)))
-      wp.printf("broadphase overflow - please increase nconmax to %u or naconmax to %u\n", nconmax, ncollision)
-
-    if nacon_in[0] > naconmax_in:
+      if opt_warn_overflow:
+        wp.printf("njmax_nnz overflow - please increase njmax_nnz to %u\n", efc_nnz)
+      overflow_out[worldid] = overflow_out[worldid] | wp.static(types.OverflowType.NJMAX_NNZ)
+
+  if ncollision_in[0] > naconmax_in:
```
thowell: nit: we could read `ncollision_in[0]` just once like before

mujoco_warp/_src/forward.py
```diff
@@ -236,28 +237,35 @@ def _next_time(
...
+    if worldid == 0 and opt_warn_overflow:
+      nconmax = int(wp.ceil(float(ncollision_in[0]) / float(nworld_in)))
+      wp.printf("broadphase overflow - please increase nconmax to %u or naconmax to %u\n", nconmax, ncollision_in[0])
+    overflow_out[worldid] = overflow_out[worldid] | wp.static(types.OverflowType.BROADPHASE)
+
+  if nacon_in[0] > naconmax_in:
```
thowell: nit: same thing with reading `nacon_in[0]`

mujoco_warp/_src/island.py
```diff
@@ -1126,12 +1128,14 @@ def _compact_dofs(
...
-      nvmax_in,
-    )
+    if warn_overflow:
+      wp.printf(
+        "nvmax overflow: world %d needs %d active DOFs but nvmax = %d (behavior undefined)\n",
+        worldid,
+        count,
+        nvmax_in,
+      )
+    overflow_out[worldid] = overflow_out[worldid] | wp.static(types.OverflowType.NVMAX)
```
thowell: is it beneficial and/or necessary to use `wp.static` here?

mujoco_warp/_src/sensor.py
```diff
@@ -2430,8 +2434,10 @@ def _contact_match(
   contactmatchid = wp.atomic_add(sensor_contact_nmatch_out[worldid], contactsensorid, 1)
 
   if contactmatchid >= opt_contact_sensor_maxmatch:
-    # TODO(team): alternative to wp.printf for reporting overflow?
-    wp.printf("contact match overflow: please increase Option.contact_sensor_maxmatch to %u\n", contactmatchid)
+    if opt_warn_overflow:
+      # TODO(team): alternative to wp.printf for reporting overflow?
```
thowell: are we considering this todo resolved by this pr?

thowell:
this is great, thanks @erikfrey!

thowell (approved):

mujoco_warp/_src/sensor.py
erikfrey: Yes, since we now set OverflowType.CONTACT_MATCH in d.overflow programmatically, this TODO is resolved. I've removed the comment.

mujoco_warp/_src/collision_convex.py
erikfrey: Done! Both `ccd_hfield_kernel_builder` and `ccd_kernel_builder` have been refactored to take `warn_overflow` as a JIT-time template parameter, enabling `wp.static(warn_overflow)` branch elimination inside the JIT-compiled kernels.

mujoco_warp/_src/collision_convex.py
erikfrey: Yes, you are completely right. I've restored the local check to `count >= MJ_MAXCONPAIR` since `count` is the index into the thread's local arrays. Global narrowphase overflow is handled at the end of the step in `_next_time`.

mujoco_warp/_src/collision_convex.py
erikfrey: Since we restored the check to the local thread `count`, I have removed the `worldid == 0` guard so that we print the warning if heightfield overflow occurs in any world.

mujoco_warp/_src/forward.py
erikfrey: Done. Optimized by caching it to a local variable.

mujoco_warp/_src/forward.py
erikfrey: Done. Optimized by caching it to a local variable.

mujoco_warp/_src/island.py
erikfrey: Warp enum support can be subtle, and using `wp.static()` guarantees the JIT compiler resolves the Python enum to a literal integer constant during compilation, avoiding potential dynamic lookup overhead or transpiler issues. I've left it as-is for consistency.

erikfrey:
Applied resolutions to address all PR review feedback:

1. **JIT Static Specialization for Printfs**:
   - Refactored the narrowphase collision builders (`ccd_hfield_kernel_builder`, `ccd_kernel_builder`) to accept `warn_overflow` as a template parameter. The nested `@wp.kernel` and `@wp.func` constructs now evaluate `wp.static(warn_overflow)` to compile out register-heavy printfs when disabled.
   - Refactored `_next_time` in `forward.py` to a nested kernel builder context `_next_time_builder(warn_overflow: bool)` to similarly static-compile the printfs.

2. **Heightfield Collision Checks**:
   - Restored the local heightfield overflow check to `count >= MJ_MAXCONPAIR` (the stack array's local limit) instead of using the global `ncollision_in[0]`.
   - Removed the `worldid == 0` check on heightfield overflow prints, as this is a local per-pair event that can happen in any world.

3. **Optimizations and Cleanups**:
   - Cached global array accesses for `ncollision_in[0]` and `nacon_in[0]` in `forward.py` into local variables.
   - Removed the resolved `TODO` about `wp.printf` alternatives in `sensor.py` (since we now have programmatic `OverflowType` tracking).

4. **Testspeed Benchmark Support**:
   - Added a new `--overflow_behavior` CLI option to `testspeed.py` (`error` vs `continue`, defaulting to `error`). 
   - Under `error` mode, it JIT-compiles the kernels with `warn_overflow=False` and uses the step callback to check `d.overflow` programmatically, printing active overflow flags and exiting immediately on failure.

mujoco_warp/_src/island.py
erikfrey: Correction: After double-checking the codebase style and transpiler behavior, Warp does natively resolve class-scoped enum constants inside kernels. The CI transpilation error was actually caused by trying to access it via the module namespace (e.g. `types.OverflowType.NVMAX`). I have imported `OverflowType` directly, removed all `wp.static()` wrappers from `OverflowType` references, and verified they transpile and run correctly.

mujoco_warp/_src/types.py
```diff
@@ -134,6 +134,30 @@ class BroadphaseFilter(enum.IntFlag):
...
+    NJMAX_NNZ: njmax_nnz overflow
+    BROADPHASE: broadphase overflow / flex broadphase overflow
+    NARROWPHASE: narrowphase overflow / flex narrowphase / contact overflow
+    CCD: CCD overflow / flex CCD overflow
+    HFIELD: height field collision overflow
+    CONTACT_MATCH: contact match sensor overflow
+    NVMAX: nvmax overflow (islands)
+  """
+
+  NEFC = 1
```
btaba: 1 << 0
1 << 1

or

enum.auto()

mujoco_warp/testspeed.py
```diff
@@ -258,6 +263,20 @@ def callback(step, step_trace, latency):
     solver_niter.append(d.solver_niter.numpy())
     trace = _sum_trace(trace, step_trace)
 
+    if _OVERFLOW_BEHAVIOR.value == "error":
+      overflows = d.overflow.numpy()
+      if np.any(overflows != 0):
+        world_ids = np.where(overflows != 0)[0]
+        n_worlds = len(world_ids)
```
btaba, line 270: maybe just groupby count?

```
counts = Counter(
    tuple(f.name for f in OverflowType if overflows[wid] & f.value) 
    for wid in world_ids
)
```

btaba (approved):

mujoco_warp/testspeed.py
erikfrey: It's a good point but I'm not entirely sure yet how it'll get used.  I'm slightly more leaning towards the idea folks will want to debug a certain world, but I could be wrong.  Let's let it sit.

PR #1464, opened 2026-06-26
Bound convex CCD kernel with grid-stride and launch_bounds

adenzler-nvidia:
The convex CCD kernel launched one thread per `naconmax` contact slot and was register-limited to a single block per SM, so most threads returned immediately on empty slots and occupancy was poor. This switches it to a grid-stride loop bounded by the live contact count and caps registers via `launch_bounds` (two blocks per SM), lifting occupancy on contact-heavy scenes without over-spilling on light ones.

Measured with nsys (CUDA 12.9) on the official aloha replay benchmarks:

| benchmark | total GPU | ccd kernel |
|---|---|---|
| aloha_clutter (SLEEP, official) | 1.18× | 1.57× |
| aloha_clutter (no sleep) | 1.37× | 1.88× |
| aloha_pot | no change | no change |

The no-CCD benchmarks (humanoid, three_humanoids, g1 flat, flex) are unaffected since the kernel never launches there. `collision_driver_test` passes.


thowell (approved):
awesome! thanks @adenzler-nvidia

PR #1465, opened 2026-06-26
_is_batched

thowell:
`io._create_array` sets the array attribute `_is_batched`. however not all arrays are constructed with this method. as a result some batched fields were not correctly marked. this pr marks all arrays that are part of `Model`, `Data`, and `RenderContext`. additionally these dataclasses' fields are checked in io_test.py `test_is_batched_metadata`.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -39,6 +39,26 @@ def _is_array_spec(typ) -> bool:
   return isinstance(typ, wp.array) or type(typ).__name__ == "_ArrayAnnotation"
 
 
+def _mark_batched(obj):
+  """Recursively set _is_batched = True on all batched warp arrays within obj."""
+  if not dataclasses.is_dataclass(obj):
+    return
+  for f in dataclasses.fields(obj):
+    val = getattr(obj, f.name, None)
+    if val is None:
+      continue
+    if dataclasses.is_dataclass(val):
+      _mark_batched(val)
+      continue
+    if not isinstance(val, wp.array):
+      continue
+    if not _is_array_spec(f.type):
+      continue
+    spec_shape = getattr(f.type, "shape", ())
+    if spec_shape and spec_shape[0] in ("*", "nworld"):
+      val._is_batched = True
+
+
 def _create_array(data: Any, spec, sizes: dict[str, int], batch_size: int = 1) -> wp.array | None:
   """Creates a warp array and populates it with data.
 
@@ -67,12 +87,6 @@ def _create_array(data: Any, spec, sizes: dict[str, int], batch_size: int = 1) -
         data = np.broadcast_to(data, target_shape).copy()
     array = wp.array(data, dtype=spec.dtype, shape=shape)
 
-  if is_batched:
-    # add private attribute for JAX to determine which fields are batched
-    array._is_batched = True
-    if array.shape[0] == 1:
-      # also set stride 0 to 0 which is expected legacy behavior (but is deprecated)
-      array.strides = (0,) + array.strides[1:]
   return array
 
 
@@ -1065,6 +1079,7 @@ def put_model(mjm: mujoco.MjModel, batch_sizes: dict[str, int] | None = None) ->
       batch_size = batch_sizes.get(f.name, 1)
       setattr(m, f.name, _create_array(getattr(m, f.name), f.type, sizes, batch_size))
 
+  _mark_batched(m)
   return m
 
 
@@ -1594,6 +1609,7 @@ def make_data(
   d.dof_cdof.fill_(-1)
   d.cdof_dof.fill_(-1)
 
+  _mark_batched(d)
   return d
 
 
@@ -1880,6 +1896,7 @@ def put_data(
 
   d.nacon = wp.array([mjd.ncon * nworld], dtype=int)
 
+  _mark_batched(d)
   return d
 
 
@@ -3840,4 +3857,5 @@ def create_render_context(
 
   bvh.build_scene_bvh(mjm, mjd, rc, nworld)
 
+  _mark_batched(rc)
   return rc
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/io.py
```diff
@@ -42,6 +42,27 @@ def _is_array_spec(typ) -> bool:
   return isinstance(typ, wp.array) or type(typ).__name__ == "_ArrayAnnotation"
 
 
+def _mark_batched(obj):
```
btaba, line 42: can we delete the _is_batch annotation on create_array then? better to keep a single codepath?

mujoco_warp/_src/io.py
```diff
@@ -42,6 +42,27 @@ def _is_array_spec(typ) -> bool:
...
+  """Recursively set _is_batched = True on all batched warp arrays within obj."""
+  if dataclasses.is_dataclass(obj):
```
btaba: how about:

```
if not dataclasses.is_dataclass(obj):
  return
# do stuff
```

you can do similar things below to reduce the number of nested if statements

mujoco_warp/_src/io.py
```diff
@@ -42,6 +42,27 @@ def _is_array_spec(typ) -> bool:
...
+        continue
+      if dataclasses.is_dataclass(val):
+        _mark_batched(val)
+      elif _is_array_spec(f.type):
+        spec_shape = getattr(f.type, "shape", ())
+        if spec_shape and spec_shape[0] in ("*", "nworld"):
+          if isinstance(val, wp.array):
+            val._is_batched = True
+            if val.shape[0] == 1:
+              try:
```
btaba: why do you need the bare except? bare excepts can lead to silent failures

do you know which fields this is now adding stride 0 to that it wasn't before, and vice versa?

mujoco_warp/_src/io.py
```diff
@@ -42,6 +42,27 @@ def _is_array_spec(typ) -> bool:
...
+                val.strides = (0,) + val.strides[1:]
```
btaba: keep the comment about the "legacy" bit compared to `_create_array`?

also why is it considered legacy? I thought (probably legacy thinking :D ) that stride 0 was a requirement for nworld>1 for Model fields?

mujoco_warp/_src/io_test.py
```diff
@@ -469,41 +470,27 @@ def test_make_put_data(self):
...
+  def test_is_batched_metadata(self, dataclass_type):
+    """Verify that all fields specified with 'nworld' or '*' are marked with _is_batched=True."""
+    fixture_outputs = test_data.fixture("pendula.xml")
+    if dataclass_type is types.RenderContext:
+      obj = io.create_render_context(fixture_outputs[0], nworld=1)
+    else:
+      obj = next(o for o in fixture_outputs if isinstance(o, dataclass_type))
+
+    for f in dataclasses.fields(dataclass_type):
+      if io._is_array_spec(f.type):
```
btaba: similar comment about reducing the number of nested ifs by using early `continue`

btaba (changes requested):

mujoco_warp/_src/io.py
thowell: done

mujoco_warp/_src/io.py
thowell: removed the legacy stride 0. before not sure which fields had this feature, probably all batched `Model` fields. now removed and no fields have this stride 0 feature.

mujoco_warp/_src/io.py
thowell: removed the legacy string 0 feature. think this is from before we switched to modulo indexing for batched `Model` fields https://github.com/google-deepmind/mujoco_warp/pull/774. with this type of indexing we shouldn't need stride 0.

mujoco_warp/_src/io_test.py
thowell: done

mujoco_warp/_src/io.py

mujoco_warp/_src/io.py
btaba: cool cool, in that case, I should remove all stride 0 logic downstream then

mujoco_warp/_src/io_jax_test.py
```diff
@@ -203,11 +203,11 @@ def test_put_model_nworld_array(self):
 
     self.assertTrue(hasattr(m1.geom_pos, "_is_batched"))
     self.assertEqual(m1.geom_pos.shape[0], 1)
-    self.assertEqual(m1.geom_pos.strides[0], 0)
+    self.assertGreater(m1.geom_pos.strides[0], 0)
     self.assertLen(m1.geom_pos.strides, m1.geom_pos.ndim)
     self.assertTrue(hasattr(m1.opt.gravity, "_is_batched"))
     self.assertEqual(m1.opt.gravity.shape[0], 1)
-    self.assertEqual(m1.opt.gravity.strides[0], 0)
+    self.assertGreater(m1.opt.gravity.strides[0], 0)
```
btaba: should this be deleted now?

https://github.com/google-deepmind/mujoco/blob/main/mjx/mujoco/mjx/warp/ffi.py#L171-L183

mujoco_warp/_src/io_test.py
```diff
@@ -1480,10 +1507,6 @@ def test_put_model_batch_sizes(self):
     np.testing.assert_allclose(m.geom_size.numpy(), np.repeat(m_default.geom_size.numpy(), 2, axis=0))
     np.testing.assert_allclose(m.mat_rgba.numpy(), np.repeat(m_default.mat_rgba.numpy(), 4, axis=0))
 
-    # Fields not listed in batch_sizes keep the shared legacy allocation.
-    self.assertEqual(m.geom_pos.shape[0], 1)
```
btaba, line 1484: wouldn't you still want to check that geom_pos has size 1 in shape[0]. FWIU you're just removing the stride 0 logic in this PR right?

thowell: good catch thanks!

refactored to check that all batched `Model` fields by default have batch size 1.

mujoco_warp/_src/io_jax_test.py
thowell: yeah think we can delete

thowell:
removed more tests that were checking stride

btaba (approved):

PR #1466, opened 2026-06-26
Bump MuJoCo Warp to v3.10.0.1

erikfrey:


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name="mujoco-warp"
-version = "3.10.0"
+version = "3.10.1"
```
thowell: should this be 3.10.0.1 to match our previous hotfixes?

uv.lock
```diff
@@ -1086,7 +1086,7 @@ wheels = [
 
 [[package]]
 name = "mujoco-warp"
-version = "3.10.0"
+version = "3.10.1"
```
thowell: same thing here

thowell (approved):

pyproject.toml
erikfrey: oh my goodness, not enough coffee this morning.  let me fix sec

PR #1468, opened 2026-06-26
Add dual-arm cloth lifting trajectory for aloha_cloth

erikfrey:
Update the Aloha cloth benchmark to use a diagonal dual-arm trajectory. The right arm now reaches across to grasp the opposite front-right corner while the left arm grabs the back-left corner, yanking and releasing the cloth to showcase self-collision and free-fall dynamics. Added checkerboard texture and material configurations to the cloth model.

Performance & Configuration Notes:
- Raised `nccdmax` from `1` to `256` in the benchmark configuration. The new diagonal yanking and dropping trajectory causes folding and self-collisions that generate significant CCD pairs. Setting it lower triggers CCD contact overflows during rollout.
- Baseline performance (with flat cloth and no control replay): 667.4 steps/second (solver iterations: 40.1 mean).
- New trajectory performance: 546.5 steps/second (solver iterations: 61.2 mean). The change is due to the increased complexity of solving the diagonal folding and self-collisions.
- Step time breakdown shows `flex_collision` remains the primary bottleneck (994.2ms out of 1829.2ms total step time, ~54.4%).


thowell:
this is great!

@erikfrey is it possible to extend this demonstration a bit to allow the cloth to fall on itself and generate self collisions?

@quagla what is your recommendation for reducing the instabilities in the cloth that occur at the boundary between the cloth and table when the corners are picked up, maybe increase the damping?

erikfrey:
When developing the trajectory (coding agents are getting pretty good at designing trajectories without teleop!), I found that the cloth was nice and stable on the table, and sliding a gripper end under the cloth worked quite well!

Where I see some issues is once the cloth is in the gripper - it tends to slip out, and if the gripper fully closes on the cloth it depenetrates in unintuitive ways.  But the motion in the video is legit - you can definitely do significant kinds of cloth manipulation with the current settings.

thowell (approved):

PR #1471, opened 2026-06-29
Require cost improvement in line search

adenzler-nvidia:
The iterative line search can stop at a bracket endpoint solely because its derivative is within `gtol`, even when that endpoint has not reduced the objective. With a large scaled tolerance, this can accept the zero-step endpoint and leave the solver at its warm start.

Require negative shifted cost for derivative-based convergence, matching the existing check for the initial Newton trial. Add a regression test covering a loose scaled tolerance where the current implementation returns a zero step instead of continuing to an improving point.

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -1202,7 +1202,11 @@ def _linesearch_iterative_kernel(ls_iterations: int, cone_type: types.ConeType, 
         swap_hi = swap_hi_hi_next or swap_hi_mid or swap_hi_lo_next
 
         # check for convergence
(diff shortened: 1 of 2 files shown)


erikfrey:
@adenzler-nvidia does this impact any of our benchmarks?

adenzler-nvidia:
I ran the full non-render benchmark suite plus focused repeats on the apparent outliers. I don’t see a meaningful performance impact: aggregate step time is effectively unchanged, and all worlds still converge.

This fixes an issue observed in internal learning workloads. The added cost-improvement requirement is also consistent with the existing check for the initial Newton step size.

yuvaltassa (approved):
Good idea!

PR #1473, opened 2026-06-29
Support per-world heightfield data

nico-bohlinger:
`hfield_data` is currently shared by all worlds, which prevents using different terrain in each world , e.g. for domain randomization of terrain.

This PR adds a batch/world-dimension to `hfield_data` and updates heightfield collision and ray casting to select the data for the current world.

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -203,7 +203,7 @@ def ccd_hfield_kernel_builder(
     hfield_nrow: wp.array[int],
     hfield_ncol: wp.array[int],
     hfield_adr: wp.array[int],
-    hfield_data: wp.array[float],
+    hfield_data: wp.array2d[float],
     pair_dim: wp.array[int],
     pair_solref: wp.array2d[wp.vec2],
     pair_solreffriction: wp.array2d[wp.vec2],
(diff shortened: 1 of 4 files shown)


thowell:
@nico-bohlinger thank you for your contribution!

it should be possible to enable per-world height fields with the existing implementation following a pattern similar to the one for [per-world meshes](https://mujoco.readthedocs.io/en/latest/mjwarp/index.html#per-world-meshes).

please let use know if you have any questions. thanks!

nico-bohlinger:
@thowell yes you are totally right that one can achieve this also by simply placing multiple hfield assets and then doing per-world routing.

My motivation is coming mainly from the fact that the API for doing so is not that user-friendly. I hoped by having a batched version of hfield_data, this allows the user to do per-world randomization of hfields (as typically done in robot locomotion learning) without needing any additional code on the user side to generate the correct amount of hfield assets and handling the routing.
But as hfields are assets and they also just use one precomputed BVH I understand if this needs to stay that way.

erikfrey:
@nico-bohlinger domain randomization intersects with environment authoring apis, each of which has a different (and quite opinionated!) recipe on how to do it.

Until the dust settles on that question, it's probably wiser for the physics engine to offer a dumb but complete way of supporting domain randomization, and leave it to the environment authoring libraries to propose user friendly interfaces.

nico-bohlinger:
@erikfrey that makes totally sense! Thanks!

PR #1475, opened 2026-07-01
Add polytope center to EPA

kbayes:
Reference from MuJoCo: [2c5b8ca](https://github.com/google-deepmind/mujoco/commit/2c5b8ca).

We store the geometric center of the polytope in EPA so the orientation of the normal in _attach_face can be corrected if pointing in wrong direction: the origin is contained in the polytope so the projection on the new face should be pointing outward.

There's no repro case for this in mjWarp, but no regressions were spotted, so it's a good numerical fallback to have.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -1074,7 +1074,7 @@ def _polytope3(
 ) -> Polytope:
   """Create polytope for EPA given a 2-simplex from GJK."""
   # set the polytope center
-  pt.center = (simplex[0] + simplex[1] + simplex[2]) / 3.0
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_gjk.py
```diff
@@ -1063,6 +1073,9 @@ def _polytope3(
   geomtype2: int,
 ) -> Polytope:
   """Create polytope for EPA given a 2-simplex from GJK."""
+  # set the polytope center
+  pt.center = (simplex[0] + simplex[1] + simplex[2]) / 3.0
```
thowell: `* wp.static(1.0 / 3.0)`?

thowell (approved):

mujoco_warp/_src/collision_gjk.py
kbayes: Sure

PR #1476, opened 2026-07-01
Optimize factorization for small inertia blocks

adenzler-nvidia:
MuJoCo's compact simple-body `M` layout is only valid while the omitted couplings stay zero. Runtime-mutable integrations can compile those bodies non-simple to preserve the full native triangular storage, but upstream routes the resulting small blocks through tile Cholesky. This change makes that opt-out practical without expanding or replacing MuJoCo's native matrix layout.

Blocks with at most six DOFs use one scalar path when their native storage is either compact diagonal or a complete lower triangle. Compact blocks use reciprocal-diagonal factorization, while full blocks always use scalar Cholesky. The structural choice is encoded in the existing block-factor address and shared by the separate and fused factor/solve paths.

Partial native CSR blocks continue to use tile Cholesky. Blocks from 7 through 64 DOFs also use tile Cholesky, and larger blocks use sparse LDL. This removes the separate runtime-simple group and the maximal-layout conversion from the earlier version of this PR. It does not allow a compact matrix to acquire structurally absent entries; callers that mutate inertia coupling must compile the affected bodies non-simple.

The benchmark uses 1,024 worlds containing 100 independent 6-DOF blocks each, for 102,400 blocks per launch. Each value is the median of seven samples with 3,000 repeated launches per sample after warmup, paired against `upstream/main` at `811485fc`.

| Native layout and runtime state | Factor | Solve | Fused factor/solve |
|---|---:|---:|---:|
| Compact, diagonal | 14.091 → 16.660 us (+18%) | 13.147 → 16.768 us (+28%) | 15.830 → 18.886 us (+19%) |
| Full, diagonal | 54.157 → 44.350 us (-18%) | 56.752 → 17.298 us (-70%) | 54.133 → 52.957 us (-2%) |
| Full, 10% coupled | 54.005 → 44.624 us (-17%) | 56.905 → 17.203 us (-70%) | 54.297 → 54.464 us (+0.3%) |
| Full, coupled | 54.490 → 43.544 us (-20%) | 56.942 → 17.284 us (-70%) | 54.897 → 54.057 us (-1.5%) |

The compact case shows the local cost of routing ordinary simple bodies through the common block kernel instead of dedicated per-DOF kernels. Full runtime-mutable blocks factor about 18% faster and solve about 70% faster than tile Cholesky, while fused factor/solve remains within about 2%.

Measurements were collected with Warp 1.14 on an NVIDIA RTX PRO 6000 Blackwell and CUDA 12.9.

Added coverage exercises compact diagonal blocks, full blocks with diagonal and coupled values across worlds, coupled small blocks, partial small CSR blocks, upload-time factor initialization, and mixed block/sparse layouts. The full suite passes with 1,013 tests and 3 skips. Pre-commit and the kernel analyzer also pass.


mujoco_warp/_src/io.py
```diff
@@ -169,6 +169,71 @@ def _m_allow_dense(mjm: mujoco.MjModel) -> bool:
   return not (mjm.ntendon and np.any(mjm.tendon_armature))
 
 
+def _m_full_rownnz(mjm: mujoco.MjModel) -> np.ndarray:
+  """Return row lengths in MuJoCo's maximal ancestor storage."""
+  return np.diff(np.append(mjm.dof_Madr, mjm.nM)).astype(np.int32)
+
+
+def _m_full_layout(mjm: mujoco.MjModel) -> dict[str, np.ndarray | int]:
```
adenzler-nvidia: This is the part of the implementation where I would especially appreciate suggestions for a better option. I reused `dof_Madr`, `mapM2M`, and the existing D CSR metadata where possible, but this helper still reconstructs the maximal ancestor columns from `dof_parentid`. Is there an existing MuJoCo structure or a cleaner mapping we should use instead?

yuvaltassa:
Not sure I like this solution. Correct me if I misunderstand, but I think this is for users that wish to make a simple body non-simple (i.e. misalign a body's inertia that was aligned at compile time)... is that right?

If yes, then I much prefer to do this at the compiler level and have it apply to MuJoCo as well. What would you say to something like `<body simple="auto/false">` where "auto" is the curent behavior and "false" disables the simple-body optimization?

adenzler-nvidia:
Thanks - yes, this covers the fact that a body might be simple while MjWarp put_model is called, but later inertia randomization makes it non-simple.

I don't like this solution either because it's really a hack. So fixing it at the compiler level is fine for me - and I would welcome parity between MuJoCo and MjWarp here. I don't think it's a good idea to have bodies randomly transition between these states, it adds a lot of complexity. So making simpleness a guaranteed contract would be useful.

adenzler-nvidia:
I took another pass based on the review feedback and substantially narrowed the approach.

The earlier version expanded MuJoCo's compact mass-matrix layout so bodies compiled as simple could become coupled at runtime. That added considerable `io.py` plumbing and treated a structural compile-time choice as runtime-mutable.

The revised version keeps MuJoCo's native layout. Integrations that need to mutate inertia coupling must compile the affected bodies non-simple, giving them full triangular storage. This PR then makes that option inexpensive by routing both compact and full native blocks of up to six DOFs through one scalar kernel. Full blocks select diagonal or Cholesky factorization from their current values; larger blocks retain the tile and sparse paths.

This removes the maximal-layout conversion, export projection, and separate runtime-simple group. I also reran the benchmarks against current `upstream/main`; the updated description includes both the compact-body overhead and the improvements for full runtime-mutable blocks.

A compiler option to preserve full storage would still be useful, but it is not blocking Newton because we can arrange for runtime-mutable bodies to be classified as non-simple during compilation.

yuvaltassa:
Done: https://github.com/google-deepmind/mujoco/commit/5618666a7ddbf6eb9bc5dac4a38acf3cc37beeb3

yuvaltassa:
One thing that's a little annoying is that bodies don't have defaults, so you need to tag this per body. That's not so bad, but I also have a followup adding body defaults.

adenzler-nvidia:
Thank you very much!

I'll test this on our side. And will think about that PR. The non-simple optimization here still makes sense, but I'm not sure whether we should accept the regression for pure simple bodies. Although I would really like to kick out some complexity here.

adenzler-nvidia:
@thowell @erikfrey how should we proceed here? this is now an independent optimization for non-simple small bodies, unifying some code.

thowell:
@adenzler-nvidia lets proceed. is the implementation up to date and ready for review?

adenzler-nvidia:
Brought the branch up to date with main and did a self-review pass: the mixed-blocks test now exercises the scalar, tile, and sparse factor paths in one model, new assertions pin the per-block routing (partial block to the gather path, coupled block to the scalar path, and compact plus triangular blocks of the same size sharing one launch), the fused sparse solve references the named sentinel instead of a bare -1, and the scalar launch widths moved into BlockDim. This is ready for review.

mujoco_warp/_src/io_test.py
```diff
@@ -1455,6 +1455,48 @@ def test_set_const_freejoint(self):
     _assert_eq(m.dof_invweight0.numpy()[0], mjm.dof_invweight0, "dof_invweight0")
     _assert_eq(m.body_invweight0.numpy()[0, 1], mjm.body_invweight0[1], "body_invweight0")
 
+  def test_set_const_full_freejoint_per_world_com(self):
+    """A full free-joint factor handles diagonal and coupled worlds."""
+    mjm = mujoco.MjModel.from_xml_string("""
```
thowell: let use `test_data.fixture` here

mujoco_warp/_src/smooth.py
```diff
@@ -1249,6 +1231,50 @@ def _factor_i_sparse(m: Model, d: Data, M: wp.array2d[float], L: wp.array2d[floa
...
+    D_out: wp.array2d[float],
+    L_out: wp.array2d[float],
+  ):
+    worldid, blk = wp.tid()
+    start = block_dof[blk]
+    size = wp.static(block_size)
+    matrix_adr = M_rowadr[start]
+
+    factor_adr = qLD_block_adr[start]
+    if factor_adr < 0:
```
thowell: can we check `Q_LD_BLOCK_{COMPACT/SPARSE}` here for improved readability?

mujoco_warp/_src/smooth.py
```diff
@@ -3029,48 +3061,63 @@ def _solve_LD_sparse(
...
+    L_in: wp.array2d[float],
+    y_in: wp.array2d[float],
+    # Out:
+    x_out: wp.array2d[float],
+  ):
+    worldid, blk = wp.tid()
+    start = block_dof[blk]
+    size = wp.static(block_size)
+    factor_adr = qLD_block_adr[start]
+    if factor_adr < 0:
```
thowell: can we check `Q_LD_BLOCK_{COMPACT/SPARSE}` here for readability?

mujoco_warp/_src/smooth.py
```diff
@@ -3200,49 +3259,113 @@ def kernel(
...
+    x_out: wp.array2d[float],
+    L_out: wp.array2d[float],
+  ):
+    worldid, blk = wp.tid()
+    start = block_dof[blk]
+    size = wp.static(block_size)
+    matrix_adr = M_rowadr[start]
+
+    factor_adr = qLD_block_adr[start]
+    if factor_adr < 0:
```
thowell: lets check `Q_LD_BLOCK_{COMPACT/SPARSE}` here

thowell (approved):
thanks @adenzler-nvidia!

PR #1480, opened 2026-07-02
Add tests where EPA convergences on first iteration

kbayes:
Reference to MuJoCo commit [6957966](https://github.com/google-deepmind/mujoco/commit/6957966).

The examples I could pull out that are affected by this guard are actually fine, so left them in as these are useful tests to guard against future regressions.


thowell (approved):

mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -507,6 +507,90 @@ def test_box_box_ccd2(self):
...
+      0.00605074502527713776,
+      0.85659545660018920898,
+      -0.51595354080200195312,
+      -0.05836316198110580444,
+      0.51538598537445068359,
+      0.85496860742568969727,
+    )
+
+    dist, _, _, _ = _geom_dist(m, d, 0, 1, multiccd=False, pos1=pos1, mat1=rot1, pos2=pos2, mat2=rot2)
+    self.assertEqual(dist, -2.515156e-06)  # -2.515764037690309e-06 actual depth
```
thowell: should we update to check almost equal?

kbayes: Oops, typo. Fixed now. Can confirm passes on CPU locally now.

PR #1481, opened 2026-07-02
elastic2d

thowell:
flex elastic2d

```
mjwarp-viewer ../mujoco/model/flex/hollow_vs_solid.xml --nconmax=512 --njmax=2048
```

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -26,6 +26,7 @@ from mujoco_warp._src import math as mjmath
 from mujoco_warp._src import render_util
 from mujoco_warp._src import sleep
 from mujoco_warp._src import smooth
+from mujoco_warp._src import support
 from mujoco_warp._src import types
 from mujoco_warp._src import warp_util
 from mujoco_warp._src.types import MJ_MINVAL
@@ -145,16 +146,29 @@ def _jtdaj_groups(mjd: mujoco.MjData) -> tuple[np.ndarray, np.ndarray]:
 
 def _get_nflexintcell(mjm: mujoco.MjModel) -> int:
   nflexintcell = 0
-  if mjm.nflex > 0 and hasattr(mjm, "flex_interp"):
+  if mjm.nflex > 0:
     for fi in range(mjm.nflex):
       order = abs(int(mjm.flex_interp[fi]))
       if order == 0:
         continue
-      if hasattr(mjm, "flex_edgeequality") and mjm.flex_edgeequality[fi] == 3:
+      if mjm.flex_edgeequality[fi] == 3:
         continue
       cx, cy, cz = mjm.flex_cellnum[fi]
       nflexintcell += int(cx) * int(cy) * int(cz)
   return nflexintcell
+
+
+def _get_nflexface(mjm: mujoco.MjModel) -> int:
+  nflexface = 0
+  if mjm.nflex > 0:
+    for fi in range(mjm.nflex):
+      order = mjm.flex_interp[fi]
+      if order >= 0:
+        continue
+      cx, cy, cz = mjm.flex_cellnum[fi]
+      nfaces = 2 * (cy * cz + cx * cz + cx * cy)
+      nflexface += int(nfaces)
+  return nflexface
 
 
 def is_sparse(mjm: mujoco.MjModel) -> bool:
@@ -344,9 +358,6 @@ def put_model(mjm: mujoco.MjModel, batch_sizes: dict[str, int] | None = None) ->
   if (mjm.opt.enableflags & mujoco.mjtEnableBit.mjENBL_SLEEP) and (mjm.eq_type == mujoco.mjtEq.mjEQ_FLEX).any():
     raise NotImplementedError("Flex equality constraints are not supported with sleeping enabled.")
 
-  if mjm.nflex > 0 and (mjm.flex_interp < 0).any():
-    raise NotImplementedError("Flex interpolation order < 0 (shell/quad elements) is not supported.")
-
   if mjm.opt.noslip_iterations > 0:
     raise NotImplementedError(f"noslip solver not implemented.")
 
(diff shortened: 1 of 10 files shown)


mujoco_warp/_src/io.py
```diff
@@ -741,15 +738,85 @@ def _check_margin(name, t1, t2, margin):
   m.eq_flex_adr = np.nonzero(mjm.eq_type == types.EqType.FLEX)[0]
   m.eq_flexstrain_adr = np.nonzero(mjm.eq_type == types.EqType.FLEXSTRAIN)[0]
   m.neq_flexstrain = m.eq_flexstrain_adr.size
-
   # Precompute flex strain Jacobian sparsity pattern
   flexstrain_J_rownnz = []
   flexstrain_J_colind = []
 
+  def _gather_face_node_index(order_abs: int, cx: int, cy: int, cz: int, face_elem_idx: int, local_idx: int) -> int:
```
quagla: can you use the function from support.py?

mujoco_warp/_src/passive.py
```diff
@@ -728,7 +728,7 @@ def _flex_bending(
       f = i
       break
 
-  bendingadr = flex_bendingadr[f]
+  bendingadr = flex_bendingadr[f] + locid * 17
```
quagla: This seems a significant difference from mujoco. Is it an intended warp optimization?

quagla (commented):
We need the bending term too for elastic2d I think. It might be working also without it though, maybe you just get an overly soft behavior.

mujoco_warp/_src/io.py

mujoco_warp/_src/passive.py
thowell: refactored to more closely follow the mujoco indexing

mujoco_warp/_src/types.py
```diff
@@ -1037,6 +1038,7 @@ class Model:
     body_gravcomp: antigravity force, units of body weight   (*, nbody)
     body_contype: OR over all geom contypes                  (nbody,)
     body_conaffinity: OR over all geom conaffinities         (nbody,)
+
```
quagla, line 1050: is this newline intentional?

mujoco_warp/_src/types.py
```diff
@@ -1436,12 +1438,18 @@ class Model:
...
+    flex_faceadr: maps each flex to its start face index                    (nflex,)
     flex_cell_map: precomputed flex cell mapping (nflexintcell,)
     flexstrain_J_rownnz: number of nonzeros in flex strain Jacobian row (neq_flexstrain,)
     flexstrain_J_rowadr: row start address in colind array (neq_flexstrain,)
     flexstrain_J_colind: column indices in sparse flex strain Jacobian (nJfs,)
     neq_flexstrain: number of flex strain equality constraints
     nJfs: number of non-zeros in sparse flex strain Jacobian
+    nflexbend_interp: number of interpolated bending edges
+    flex_bend_interp_map: mapping of interpolated bending edges to flex and local
+                          edge indices (nflexbend_interp, 2)
```
quagla: weird indentation

mujoco_warp/_src/support.py
```diff
@@ -548,6 +548,55 @@ def jac_dof(
   return jacp, jacr
 
 
+@wp.func
+def apply_force_torque(
```
quagla: should we use function names that are easier to map to mujoco public functions?

mujoco_warp/_src/smooth.py
```diff
@@ -503,6 +503,130 @@ def kinematics(m: Model, d: Data):
   )
 
 
+@wp.kernel
+def _flex_precompute_faces(
```
quagla: also here the function name doesn't tell much

quagla (commented):
I assume you loaded hollow_vs_solid and compared the behavior with mujoco?

thowell:
added additional tests

mujoco_warp/_src/types.py
thowell: removed

mujoco_warp/_src/types.py
thowell: updated to align local shapes

mujoco_warp/_src/support.py
thowell: removed this function -- part of a refactor to use the existing `support.apply_ft`

mujoco_warp/_src/smooth.py
thowell: renamed to `_flex_face_kinematics`

mujoco_warp/_src/support.py
thowell: updated now (function and a corresponding test removed)

thowell:
@quagla

yes ran 

```
mjwarp-viewer ../mujoco/model/flex/hollow_vs_solid.xml --nconmax=2048 --njmax=8192
```

and the simulation looks reasonable

edit:
note sometimes the mujoco warp version of the flex on the left falls to the right whereas in the mujoco simulation the object (always?) falls to the left?

mujoco_warp/_src/types.py
```diff
@@ -2181,6 +2192,9 @@ class Data:
     flex_aabb_max: dynamic flex object bounding box max         (nworld, nflex, 3)
     flexnode_xpos: cartesian flex node positions                (nworld, nflexnode, 3)
     overflow: overflow bitmask (OverflowType)                   (nworld,)
+    face_xpos: cartesian flex face positions                    (nworld, nflexface, 9, 3)
+    face_quat: cartesian flex face orientations                 (nworld, nflexface, 4)
+    face_gidx: flex face node indices                           (nflexface, 9)
```
quagla: Why is face_gidx in Data?

mujoco_warp/_src/support.py
```diff
@@ -67,7 +67,7 @@ def next_act(
 @wp.func
 def mat33_to_quat_polar(F: wp.mat33) -> wp.quat:
   cell_quat = wp.quat(0.0, 0.0, 0.0, 1.0)
-  for _iter in range(10):
+  for _iter in range(50):
```
quagla, line 70: Maybe we should add a test in support_test. You can take those from mujoco or just compare values.

mujoco_warp/_src/types.py
thowell: moved to `Model` and renamed to `flex_face`

mujoco_warp/_src/support.py
thowell: added a test to support_test.py

mujoco_warp/_src/smooth.py
```diff
@@ -503,6 +503,119 @@ def kinematics(m: Model, d: Data):
...
+    else:
+      face_xpos_out[worldid, face_id, local_idx] = wp.vec3(0.0)
+
+  normal = wp.cross(t1, t2)
+
+  size01 = cy * cz
+  size23 = cx * cz
+  size45 = cx * cy
+
+  face_type = 0
```
quagla: Should you use the get_face_metadata function here?

quagla (approved):
LGTM just it would be really nice to have one test in constraint_test to confirm equivalence with mujoco to catch if one day we change the formulation in mujoco (not planning to but it may happen one day).

thowell:
added `test_flex_interpolated` to constraint_test.py

PR #1485, opened 2026-07-03
Minor renaming in GJK

kbayes:
* Improved S1D and linear_combine a little
* Rename xnorm -> xnorm2 as it's the x_k normal squared
* Rename coordinates to lmbda (lambda is a Python keyword) to reflect C MuJoCo and the literature naming convention. Coordinates is too vague and long of a name.
* Move convergence check before the cut off check. This is now same as C MuJoCo and rationally makes more sense.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -570,8 +570,7 @@ def _S1D(s1: wp.vec3, s2: wp.vec3) -> wp.vec2:
   p_o = _project_origin_line(s1, s2)
 
   # find the axis with the largest projection "shadow" of the simplex
-  mu = s1[0] - s2[0]
-  mu_max = mu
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_gjk.py
```diff
@@ -573,13 +570,19 @@ def _S1D(s1: wp.vec3, s2: wp.vec3) -> wp.vec2:
   p_o = _project_origin_line(s1, s2)
 
   # find the axis with the largest projection "shadow" of the simplex
-  mu_max = 0.0
+  mu = s1[0] - s2[0]
+  mu_max = mu
```
thowell: would it make sense to write the result above to `mu_max` and remove the `mu` above since `mu` is overwritten after `index` below?

thowell (approved):

PR #1488, opened 2026-07-03
Guard tile Cholesky against float32-indefinite Hessians with a pivot floor

### LINKED ISSUE #1415: block_cholesky / tile_cholesky lack pivot floor, leads to NaNs
## Summary

With `solver=newton`, the constraint solver builds the Hessian

```
H = qM + Σ_k efc_D[k] · J[k]ᵀ J[k]
```

and factors it with a Cholesky (`block_cholesky` for `nv > 32`, `tile_cholesky` for `nv ≤ 32`, in `mujoco_warp/_src/solver.py`). Both the build and the Cholesky run in **float32**, with **no pivot floor**. Reference MuJoCo's `mju_cholFactor` floors each pivot with a `mindiag`; the Warp port dropped it.

 With `efc_D > 0` on every active row, `H` is **mathematically SPD**. But on stiff-contact states (large `efc_D`; `impratio` inflates the tangential rows of elliptic contacts), `‖JᵀD J‖` reaches `~1e8`, so the **float32 assembly** of `H` carries an absolute error `~ ‖H‖·eps32` that **swamps H's true small eigenvalues** (the rigid-body dynamics modes, `~3e-4`). The float32 `H` flips numerically **indefinite**, and the unguarded Cholesky takes `sqrt()` of a non-positive pivot → **NaN**, which then propagates `qfrc_constraint → qacc → qpos/qvel`. `solver=cg` (matrix-free, no Hessian factorization) is **clean** on the exact same state.

Tagging @ooctipus who also experienced the same issue independently.

## Expected vs. actual

- **Expected:** a mathematically-SPD constraint Hessian factors without producing NaN, as in reference MuJoCo (whose `mju_cholFactor` floors pivots with `mindiag`).
- **Actual:** float32 `block_cholesky` / `tile_cholesky` return NaN; the NaN propagates into `qacc`/`qpos`/`qvel` and corrupts the step.

## Root cause

`block_cholesky.py` / the `tile_cholesky` path do an **unguarded** factorization in float32. For a Hessian with a wide eigenvalue range — large `‖JᵀD J‖` (≈ `efc_D_max · ‖J‖²`) over small dynamics modes — the float32 assembly error (`~ ‖H‖·eps32`) exceeds the smallest true eigenvalue, so the stored matrix is indefinite and the factor square-roots a non-positive pivot.

## Reproduction (pure numpy, self-contained)

The same `H = qM + Σ efc_D·JᵀJ` is SPD when assembled in float64 but indefinite when accumulated in float32, so the float32 Cholesky fails:

```python
import numpy as np
rng = np.random.default_rng(0); nv = 35          # nv > 32 -> the blocked path

# qM: SPD mass matrix with a fragile small dynamics mode (3e-4) up to O(1)
q, _ = np.linalg.qr(rng.standard_normal((nv, nv)))
qM = (q * np.linspace(3e-4, 1.0, nv)) @ q.T

# stiff contacts: rank-1 efc_D * J Jᵀ terms, ||Jᵀ D J|| ~ 1e8
J = rng.standard_normal((12, nv)); D = np.geomspace(2e5, 2e7, 12)

H64 = qM.copy()                                  # assemble in float64
for k in range(12):
    H64 += D[k] * np.outer(J[k], J[k])

H32 = qM.astype(np.float32)                      # accumulate in float32 (what the solver does)
for k in range(12):
    jk = J[k].astype(np.float32)
    H32 = (H32 + np.float32(D[k]) * np.outer(jk, jk)).astype(np.float32)
H32 = 0.5 * (H32 + H32.T)

print("eigmin, float64 assembly   :", np.linalg.eigvalsh(H64).min())   # +9.7e-02  (SPD)
print("eigmin, float32 accumulation:", np.linalg.eigvalsh(H32).min())  # -1.7e+01  (INDEFINITE)
np.linalg.cholesky(H32)   # -> LinAlgError (numpy);  block_cholesky -> NaN (unguarded sqrt)
```

```
eigmin, float64 assembly    : +0.0973
eigmin, float32 accumulation : -16.51
numpy.linalg.LinAlgError: Matrix is not positive definite
```

A fuller standalone script is attached — it adds **(a)** a per-pivot `mindiag` floor that recovers the factor, and **(b)** a dump of a *real* `ctx.h` (48×48, `nv=35`) from a contact-rich state where the float32 factor NaNs (`eigmin = −1.48`, `efc_D` up to `2700`, `‖JᵀD J‖ ≈ 8.6e7`) alongside a clean state from the same scene (`efc_D ≈ 8`, `eigmin = +3e-4`, factors fine). Both halves run with only numpy.

[solver_hessian_env0_env221.txt](https://github.com/user-attachments/files/28656097/solver_hessian_env0_env221.txt)
[repro_chol_floor_standalone.py](https://github.com/user-attachments/files/28656101/repro_chol_floor_standalone.py)

## Proposed fix

Add a **scale-relative** `mindiag` pivot floor to `block_cholesky` and `tile_cholesky`, matching `mju_cholFactor` (clamp each pivot to `max(pivot, mindiag)` with `mindiag` proportional to the matrix scale, e.g. `~ eps · max_diag`). A small **absolute** floor is insufficient — when `‖H‖` is large the floor is float32-swallowed; the clamp must scale with the matrix. 

**I've validated a diagonal regularization injected before the factor that eliminates the NaN at negligible cost**; a per-pivot relative `mindiag` (as upstream MuJoCo does) is the more robust form. (A float64 assembly/factor of the constraint Hessian, as reference MuJoCo uses, would also remove the root cause.)

## Environment

- mujoco-warp: observed on `3.5.0.2`; `block_cholesky.py` is byte-identical in `3.6.0`, so the gap is present there too
- mujoco: `3.5.0` / `3.6.0`
- warp-lang: `1.12.0`
- newton (as the calling solver): `git@2684d75` (the `solver=newton` MJWarp path)

elliot-at-liminalnook:
Fixes #1415.

The Newton solver builds the constraint Hessian H = M + JᵀDJ and factors it with the tile Cholesky in float32, with no pivot floor. On stiff-contact states the float32 assembly error (~‖H‖·eps32) swamps H's smallest eigenvalues, so a Hessian that is SPD in float64 becomes numerically indefinite in float32 (eigmin -4.28 on the test matrix here). The unguarded factorization then square-roots a negative pivot and the NaNs propagate through qfrc_constraint into qacc and qpos. Reference MuJoCo avoids this via the mindiag pivot floor in mju_cholFactor, which the Warp port dropped.

The design keeps the fast path untouched: the existing unguarded tile factorization runs first, and only if its diagonal comes back non-finite do we rebuild the block and refactor element-wise with a scale-relative pivot floor eps32·max|diag|. Both call sites are guarded, the blocked factorize-solve for nv > 32 and the single-tile `_update_gradient_cholesky` for nv <= 32.

One numerical finding worth stating explicitly, because it drove a deliberate deviation from mju_cholFactor: implementing C's literal "floor the pivot, then divide the row by it" overflows float32. Consecutive floored pivots amplify the row entries quadratically, and on a badly indefinite float32 matrix that runs away to inf well before the factor completes. So instead of dividing by a floored pivot, this treats a floored pivot as rank deficient: the pivot becomes sqrt(mindiag), the rest of its row is zeroed, and the direction is decoupled from the trailing blocks. Numerically that direction then behaves like a spring of stiffness mindiag rather than a hard constraint, which is a reasonable degradation for a mode that float32 cannot resolve anyway. On well-conditioned matrices the guard never fires and parity with a float64 reference solve is ~1.4e-7; the test also asserts that the pre-fix path still NaNs on the indefinite matrix, so the guard doesn't quietly become dead code.

One caveat I want to be honest about: I've only been able to validate this on the CPU/LLVM Warp backend. The repair branch writes the tile element-wise and is only correct because every lane computes identical values in identical order; I'd appreciate a reviewer with CUDA sanity-checking that the per-lane tile writes in the repair path behave as intended on GPU. The fast path is unchanged, so this only concerns the newly added repair branch.

I prepared this change with AI assistance (Claude); per AGENTS.md I have not added it as a commit co-author (it cannot sign the CLA), and I've reviewed the change and take responsibility for it.

### DIFF
--- a/mujoco_warp/_src/block_cholesky.py
+++ b/mujoco_warp/_src/block_cholesky.py
@@ -14,8 +14,105 @@
 # ==============================================================================
 
 from functools import lru_cache
+from typing import Any
 
 import warp as wp
+
+from mujoco_warp._src.types import MJ_MINVAL
+
+# float32 machine epsilon; the pivot floor is scaled relative to the matrix magnitude,
+# mirroring the mindiag argument of MuJoCo's mju_cholFactor (issue #1415)
+_FLOAT32_EPS = 1.1920929e-07
+
+
+@wp.func
+def tile_diag_is_finite(t: Any, size: int) -> bool:
+  """Returns True if the leading size x size diagonal of tile t is all finite."""
+  finite = bool(True)
+  for i in range(size):
+    if not wp.isfinite(t[i, i]):
+      finite = False
+  return finite
+
+
+@wp.func
+def cholesky_mindiag(A: wp.array2d[float], matrix_size: int) -> float:
+  """Scale-relative pivot floor for A: eps32 * max |diag(A)|, at least MJ_MINVAL."""
+  max_diag = float(0.0)
+  for i in range(matrix_size):
+    max_diag = wp.max(max_diag, wp.abs(A[i, i]))
+  return wp.max(_FLOAT32_EPS * max_diag, float(MJ_MINVAL))
+
+
+@wp.func
+def tile_load_elementwise(t: Any, A: wp.array2d[float], size: int):
+  """Reloads the leading size x size block of A into tile t element-wise."""
+  for a in range(size):
+    for b in range(size):
+      t[a, b] = A[a, b]
+
+
+@wp.func
+def rebuild_diagonal_block(t: Any, A: wp.array2d[float], U: wp.array2d[float], k: int, block_size: int):
+  """Reloads diagonal block A[k:k+bs, k:k+bs] into tile t.
+
+  Reapplies the Schur complement updates from the already-factorized block rows
+  of U element-wise.
+  """
+  for a in range(block_size):
+    for b in range(block_size):
+      s = A[k + a, k + b]
+      for r in range(k):
+        s -= U[r, k + a] * U[r, k + b]
+      t[a, b] = s
+
+
+@wp.func
+def tile_cholesky_floored_inplace(t: Any, size: int, mindiag: float) -> wp.int64:
+  """In-place upper Cholesky of the leading size x size block of tile t.
+
+  Each pivot is floored at mindiag like MuJoCo's mju_cholFactor, so factorizing
+  a numerically indefinite matrix yields a finite factor instead of NaNs (#1415).
+
+  A floored pivot is treated as rank deficient: it is set to sqrt(mindiag) and
+  the remainder of its row is zeroed, decoupling the direction (a spring of
+  stiffness mindiag). Dividing the row by a floored pivot instead (as
+  mju_cholFactor does in float64) amplifies consecutive deficient rows
+  quadratically and overflows float32. Returns a bitmask of the deficient rows
+  so callers can decouple them from trailing blocks as well.
+
+  Element-wise and redundant across the block's threads; this is only correct
+  because every thread computes identical values in identical order.
+  """
+  deficient = wp.int64(0)
+  for j in range(size):
+    s = t[j, j]
+    for k in range(j):
+      s -= t[k, j] * t[k, j]
+    if s >= mindiag:
+      piv = wp.sqrt(s)
+      t[j, j] = piv
+      inv = 1.0 / piv
+      for i in range(j + 1, size):
+        s2 = t[j, i]
+        for k in range(j):
+          s2 -= t[k, j] * t[k, i]
+        t[j, i] = s2 * inv
+    else:  # small, negative or NaN pivot
+      deficient |= wp.int64(1) << wp.int64(j)
+      t[j, j] = wp.sqrt(mindiag)
+      for i in range(j + 1, size):
+        t[j, i] = 0.0
+  return deficient
+
+
+@wp.func
+def tile_zero_deficient_rows(t: Any, deficient: wp.int64, size: int):
+  """Zeroes the rows of tile t flagged in the deficient bitmask."""
+  for j in range(size):
+    if ((deficient >> wp.int64(j)) & wp.int64(1)) != wp.int64(0):
+      for i in range(size):
+        t[j, i] = 0.0
 
 
 @lru_cache(maxsize=None)
@@ -51,6 +148,17 @@ def create_blocked_cholesky_factorize_solve_func(block_size: int, matrix_size_st
         wp.tile_matmul(wp.tile_transpose(U_block), y_block, rhs_view, alpha=-1.0)
 
       wp.tile_cholesky_inplace(A_kk_tile, fill_mode="upper")
+
+      # float32 roundoff in the assembly of H can push its smallest eigenvalues
+      # negative when ||H|| is large; the unguarded factorization above then
+      # produces NaNs. Detect this and refactor the block with a scale-relative
+      # per-pivot floor, cf. mju_cholFactor's mindiag (issue #1415). The rebuild
+      # is element-wise, so no cooperative tile op runs inside the branch.
+      deficient = wp.int64(0)
+      if not tile_diag_is_finite(A_kk_tile, block_size):
+        rebuild_diagonal_block(A_kk_tile, A, U, k, block_size)
+        deficient = tile_cholesky_floored_inplace(A_kk_tile, block_size, cholesky_mindiag(A, matrix_size))
+
       wp.tile_store(U, A_kk_tile, offset=(k, k), bounds_check=False, aligned=True)
 
       wp.tile_lower_solve_inplace(wp.tile_transpose(A_kk_tile), rhs_view)
@@ -70,6 +178,11 @@ def create_blocked_cholesky_factorize_solve_func(block_size: int, matrix_size_st
           wp.tile_matmul(wp.tile_transpose(U_jk_tile), U_ji_tile, A_ki_tile, alpha=-1.0)
 
         wp.tile_lower_solve_inplace(wp.tile_transpose(A_kk_tile), A_ki_tile)
+
+        # decouple rank-deficient rows from the trailing blocks (issue #1415)
+        if deficient != wp.int64(0):
+          tile_zero_deficient_rows(A_ki_tile, deficient, block_size)
+
         wp.tile_store(U, A_ki_tile, offset=(k, i), bounds_check=False, aligned=True)
 
     for i in range(matrix_size - block_size, -1, -block_size):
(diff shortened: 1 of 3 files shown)


elliot-at-liminalnook:
A few notes on how this PR relates to the discussion on #1415, since the thread had converged on a tradeoff this design tries to escape:

**On the cuSolverDx constraint** (@adenzler-nvidia): this PR deliberately does *not* modify `wp.tile_cholesky` or ask for a warp-level pivot floor. The fast path runs the existing unguarded factorization exactly as today; afterwards a cheap diagonal scan decides whether the block needs repair, and only failing blocks are rebuilt element-wise and refactored with a scalar floored Cholesky inside the tile. cuSolverDx keeps its SPD assumption; the repair path never runs on healthy matrices.

**On the performance concern with always-on diagonal regularization**: because detection is post-hoc, the healthy-path cost is unchanged — the well-conditioned parity test asserts the factor is bit-identical pre/post-patch, and `mindiag` is hoisted so the added scan is O(nv) once per matrix. Stiff-contact states that currently NaN pay the scalar repair; everything else pays nothing. That seemed like a better trade than a few % on every step of every healthy env.

**On matching `mju_cholFactor` literally** (the issue's proposed fix): we found that porting the C floor-then-divide recipe directly *overflows in float32* — consecutive floored pivots amplify the trailing updates quadratically, and we hit `inf` on the issue's own regime. So floored pivots are instead treated as rank-deficient: pivot = √mindiag, row decoupled from trailing blocks (spring-of-stiffness-mindiag semantics). The scale-relative floor itself is as the issue prescribes: `eps32 · max|diag(H)|`, clamped at `MJ_MINVAL`.

**Just pushed a strengthening** (d79f0b4): detection now triggers on *finite-but-below-floor* pivots, not only non-finite values — rank deficiency can factor to finite garbage without a NaN, and this also covers backends whose indefinite-input behavior is finite garbage rather than NaN. New regression test included where the unguarded factor is fully finite but a pivot sits far below the floor.

**One question we can't answer without CUDA hardware** (validation here is CPU/LLVM only, as flagged in the PR body): what does the cuSolverDx-backed `tile_cholesky` actually produce on an indefinite input — NaNs, finite garbage, or an error path? The health check should now catch either of the first two, but a confirmation of the failure mode (and a sanity check of the repair branch's per-lane tile writes on CUDA) would be very welcome.

elliot-at-liminalnook:
Update: we got A100 time and can answer our own cuSolverDx question. On sm_80 (warp 1.14, CUDA 12.9), the cuSolverDx-backed tile factorization produces **NaNs** on indefinite input — the in-test unguarded assertion passes on GPU, so the failure mode matches CPU/LLVM. All 4 tests in `block_cholesky_test.py` pass on CUDA, including the finite-garbage-pivot case, i.e. the health check triggers on hardware and the scalar repair branch's per-lane tile writes behave on a real device. Fast-path overhead of the health check was not measurable above run variance on a contact-rich 4096-env benchmark (82–102k env-steps/s spread across cold/warm runs vs 92k single-sample unfixed). The CPU-only caveat in the PR body no longer applies.

adenzler-nvidia:
Thanks - I should have written in the comments that cuSolverDx uses NaN as a signaling mechanism.

Without going too deep into the code yet things look correct on the surface, but I wonder how much this impacts performance. Given that you got your hands on an A100, are you able to run the benchmarks? Otherwise I can also give generate some blackwell numbers next week.

elliot-at-liminalnook:
Benchmarks run on the A100 (80GB PCIe, warp 1.14.0, CUDA 12.9, mujoco 3.10.0), `mjwarp-testspeed`, nworld 8192, nstep 1000, 3 reps each, medians reported. Base = merge-base of this branch with main (edbc49b), PR = this branch's head. Chose the two models so both call sites are covered: humanoid hits the nv≤32 tile path, unitree_g1 (scene_flat, `--njmax 300`) the blocked path.

| model | path | base steps/s | PR steps/s | delta |
|---|---|---|---|---|
| humanoid | tile (nv=27) | 2,806,428 | 2,786,619 | **−0.71%** |
| unitree_g1 | blocked (nv≈35) | 1,454,802 | 1,438,439 | **−1.12%** |

Rep spread within each config is ≤0.2%, so the deltas are real. The cost is the always-on part — the `mindiag` scan plus the factor-diagonal health check per solve iteration; the repair branch never fires on these healthy benchmarks (verified: converged worlds 8192/8192 on both). For comparison against the alternative discussed in #1415, this sits below the "few %" you measured for always-on diagonal regularization, with the difference that the factorization itself is untouched on the healthy path.

If ~1% is more than this should cost, two easy directions: (a) check the diagonal only after the final solver iteration rather than every iteration (a NaN'd factor still gets caught before its solve is consumed), or (b) gate the check behind an option so it's opt-in for stiff-contact workloads. Happy to implement either — and Blackwell numbers would still be a welcome cross-check!

ooctipus:
We ran into #1415 in an RL training workload (large batched manipulation scenes with stiff contact materials, elliptic cone, impratio=10, solver=newton) and tried this fix. First off, thanks — it does exactly what it promises, the NaNs are gone.

But a heads-up on the perf assumption: in our setup the float32 indefiniteness isn't an occasional event, it's chronic. With light bodies and stiff contacts (efc_D up in the 1e5–1e6 range, impratio inflating the friction rows on top) the assembled Hessian fails the health check on basically every solver iteration, so the element-wise rebuild+refactor becomes the steady-state path rather than a rare fallback.

Microbenchmark of the nv<=32 single-tile path (8192 worlds, nv=32, RTX 5090, warp 1.15 dev): on healthy matrices the guarded kernel is ~1.05x the unguarded one, which matches the ~1% number reported here. On indefinite matrices it's ~11.6x per launch. End to end, our training throughput dropped ~8x on a pyramidal-cone config and ~30x with elliptic+impratio=10 (roughly 1k env-steps/s where we normally see tens of thousands). I suspect part of the end-to-end gap beyond the raw kernel cost is that chronically floored, rank-deficient factors give poor search directions, so the Newton solver stops early-exiting and burns its full iteration budget.

One caveat: we measured this on a backport of the patch to the 3.8.1 solver structure (the lower-triangular split factorize/solve), not this branch as-is. The guard structure is the same though — unguarded fast path, then element-wise rebuild and refactor inside the branch — so I'd expect the repair cost to carry over.

Might be worth parallelizing the repair across the block's threads instead of the redundant element-wise loop, or folding the pivot floor into the primary factorization so there's no second pass, at least for workloads where indefiniteness is the common case rather than the exception. For now we've switched these scenes to solver=cg, which avoids the factorization entirely.

PR #1489, opened 2026-07-03
Add reset_data_keyframe to mirror mj_resetDataKeyframe in MuJoCo

sibocw:
Currently, `mujoco_warp.reset_data` mirrors `mujoco.mj_resetData`, but there's no `mujoco_warp.reset_data_keyframe` that mirrors `mujoco.mj_resetDataKeyframe`.

This PR aims to implement it by repeating exactly what `mj_resetDataKeyframe` does on the CPU: calling `mj_resetData`, then re-setting values defined in the keyframe.

The complication is that `nkey` and `key_{time,qpos,qvel,act,mpos,mquat,ctrl}` are not currently included in `mujoco_warp.Model`. Therefore, the reset-to-keypoint function would have to either
- receive the `MjModel` object as an additional input (this is the state after d599780), or
- rely on keyframe-related attributes being added to `Model` (this is the current state of this PR). The dynamic attribute setting logic in `put_model` automatically populates these fields, so we don't need to touch `put_model`.

Would this be something useful?

PS. I guess we could in principle also allow different worlds to be reset to different keyframes, but IDK if this is a "real" use case.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -2729,7 +2729,6 @@ def reset_data(m: types.Model, d: types.Data, reset: Optional[wp.array] = None):
 
 
 def reset_data_keyframe(
-  mjm: mujoco.MjModel,
   m: types.Model,
   d: types.Data,
   key: int,
@@ -2737,23 +2736,20 @@ def reset_data_keyframe(
 ):
   """Reset data, set fields from specified keyframe; optionally by world.
 
-  Unlike `reset_data`, this function requires the host-side MuJoCo model as an argument
-  becasue the keyframe data are not tracked by `types.Model`.
-
   Args:
-    mjm: The host-side MuJoCo model.
     m: The model containing kinematic and dynamic information (device).
     d: The data object containing the current state and output arrays (device).
-    key: The keyframe index to initialize the data with. An IndexError is raised if
-      key<0 or key>=mjm.nkey.
+    key: The keyframe index to initialize the data with.
     reset: Per-world bitmask. Reset if True.
   """
   reset_data(m, d, reset)
 
   @wp.kernel(module="unique", enable_backward=False)
   def reset_time(
+    # Model:
+    key_time: wp.array[float],
     # In:
-    target_time: float,
+    key: int,
     reset_in: wp.array[bool],
     # Data out:
     time_out: wp.array[float],
@@ -2764,12 +2760,14 @@ def reset_data_keyframe(
       if not reset_in[worldid]:
         return
 
-    time_out[worldid] = target_time
+    time_out[worldid] = key_time[key]
 
   @wp.kernel(module="unique", enable_backward=False)
   def reset_qpos(
+    # Model:
+    key_qpos: wp.array2d[float],
     # In:
-    target_qpos: wp.array[float],
+    key: int,
     reset_in: wp.array[bool],
     # Data out:
     qpos_out: wp.array2d[float],
@@ -2780,12 +2778,14 @@ def reset_data_keyframe(
       if not reset_in[worldid]:
         return
 
-    qpos_out[worldid, qid] = target_qpos[qid]
+    qpos_out[worldid, qid] = key_qpos[key, qid]
 
   @wp.kernel(module="unique", enable_backward=False)
   def reset_qvel(
+    # Model:
+    key_qvel: wp.array2d[float],
     # In:
-    target_qvel: wp.array[float],
+    key: int,
     reset_in: wp.array[bool],
     # Data out:
     qvel_out: wp.array2d[float],
@@ -2796,12 +2796,14 @@ def reset_data_keyframe(
       if not reset_in[worldid]:
         return
 
-    qvel_out[worldid, vid] = target_qvel[vid]
+    qvel_out[worldid, vid] = key_qvel[key, vid]
 
   @wp.kernel(module="unique", enable_backward=False)
   def reset_activation(
+    # Model:
+    key_act: wp.array2d[float],
     # In:
-    target_act: wp.array[float],
+    key: int,
     reset_in: wp.array[bool],
     # Data out:
     act_out: wp.array2d[float],
@@ -2812,16 +2814,16 @@ def reset_data_keyframe(
       if not reset_in[worldid]:
         return
 
-    act_out[worldid, aid] = target_act[aid]
+    act_out[worldid, aid] = key_act[key, aid]
 
   @wp.kernel(module="unique", enable_backward=False)
   def reset_mocap(
     # Model:
     body_mocapid: wp.array[int],
+    key_mpos: wp.array2d[wp.vec3],
+    key_mquat: wp.array2d[wp.quat],
     # In:
-    target_mpos: wp.array[wp.vec3],
-    target_mquat: wp.array[wp.quat],
-    # In:
+    key: int,
     reset_in: wp.array[bool],
     # Data out:
     mocap_pos_out: wp.array2d[wp.vec3],
@@ -2836,13 +2838,15 @@ def reset_data_keyframe(
     mocapid = body_mocapid[bodyid]
 
     if mocapid >= 0:
-      mocap_pos_out[worldid, mocapid] = target_mpos[mocapid]
-      mocap_quat_out[worldid, mocapid] = target_mquat[mocapid]
+      mocap_pos_out[worldid, mocapid] = key_mpos[key, mocapid]
+      mocap_quat_out[worldid, mocapid] = key_mquat[key, mocapid]
 
   @wp.kernel(module="unique", enable_backward=False)
   def reset_control(
+    # Model:
+    key_ctrl: wp.array2d[float],
     # In:
-    target_ctrl: wp.array[float],
+    key: int,
     reset_in: wp.array[bool],
     # Data out:
     ctrl_out: wp.array2d[float],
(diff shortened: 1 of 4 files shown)


thowell:
@erikfrey what do we think about adding `reset_data_keyframe` to the api?

erikfrey:
@thowell I think it's appropriate to add this to the mjwarp API.  I bet that env frameworks would call it for smarter resetting.

thowell:
@sibocw the implementation looks great thanks for putting this together. lets update the implementation to set keyframes per world.

instead of batching all of the keyframe fields in `Model` we can have `reset_data_keyframe` accept either an `int` or `wp.array(dtype=int)`. if an `int` is provide this value is used to set all worlds to the same keyframe. if `wp.array(dtype=int)` is provided it should have size `nworld` and will set keyframes per world. in the latter case if the keyframe provided for the world is `< 0` or `>=nkey` then we skip resetting the world.

sibocw:
@thowell How about this? I also noticed that there's no type/shape check for the per-world `reset` bitmask in `reset_data`, so I added it, along with checks for the target `key` in `reset_data_keyframe`. Let me know if you want me to get rid of them.

Thanks!

sibocw:
@thowell Have you had a chance to take a look yet? I rebased it to match the current main, all tests passing again on my local machine now.

thowell:
@sibocw thanks for updating the pr!

it would be great if we could combine the current functionality of the `key` and `reset` arguments into a single argument for simplicity.

sibocw:
@thowell good idea! How about now?

mujoco_warp/_src/io.py
```diff
@@ -2622,7 +2625,14 @@ def reset_sleep(
     if elemid < nv:
       dof_awake_ind_out[worldid, elemid] = elemid
 
-  reset_input = reset or wp.ones(d.nworld, dtype=bool)
+  if reset is None:
+    reset_input = wp.ones(d.nworld, dtype=bool)
+  elif isinstance(reset, wp.array):
```
thowell, line 2631: lets additionally check the dtype for bool

mujoco_warp/_src/io.py
```diff
@@ -2728,6 +2738,228 @@ def reset_sleep(
...
+      dtype is not int.
+  """
+  # Resolve target keyframe index
+  if isinstance(key, wp.array):
+    if key.shape != (d.nworld,):
+      raise ValueError(f"key array must have shape ({d.nworld},), got {key.shape}.")
+    if not wp.types.type_is_int(key.dtype):
+      raise ValueError(f"key array must be of integer type, got {key.dtype}.")
+    key_input = key
+  elif isinstance(key, int):
```
thowell: lets modify the check to: `isinstance(key, (int, np.integer))` and then cast `key = int(key)`

mujoco_warp/_src/io.py
```diff
@@ -2728,6 +2738,228 @@ def reset_sleep(
...
+    dim=d.nworld,
+    inputs=[m.nkey, key_input],
+    outputs=[reset_mask],
+  )
+
+  # Call normal reset using the mask
+  reset_data(m, d, reset_mask)
+
+  # Set time, qpos, qvel, act, mocap_pos, mocap_quat, ctrl from keyframes
+  @wp.kernel(module="unique", enable_backward=False)
```
thowell, line 2807: lets simplify to a single kernel that launches over `nworld` and loops over the keyframe arrays values. something like:

```
@wp.kernel(module="unique", enable_backward=False)
def reset_keyframe_data(
  # Model:
  nq: int,
  nv: int,
  na: int,
  nu: int,
  nmocap: int,
  key_time: wp.array[float],
  key_qpos: wp.array2d[float],
  key_qvel: wp.array2d[float],
  key_act: wp.array2d[float],
  key_ctrl: wp.array2d[float],
  key_mpos: wp.array2d[wp.vec3],
  key_mquat: wp.array2d[wp.quat],
  # In:
  key_in: wp.array[int],
  reset_in: wp.array[bool],
  # Data out:
  time_out: wp.array[float],
  qpos_out: wp.array2d[float],
  qvel_out: wp.array2d[float],
  act_out: wp.array2d[float],
  ctrl_out: wp.array2d[float],
  mocap_pos_out: wp.array2d[wp.vec3],
  mocap_quat_out: wp.array2d[wp.quat],
):
  worldid = wp.tid()
  if wp.static(reset_in is not None):
    if not reset_in[worldid]:
      return
  key = key_in[worldid] if wp.static(is_array) else key_scalar
  time_out[worldid] = key_time[key]
  for i in range(nq):
    qpos_out[worldid, i] = key_qpos[key, i]
  for i in range(nv):
    qvel_out[worldid, i] = key_qvel[key, i]
  for i in range(na):
    act_out[worldid, i] = key_act[key, i]
  for i in range(nu):
    ctrl_out[worldid, i] = key_ctrl[key, i]
  for i in range(nmocap):
    mocap_pos_out[worldid, i] = key_mpos[key, i]
    mocap_quat_out[worldid, i] = key_mquat[key, i]
```

mujoco_warp/_src/io_test.py
thowell, line 1: lets add the following tests:


1. **Per-world reset coverage for all state fields**: In `test_reset_data_keyframe_per_world`, assert that all keyframe fields (`time`, `qpos`, `qvel`, `act`, `ctrl`, `mocap_pos`, `mocap_quat`) are correctly applied or preserved per world, rather than just `qpos`.
2. **Models without keyframes (`nkey == 0`)**: Test that `reset_data_keyframe` properly raises `ValueError` when given a scalar `key`, and skips resetting when given a `wp.array`.
3. **NumPy scalar integer inputs**: Test passing NumPy integer types (e.g., `key = np.int32(0)` or `np.int64(0)`).
4. **Invalid `dtype` for `reset` in `reset_data`**: Test that passing a non-boolean array (e.g., `wp.zeros(..., dtype=float)` or `dtype=int`) raises a `ValueError`.

mujoco_warp/_src/types.py
```diff
@@ -1323,6 +1324,13 @@ class Model:
     sensor_interval: sensor interval and phase               (nsensor, 2)
     plugin: globally registered plugin slot number           (nplugin,)
     plugin_attr: config attributes of geom plugin            (nplugin, _NPLUGINATTR)
+    key_time: keyframe time                                  (nkey,)
+    key_qpos: keyframe qpos                                  (nkey, nq)
+    key_qvel: keyframe qvel                                  (nkey, nv)
+    key_act: keyframe act                                    (nkey, na)
+    key_mpos: keyframe mocap pos                             (nkey, nmocap, 3)
+    key_mquat: keyframe mocap quat                           (nkey, nmocap, 4)
+    key_ctrl: keyframe ctrl                                   (nkey, nu)
```
thowell: lets update the size indentation to align with the size alignments above

mujoco_warp/_src/io_test.py
```diff
@@ -1072,6 +1072,182 @@ def test_reset_data_world(self):
...
+          <geom type="sphere" size="1"/>
+          <joint type="slide"/>
+        </body>
+      </worldbody>
+      <keyframe>
+        <key name="k0" qpos="0.5"/>
+      </keyframe>
+    </mujoco>
+    """
+    mjm = mujoco.MjModel.from_xml_string(_MJCF)
```
thowell: lets update to use `test_data.fixture` and inline the xml

mujoco_warp/_src/io_test.py
```diff
@@ -1072,6 +1072,182 @@ def test_reset_data_world(self):
...
+          <joint type="slide"/>
+        </body>
+      </worldbody>
+      <keyframe>
+        <key name="k0" qpos="0.5"/>
+        <key name="k1" qpos="0.6"/>
+      </keyframe>
+    </mujoco>
+    """
+    mjm = mujoco.MjModel.from_xml_string(_MJCF)
```
thowell: lets update to use `test_data.fixture` and inline the xml

mujoco_warp/_src/io_test.py
```diff
@@ -1072,6 +1072,182 @@ def test_reset_data_world(self):
...
+          <joint type="slide"/>
+        </body>
+      </worldbody>
+      <keyframe>
+        <key name="k0" qpos="0.5"/>
+      </keyframe>
+    </mujoco>
+    """
+    key = 0
+    mjm = mujoco.MjModel.from_xml_string(_MJCF)
```
thowell: lets update to use `test_data.fixture` and inline the xml

mujoco_warp/_src/io_test.py
```diff
@@ -1072,6 +1072,182 @@ def test_reset_data_world(self):
...
+      <keyframe>
+        <key name="k0" time="0.5" qpos="0.3" qvel="0.4" act="0.6" ctrl="0.7"
+             mpos="0.1 0.2 0.3" mquat="0.7071068 0.7071068 0 0"/>
+      </keyframe>
+    </mujoco>
+    """
+    reset_datafield = ["time", "qpos", "qvel", "act", "ctrl", "mocap_pos", "mocap_quat"]
+    key = 0
+
+    mjm, mjd, m, d = test_data.fixture(xml=_MJCF, keyframe=key)
```
thowell: lets inline the xml

mujoco_warp/_src/io_test.py
```diff
@@ -1072,6 +1072,182 @@ def test_reset_data_world(self):
...
+    <mujoco>
+      <worldbody>
+        <body>
+          <geom type="sphere" size="1"/>
+          <joint type="slide"/>
+        </body>
+      </worldbody>
+    </mujoco>
+    """
+    mjm = mujoco.MjModel.from_xml_string(_MJCF)
```
thowell: lets update to use `test_data.fixture` and inline the xml

thowell:
@sibocw thanks for updating the pr! left some comments. please let us know if you have any questions. thanks!

mujoco_warp/_src/io.py
sibocw: do we want to additionally tolerate int?

mujoco_warp/_src/io_test.py
```diff
@@ -1072,6 +1071,266 @@ def test_reset_data_world(self):
     _assert_eq(d.qvel.numpy()[0], 1.0, "qvel[0]")
     _assert_eq(d.qvel.numpy()[1], 2.0, "qvel[1]")
```
sibocw, line 1072: I additionally inlined the XML for the existing `test_reset_data_world` for consistency

mujoco_warp/_src/io_test.py
sibocw, line 994: Personally I don't quite like the variable name `arr` because it sounds like it's an array while it's actually the string name of a data field. This appears a couple more times in this file.

But following this convention, I named the iterating variable `arr` for the new tests too.

Do you want to keep `arr` or maybe change it to something else?

mujoco_warp/_src/types.py
```diff
@@ -1323,6 +1324,13 @@ class Model:
     sensor_interval: sensor interval and phase               (nsensor, 2)
     plugin: globally registered plugin slot number           (nplugin,)
     plugin_attr: config attributes of geom plugin            (nplugin, _NPLUGINATTR)
+    key_time: keyframe time                                  (nkey,)
+    key_qpos: keyframe qpos                                  (nkey, nq)
+    key_qvel: keyframe qvel                                  (nkey, nv)
+    key_act: keyframe act                                    (nkey, na)
+    key_mpos: keyframe mocap pos                             (nkey, nmocap, 3)
+    key_mquat: keyframe mocap quat                           (nkey, nmocap, 4)
+    key_ctrl: keyframe ctrl                                  (nkey, nu)
```
sibocw, line 1333: Just a little bit of oddity to point out here: for the keyframe variables, I put `key_ctrl` after `key_mpos` and `key_mquat`, following the [order in the C `mjModel_` struct](https://github.com/google-deepmind/mujoco/blob/4929f2cd998ca83471ae5ec33d20ca210e75b791/include/mujoco/mjmodel.h#L864-L871).

However, in the mjw `Data` class, `ctrl` appears before `mocap_pos/quat`. This also mirrors [the C `mjData_` struct](https://github.com/google-deepmind/mujoco/blob/4929f2cd998ca83471ae5ec33d20ca210e75b791/include/mujoco/mjdata.h#L178-L186).

The result is that the kernel analyzer expects the parameter order `[..., key_mpos, key_mquat, key_ctrl]` for the keyframe stuff, but `[..., ctrl_out, mocap_pos_out, mocap_quat_out]` for the actual data buffers.

I think reordering things in the C structs could result in a catastrophic mess, so I kept it as is here.

mujoco_warp/_src/io.py
```diff
@@ -2622,7 +2626,16 @@ def reset_sleep(
...
       dof_awake_ind_out[worldid, elemid] = elemid
 
-  reset_input = reset or wp.ones(d.nworld, dtype=bool)
+  if reset is None:
+    reset_input = wp.ones(d.nworld, dtype=bool)
+  elif isinstance(reset, wp.array):
+    if reset.shape != (d.nworld,):
+      raise ValueError(f"reset array must have shape ({d.nworld},), got {reset.shape}.")
+    if reset.dtype != wp.bool:
+      raise ValueError(f"reset array must be of bool type, got {reset.dtype}.")
```
sibocw: do we want to additionally tolerate int?

sibocw (commented):
Thanks @thowell for the feedback! I added a review myself with 4 questions/comments.

mujoco_warp/_src/io_test.py
thowell: thanks!

mujoco_warp/_src/io_test.py
thowell: for this pr lets stick with `arr` to match the convention in the existing tests. please feel free to open an issue/pr about the `arr` naming convention and we can follow up in a separate issue/pr. thanks!

mujoco_warp/_src/types.py
thowell: lets match the mujoco field ordering. please feel free to open a pr/issue with mujoco about the field ordering. thanks!

mujoco_warp/_src/io.py
thowell: lets update to allow `int`. thanks!

sibocw: implemented now!

thowell (approved):
great work @sibocw, thanks!

PR #1490, opened 2026-07-03
Replace O(n²) flex contact dedup with sort-based O(n log n)

quagla:
## Summary

Replace the O(n²) `_filter_flex_candidates` kernel with a sort-based O(n log n) approach for deduplicating flex collision contacts.

## Problem

The existing dedup kernel launches with `dim=naconmax` and each thread loops over all `naconmax` candidates, resulting in `naconmax²` comparisons. For aloha cloth with `naconmax=115,200`, that is 13.3 billion operations per step.

## Solution

Three-stage sort-based deduplication:
1. **`_compute_filter_key`**: Compute a sort key per candidate encoding `(worldid, flex_id, geom_id)`
2. **`radix_sort_pairs`**: Sort candidates by key so duplicates are contiguous
3. **`_filter_flex_candidates_sorted`**: Compare only within same-key groups (neighbors in sorted order)

This reduces complexity to O(n log n) for the sort + O(n * k) for the filter, where k is the average group size (typically 5-20).

## Performance (RTX A4000)

| Benchmark | Before | After | Speedup |
|-----------|--------|-------|---------|
| Cloth (32 worlds) | 1,098 sps | 3,723 sps | **3.4x** |
| Aloha cloth (32 worlds) | 162 sps | 1,368 sps | **8.4x** |

Correctness verified: exact match of active contacts (27,487/27,487) between old and new filter.

### DIFF
--- a/mujoco_warp/_src/collision_flex.py
+++ b/mujoco_warp/_src/collision_flex.py
@@ -2643,6 +2643,108 @@ def _flex_narrowphase_tet_detect(
 
 
 @wp.kernel
+def _compute_filter_key(
+  # Model:
+  ngeom: int,
+  nflex: int,
+  # In:
+  ncand: wp.array[int],
+  cand_geom: wp.array[wp.vec2i],
+  cand_flex: wp.array[wp.vec2i],
+  cand_worldid: wp.array[int],
+  # Out:
+  key_out: wp.array[int],
+  val_out: wp.array[int],
+):
+  """Compute sort key for candidate grouping.
+
+  Groups candidates by (worldid, flex_id, geom_id) so that duplicates
+  are contiguous after sorting. Self-collision contacts (geom_id < 0)
+  are mapped to a sentinel value (ngeom).
+  """
+  i = wp.tid()
+  if i >= ncand[0]:
+    key_out[i] = 2147483647  # INT_MAX: sort unused entries to end
+    val_out[i] = i
+    return
+
+  worldid = cand_worldid[i]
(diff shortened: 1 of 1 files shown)


mujoco_warp/_src/collision_flex.py
```diff
@@ -2743,6 +2642,110 @@ def _flex_narrowphase_tet_detect(
...
+  """Compute sort key for candidate grouping.
+
+  Groups candidates by (worldid, flex_id, geom_id) so that duplicates
+  are contiguous after sorting. Self-collision contacts (geom_id < 0)
+  are mapped to a sentinel value (ngeom).
+  """
+  i = wp.tid()
+  limit = ncand[0]
+  if i >= limit:
+    key_out[i] = 2147483647
```
thowell: lets add a comment for this number

mujoco_warp/_src/collision_flex.py
```diff
@@ -2743,6 +2642,110 @@ def _flex_narrowphase_tet_detect(
...
+):
+  """Compute sort key for candidate grouping.
+
+  Groups candidates by (worldid, flex_id, geom_id) so that duplicates
+  are contiguous after sorting. Self-collision contacts (geom_id < 0)
+  are mapped to a sentinel value (ngeom).
+  """
+  i = wp.tid()
+  limit = ncand[0]
+  if i >= limit:
```
thowell: lets inline `ncand[0]`

thowell (approved):
AWESOME

PR #1496, opened 2026-07-06
flex-flex collisions

thowell:
add flex-flex collisions

- additionally remove flex internal collisions

---

```
<mujoco model="flex_flex">
  <include file="scene.xml"/>
  <option solver="CG" tolerance="1e-6" timestep="0.001" integrator="Euler"/>
  <size nconmax="50000" njmax="50000"/>

  <worldbody>
    <camera pos="0 -1.5 0.5" euler="70 0 0 "/>

    <!-- Cube 1: Resting on the plane -->
    <flexcomp name="cube1" type="grid" count="8 8 8" spacing="0.07 0.07 0.07" pos="-0.2 0 0.27"
              radius="0.001" dim="3" mass="5.0" dof="trilinear">
      <contact selfcollide="none" internal="false"/>
      <elasticity young="1e4" damping="0.01" poisson="0.1"/>
    </flexcomp>

    <!-- Cube 2: Falling from above onto the corner of Cube 1 -->
    <flexcomp name="cube2" type="grid" count="8 8 8" spacing="0.07 0.07 0.07" pos="-0.2 0 1.0"
              radius="0.001" dim="3" mass="5.0" dof="trilinear">
      <contact selfcollide="none" internal="false"/>
      <elasticity young="1e4" damping="0.01" poisson="0.1"/>
    </flexcomp>
  </worldbody>
</mujoco>
```

```
mjwarp-viewer flex-flex-collision.xml   --nconmax=16000   --njmax=16000
```

https://github.com/user-attachments/assets/d67baedb-70b1-4930-8195-7bc76f28281a

```
<mujoco model="flex_flex">
  <include file="scene.xml"/>
  <option solver="CG" tolerance="1e-6" timestep="0.001" integrator="Euler"/>
  <size nconmax="50000" njmax="50000"/>

  <worldbody>
    <camera pos="0 -1.5 0.5" euler="70 0 0 "/>

    <!-- Cube 1: Resting on the plane -->
    <flexcomp name="cube1" type="grid" count="8 8 8" spacing="0.07 0.07 0.07" pos="-0.2 0 0.27"
              radius="0.001" dim="3" mass="5.0" dof="trilinear">
      <contact selfcollide="none" internal="false"/>
      <elasticity young="1e4" damping="0.01" poisson="0.1"/>
    </flexcomp>

    <!-- Cube 2: Falling from above onto the corner of Cube 1 -->
    <flexcomp name="cube2" type="grid" count="8 8 8" spacing="0.07 0.07 0.07" pos="0.0 0 1.0"
              radius="0.001" dim="3" mass="5.0" dof="trilinear">
      <contact selfcollide="none" internal="false"/>
      <elasticity young="1e4" damping="0.01" poisson="0.1"/>
    </flexcomp>
  </worldbody>
</mujoco>
```

```
mjwarp-viewer ../flex-flex-collision.xml   --nconmax=16000   --njmax=16000
```

https://github.com/user-attachments/assets/ee8a2d45-e0e9-4517-882c-4155f123ba1b


thowell:
@quagla the implementation is updated and it looks like the flex-flex collisions are working correctly now.

2 notes:
- there is an issue with resets + warmstarts + flex + the viewer -- not sure what the issue is but can follow up in a future pr
- the existing sap broadphase is silently overflowing https://github.com/google-deepmind/mujoco_warp/blob/130c95b042cf9b2b7171a0e5068a09804bb49810/mujoco_warp/_src/collision_core.py#L572 as a result we should probably expect some performance regression on the aloha cloth benchmark since we likely need to increase `nconmax` / `naconmax` to avoid overflow. can follow up with changes related to that benchmark in a future pr

quagla (approved):

mujoco_warp/_src/io.py
```diff
@@ -1622,6 +1711,31 @@ def _initial_body_awake(mjm: mujoco.MjModel, nworld: int, init_asleep: bool) ->
...
+
+  for fi in range(mjm.nflex):
+    # Check elements
+    start_elem = mjm.flex_elemadr[fi]
+    end_elem = mjm.flex_elemadr[fi + 1] if fi < mjm.nflex - 1 else mjm.nflexelem
+    num_elems = end_elem - start_elem
+    if num_elems >= FLEX_KEY_ELEM_LIMIT:
+      raise ValueError(f"Flex object {fi} has too many elements: {num_elems} (limit is {FLEX_KEY_MAX_ELEM:,}).")
+
+    # Check vertices
```
quagla: Vertices or nodes?

mujoco_warp/_src/collision_flex.py
```diff
@@ -2801,260 +2788,563 @@ def _filter_flex_candidates_sorted(
...
+          flex_gap[flex2],
+        )
+
+      if cand_vert[i][0] >= 0 and flex_dim[flex1] == 3:
+        condim = 1
+      else:
+        condim = mixed_condim
+
+      if flex_dim[flex1] == 3 and flex_dim[flex2] == 3:
+        elem1 = cand_elem[i][0]
```
quagla: Can you guard elem1 and elem2 to check they are not -1?

mujoco_warp/_src/io.py
```diff
@@ -284,6 +286,71 @@ def _filter_tri_geoms(
...
+    shell_start = mjm.flex_shelldataadr[fi]
+    shell_num = mjm.flex_shellnum[fi]
+
+    if shell_num == 0:
+      continue
+
+    shell_data = mjm.flex_shell[shell_start : shell_start + shell_num * 3].reshape(-1, 3)
+
+    vert0 = mjm.flex_vert0[vert_start : vert_start + vert_num]
+    shell_normals = []
```
quagla: Do you only need the undeformed normals? Don't you need to update them at every step?

mujoco_warp/_src/collision_flex.py
```diff
@@ -2801,260 +2788,563 @@ def _filter_flex_candidates_sorted(
...
+          flex_friction[flex2],
+          flex_gap[flex2],
+        )
+
+      if cand_vert[i][0] >= 0 and flex_dim[flex1] == 3:
+        condim = 1
+      else:
+        condim = mixed_condim
+
+      if flex_dim[flex1] == 3 and flex_dim[flex2] == 3:
```
quagla: I think you also want flex1 != flex2?

mujoco_warp/_src/collision_flex.py
thowell: done

mujoco_warp/_src/io.py
thowell: updated to 'nodes'

mujoco_warp/_src/collision_flex.py
thowell: fixed

mujoco_warp/_src/io.py
thowell: removed this feature

thowell:
@kbayes please take a look at the changes to `_polytope3`. think there is a bug where a vector is not normalized. thanks!

kbayes:
@thowell oh yes! I had this fixed in an old PR that I closed but forgot to add the fix back in. Thanks!

mujoco_warp/_src/collision_flex.py
```diff
@@ -2781,20 +2393,31 @@ def _compute_filter_key(
...
+    f1 = wp.min(flex1, flex_id)
+    f2 = wp.max(flex1, flex_id)
+    group = ngeom * nflex + f1 * nflex + f2
+
+  group_id = worldid * (ngeom * nflex + nflex * nflex) + group
+  is_self = int(flex1 == flex_id) if geom_id < 0 else 0
 
-  key_out[i] = worldid * (nflex + 1) * (ngeom + 2) + flex_id * (ngeom + 2) + g
+  group_key = (wp.int64(group_id) << wp.int64(1)) | wp.int64(is_self)
+  elem_key = (wp.int64(elem1) << wp.int64(16)) | wp.int64(elem2)
```
quagla: elem1 is -1 for every geom-elem and vertex candidate (_write_candidate sets cand_elem = (-1, elemid)), so int64(-1) << 16 = 0xFFFF_FFFF_FFFF_0000 and the OR on line 2420 sets bits 16–63 unconditionally. Every geom-flex candidate ends up with key >> 32 == -1 regardless of world/geom/flex:

group_id=7,     elem=(-1,5)  -> key=-65531, group=-1
group_id=12345, elem=(-1,5)  -> key=-65531, group=-1
group_id=3,     elem=(10,20) -> key=30065426452, group=7

Since _find_group_starts (2721) and the neighbour scan (2461/2481) both compare on key >> 32, all geom-flex candidates across all worlds collapse into one group — O(n²) scan, and cross-world dedup once positions land within epsilon. The old key was correct here. Suggest elem_key = ((wp.int64(elem1 + 1) & 0xFFFF) << 16) | (wp.int64(elem2 + 1) & 0xFFFF). Note the 16-bit field also wraps above 65535 elements/flex.

mujoco_warp/_src/collision_flex.py
```diff
@@ -1882,71 +1509,83 @@ def _flex_selfcollision_narrowphase(
...
-    margin = 0.0
-
     contact_dist, contact_pos, contact_normal = collision_primitive_core.capsule_capsule(
-      cap1_pos, cap1_axis, radius, cap1_half_len, cap2_pos, cap2_axis, radius, cap2_half_len, margin
+      cap1_pos, cap1_axis, radius1, cap1_half_len, cap2_pos, cap2_axis, radius2, cap2_half_len, mix_margin
     )
 
     for c in range(2):
       d_val = contact_dist[c]
       if d_val < 0.0:
```
quagla: capsule_capsule returns absolute distances and already filters at <= mix_margin (1575), so this drops every contact with 0 <= dist < margin. The GJK branch below handles it correctly (1645/1666). C keeps these and makes them active constraints (includemargin = margin in mj_collideElems). Should be if d_val < mix_margin:.

mujoco_warp/_src/collision_flex.py
```diff
@@ -1743,7 +1199,7 @@ def _flex_sap_project(
     aabb_max = wp.max(aabb_max, p3)
 
   radius = flex_radius[flexid]
-  rbound = 2.0 * radius
+  rbound = radius
```
quagla: AABBs are inflated by flex_radius only. C's flexelem_aabb also carries radius (engine_core_smooth.c:655-657) but mj_collideElems then applies filterBox(..., margin + gap), and intermediate BVH nodes use filterBox(..., margin). With margin plumbed into narrowphase but not here, flex-flex pairs separated by more than 2*radius but less than margin+gap get pruned at broadphase. Moot for self-collision (margin forced to 0), not for flex-flex.

mujoco_warp/_src/flex_test.py
```diff
@@ -2594,6 +2391,161 @@ def test_insidesite_flex_body(self, nworld):
...
+            </flexcomp>
+          </worldbody>
+        </mujoco>
+        """,
+        False,
+      ),
+    ],
+    nworld=[1, 2],
+  )
+  def test_flex_flex_collisions(self, test_case, nworld):
```
quagla, line 2730: These assert only nacon > 0 / == 0. The FlexInternalCollisionTest this PR deletes contained the only tests that ran mujoco.mj_collision and compared dist/pos/normal/ids, so flex-flex now ships with no numerical parity coverage. A two-tet and a two-cloth case against mj_collision would have caught both margin issues above.

mujoco_warp/_src/io.py
```diff
@@ -1517,6 +1520,16 @@ def _eq_bodies(i):
...
           )
 
+    # flex-flex collision
+    for fj in range(fi + 1, mjm.nflex):
+      fct_j = mjm.flex_contype[fj]
+      fca_j = mjm.flex_conaffinity[fj]
+      if (fct & fca_j) or (fct_j & fca):
+        dim_i = mjm.flex_dim[fi]
+        dim_j = mjm.flex_dim[fj]
+        nnz_flex_flex = (dim_i + dim_j + 2) * 3
```
quagla: (dim_i + dim_j + 2) * 3 assumes vertex dofs. The geom-flex branch immediately above handles flex_interp != 0 explicitly; the elem-elem Jacobian for interpolated flexes expands to up to 8 node bodies per side in constraint.py (~48 columns, more with non-trivial ancestors). The demo model in the PR description is dof="trilinear", so this is the primary path. Only bites when the flex-flex term is the max, but it's an under-estimate.

mujoco_warp/_src/collision_flex.py
```diff
@@ -3333,492 +3144,559 @@ def flex_collision(m: Model, d: Data, ctx):
...
+  nworldelem = d.nworld * nelem
 
-      # Step 2: Sort
-      wp.utils.segmented_sort_pairs(
-        sap_lower.reshape((-1, nelem)),
-        sap_sort_index.reshape((-1, nelem)),
-        nworldelem,
-        sap_seg_index,
-      )
+  direction = wp.vec3(0.5935, 0.7790, 0.1235)
```
quagla: Same direction vector, same inflation, same kernel as _detect_self_flex_candidates (3227) — two full segmented sorts over nworld × nflexelem plus two sets of wp.empty per step. The sweeps differ only by the flexid1 == flexid2 test, so one sort could feed both.

mujoco_warp/_src/collision_flex.py
```diff
@@ -1772,84 +1228,257 @@ def _flex_sap_project(
...
+      flexid2 = flex_elemflexid[elem2]
 
-  e1 = elem1_global - elem_adr
-  e2 = elem2_global - elem_adr
-  elem_data_idx1 = flex_elemdataadr[flexid] + e1 * (dim + 1)
-  elem_data_idx2 = flex_elemdataadr[flexid] + e2 * (dim + 1)
-  v1_indices = _get_element_vertices(flex_elem, dim, elem_data_idx1)
-  v2_indices = _get_element_vertices(flex_elem, dim, elem_data_idx2)
+      # Self-collision: must be same flex and selfcollide enabled
+      if flexid1 != flexid2 or flex_selfcollide[flexid1] == 0:
```
quagla: C additionally requires flex_contype[f] & flex_conaffinity[f] for self-collision (engine_collision_driver.c:831). !flex_rigid is covered implicitly by the shared-body test, but the bitmask one isn't.

thowell: fixed

mujoco_warp/_src/io.py
thowell: updated

mujoco_warp/_src/collision_flex.py
thowell: fixed

mujoco_warp/_src/collision_flex.py
thowell: fixed

mujoco_warp/_src/collision_flex.py
thowell: fixed

mujoco_warp/_src/flex_test.py
thowell: fixed

mujoco_warp/_src/collision_flex.py
thowell: its intentional to have separate broadphase calls for different flex collision types in order to help avoid overflow. added a todo to investigate combining the calls. additionally, refactored this code a bit to reduce code duplication.

thowell:
@quagla thanks for the review! the comments have been addressed. 

additionally added a test for flex-flex trilinear collisions `test_flex_flex_collision_trajectory`.

note: also added `nconmax` and `njmax` as arguments to `test_data.fixture`

quagla (approved):

PR #1498, opened 2026-07-06
Remove duplicates in collision_primitive_core, import from math instead

etaoxing:


thowell (approved):
thanks @etaoxing!

etaoxing:
for for https://github.com/google-deepmind/mujoco_warp/pull/1535

PR #1501, opened 2026-07-07
Fix sparse qfrc_constraint zeroing for early-converged worlds

### LINKED ISSUE #1500: Sparse solver zeroes qfrc_constraint for worlds that converge early
When the constraint Jacobian is sparse, `_update_constraint` clears the whole `qfrc_constraint` array on every solver iteration and then only fills back in the worlds that haven't converged yet. So if a world converges before the rest of the batch, it gets cleared once and never refilled, and finishes the step at zero even though its constraints are active. The dense path doesn't do the blanket clear (it just overwrites each DOF), which is why it's fine.

This doesn't affect `qacc`, since that's already settled by the time a world converges. It only shows up if you actually read `qfrc_constraint`, e.g. force sensors going through `rne_postconstraint`. Newton happens to dodge it because the stable-fast path recomputes `qfrc_constraint` after the solve, but CG, PGS, elliptic cone, and compact have no such fallback.

Easy to see with a batch of 2048 humanoids started from perturbed states: ~2042 worlds have active constraints, but only a handful come back with a nonzero `qfrc_constraint`.

Originally reported in mjlab: mujocolab/mjlab#1086.

kevinzakka:
Fixes #1500.

### DIFF
--- a/mujoco_warp/_src/solver_test.py
+++ b/mujoco_warp/_src/solver_test.py
@@ -684,34 +684,57 @@ class SolverTest(parameterized.TestCase):
 
     self.assertTrue(total_any_changes, "no state changes detected across any keyframe")
 
-  @parameterized.parameters(SolverType.CG, SolverType.NEWTON)
-  def test_qfrc_constraint_early_convergence(self, solver_):
-    """Sparse qfrc_constraint must survive for worlds that converge before the batch."""
-    mjm, mjd, m, _ = test_data.fixture(
-      "humanoid/humanoid.xml",
-      keyframe=0,
-      overrides={"opt.jacobian": mujoco.mjtJacobian.mjJAC_SPARSE, "opt.solver": solver_},
-    )
+  def test_qfrc_constraint_early_convergence(self):
+    """Sparse qfrc_constraint must survive for a world that converges before the batch.
+
+    The sparse solve clears qfrc_constraint each iteration and only rewrites
+    unconverged worlds; a world that converges early must keep its value rather
+    than be zeroed by a later iteration's clear. Uses CG, which has no post-solve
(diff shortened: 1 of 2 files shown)


thowell:
thanks @kevinzakka! should be fixed in https://github.com/google-deepmind/mujoco_warp/pull/1493 (merged)

kevinzakka:
Hi @thowell, #1493's post-solve recovery covers the Newton stable-fast path, but the underlying sparse `zero_()` still wipes early-converged worlds for CG/PGS/elliptic/compact which have no recovery. The regression test in my PR fails on CG on top of #1493 but passes on Newton.

mujoco_warp/_src/solver_test.py
```diff
@@ -684,6 +684,35 @@ def _run_solver(d, update_fn, track=False):
 
     self.assertTrue(total_any_changes, "no state changes detected across any keyframe")
 
+  @parameterized.parameters(SolverType.CG, SolverType.NEWTON)
```
thowell: lets simplify this test to something that just tests that there are no zero values in `qfrc_constraint` when there shouldn't be after calling `solve`?

maybe something like: use `nworld=2`, provide 1 world with a converged solution and the other with an unconverged solution, run solve for 1 iteration, then 1 world is converged and 1 is not converged, check that all `qfrc_constraint` values are non-zero?

kevinzakka: Thanks Taylor, went ahead and simplified to a deterministic two-world CG setup, with the caveat that it needs staggered convergence across ≥2 iterations and a non-recovery solver like CG to actually reproduce.

thowell (approved):
thanks @kevinzakka!

PR #1504, opened 2026-07-07
Update plane texture formula to match mujoco

bd-pmorais:
I noticed that infinite plane tessellation differed between mujoco and the warp renderer. This PR brings the warp renderer to parity with mujoco by adding in the missing scale factor and offset. 

Note that for finite planes with `texuniform=false` the renderers still wont exactly match because mujoco divides by geom_size in that case but that would be a bigger change (warp doesn't have `mat_texuniform` exposed) and this covers the more common default case. 

side_by_side

### DIFF
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -78,12 +78,16 @@ def sample_texture(
   mesh_id: int,
 ) -> wp.vec3:
   uv = wp.vec2(0.0, 0.0)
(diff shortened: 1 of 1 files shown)


thowell:
@bd-pmorais thanks for contributing this fix! please sign the cla. thanks!

thowell (approved):

thowell:
thanks @bd-pmorais!

PR #1505, opened 2026-07-08
Newton fast path: rebuild the gradient on friction transitions and linesearch stagnation

adenzler-nvidia:
The stable-state fast path from #1493 skips recomputing the gradient, search direction, and Cholesky solve for worlds whose constraint states did not change during an iteration, on the premise that such a step stayed exactly quadratic and the cached quantities remain valid up to a scalar. Two kinds of change escaped that detection, so the affected worlds kept iterating with a stale gradient.

First, friction transitions. The fast path watched only the counter of rows whose QUADRATIC state flipped, which is the right set for the incremental Hessian update but not for the gradient: a frictionloss row crossing between its two linear force zones changes its force by 2*frictionloss while the quadratic flag and the Hessian stay put. The stale gradient desyncs qfrc_constraint from J^T efc_force and can leave worlds unable to converge. The fix tracks any per-row state change in a second per-world counter and gates all gradient/search/solve skips on it, while the quadratic-flip list keeps driving the incremental Hessian update; worlds whose only transitions were between the friction linear zones rebuild the gradient but reuse the cached factorization. The two counters coincide on models without friction rows.

Second, linesearch stagnation. After skipped iterations the search direction is still the one computed at the last rebuild, and the linesearch's float32 bracketing arithmetic cannot resolve steps below eps times its gross derivative magnitude over the ray curvature. A world whose tolerance lies below that floor accepts noise-sized steps forever and runs to the iteration cap, and since the solver iterates under one conditional graph, that world holds the whole batch for all 100 iterations. This could not happen before the fast path existed, because rebuilding every iteration re-anchors the arithmetic at the current gradient scale. The linesearch now computes its per-world noise floor (eps * |q1| / p''(0), never below 8 ulps of the unit step, with the gross derivative magnitude bounded via Cauchy-Schwarz from sums the linesearch already reduces, so no per-row work is added) and flags a world whose accepted step falls below it; the flag counts as a state change, so the next iteration rebuilds from the actual constraint state and either resumes real progress or terminates through the normal convergence criteria. The trigger can only add a rebuild, never end a solve early. The easiest reproduction is friction plus disabled warmstart at default tolerance: every solve then starts far from the solution, and without the fix the friction-chain regression test spins to the cap on 79 of 100 steps (2 of 100 with it, matching what full per-iteration rebuilds also produce through ordinary friction chatter). The tracking is compiled out of configurations that never skip rebuilds (elliptic, CG), whose linesearch is unchanged.

The regression test rolls out a friction chain with warmstart disabled and asserts both properties: qfrc_constraint stays consistent with J^T efc_force each step (fails at relative error 1.07 without the transition fix), and the number of cap-hitting steps stays bounded. The change also renames the change counters so the two sets are self-documenting, and fuses the per-iteration counter zeroing into one kernel.

Benchmarks on RTX PRO 6000: total GPU kernel time under nsys (CUDA 12.9) over 1000-step mjwarp-testspeed rollouts, main da4d7ba6 vs this PR measured back to back per benchmark:

| benchmark | total GPU time | solver iterations/step |
|---|---|---|
| three_humanoids | 4854 -> 4755 ms (-2.0%) | 2.58195 -> 2.58156 |
| unitree_g1_flat | 2223 -> 2243 ms (+0.9%) | 4.36732 -> 4.36736 |
| humanoid | 964 -> 965 ms (+0.05%) | 1.94396 -> 1.94421 |
| aloha_pot | 2793 -> 2732 ms (-2.2%) | 1.44555 -> 1.44552 |
| aloha_clutter | 20756 -> 20786 ms (+0.1%) | 1.06712 -> 1.06712 |

Solver iteration statistics are identical to four decimals everywhere. The aloha_pot row is a built-in control: its two builds compile identical kernels (elliptic disables the fast path), so its -2.2% is pure per-pair scatter — on trajectory-sensitive benchmarks the conditional solver graph re-rolls its iteration count from run to run (pot executed the linesearch 6833 vs 6702 times), and total GPU time inherits that, which also covers three_humanoids' -2.0%. The trajectory-stable benchmarks read +0.05% (humanoid) and +0.1% (aloha_clutter), and unitree_g1_flat's repeated pairs across the day (-0.4%, -0.1%, +0.9%) bracket zero. The only per-kernel residuals attributable to the change are +0.1us on the linesearch median (noise-floor epilogue) and +0.4us on the incremental-Cholesky gate median (the second counter read).

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -97,6 +97,7 @@ def _create_solver_context(m: types.Model, d: types.Data) -> SolverContext:
     hfactor=wp.empty((nworld, nv_pad, nv_pad), dtype=float) if alloc_hfactor else wp.empty((nworld, 0, 0), dtype=float),
     changed_efc_ids=wp.empty((nworld, njmax), dtype=int) if alloc_h else wp.empty((nworld, 0), dtype=int),
     changed_efc_count=wp.empty((nworld,), dtype=int) if alloc_h else wp.empty((0,), dtype=int),
+    state_changed_count=wp.empty((nworld,), dtype=int) if alloc_h else wp.empty((0,), dtype=int),
   )
 
 
@@ -1553,6 +1554,7 @@ def _update_constraint_efc(track_changes: bool):
     # Out:
     changed_ids_out: wp.array2d[int],
     changed_count_out: wp.array[int],
+    state_changed_count_out: wp.array[int],
   ):
     worldid, efcid = wp.tid()
 
@@ -1562,9 +1564,9 @@ def _update_constraint_efc(track_changes: bool):
     if ctx_done_in[worldid]:
       return
 
-    # Read old QUADRATIC status before overwriting
+    # Read old state before overwriting
     if wp.static(TRACK_CHANGES):
-      old_quad = efc_state_out[worldid, efcid] == types.ConstraintState.QUADRATIC.value
+      old_state = efc_state_out[worldid, efcid]
 
     efc_D = efc_D_in[worldid, efcid]
     Jaref = ctx_Jaref_in[worldid, efcid]
@@ -1630,10 +1632,15 @@ def _update_constraint_efc(track_changes: bool):
     efc_state_out[worldid, efcid] = new_state
 
     if wp.static(TRACK_CHANGES):
+      old_quad = old_state == types.ConstraintState.QUADRATIC.value
       new_quad = new_state == types.ConstraintState.QUADRATIC.value
       if old_quad != new_quad:
         idx = wp.atomic_add(changed_count_out, worldid, 1)
         changed_ids_out[worldid, idx] = efcid
+      # LINEARNEG <-> LINEARPOS friction transitions change the force without
+      # changing the quadratic flag (or H); the fast path must still see them.
+      if old_state != new_state:
+        wp.atomic_add(state_changed_count_out, worldid, 1)
 
   return kernel
 
@@ -1858,12 +1865,12 @@ def _update_constraint(
     _update_constraint_efc(track_changes),
     dim=(d.nworld, d.njmax),
     inputs=efc_inputs,
-    outputs=[d.efc.force, d.efc.state, ctx.changed_efc_ids, ctx.changed_efc_count],
+    outputs=[d.efc.force, d.efc.state, ctx.changed_efc_ids, ctx.changed_efc_count, ctx.state_changed_count],
   )
 
   # qfrc_constraint = efc_J.T @ efc_force. Fast-path worlds with no state flips
   # skip the rebuild; the public value is recovered after the solve.
-  changed = ctx.changed_efc_count if stable_fast else d.nefc
+  changed = ctx.state_changed_count if stable_fast else d.nefc
   if m.is_sparse:
     d.qfrc_constraint.zero_()
     wp.launch(
@@ -2749,6 +2756,7 @@ def _update_gradient_cholesky_blocked_skip_unchanged(tile_size: int, matrix_size
     ctx_grad_in: wp.array3d[float],
     ctx_h_in: wp.array3d[float],
     changed_count_in: wp.array[int],
+    state_changed_count_in: wp.array[int],
     ctx_hfactor: wp.array3d[float],
     # Out:
     ctx_Mgrad_out: wp.array3d[float],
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -52,8 +52,10 @@ def create_inverse_context(m: types.Model, d: types.Data) -> InverseContext:
     Jaref=wp.empty((nworld, njmax), dtype=float),
     search_dot=wp.empty((nworld,), dtype=float),
     done=wp.empty((nworld,), dtype=bool),
-    changed_efc_ids=wp.empty((nworld, 0), dtype=int),
-    changed_efc_count=wp.empty((0,), dtype=int),
+    quad_changed_ids=wp.empty((nworld, 0), dtype=int),
+    quad_changed_count=wp.empty((0,), dtype=int),
+    state_changed_count=wp.empty((0,), dtype=int),
+    ray_exhausted=wp.empty((0,), dtype=bool),
```
thowell: would it make sense to call this something like 'ls_exhausted' or 'step_size_exhausted'?

mujoco_warp/_src/solver_test.py
```diff
@@ -685,6 +685,69 @@ def _run_solver(d, update_fn, track=False):
...
+    m = mjw.put_model(mjm)
+    nworld = 64
+    d = mjw.put_data(mjm, mjd, nworld=nworld)
+    rng = np.random.default_rng(3)
+    qvel = rng.uniform(-8.0, 8.0, size=(nworld, mjm.nv)).astype(np.float32)
+    wp.copy(d.qvel, wp.array(qvel, dtype=float))
+
+    max_rel = 0.0
+    cap_steps = 0
+    for _ in range(100):
```
thowell: can we reduce the number of steps to 10 or 20 in order to reduce the amount of time the test takes to run?

mujoco_warp/_src/types.py
```diff
@@ -2354,16 +2356,21 @@ class SolverContext:
   quad: wp.array2d[wp.vec3]
   alpha: wp.array[float]
   grad_scale: wp.array[float]
+  # rows with any piecewise-state change this iteration (gradient rebuild needed)
```
thowell: lets remove this comment

mujoco_warp/_src/types.py
```diff
@@ -2354,16 +2356,21 @@ class SolverContext:
...
+  state_changed_count: wp.array[int]
   improvement: wp.array[float]
+  # accepted linesearch step was below its arithmetic noise floor (stale ray exhausted)
```
thowell: lets remove this comment

mujoco_warp/_src/types.py
```diff
@@ -2354,16 +2356,21 @@ class SolverContext:
...
   prev_grad: wp.array2d[float]
   prev_Mgrad: wp.array2d[float]
   beta: wp.array[float]
   beta_den: wp.array[float]
   h: wp.array3d[float]
   hfactor: wp.array3d[float]
   # Incremental Hessian update (Newton only)
-  changed_efc_ids: wp.array2d[int]
-  changed_efc_count: wp.array[int]
+  # rows whose QUADRATIC membership flipped this iteration (Hessian delta needed)
```
thowell: lets remove this comment (and the line above)

thowell (approved):

mujoco_warp/_src/solver.py
adenzler-nvidia: Renamed to ls_exhausted.

mujoco_warp/_src/solver_test.py
adenzler-nvidia: Reduced to 20 steps with the cap bound scaled to 5; verified it still fails without the fix (rel 1.07) and runs in ~9s.

mujoco_warp/_src/types.py
adenzler-nvidia: Removed.

mujoco_warp/_src/types.py
adenzler-nvidia: Removed.

mujoco_warp/_src/types.py
adenzler-nvidia: Removed both lines.

adenzler-nvidia:
Applied all five review comments in 2c92d02d: renamed ray_exhausted to ls_exhausted, removed the context field comments in types.py, and shortened the friction-transition regression test to 20 steps with the cap bound scaled to 5 (verified it still fails without the fix). Nothing was skipped.

thowell:
lgtm, thanks!

PR #1506, opened 2026-07-08
Fuse blocked Cholesky forward solve

adenzler-nvidia:
The blocked Cholesky path currently factors the Hessian and then performs a separate forward triangular solve for the gradient. This change places the right-hand side in an augmented matrix column so the existing panel factorization computes the forward solution, leaving only the backward solve. Cached-factor worlds continue to use the ordinary solve-only path. Dense Newton allocations reserve one additional padded degree of freedom when needed, so dimensions at a tile boundary pay for one additional tile. Focused tests cover padding, factor output, the combined solution, and reuse of the resulting factor with a different right-hand side.

Benchmarks used 8,192 worlds on an RTX PRO 6000 Blackwell.

| Benchmark | Baseline | Augmented | Improvement |
| --- | ---: | ---: | ---: |
| Isolated factor/solve, padded size 48 | 71.14 µs | 68.02 µs | 4.6% |
| Isolated factor/solve, padded size 96 | 340.55 µs | 323.67 µs | 5.2% |
| G1 testspeed GPU runtime | 2.620 s | 2.531 s | 3.4% |
| G1 testspeed solve | 149.83 µs | 141.95 µs | 5.3% |
| Three-humanoid testspeed GPU runtime | 2.836 s | 2.810 s | 0.9% |
| Three-humanoid testspeed solve | 495.97 µs | 487.93 µs | 1.6% |

Solve-only performance was neutral. Repeated clean three-humanoid pairs varied from 0.1-0.9% lower total runtime and 0.6-1.6% lower solve time.


thowell (approved):
nice!

PR #1507, opened 2026-07-09
Enable the stable-state fast path under the compact sleep solve

adenzler-nvidia:
The stable-state fast path from #1493 was excluded under the compact sleep solve, so enabling SLEEP silently gave up the fast-path win, and the sleep path also kept the qfrc_constraint defect that #1493 fixed elsewhere (the per-iteration zeroing wipes the public value for worlds that converge early). The exclusion turns out to be unnecessary: the compact solve is self-contained in compacted coordinates -- the dof mapping is fixed within a solve, the solver context is rebuilt every solve, and the shallow-replaced data maps qfrc_constraint to the compacted buffer that _compact_scatter reads back -- so the fast path's within-solve staleness needs no special handling. The first commit drops the exclusion, extends the post-solve qfrc_constraint recovery to the compact path, and collapses the now-redundant _stable_fast predicate into _use_incremental. The ray-exhaustion rebuild from #1505 covers the wake transients that originally motivated caution: freshly woken islands start far from the solution, which is exactly the stall-prone regime.

The second commit extends what the fast path skips. The linesearch recomputes mv = M @ search and jv = J @ search every iteration, but both are pure functions of the search direction, and M and J are constant within a solve -- so worlds the fast path left on their ray can reuse last iteration's values. _solve_search_update is the kernel that decides whether a world keeps its search vector, so it records that fact (together with done, which subsumes the skip) in a per-world search_unchanged mask, and the mv/jv launches consume the mask through their existing per-world skip mechanism; solve init clears it since a new solve computes a new search. No public API changes.

Originally stacked on #1505, which has since merged; upstream/main is merged into this branch, so the diff is now just this PR's delta. That merge also brought in #1509, which replaced the compact solve's dense gathered M and J products with full-coordinate sparse walks -- shrinking the per-iteration cost pool the mv/jv reuse targets, so the numbers below are smaller than this PR's original round.

Correctness: qfrc_constraint stays consistent with J^T efc_force at 1e-6-class residuals across all worlds on a sleeping clutter scene (pyramidal override, 2048 worlds) and on unitree_g1_flat, humanoid, and three_humanoids with SLEEP forced on (8192 worlds each, sparse and dense models). After the upstream merge the full test suite passes, and the sleep solver's iteration counts match the non-sleep solver's (humanoid with SLEEP forced on: mean niter 1.375 vs 1.376 non-sleep over 8192 x 1000 steps). The friction-chain fast-path regression test now also runs under the compact solve (SLEEP + nvmax).

Benchmarks on RTX PRO 6000: total GPU kernel time under nsys (CUDA 12.9, graph-node kernel tracing, profiled range excludes JIT and warmup), 1000-step rollouts through the testspeed harness (the clutter replay is 2751 steps), current upstream/main vs this branch measured as mirrored back-to-back pairs per configuration; entries are medians of within-pair deltas across 2-5 pairs. Standard configurations, where the fast path already runs natively and mean solver iterations are ~1.4, so there is rarely a second iteration to reuse in:

| benchmark | total GPU time |
|---|---|
| three_humanoids | -0.1% |
| humanoid | +0.5% |
| unitree_g1_flat | -0.3% |
| aloha_pot | 0.0% |

aloha_pot is elliptic, where this change compiles inert code; its five pairs spanned -3.3% to +3.7%, which reads as this host's run-to-run noise floor rather than the change.

Configurations where the fast path newly operates (SLEEP forced on, everything awake, so every solve runs compacted):

| config | total GPU time | steps/s |
|---|---|---|
| three_humanoids + SLEEP | -1.2% | ~0% |
| humanoid + SLEEP | -2.0% | +2.5% |
| unitree_g1_flat + SLEEP | -1.6% | +1.4% |
| aloha_clutter pyramidal (sleeping) | +0.1% | +0.3% |

The wins concentrate where the fast path skips work: on humanoid + SLEEP the Newton solve from the cached factor (-20%), gradient rebuild (-13%), qfrc scatter (-10%), and mv (-6%); on unitree_g1 + SLEEP and three_humanoids + SLEEP the cached-factor solve (-15% and -18%) and the qfrc scatter (-25% on three_humanoids). steps/s on the free-running three_humanoids does not resolve under trajectory divergence (samples span -1.3% to +2.4%); the GPU kernel time column is the controlled measurement. The sleeping clutter case is parity as expected: with most trees asleep the compacted systems are small and the solver is a minor share of the step. As before, for scenes where nothing sleeps, running without SLEEP remains much faster, with or without this PR.


mujoco_warp/_src/solver.py
```diff
@@ -92,6 +92,7 @@ def _create_solver_context(m: types.Model, d: types.Data) -> SolverContext:
     grad_scale=wp.empty((nworld,), dtype=float),
     improvement=wp.empty((nworld,), dtype=float),
     ls_exhausted=wp.zeros((nworld,), dtype=bool),
+    search_unchanged=wp.zeros((nworld,), dtype=bool),
```
thowell: can we use `wp.empty`?

mujoco_warp/_src/support.py
```diff
@@ -39,7 +39,7 @@ def next_act(
   # Model:
   opt_timestep: float,  # kernel_analyzer: ignore
   actuator_dyntype: int,  # kernel_analyzer: ignore
-  actuator_dynprm: vec10,  # kernel_analyzer: ignore
+  actuator_dynprm: vec10f,  # kernel_analyzer: ignore
```
thowell: is this change necessary?

mujoco_warp/_src/types.py
```diff
@@ -2388,6 +2388,8 @@ class SolverContext:
   state_changed_count: wp.array[int]
   improvement: wp.array[float]
   ls_exhausted: wp.array[bool]
+  # search vector kept from last iteration (or done): mv/jv are still valid
```
thowell: lets remove this comment

mujoco_warp/_src/solver_test.py
```diff
@@ -889,6 +894,69 @@ def test_fast_path_friction_transitions(self):
...
+        </body>
+      </body>
+    </body>
+  </worldbody>
+</mujoco>
+"""
+
+
+class SolverFastPathTest(parameterized.TestCase):
+  def test_fast_path_friction_transitions(self):
```
thowell: we should test the compact solver path by enabling sleeping and setting nvmax?

mujoco_warp/_src/solver.py
adenzler-nvidia: Done -- the solve-init clear initializes it, so the allocation does not need the memset.

mujoco_warp/_src/support.py
adenzler-nvidia: No -- leftover from an intermediate rebase during development. Reverted, support.py now matches main exactly.

mujoco_warp/_src/types.py
adenzler-nvidia: Done.

mujoco_warp/_src/solver_test.py
adenzler-nvidia: Done -- parameterized the test over the compact path (SLEEP + nvmax). The upstream merge had also left a stale duplicate of this test class shadowing the reviewed 20-step version; removed it.

adenzler-nvidia:
All comments addressed: search_unchanged is now allocated with wp.empty, the stray vec10f rename is reverted so support.py matches main again, the field comment is removed, and the fast-path regression test is parameterized to also run under the compact solve (SLEEP + nvmax). While doing that I found the upstream merge had left a duplicate of the test class shadowing the reviewed 20-step version -- the stale copy is removed.

The PR description is also updated: upstream/main (now including #1509) is merged into the branch and the benchmarks are re-run against current main. The wins are smaller than the original round because #1509's full-coordinate sparse M and J products removed most of the dense per-iteration work the mv/jv reuse targeted: humanoid + SLEEP -2.0% GPU time / +2.5% steps/s, unitree_g1 + SLEEP -1.6% / +1.4%, three_humanoids + SLEEP -1.2%, sleeping clutter and the standard configurations at parity.

thowell (approved):

PR #1503, opened 2026-07-07
Add numerical robustness improvements to GJK

kbayes:
The main changes are:

1. Do a direction fallback when norm becomes too small,
 2. Recorrect barycentric coordinates to be exactly a partition of unity, and
 3. Some exit condition improvements.

*NOTE*: One of the tests was actually incorrect and this PR corrects it.

Benchmark Performance (a little regression is expected due to the support fallback):

 | Benchmark / Task | Before | After | Narrowphase Change (%) |
  | :--- | :--- | :--- | :--- |
  | **aloha_pot** | 81.38836 | 85.13249 | +4.60% |
  | **aloha_sdf** | 39.90975 | 40.09000 | +0.45% |
  | **aloha_cloth** | 3839.18790 | 4036.44060 | +5.14% |
  | **aloha_clutter[0]** | 1332.36986 | 1373.55125 | +3.09% |
  | **aloha_clutter[1]** | 63.47256 | 65.69284 | +3.50% |
  | **cloth** | 45.53600 | 45.92000 | +0.84% |
  | **franka_emika_panda** | 0.04578 | 0.04756 | +3.89% |
  | **humanoid** | 0.18512 | 0.18262 | -1.35% |
  | **three_humanoids** | 0.18125 | 0.18362 | +1.31% |
  | **myoarm** | 8.98313 | 9.50987 | +5.86% |
  | **unitree_g1_flat** | 0.20450 | 0.17312 | -15.34% |
  | **unitree_g1_hfield** | 209.68875 | 214.97988 | +2.52% |

PR #1509, opened 2026-07-09
Use the full-coordinate sparse M and J in the compact sleep solve

adenzler-nvidia:
The compact sleep solve gathered M and J into dense nvmax_pad-sized blocks (cM, cJ) and ran dense products and assemblies against them, so every world paid worst-case padded work regardless of how many DOFs are awake: with everything awake the padding multiplies real work, and mostly-sleeping worlds still pay nvmax-wide dense operations for a handful of active DOFs, since nvmax is sized for the worst world in the batch.

This series moves the compact Newton path onto the original full-coordinate sparse structures instead. Kernels launch at compact width and follow the compaction maps -- cdof_dof selects a thread's global DOF, and row columns remap back through dof_cdof -- reading the static sparse M and J structures the pipeline already maintains, and skipping unmapped entries exactly as the gathers did (M can carry cross-tree entries via tendon armature; a sleeping column's gathered value is zero). Work becomes proportional to actual nonzeros instead of padded dense widths. The dense Hessian and its blocked factorization at compact width remain -- that is what compaction is for -- including the identity padding for per-world unused rows.

The commits, in order: the M products (mv = M @ search in the linesearch, the compacted Ma = M @ qacc at solve init); the J products (jv, the compacted qfrc_constraint, the Jaref initialization); the Hessian assembly (h initialized from the sparse M through M_elemid, JTDAJ added by the grouped sparse assembly writing remapped indices); the incremental Hessian update through the same maps, which removes the last consumer of the gathered cJ on this path so its gather is skipped; and finally a pure refactor folding the compact variants into the native kernels behind a compact factory static, so both pipelines share one implementation and the native variants compile to the same code as before. Dense full models keep the gathered paths (they have no sparse J structures); elliptic and CG keep the gathered assembly. No public API changes; the compact solver context carries references to the full-coordinate Model and Data.

Correctness, re-verified after every commit: qfrc_constraint stays consistent with J^T efc_force at 1e-7-class residuals on three_humanoids and humanoid with SLEEP forced on (8192 worlds; sparse and dense full models), and bit-exact on sleeping clutter scenes in pyramidal and elliptic configurations (2048 worlds); solver iteration counts are unchanged everywhere.

Benchmarks on RTX PRO 6000 (nsys total GPU kernel time, 1000-step mjwarp-testspeed rollouts, main vs this branch, mirrored pairs in one session):

| config | total GPU time | steps/s |
|---|---|---|
| three_humanoids + SLEEP (all awake, nvmax=81) | -42% | 582K -> 1.01M (+73%) |
| sleeping clutter, pyramidal (~92% of DOFs asleep) | -19% | 232K -> 308K (+33%) |
| fully settled clutter, no actuation | | 682K -> 791K (+16%) |

Even fully-settled scenes win: settling micro-wakes keep small islands active each step, and the work now scales with them rather than with nvmax. The sleeping case's largest single win is the incremental Hessian update, whose dense variant paid O(nvmax_pad^2) per changed row across every wake transient. Native configurations are unchanged (the compact factory statics fold away; verified at parity).

Unrelated: two collision_gjk_test failures appear in local GPU runs of this branch; they bisect to b5edf8ec (#1503) on main.


mujoco_warp/_src/solver.py
```diff
@@ -3825,6 +3913,73 @@ def _init_compact_inertia(
...
+
+def _mul_m_compact_aware(m: types.Model, d: types.Data, ctx: SolverContext | InverseContext, res, vec, skip):
+  """M @ vec: full-coordinate sparse walk under compact, support.mul_m natively."""
+  dfull = ctx.compact_d_full
+  if dfull is not None:
+    wp.launch(
+      _mul_m_sparse_compact,
+      dim=(d.nworld, m.nv),
+      inputs=[
+        m.M_mulm_rowadr,
```
thowell: should we get `mfull = ctx.compact_m_full` and use `mfull` instead of `m` here and in the next 2 lines for clarity?

thowell (approved):
great stuff! thanks @adenzler-nvidia

mujoco_warp/_src/solver.py
adenzler-nvidia: Done — added the `mfull` alias in 8879f24f.

PR #1511, opened 2026-07-10
Fix GJK termination and distance reporting for float32

adenzler-nvidia:
Since #1503, test_sphere_mesh_margin and test_box_box_ccd fail on CUDA (7.1mm distance error and lost multiccd contacts respectively). CI stays green because the runners are CPU-only. This PR fixes the failures; the mechanism was isolated with per-exit instrumentation on a sphere-mesh case with an exact analytic answer.

Three things interact. First, the no-progress check compares x_next to x_k against mjMINVAL (1e-15), which is calibrated for float64: in float32 it sits below one ulp of any relevant magnitude, so it can only fire on bitwise equality. Since the Frank-Wolfe termination runs with epsilon = 0 for discrete pairs, this check is the only exit for separated discrete geoms, and on CUDA (where FMA contraction leaves a one-ulp oscillation at the fixed point) those pairs run all the way to iteration exhaustion. The CPU backend happens to land on an exact fixed point and terminates at iteration 4, which is why the failures are GPU-only. The check now scales the caller's tolerance by the iterate's magnitude.

Second, on exhaustion the reported distance comes from xnorm, which is assigned before x_k's final update, so it is one iteration staler than the witness points. The distance is now recomputed from the final iterate, keeping the exact zero on the origin-enclosed exit (the loop forces xnorm = 0 there, and the dist > tolerance gate in front of EPA relies on intersecting pairs not reporting a cancellation residual).

Third, the barycentric renormalization added in #1503 redistributes rounding residuals into the largest coordinate every iteration. As far as I can tell the coordinates already sum to one by construction (lambda_i = C4_i / m_det with m_det the cofactor sum), and MuJoCo's subdistance doesn't renormalize, so the block only perturbs lambda at rounding level; doing that every iteration keeps the iterate jiggling, the Frank-Wolfe gap never meets its tolerance, and the sphere-mesh case runs to exhaustion (35 iterations instead of 4). It also isn't neutralized by the other two fixes: with both in place, restoring the renormalization still fails two tests on CUDA. This PR removes it, but I may be missing the case that motivated it, so if there is one, I'm happy to rework this part.

The direction heuristic and termination changes from #1503 are kept. All 24 collision_gjk tests and the full suite pass on both CUDA and CPU, including the MJC-aligned test_box_box_max2 value.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -273,8 +273,12 @@ def _linear_combine(n: int, scl: wp.vec4, mat: mat43) -> wp.vec3:
 
 
 @wp.func
-def _almost_equal(v1: wp.vec3, v2: wp.vec3) -> bool:
-  return wp.abs(v1[0] - v2[0]) < MINVAL and wp.abs(v1[1] - v2[1]) < MINVAL and wp.abs(v1[2] - v2[2]) < MINVAL
+def _almost_equal(v1: wp.vec3, v2: wp.vec3, tolerance: float) -> bool:
+  # no-progress check at the caller's tolerance: mjMINVAL (1e-15) is calibrated
+  # for float64 and can only fire on bitwise equality in float32, making
+  # termination depend on reaching an exact fixed point
+  tol = tolerance * wp.max(1.0, wp.max(wp.abs(v1[0]), wp.max(wp.abs(v1[1]), wp.abs(v1[2]))))
+  return wp.abs(v1[0] - v2[0]) < tol and wp.abs(v1[1] - v2[1]) < tol and wp.abs(v1[2] - v2[2]) < tol
 
 
 @wp.func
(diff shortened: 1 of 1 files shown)


kbayes:
@adenzler-nvidia How are you repro these failures? I can't repro them.

adenzler-nvidia:
The unittests are failing on latest main with my RTX 6000 Pro Blackwell, it looks like part of this is compiler and GPU-dependent.

kbayes:
@adenzler-nvidia when I was doing a stack trace, I realized there was actually a bug in another exit criteria that should have caught it. Can you see if #1514 fixes your breakages?

kbayes:
@adenzler-nvidia I'm in the process of getting a RTX 6000 Pro Blackwell, once that's through, I'll get back to you.

adenzler-nvidia:
Superseded by #1514, which covers all three fixes — verified all 24 collision_gjk tests pass on CUDA with it.

PR #1512, opened 2026-07-10
Preserve small elliptic line-search deltas

adenzler-nvidia:
Newton line search compares candidate costs against the cost at alpha zero. For elliptic contacts, this was implemented by evaluating both absolute costs independently in float32 and then subtracting them. When the absolute cone cost is large, a valid improvement can be smaller than its float32 unit in the last place, causing the subtraction to round to zero. The line search can then treat an improving step as unchanged, which caused unstable convergence in the friction-ramp scenario.

Compute elliptic cost changes directly from the piecewise cone objective instead of subtracting nearly equal absolute values. The alpha-zero evaluation classifies the reference zone once and returns the reference cost, tangential norm, cone residual, and constraint state. Within the quadratic zone, the change is `delta_C = alpha * (q1 + alpha * q2)`. Within the middle cone, the tangential norm change is rationalized as `delta_T = delta_T_squared / (T + T0)`, followed by `delta_r = alpha * v0 - mu * delta_T` and `delta_C = 0.5 * dm * delta_r * (2 * r0 + delta_r)`.

Crossings between the middle cone and quadratic zone use the exact identity `Q - C = 0.5 * dm * (mu * N + T)^2`. A cone-to-quadratic step adds this gap at alpha zero to the shifted quadratic, while a quadratic-to-cone step subtracts the gap at the candidate point. A step from the satisfied zone into the quadratic zone evaluates the candidate quadratic as a scaled sum of squares. These forms cover the full piecewise surface without subtracting large absolute costs.

The candidate gradient and Hessian continue to use the original piecewise elliptic-cone equations. The alpha-zero derivative path reuses the same classification, and the batched three-alpha path computes the reference terms once. Computations used only by limit and friction constraints remain inside those branches so elliptic contacts and equality constraints avoid unnecessary arithmetic.

The regression test uses binary-exact inputs whose absolute costs round to the same float32 value while their true changes are `0.5` or `-0.5`. It covers same-zone middle-cone and quadratic evaluation, both cone/quadratic transition directions, and the satisfied-to-quadratic path, together with the expected gradients and Hessians. The solver tests and Newton friction-ramp regression continue to pass, and fixed-work profiling shows no line-search performance regression.

Performance was measured in a fresh A/B against merge base `eb35c2d8` and PR head `ca5113fa`. Values are medians of three interleaved runs, with the GPU idle before every sample. All workloads used their recorded trajectories (`lift_pot.npz`, `pick_clutter.npz`, and `shuffle_dance.npz`), and the Unitree workloads explicitly set `opt.cone=elliptic`. All configured worlds converged.

| Benchmark | Worlds | Base step (us) | PR step (us) | Step change | Base solve (us) | PR solve (us) | Solve change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| `aloha_pot` | 8192 | 355.1 | 351.1 | -1.12% | 149.8 | 145.5 | -2.86% |
| `aloha_clutter` | 2048 | 4157.7 | 3898.0 | -6.25% | 1636.8 | 1488.8 | -9.04% |
| `unitree_g1_flat` | 8192 | 544.6 | 536.2 | -1.54% | 406.8 | 399.6 | -1.78% |
| `unitree_g1_hfield` | 8192 | 765.7 | 777.3 | +1.52% | 461.5 | 459.5 | -0.43% |

The solver time improves in all four workloads. The 1.52% hfield end-to-end change is outside the solver and within the observed total-step variability.

The flaky Newton ANymal elliptic reset failure was also reproduced against the exact Newton PR revision from [this CI job](https://github.com/newton-physics/newton/actions/runs/29086153345/job/86340621655). The MuJoCo Warp merge base failed 7 of 30 runs at the same post-reset 100-iteration assertion, while the PR head passed 30 of 30 runs under the same Newton revision and dependencies. Newton's locked `mujoco-warp==3.10.0.1` release reproduced the failure in 1 of 10 runs.

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -260,97 +260,52 @@ def _eval_frictionloss_pt_3alphas(
 
 
 @wp.func
-def _eval_elliptic(
-  # In:
-  mu: float,
-  quad: wp.vec3,
-  quad1: wp.vec3,
-  quad2: wp.vec3,
-  alpha: float,
-) -> wp.vec3:
-  u0 = quad1[0]
-  v0 = quad1[1]
-  uu = quad1[2]
-  uv = quad2[0]
-  vv = quad2[1]
-  dm = quad2[2]
-
-  # compute N, Tsqr
-  N = u0 + alpha * v0
-  Tsqr = uu + alpha * (2.0 * uv + alpha * vv)
-
-  # no tangential force: top or bottom zone
-  if Tsqr <= 0.0:
-    # bottom zone: quadratic cost
-    if N < 0.0:
-      return _eval_pt(quad, alpha)
-
-    # top zone: nothing to do
-  # otherwise regular processing
-  else:
-    # tangential force
-    T = wp.sqrt(Tsqr)
-
-    # N >= mu * T : top zone
-    if N >= mu * T:
-      # nothing to do
-      pass
-    # mu * N + T <= 0 : bottom zone
-    elif mu * N + T <= 0.0:
-      return _eval_pt(quad, alpha)
-
-    # otherwise middle zone
-    else:
-      # derivatives
-      N1 = v0
-      T1 = (uv + alpha * vv) / T
-      T2 = vv / T - (uv + alpha * vv) * T1 / (T * T)
-
-      # add to cost
-      cost = wp.vec3(
-        0.5 * dm * (N - mu * T) * (N - mu * T),
-        dm * (N - mu * T) * (N1 - mu * T1),
-        dm * ((N1 - mu * T1) * (N1 - mu * T1) + (N - mu * T) * (-mu * T2)),
-      )
-
-      return cost
-
-  return wp.vec3(0.0, 0.0, 0.0)
-
-
-@wp.func
 def _eval_elliptic_reference(
   # In:
   mu: float,
   quad: wp.vec3,
   quad1: wp.vec3,
   quad2: wp.vec3,
-) -> tuple[float, float, int]:
+) -> tuple[float, float, float, int]:
   u0 = quad1[0]
   uu = quad1[2]
   dm = quad2[2]
 
   if uu <= 0.0:
     if u0 < 0.0:
-      return quad[0], 0.0, int(types.ConstraintState.QUADRATIC)
-    return 0.0, 0.0, int(types.ConstraintState.SATISFIED)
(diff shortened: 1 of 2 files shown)


yuvaltassa:
Seems sensible! 
I am testing this in C and will get back to you 🙂

thowell:
https://github.com/google-deepmind/mujoco/commit/892d889793a19fdf2ed06d10d0de23a323223394

mujoco_warp/_src/solver.py
```diff
@@ -260,73 +260,88 @@ def _eval_frictionloss_pt_3alphas(
...
-      N1 = v0
-      T1 = (uv + alpha * vv) / T
-      T2 = vv / T - (uv + alpha * vv) * T1 / (T * T)
+@wp.func
+def _eval_elliptic_alpha_zero(mu: float, quad: wp.vec3, quad1: wp.vec3, quad2: wp.vec3) -> wp.vec3:
+  cost0, T0, r0, state0 = _eval_elliptic_reference(mu, quad, quad1, quad2)
+  if state0 == int(types.ConstraintState.QUADRATIC):
+    return _eval_pt(quad, 0.0)
+  if state0 == int(types.ConstraintState.CONE):
+    T1 = quad2[0] / T0
```
thowell: precompute `1.0 / T0`?

mujoco_warp/_src/solver.py
```diff
@@ -335,21 +350,42 @@ def _eval_elliptic_cost(
...
   else:
     T = wp.sqrt(Tsqr)
     if N >= mu * T:
       pass
     elif mu * N + T <= 0.0:
-      return _eval_cost(quad, alpha)
+      return _eval_elliptic_quadratic_shifted(mu, quad, alpha, N, Tsqr, u0, T0, dm, state0)
     else:
-      return 0.5 * dm * (N - mu * T) * (N - mu * T)
+      T1 = (uv + alpha * vv) / T
```
thowell: precompute `1.0 / T`?

thowell (approved):
thanks @adenzler-nvidia!

mujoco_warp/_src/solver.py
adenzler-nvidia: Done in b47831da.

mujoco_warp/_src/solver.py
adenzler-nvidia: Done in b47831da.

PR #1514, opened 2026-07-10
Fix bug with GJK exit criteria with norm

kbayes:
Exit criteria was using the Freshman's dream. This is now fixed which allows for some extra simplification. 

Also remove recomputing lambda values due to regression.

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -701,8 +701,6 @@ def gjk(
 
     # remove vertices from the simplex no longer needed
     n = int(0)
-    lmbdamax = float(-1.0)
-    imax = int(-1)
     for i in range(4):
       if lmbda[i] == 0.0:
         continue
@@ -713,8 +711,6 @@ def gjk(
       simplex_index1[n] = simplex_index1[i]
       simplex_index2[n] = simplex_index2[i]
       lmbda[n] = lmbda[i]
-      imax = wp.where(lmbda[n] > lmbdamax, n, imax)
-      lmbdamax = wp.where(lmbda[n] > lmbdamax, lmbda[n], lmbdamax)
       n += int(1)
 
     # SHOULD NOT OCCUR
@@ -725,14 +721,6 @@ def gjk(
     if n == 4:
       xnorm = 0.0
       break
-
-    # clamp barycentric coordinates to an exact partition of unity
-    lmbda[imax] = 1.0
-    acc = float(0.0)
-    for i in range(n):
-      if i != imax:
-        acc += lmbda[i]
-    lmbda[imax] -= acc
 
     # get the next iteration of x_k
     x_k = _linear_combine(n, lmbda, simplex)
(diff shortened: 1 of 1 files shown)


adenzler-nvidia:
I ran this PR's head on GPU (RTX PRO 6000, CUDA 12.9) since CI only exercises the CPU backend — the two CUDA test failures from #1503 are still present, though the margin one gets much closer:

```
FAILED GJKTest::test_sphere_mesh_margin   -0.0009959713 != -0.001 within 7 places  (was 7.1mm off on main)
FAILED GJKTest::test_box_box_ccd          1 contact instead of 4
```

Recomputing `xnorm` together with `x_k` fixes the big error (on main, `result.dist` was one iteration staler than the witness points whenever the loop exhausted its iterations). What remains is the renormalization block: the barycentric coordinates already sum to one by construction (`lambda_i = C4_i / m_det` with `m_det` the cofactor sum), so the block only redistributes rounding residuals into the largest coordinate — but doing that every iteration perturbs the iterate enough on CUDA that the Frank-Wolfe gap never meets its tolerance. Instrumenting the exits on the sphere-mesh case: without the renormalization GJK terminates via the FW gap at iteration 4 with the exact answer; with it, the loop runs all 35 iterations and lands 4e-6 off, and the perturbed witness points also degrade the box-box multiccd manifold (the 1-vs-4 contact failure). I also checked the coupling directly: with fresh distance reporting and working exit criteria in place, restoring just the renormalization still fails these two tests on CUDA.

#1511 fixes the same failures from the other end (removes the renormalization, scales the no-progress check for float32 — with mjMINVAL it can only fire on bitwise equality in float32, which is also why these failures are CUDA-only) but doesn't have the norm-vs-squared-tolerance fix from this PR. Both fixes look relevant: the criterion scaling here, and the renormalization/no-progress changes in #1511 to make the CUDA suite pass.

kbayes:
@thowell @adenzler-nvidia this is fix now on Blackwell. The culprit was indeed recomputing of the barycentric coordinates (Different LLMs give me different views on it, so don't know what the real story is). I simplified and fix the exit logic while I was at it. 

I also reran the script from #1430, and  output is better (though minor regression in 50m case) Loaded 10m in the viewer and looks a lot better now with very minimal flickering.

```
 fixed box |       MuJoCo (C) |   mjWarp default |   mjWarp NATIVECCD off
 half-size |    ncon  pen(mm) |    ncon  pen(mm) |          ncon  pen(mm)
----------------------------------------------------------------------------
      1 m  |     4    0.1078 |     4    0.1081 |         4     0.1081
     10 m  |     4    0.1078 |     4    0.2646 |         4     0.1081
     25 m  |     4    0.1078 |     4    0.1081 |         4     0.1081
     50 m  |     4    0.1078 |     4    0.1084 |         4     0.1081
    100 m  |     4    0.1078 |     4    0.1081 |         4     0.1081
```

thowell:
lgtm. will leave to @adenzler-nvidia to approve. thanks @kbayes!

adenzler-nvidia (approved):
Thank you! Verified on CUDA that this fixes the test_sphere_mesh_margin and test_box_box_ccd failures from #1503 — all 24 collision_gjk tests pass. This supersedes #1511, closing that one.

PR #1516, opened 2026-07-12
Add fast math and clear fill optimization

StafaH:
Enable fast math for the render megakernel and avoid clearing RGB and depth output buffers before every render. 

| Scene | Before | After | Change |
| --- | ---: | ---: | ---: |
| Primitives (2,048 worlds) | 1.090M FPS | 1.376M FPS | +26.2% |
| Humanoid (2,048 worlds) | 427k FPS | 479k FPS | +12.2% |

The attached before-after compares a 512×512 lit, shadowed frame before and after fast math. Only 98 of 262,144 pixels differ, by at most one 8-bit channel value.

image

### DIFF
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -551,8 +551,6 @@ def render(m: Model, d: Data, rc: RenderContext):
     d: The data on device.
     rc: The render context on device.
   """
-  rc.rgb_data.fill_(rc.background_color)
-  rc.depth_data.fill_(0.0)
   rc.seg_data.fill_(wp.vec2i(-1, -1))
 
   # Specialize the ray-cast helpers to the geom types present in the scene so the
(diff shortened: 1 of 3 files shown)


mujoco_warp/_src/render.py
```diff
@@ -735,6 +736,8 @@ def _render_megakernel(
           skybox_color[2] * 255.0,
           255.0,
         )
+      elif render_rgb[camid]:
+        rgb_out[worldid, rgb_adr[camid] + rayid_local] = background_color
```
thowell: can we remove `background_color = rc.background_color` above and instead of `background_color` use `wp.static(rc_static["background_color"])` here?

thowell (approved):
great optimizations!

mujoco_warp/_src/render.py
StafaH: Done. Had to make a small modification so that background color is fulled into rc_static since its a uint32

thowell (approved):
thanks @StafaH!

PR #1519, opened 2026-07-13
ccd mesh warmstart

thowell:
update mesh support function warmstarts
- correctly initialize distance
- return geoms from gjk with warmstart indices to epa

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -67,6 +67,8 @@ class GJKResult:
   simplex2: mat43
   simplex_index1: wp.vec4i
   simplex_index2: wp.vec4i
+  index1: int
+  index2: int
 
 
 @wp.struct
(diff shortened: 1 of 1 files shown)


kbayes:
Ok, I did some benchmarking on the aloha cluster and there seems to be a minor speedup with the combine changes.

mujoco_warp/_src/collision_gjk.py
```diff
@@ -639,7 +646,7 @@ def gjk(
   geomtype2: int,
   cutoff: float,
   is_discrete: bool,
-) -> GJKResult:
+) -> Tuple[GJKResult, Geom, Geom]:
```
kbayes: Is it possible to add the cache indices to GJKResult, and set those in the return Geoms down below? I think that might be more clear.

mujoco_warp/_src/collision_gjk.py
```diff
@@ -170,6 +170,8 @@ def support(geom: Geom, geomtype: int, dir: wp.vec3) -> SupportPoint:
       edge_localid = geom.graphadr + 2 + 2 * numvert
       prev = int(-1)
       imax = wp.where(geom.index > -1, geom.index, 0)
+      start_vert_idx = geom.graph[vert_globalid + imax]
+      max_dist = wp.dot(local_dir, geom.vert[geom.vertadr + start_vert_idx])
```
kbayes: Probably can just get rid of the start_vert_idx in favor of one line, unless you're worried about line length.

mujoco_warp/_src/collision_gjk.py

mujoco_warp/_src/collision_gjk.py

kbayes:
@thowell Looks great! Thanks!

kbayes (approved):

PR #1520, opened 2026-07-13
Add Newton decrement termination

adenzler-nvidia:
## Summary

Add the Newton decrement as an outer convergence criterion for the constraint solver.

After an accepted line-search step, the solver has already rebuilt the gradient and Hessian and computed the next Newton direction. The decrement

$$
\frac{1}{2} g^T H^{-1} g
$$

is the local quadratic model's predicted remaining improvement. Terminating when this value is below the solver tolerance avoids performing another outer Newton iteration solely to observe a correspondingly small actual improvement.

## Motivation

The existing outer termination criteria use the improvement from the previously accepted step and the Euclidean gradient norm. Neither criterion uses the newly computed Newton direction.

Consequently, a solve can reach a point where the remaining model improvement is below tolerance while the previous improvement and gradient norm remain above tolerance. The solver then performs another line search, rebuilds the constraint derivatives, and only afterward terminates based on the resulting actual improvement.

The line-search convergence criteria do not replace this check. They determine whether the one-dimensional minimization along the current direction has converged; they do not determine whether another outer Newton direction is worthwhile.

## Implementation

- Compute `g^T H^-1 g` while producing the Newton search direction.
- Apply the same `nv` and mean-inertia normalization used by the existing outer termination criteria.
- Fuse the search direction, search norm, and decrement reductions into the small and blocked Cholesky solve epilogues.
- Share the blocked Cholesky implementation between ordinary solves and the compile-time Newton epilogue specialization.
- Preserve the cached-factor and unchanged-search fast paths, including rescaling the cached decrement when the gradient changes only by a scalar.
- Stop allocating and writing the intermediate `Mgrad` array for Newton. The CG solver continues to use `Mgrad` unchanged.

## Validation

- Add a unit test where actual improvement and gradient norm are above tolerance but the Newton model improvement is below tolerance.
- Extend blocked Cholesky coverage to validate the ordinary solve, augmented solve, fused Newton epilogue, and cached-factor solve against NumPy.
- Run the complete CUDA solver and support test suites: 86 tests pass.
- Run all pre-commit checks.

## Performance

Measured with `mjwarp-testspeed`, elliptic cones, 8,192 worlds, and 1,000 steps. Results are medians of four interleaved runs, with each variant occupying each ordinal position once. G1 uses `shuffle_dance.npz`; Aloha Clutter uses `pick_clutter.npz`. No prerecorded trajectory is available for the humanoid benchmarks.

| Benchmark | Upstream | Decrement, unfused | Final | Solver iterations, upstream to final | Final vs upstream |
|---|---:|---:|---:|---:|---:|
| Humanoid | 2.542 s | 2.299 s | 2.282 s | 1.644 to 1.345 | -10.2% |
| G1 replay | 4.337 s | 3.950 s | 3.890 s | 4.030 to 3.322 | -10.3% |
| Three humanoids | 15.606 s | 13.742 s | 13.767 s | 3.547 to 2.815 | -11.8% |
| Aloha Clutter replay | 12.094 s | 11.634 s | 11.633 s | 1.106 to 1.104 | -3.8% |

The unfused variant isolates the convergence change by computing the search direction, search norm, and decrement in a separate kernel. Final adds the fused tile epilogue and removes the Newton `Mgrad` workspace. The complete runs behind the medians are:

| Benchmark | Variant | Run 1 | Run 2 | Run 3 | Run 4 | Median |
|---|---|---:|---:|---:|---:|---:|
| Humanoid | Upstream | 2.556676 | 2.527121 | 2.571986 | 2.360423 | 2.541898 |
| Humanoid | Unfused | 2.329163 | 2.344091 | 2.246855 | 2.268954 | 2.299058 |
| Humanoid | Final | 2.278954 | 2.285754 | 2.298831 | 2.246935 | 2.282354 |
| G1 replay | Upstream | 4.377498 | 4.154805 | 4.634156 | 4.296581 | 4.337040 |
| G1 replay | Unfused | 3.940362 | 3.960360 | 4.069016 | 3.926335 | 3.950361 |
| G1 replay | Final | 3.895126 | 3.884735 | 4.237712 | 3.879237 | 3.889930 |
| Three humanoids | Upstream | 15.439896 | 15.631627 | 15.580172 | 15.642628 | 15.605900 |
| Three humanoids | Unfused | 13.707339 | 13.741917 | 13.793112 | 13.741285 | 13.741601 |
| Three humanoids | Final | 13.849973 | 13.677723 | 13.788575 | 13.744602 | 13.766589 |
| Aloha Clutter replay | Upstream | 11.277973 | 12.127685 | 12.059470 | 12.237110 | 12.093578 |
| Aloha Clutter replay | Unfused | 11.307978 | 11.694572 | 11.647998 | 11.620374 | 11.634186 |
| Aloha Clutter replay | Final | 11.399335 | 11.795854 | 11.945521 | 11.470001 | 11.632928 |

CUDA 12.9 Nsight Systems profiles cover the same three variants over 100 steps:

| Benchmark | Upstream GPU time | Unfused | Final | Final vs upstream |
|---|---:|---:|---:|---:|
| Humanoid | 208.8 ms | 188.9 ms | 186.3 ms | -10.8% |
| G1 replay | 518.0 ms | 455.1 ms | 459.9 ms | -11.2% |
| Three humanoids | 1331.1 ms | 1198.0 ms | 1185.4 ms | -10.9% |
| Aloha Clutter replay | 1083.9 ms | 1068.2 ms | 1068.7 ms | -1.4% |

### `Mgrad` ablation

The earlier fused-plus-`Mgrad` measurements used the pre-tile scalar epilogue and therefore did not isolate the `Mgrad` write. The corrected variant is identical to Final except that it allocates `Mgrad` and stores the solved vector to it. Four paired runs, with each variant running first twice, show no resolvable end-to-end effect:

| Benchmark | Tile + `Mgrad` runs | Median | Final runs | Median | Final vs `Mgrad` |
|---|---|---:|---|---:|---:|
| Humanoid | 2.207159 / 2.280833 / 2.311542 / 2.319607 | 2.296187 | 2.294792 / 2.284424 / 2.295558 / 2.284665 | 2.289728 | -0.28% |
| G1 replay | 4.196580 / 4.186073 / 4.228789 / 4.231837 | 4.212684 | 4.240905 / 4.224080 / 4.242891 / 4.209969 | 4.232493 | +0.47% |
| Three humanoids | 13.625806 / 13.707099 / 13.788066 / 13.731764 | 13.719431 | 13.733249 / 13.743454 / 13.665568 / 13.727870 | 13.730560 | +0.08% |
| Aloha Clutter replay | 11.970563 / 12.093601 / 11.887607 / 11.890023 | 11.930293 | 12.047704 / 12.025934 / 11.933399 / 11.935090 | 11.980512 | +0.42% |

Matching Nsight profiles show that removing the write consistently reduces the changed Cholesky kernel, while whole-profile timing remains sensitive to persistent desktop graphics activity:

| Benchmark | `Mgrad` GPU total | Final GPU total | `Mgrad` Cholesky avg | Final Cholesky avg | Final kernel delta |
|---|---:|---:|---:|---:|---:|
| Humanoid | 189.532 ms | 186.004 ms | 26.639 us | 23.948 us | -10.1% |
| G1 replay | 454.826 ms | 419.452 ms | 50.629 us | 49.490 us | -2.3% |
| Three humanoids | 1032.031 ms | 1026.428 ms | 198.085 us | 195.541 us | -1.3% |
| Aloha Clutter replay | 940.726 ms | 939.491 ms | 223.269 us | 221.742 us | -0.7% |

The iteration reduction provides the main end-to-end improvement. Fusion removes the standalone reduction launch. Removing `Mgrad` eliminates an unused Newton workspace and output store; its whole-step effect is below benchmark noise, while the isolated kernel consistently moves in the expected direction.

### DIFF
--- a/mujoco_warp/_src/block_cholesky.py
+++ b/mujoco_warp/_src/block_cholesky.py
@@ -19,6 +19,32 @@ import warp as wp
 
 
 @lru_cache(maxsize=None)
+def _create_solve_epilogue_func(matrix_size_static: int, vector_size_static: int):
+  @wp.func
+  def solve_epilogue_func(
+    # In:
+    solution_tile: wp.tile[float, matrix_size_static, 1],
+    b: wp.array2d[float],
+    thread_idx: int,
+    # Out:
+    search_out: wp.array2d[float],
+  ):
+    local_search_dot = float(0.0)
+    local_decrement = float(0.0)
+    for i in range(thread_idx, vector_size_static, wp.block_dim()):
+      solution = wp.tile_extract(solution_tile, i, 0)
+      search_out[i, 0] = -solution
+      local_search_dot += solution * solution
+      local_decrement += b[i, 0] * solution
+
+    search_dot = wp.tile_reduce(wp.add, wp.tile(local_search_dot, preserve_type=True))[0]
+    decrement = wp.tile_reduce(wp.add, wp.tile(local_decrement, preserve_type=True))[0]
+    return wp.vec2(search_dot, decrement)
+
+  return solve_epilogue_func
+
+
+@lru_cache(maxsize=None)
 def create_blocked_cholesky_factorize_solve_func(block_size: int, matrix_size_static: int):
   @wp.func
   def blocked_cholesky_factorize_solve_func(
@@ -93,7 +119,13 @@ def create_blocked_cholesky_factorize_solve_func(block_size: int, matrix_size_st
 
 
 @lru_cache(maxsize=None)
-def create_blocked_cholesky_augmented_factorize_solve_func(block_size: int, matrix_size_static: int):
+def _create_blocked_cholesky_augmented_factorize_solve_func(
+  block_size: int,
+  matrix_size_static: int,
+  with_newton_epilogue: bool,
+  vector_size_static: int,
+):
+  WITH_NEWTON_EPILOGUE = with_newton_epilogue
   border_size = block_size - 1
 
   @wp.func
@@ -102,9 +134,11 @@ def create_blocked_cholesky_augmented_factorize_solve_func(block_size: int, matr
     A: wp.array2d[float],
     b: wp.array2d[float],
     matrix_size: int,
+    thread_idx: int,
     # Out:
     U: wp.array2d[float],
-    x: wp.array2d[float],
+    # Out:
+    result_out: wp.array2d[float],
   ):
     """Factor A with b as an augmented border and reuse it as the forward solution."""
     rhs_tile = wp.tile_zeros(shape=(matrix_size_static, 1), dtype=float, storage="shared")
@@ -170,21 +204,35 @@ def create_blocked_cholesky_augmented_factorize_solve_func(block_size: int, matr
       )
       wp.tile_upper_solve_inplace(U_tile, tmp_tile)
 
-    wp.tile_store(x, rhs_tile, offset=(0, 0), bounds_check=False)
+    sums = wp.vec2(0.0)
+    if wp.static(WITH_NEWTON_EPILOGUE):
+      sums = wp.static(_create_solve_epilogue_func(matrix_size_static, vector_size_static))(rhs_tile, b, thread_idx, result_out)
+    else:
+      wp.tile_store(result_out, rhs_tile, offset=(0, 0), bounds_check=False)
+
+    return sums
 
   return blocked_cholesky_augmented_factorize_solve_func
 
 
 @lru_cache(maxsize=None)
-def create_blocked_cholesky_solve_func(block_size: int, matrix_size_static: int):
+def _create_blocked_cholesky_solve_func(
+  block_size: int,
+  matrix_size_static: int,
+  with_newton_epilogue: bool,
+  vector_size_static: int,
+):
+  WITH_NEWTON_EPILOGUE = with_newton_epilogue
+
   @wp.func
   def blocked_cholesky_solve_func(
     # In:
     U: wp.array2d[float],
     b: wp.array2d[float],
     matrix_size: int,
+    thread_idx: int,
     # Out:
-    x: wp.array2d[float],
+    result_out: wp.array2d[float],
   ):
     """Block Cholesky solve.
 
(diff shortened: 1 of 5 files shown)


adenzler-nvidia:
some of the data does not make any sense because my agent combined runs from different times for the ablation. The end-to-end numbers are validated but I need to rerun the ablation.

yuvaltassa:
Ha, I've wanted to do something like this for a long time 🙂 Will evaluate this closely.

There's actually another, maybe related, (possibly identical!) idea that I remember having, I think sometimes we can do zero iterations? Imagine a well-warmstarted static scene. After building the Hessian but before factorizing, I should be able to determine convergence. You see it?

This is just from clocking your description on a phone, I've not read a LoC. Will look carefully soon!

yuvaltassa:
Early results on CPU are promising. The thing I alluded to (zero iterations) is not the same but is a natural extension.

yuvaltassa:
Too much GPU arcana in this PR for me to review, I'll leave that to @thowell , but as you can see from the above, I support this idea 🙂

mujoco_warp/_src/solver.py
```diff
@@ -2079,10 +2076,12 @@ def kernel(
         if sigma != 0.0:
           ratio = new_sigma / sigma
         ctx_grad_dot_out[worldid] *= ratio * ratio
+        ctx_newton_decrement_out[worldid] *= ratio * ratio
```
thowell: compute `ratio * ratio` once and reuse?

mujoco_warp/_src/solver.py
```diff
@@ -23,8 +23,10 @@
 from mujoco_warp._src import smooth
 from mujoco_warp._src import support
 from mujoco_warp._src import types
-from mujoco_warp._src.block_cholesky import create_blocked_cholesky_augmented_factorize_solve_func
-from mujoco_warp._src.block_cholesky import create_blocked_cholesky_solve_func
+from mujoco_warp._src.block_cholesky import _solve_search_sums
```
thowell: lets remove the `_` prefix since this is imported from another module

mujoco_warp/_src/block_cholesky.py
```diff
@@ -18,6 +18,29 @@
 import warp as wp
 
 
+@wp.func
+def _solve_search_sums(grad: float, solution: float):
+  return wp.vec2(solution * solution, grad * solution)
+
+
+@lru_cache(maxsize=None)
+def _create_solve_epilogue_func(matrix_size_static: int, vector_size_static: int):
```
thowell: what do we think about using something like 'check_decrement' or 'decrement_termination` throughout instead of 'epilogue' to hopefully be a bit more descriptive?

thowell (approved):
thanks @adenzler-nvidia!

yuvaltassa:
@adenzler-nvidia @thowell you might want to consider porting this to MJWarp, but I'm not sure how valuable it is. It's basically a weak form of sleeping. 

https://github.com/google-deepmind/mujoco/commit/c69ef03083f8badc1ecc29630728572c47c29341

mujoco_warp/_src/solver.py
adenzler-nvidia: Done in 9f289cc4 — hoisted into `ratio_sq`.

mujoco_warp/_src/solver.py
adenzler-nvidia: Done in 9f289cc4 — renamed to `solve_search_sums`.

mujoco_warp/_src/block_cholesky.py
adenzler-nvidia: Renamed in 9f289cc4: `_create_solve_epilogue_func` → `_create_newton_decrement_func` and `with_newton_epilogue` → `with_newton_decrement`. Went with plain `newton_decrement` since this helper computes the decrement (and search sums) — the termination check itself lives in the solver kernel.

PR #1521, opened 2026-07-15
Fuse elliptic curvature into sparse Hessian assembly

adenzler-nvidia:
Sparse Newton currently assembles ordinary constraint curvature and elliptic cone curvature in separate JTDAJ and JTCJ kernels even though both traverse the same contact Jacobian support and update the same upper-triangular Hessian block. The extra elliptic pass repeats Jacobian loads and atomics, adds a kernel launch, and requires a model-side support-pair bound solely to size that launch.

This change folds elliptic contact curvature into the grouped sparse JTDAJ assembly. One warp handles a constraint group, prepares its cone curvature once, shares it across the lanes, and accumulates each Hessian entry while the Jacobian rows are already resident. Ordinary constraints retain the diagonal-D path, and the pyramidal kernel is compile-time specialized without elliptic work. The regular and compact sparse Newton paths use the same implementation; the dense solver path is unchanged.

The generated kernel is specialized from the model's maximum contact dimension. Three-dimensional cones retain a six-coefficient scalar fast path. Higher-dimensional specializations share the sixteen common cone terms and evaluate each Hessian entry directly from the cone Hessian's diagonal-plus-rank-one structure, reducing the contraction from quadratic to linear work in the contact dimension. Exact condim 3/4/6 paths remain statically unrolled while mixed contact dimensions dispatch at runtime. This removes the expanded ten- and twenty-one-coefficient tables, their preparation stage, and the obsolete standalone sparse JTCJ kernels and launch-sizing metadata.

The solver test reconstructs the elliptic `J^T C J` contribution with an independent NumPy reference and compares it to the fused Hessian for maximum dimension 3 and mixed 3/4/6 models. It also covers elliptic compact-solver equivalence in addition to the existing pyramidal case.

Performance was measured on an RTX PRO 6000 Blackwell with 8,192 worlds. Values are medians of three interleaved runs after unmeasured warmups, comparing current `upstream/main` (`c50dd837`) with the final candidate (`1ca3be65`). G1 and Aloha use prerecorded trajectories, production scenes run 1,000 steps, synthetic fixed-contact scenes run 500 steps, and all 8,192 worlds converged.

| Benchmark | Main | Fused | Difference |
|---|---:|---:|---:|
| G1, elliptic | 4.129 s | 3.485 s | 15.6% faster |
| Three humanoids, elliptic | 13.501 s | 8.785 s | 34.9% faster |
| Aloha clutter, elliptic | 11.829 s | 11.946 s | neutral (+1.0%) |
| G1, pyramidal control | 2.329 s | 2.354 s | neutral (+1.0%) |
| Synthetic `condim=3` | 2.094 s | 1.319 s | 37.0% faster |
| Synthetic `condim=4` | 2.306 s | 1.391 s | 39.7% faster |
| Synthetic `condim=6` | 4.870 s | 2.339 s | 52.0% faster |

The independent maximum-dimension 3/4/6 Hessian tests and compact-solver equivalence tests pass on CUDA. Pre-commit, including kernel-analyzer, also passes.


adenzler-nvidia:
@thowell @erikfrey this got a bit messy, but I don't think there is a good way to reduce all of the branching and specialization. could you guys give me an indication of how mergeable this is?

thowell:
@adenzler-nvidia lets proceed. please let me know when this is ready for review and i will take a look. thanks!

adenzler-nvidia:
Thanks @thowell - should be ready for a first pass of reviews.

mujoco_warp/_src/io.py
```diff
@@ -573,13 +573,6 @@ def _check_friction(name: str, id_: int, condim: int, friction, checks):
       dofid = mjm.dof_parentid[dofid]
   m.body_isdofancestor = body_isdofancestor
 
-  # Upper bound on a contact's Jacobian support-pair count, to size the elliptic-cone JTCJ
```
thowell, line 576: nice!

mujoco_warp/_src/solver.py
```diff
@@ -2392,351 +2393,6 @@ def kernel(
   return kernel
 
 
-# TODO(thowell): combine with JTDAJ ?
```
thowell, line 2395: awesome!

mujoco_warp/_src/solver.py
```diff
@@ -3087,25 +2743,168 @@ def _cholesky_factorize_solve(
...
 # ---------------------------------------------------------------------------
-_JTDAJ_THREADS_PER_GROUP = 32  # one warp per group, so its J reads coalesce
-_JTDAJ_OVERSUBSCRIBE_WAVES = 6  # grid-stride depth; short per-warp chains load-balance groups
+_JTDAJ_THREADS_PER_GROUP = 32
+_JTDAJ_OVERSUBSCRIBE_WAVES = 6
 
 
 @cache_kernel
-def _JTDAJ_sparse(compact: bool):
+def _JTDAJ_sparse(compact: bool, cone_type: types.ConeType, max_condim: int):
```
thowell: should we consider renaming `JTDAJ` to something like `JTDACJ` now?

mujoco_warp/_src/solver.py
```diff
@@ -3087,25 +2743,168 @@ def _cholesky_factorize_solve(
...
+          u = ctx_Jaref_in[worldid, efcid] * scale
+          terms[dim] = u
+          terms[6 + dim] = scale
+          tt += u * u
+
+      t = wp.max(wp.sqrt(tt), types.MJ_MINVAL)
+      ttt = wp.max(t * t * t, types.MJ_MINVAL)
+      mu_over_t = math.safe_div(mu, t)
+      mu_n_over_ttt = mu * math.safe_div(n, ttt)
+      tangent_diag = mu2 - mu * math.safe_div(n, t)
```
thowell: `mu * math.safe_dev(n, t)` -> `n * mu_over_t`?

thowell (approved):
thanks @adenzler-nvidia!

mujoco_warp/_src/solver.py
erikfrey: Renamed `_JTDAJ_sparse` to `_JTDACJ_sparse` (and retained `_JTDAJ_sparse = _JTDACJ_sparse` alias).

mujoco_warp/_src/solver.py
erikfrey: Updated `tangent_diag = mu2 - n * mu_over_t` to reuse `mu_over_t`.

erikfrey:
Applied review suggestions from @thowell:
- Renamed `_JTDAJ_sparse` kernel to `_JTDACJ_sparse` (with `_JTDAJ_sparse` alias retained for backward compatibility).
- Simplified `tangent_diag` calculation in `_JTDACJ_sparse` to `mu2 - n * mu_over_t`, reusing the precomputed `mu_over_t` term.

All unit tests and pre-commit checks pass.

thowell (approved):
lgtm, thanks @adenzler-nvidia and thanks @erikfrey!

PR #1522, opened 2026-07-16
Use npolygonmax and nmeshdegmax from MuJoCo C

kbayes:
These fields are newly added to the C code and we can just use them directly.


erikfrey:
@kbayes this is a nice PR - I brought it up to date with head and it looks like all tests are passing.  Do you have any concerns with merging it?

kbayes:
@thowell If mjWarp is caught up with the MuJoCo release then this should be safe to merge.

thowell:
@kbayes minimum mujoco version is v3.11.0 now so we should be good

kbayes:
@thowell FYI, this is ready for review.

thowell (approved):

thowell:
thanks @kbayes!

PR #1523, opened 2026-07-16
_solve_init_dof

thowell:
fix bug where `qfrc_constraint` wasn't correctly initialized to zero when `nefc==0` and `nefc>0` at the previous timestep

### DIFF
--- a/mujoco_warp/_src/inverse.py
+++ b/mujoco_warp/_src/inverse.py
@@ -119,12 +119,27 @@ def discrete_acc(m: Model, d: Data, qacc: wp.array2d[float]):
   smooth.solve_m(m, d, qacc, qfrc)
 
 
+@wp.kernel
+def _zero_qfrc_constraint_nefc(nefc_in: wp.array[int], qfrc_constraint_out: wp.array2d[float]):
+  worldid, dofid = wp.tid()
+  if nefc_in[worldid] == 0:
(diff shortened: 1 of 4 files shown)


adenzler-nvidia (approved):
LGTM. Verified the new test fails on main and passes here, and profiled the copy→kernel swap on humanoid at 8192 worlds: the fused kernel costs ~2.7µs/step vs ~1.9µs for the D2D memcpy it replaces, so no measurable step-time impact. The sparse-only zeroing is right: the dense init kernel stores its accumulated sum unconditionally, and the sleep/compact path also comes out clean (checked empirically).

One gap worth covering here or in a follow-up: the inverse dynamics path has the same staleness. inv_constraint → init_context → _update_constraint uses the same nefc-gated sparse zeroing, and the new kernel only runs in _solve. So a contact step followed by mj_inverse at nefc==0 keeps the old contact forces in qfrc_constraint, and qfrc_inverse absorbs them (reproduced on this branch: sphere on plane, teleport up, mjw.inverse → max|qfrc_constraint| stays 34.3). Zeroing per-world in inv_constraint next to the existing njmax==0 zero would cover it.

thowell:
updated with a similar `qfrc_constraint` initialization and test for the inverse dynamics

adenzler-nvidia:
looks good, thanks for the fix!

Issue #1517, opened 2026-07-13
MuJoCo Warp Generates Nans with sliding friction=0.001


gyeomannvidia:
Reproduction script here:

"""
Reproduce NaN in mujoco_warp: rigid box-box contact with zero sliding friction
and condim=3.

Two free boxes stacked on a ground plane.  The top box-bottom box contact has
sliding friction = max(box_mu, box_mu) = box_mu, condim=3.  When box_mu is too
small the friction cone is degenerate and qpos goes NaN.

Usage:
    uv run python repro_nan_friction.py
"""

import mujoco
import numpy as np

import mujoco_warp as mjw

# ---------------------------------------------------------------------------
# Scene
# ---------------------------------------------------------------------------

MJCF_TEMPLATE = """
<mujoco model="nan_friction_repro">
  <option gravity="0 0 -10" timestep="{timestep:.10f}"/>

  <worldbody>
    
    <geom name="ground" type="plane" pos="0 0 0" size="10 10 0.1"
          friction="1 0.005 0.0001" condim="3"/>

    
    <body name="boxB" pos="0 0 0.2">
      <freejoint/>
      <inertial mass="1.0" pos="0 0 0" diaginertia="1.0 1.0 1.0"/>
      <geom name="geomB" type="box" size="0.2 0.2 0.2"
            friction="{mu:.10g} 0.005 0.0001" condim="3"/>
    </body>

    
    <body name="boxA" pos="0 0 0.5">
      <freejoint/>
      <inertial mass="1.0" pos="0 0 0" diaginertia="1.0 1.0 1.0"/>
      <geom name="geomA" type="box" size="0.1 0.1 0.1"
            friction="{mu:.10g} 0.005 0.0001" condim="3"/>
    </body>
  </worldbody>
</mujoco>
"""

DT = 1.0 / 240.0
N_STEPS = 30

# ---------------------------------------------------------------------------
# Helper
# ---------------------------------------------------------------------------

def run_scene(mu: float, n_steps: int = N_STEPS, verbose: bool = True) -> dict:
    """
    Load the scene with the given sliding friction mu, step n_steps times, and
    return a result dict with keys:
        nan_step  - first step that produced a NaN (-1 if none)
        nan_body  - name of the body (or 'none')
        qpos_final - final qpos array (may contain NaN)
    """
    xml = MJCF_TEMPLATE.format(timestep=DT, mu=mu)

    mjm = mujoco.MjModel.from_xml_string(xml)
[...]

PR #1525, opened 2026-07-16
minimum compliance

thowell:
introduces a minimum compliance setting `MIN_COMPLIANCE` that addresses https://github.com/google-deepmind/mujoco_warp/issues/1517

### DIFF
--- a/mujoco_warp/_src/constraint.py
+++ b/mujoco_warp/_src/constraint.py
@@ -90,6 +90,7 @@ def _efc_row(
   pos_aref: float,
   pos_imp: float,
   invweight: float,
+  invweight_ref: float,
   solref: wp.vec2,
   solimp: vec5,
   margin: float,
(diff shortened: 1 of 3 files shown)


adenzler-nvidia:
Could this be a relative floor instead, to make it mass-invariant? Compliance scales like 1/mass while the conditioning this guards against doesn't — e.g. this PR's test scene with masses rescaled to 1e-3 still NaNs at mu=1e-4, and at mass=1000/mu=0.1 the floor softens contacts that were fine before. Something like flooring against the invweight from before the pyramidal `2*mu^2` scaling:

```python
# _efc_row, with invweight_ref = invweight before the pyramidal friction scaling
D_out[worldid, efcid] = 1.0 / wp.max(invweight * (1.0 - imp) / imp, MIN_COMPLIANCE_RATIO * invweight_ref)
```

with `MIN_COMPLIANCE_RATIO = 1e-4` (dimensionless). At default solimp this only binds for pyramidal rows with mu <~ 0.02. Tried an equivalent version on this branch: the mass {1e-3, 1, 1e3} x mu {1e-4, 1e-3, 0.1, 1} matrix runs clean, with the heavy/moderate-mu cases keeping their original stiffness.

thowell:
@adenzler-nvidia changed to a minimum compliance ratio as recommended.

adenzler-nvidia:
Thanks for the quick turnaround! Verified on the current head (0d5280bc): the mass {1e-3, 1, 1e3} x mu {1e-5 .. 1} matrix runs clean, floored D scales correctly with mass, resting heights match C, and the mass=1000/mu=0.1 case keeps its unclamped stiffness. Since this intentionally diverges from C's `mjMINVAL` floor for pyramidal rows below mu ~ 0.02, it would be good to have a MuJoCo-side expert confirm the semantics are acceptable (and whether C's single-precision build wants the same treatment).

yuvaltassa:
Interesting. I mean a longstanding thought is that mjMINVAL should be precision-dependent...
let me run some experiments and get back to you

PR #1529, opened 2026-07-17
Gate the EPA horizon warning behind warn_overflow and record it in Data.overflow

adenzler-nvidia:
#1461 added the warn_overflow option to control the legacy in-kernel stdout warnings, routing overflow events to the per-world Data.overflow bitmask instead. The EPA horizon warning in collision_gjk.py was not covered, so it still prints even when warn_overflow is disabled, and the event leaves no record in Data.overflow.

This change threads warn_overflow through ccd, epa_phase and _epa to gate the printf, and adds an EPA_HORIZON bit to OverflowType that is set whenever the horizon buffer runs out of space, matching the other overflow warnings. The box-box horizon test now also runs with an undersized horizon buffer to check the new bit.

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -475,6 +475,8 @@ def ccd_hfield_kernel_builder(
             epa_norm2,
             epa_horizon,
             wp.static(warn_overflow),
+            worldid,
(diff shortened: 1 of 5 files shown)


thowell (approved):
thanks @adenzler-nvidia!

PR #1533, opened 2026-07-18
Extend deterministic mode to the full pipeline: smooth dynamics, forces, sensors, derivatives, and remaining solver paths

johnnynunez:
Stacked on #1425 (includes its commits; will rebase to contain only these commits as the stack lands).

opt.deterministic (#1422, #1425) makes contact ordering, constraint row allocation, and the sparse-solver reductions bitwise reproducible, but the rest of the pipeline still accumulates through scheduling-ordered float atomics, so stepping stays nondeterministic whenever those paths are exercised. Branching-tree models diverge from step 0 through the smooth stage (subtree_com, the CRB pass, RNE bias forces, the sparse LDL factorization), and gravity compensation, flex elasticity, tendon lengths, actuator moments, touch/force/torque/subtree sensors, energy, implicit-integrator derivatives, dense solves with nv above the tile width, the elliptic cone Hessian, the CG flex preconditioner, and the sleep compact solve each carry their own order-sensitive accumulation.

This PR gives every one of those accumulations a deterministic gather variant behind the same wp.static(deterministic) factory pattern the stack already uses, so the default path compiles to the same kernels as before. Tree fan-ins become children gathers over a precomputed body-children CSR walked level by level; the sparse LDL factorization and fused solve gather per row slot in ascending source order; tendon lengths and Jacobians walk each tendon's wrap path serially; actuator moments use a statically derived CSR layout, since row widths are structural for every transmission type; qfrc_actuator and passive tendon forces gather per dof through a dof-to-tendon transpose; gravity compensation gathers per dof over subtree body lists; flex elastic and bending forces materialize per-element partials into static slots gathered per body; touch, tactile, and contact-match sensors scan the world's sorted contacts serially; energy reduces serially per world; qDeriv gathers per mass-matrix element so implicit and implicitfast are bitwise; dense solves pin the per-efc thread count and pad the contact-Jacobian tile to nv_pad; the incremental-Newton flag design extends to dense layouts; the elliptic JTCJ term gets merge-intersection gathers; and the deterministic column index extends to the sleep compact solve, so deterministic mode survives sleep and wake churn.

Two default-path bugfixes ride along: the dense elliptic JTCJ accumulated cross-block contributions with a non-atomic read-modify-write, losing updates under contention, and equality-flexstrain JTDAJ blocks were stamped nrow=6 while each eigenmode emits one row, double-counting the Hessian on sparse Newton.

The tradeoff is deterministic-mode cost: the gathers serialize work the atomic path spreads across threads, and this PR does not yet include a throughput characterization like the one in #1425. The default path is unchanged and the full suite passes (1266 passed, 18 skipped). Determinism tests grow to 71 and now cover branching trees in both jacobian layouts, nv=50 dense, gravity compensation, fixed-tendon actuation, implicit integrators, sensordata and energy, flex full-state stepping, CG with the flex preconditioner, the elliptic cone, sleep wake churn, and long-horizon graph-captured replay.


ruziniuuuuu:
I found a deterministic contact-allocation bug when contacts come from an external collision producer: the compacted contact buffer can interleave world IDs, while the current scan assumes one contiguous range per world. The failure is visible in a synthetic four-world test: permuting an otherwise valid contact buffer changes nefc from [8, 8, 8, 8] to [24, 24, 24, 24].

I opened a small stacked PR against this branch: https://github.com/johnnynunez/mujoco_warp/pull/3

It replaces the global contact scan/world-boundary localization with the existing per-world 2D scan and covers interleaved world IDs. The focused contact, flex, and constraint suites pass (83 passed, 2 skipped), and a downstream external-contact four-world first-step regression is finite with valid per-world nefc.

PR #1534, opened 2026-07-19
Extract reusable constraint kbi and efc_D computations

etaoxing:
The k, b, and impedance calculations were embedded in `_efc_row`, which made them unavailable to code that needs the same solver-parameter conversion. This change extracts `_contact_kbimp` and `_efc_D` as reusable Warp functions. Active impedance regions are evaluated with branches so inactive expressions do not introduce floating-point exceptions or invalid adjoints. Swaps `wp.where` selections for real branches to avoid the NaN parameter adjoints warp produced by differentiating the unselected operand.

---

Summary results:
  - 4,416 values across 552 cases matched bit-for-bit.
  - Humanoid: 2.6276 s legacy, 2.6279 s PR (−0.01%).
  - G1: 4.6654 s legacy, 4.6661 s PR (−0.01%).
  - Every run completed all 8,192 worlds.

The A/B check called `_efc_row` directly with deterministic inputs derived from both benchmark models and compared all eight output fields as raw 32-bit values. All 4,416 values across 552 cases matched `origin/main` (`46353db`) bit-for-bit.

Performance was measured with the standard `mjwarp-testspeed` runner on an NVIDIA L40 with 8,192 worlds and 1,000 steps. Values are means over 10 interleaved runs after 1 unmeasured warmup run, comparing `origin/main` (`46353db`) with this PR (`e4f3548`). Both workloads use `opt.cone=elliptic`; G1 uses the prerecorded `shuffle_dance.npz` trajectory and Humanoid uses its standard keyframe rollout.
Every run converged all 8,192 worlds.

| Benchmark | `origin/main` | This PR | Difference |
|---|---:|---:|---:|
| Humanoid, elliptic | 2.628 s | 2.628 s | -0.01% |
| G1 flat, elliptic, prerecorded | 4.665 s | 4.666 s | -0.01% |


etaoxing:
for https://github.com/google-deepmind/mujoco_warp/pull/1535

thowell:
@etaoxing lets run the humanoid benchmark before and after the proposed changes to confirm that there are no performance regressions. thanks!

thowell (approved):
thanks @etaoxing!

PR #1535, opened 2026-07-19
hybrid analytic differentiability

### LINKED ISSUE #500: Differentiability

etaoxing:
w/ @eric-heiden, @mmacklin 

initial report: https://etaoxing.com/reports/diff_mjw/diff_mjw.html

note: there are currently changes in `collision_primitive_core.py`, `constraint.py`, `solver.py` that have already / still need to be merged first. these are just refactors that do not modify the forward pass.


yuvaltassa:
This looks very promising. I guess you are computing d(qfrc_constraint)/d(qvel) analytically? I have code that does this in C that I've never published.

What I don't understand is how you do dervatives wrt position, since that includes the narrowphase collisions. Can you explain?

etaoxing:
@yuvaltassa thanks for the questions, nice to hear that you've done that before!

Yes, d(qfrc_constraint)/d(qvel) computed analytically with the adjoint method: one linear solve against the Newton solver's Hessian at the converged qacc from the forward pass (implicit function theorem), then that multiplier is backpropagated through the constraint force law to each input. "Hybrid analytic", since this is done by combining implicit differentiation, autodiff, and custom VJPs (without materializing the full d(qfrc_constraint)/d(qvel) matrix).

For position derivatives, we freeze the discrete narrowphase choices (contact set, contact-point slot assignment, witness feature) and differentiate only the smooth contact geometry: `contact_qpos_vjp()` in `collision_adjoint.py` replays the narrowphase under the frozen witness and autodiffs the replay (`_narrowphase_recompute()`) for derivatives of dist, pos, and the contact frame with respect to the geom poses. Those then chain back to qpos through the kinematic Jacobians and quaternion lift.

yuvaltassa:
Regarding the contact-related position derivatives, I hope you understand that this approximation is quite bad in many situations. Consider a ball rolling on a plane. The zeroth order approximation tells you that the witness points are moving not at all. The approximation you're using tells you that they are moving apart vertically. In fact they are moving linearly sideways and vertically not at all 🙂. That said, you don't really have a good alternative right now.  BTW Trinkle had a great paper about this maybe ~13 years ago, let me know if you find it.

Do you approximate the CRB (inertia) derivatives or not? IIRC the propagation algorithm for the exact version was non trivial.

Regarding the velocity derivatives, I didn't really understand your reply. I'll post what I did in the next comment, perhaps we are saying exaclty the same thing.

yuvaltassa:
# MuJoCo: Analytical Derivative of Constraint Force with Respect to Velocity

We describe the math behind `mjd_constraint_vel` — the analytical computation of $`\partial(\texttt{qfrc\_constraint}) / \partial v`$, an $`n_v \times n_v`$ dense matrix.

## 1. Setup and notation

Consider MuJoCo's constrained dynamics. The state is $`(q, v)`$ with $`v \in \mathbb{R}^{n_v}`$. The constraint solver finds an acceleration $`\dot{v}^*`$ by minimizing a cost that balances the unconstrained acceleration against constraint satisfaction:

```math
\dot{v}^*(v) = \arg\min_x\; C(x, v)
```

where

```math
C(x, v) = \tfrac{1}{2}\|x - M^{-1}(\tau(v) - c(v))\|_M^2 + s(Jx - a_\text{ref}(v))
```

Here:
- $M$ is the joint-space inertia matrix ($n_v \times n_v$, symmetric positive definite)
- $J$ is the constraint Jacobian ($n_\text{efc} \times n_v$)
- $\tau(v)$ are applied/actuator forces, $c(v)$ are bias forces (Coriolis + gravity)
- $a_\text{ref}(v)$ is the reference acceleration (encodes constraint targets)
- $`s(\cdot)`$ is a piecewise-quadratic penalty (quadratic for active constraints, flat for inactive)

The **constraint force in joint space** is:

```math
\texttt{qfrc\_constraint} = J^T f = M \dot{v}^* + c - \tau
```

We want its derivative with respect to $`v`$.

## 2. Implicit function theorem

At the solver solution, the first-order optimality condition is:

```math
\nabla_x C(\dot{v}^*, v) = 0
```

Differentiating with respect to $`v`$ (implicit function theorem), assuming the active set is locally constant:

```math
\frac{\partial^2 C}{\partial x^2} \frac{\partial \dot{v}^*}{\partial v}
+ \frac{\partial^2 C}{\partial x \, \partial v} = 0
```

Solving:

```math
\frac{\partial \dot{v}^*}{\partial v}
= -\left(\frac{\partial^2 C}{\partial x^2}\right)^{-1}
   \frac{\partial^2 C}{\partial x \, \partial v}
```

## 3. The two Hessian terms

### 3.1 Primal Hessian $\partial^2 C / \partial x^2$

This is the solver's Newton Hessian:

```math
H = M + J^T D_\text{active}\, J
```

where $`D_\text{active} = \text{diag}(D_1, \dots, D_{n_\text{efc}})`$ collects the per-constraint curvatures. $`D_i = 1/R_i`$ for constraints that are in the **quadratic** regime of the penalty; $`D_i = 0`$ for inactive constraints (see Section 5 for the full active-set logic).

### 3.2 Cross-Hessian $\partial^2 C / \partial x \, \partial v$

The gradient of $C$ with respect to $x$ is:

```math
\nabla_x C = M x - (\tau - c) + J^T D_\text{active}(Jx - a_\text{ref})
```

Differentiating with respect to $`v`$: the smooth forces contribute $`D_\text{smooth} = \partial(\tau - c)/\partial v`$ (the existing `qDeriv`), and $`a_\text{ref}`$ depends on $`v`$ through the damping term: $`\partial a_\text{ref} / \partial v = -B J`$, where $`B = \text{diag}(b_1, \dots, b_{n_\text{efc}})`$ collects the per-constraint damping coefficients (the `B` component of `efc_KBIP`). Therefore:

```math
\frac{\partial^2 C}{\partial x \, \partial v}
= -D_\text{smooth} + J^T D_\text{active}\, B\, J
```

### 3.3 Additional cross-Hessian term: $\dot{J}v$ bilinear correction

For constraints where $`a_\text{ref}`$ includes a $`\dot{J} v`$ term (connect and weld equality constraints), there is an additional contribution to the cross-Hessian. The quantity $`\dot{J} v`$ is **bilinear** in $`v`$ — the matrix $`\dot{J}`$ itself depends on $`v`$. This means:

```math
\frac{\partial (\dot{J} v)}{\partial v} = \dot{J} + \frac{\partial \dot{J}}{\partial v} v
```

The first term ($`\dot{J}`$) is the standard Jacobian-dot computed by `mj_jacDot`. The second term is a **bilinear correction** that must be added column-by-column. This correction contributes to $`J^T D_\text{active} B J`$ (it effectively modifies the "$B J$" factor for equality constraint rows). See Section 6 for details.

## 4. Final result

Combining the two Hessians:

```math
\boxed{
\frac{\partial \dot{v}^*}{\partial v}
= H^{-1}\bigl(D_\text{smooth} - J^T D_\text{active}\, B\, J\bigr)
}
```

where $`H = M + J^T D_\text{active}\, J`$.

The constraint force derivative we seek is then:

```math
\boxed{
\frac{\partial\, (J^T f)}{\partial v}
= M \frac{\partial \dot{v}^*}{\partial v} - D_\text{smooth}
}
```

This follows from the equations of motion: $`J^T f = M \dot{v}^* + c - \tau`$, so $`\partial(J^T f)/\partial v = M\, \partial \dot{v}^*/\partial v - D_\text{smooth}`$.

## 5. Active-set determination

The diagonal $`D_\text{active}`$ has a nonzero entry $`D_i`$ only when constraint $i$ is in the quadratic regime of the penalty. The rules differ by constraint type:

| Constraint type | Active condition ($`D_i \neq 0`$) |
|---|---|
| **Equality** | Constraint state is `QUADRATIC` |
| **Joint/tendon limit** | Post-solver constraint acceleration $`\text{jar}_i < 0`$ (limit engaged) |
| **DOF/tendon friction** | $`\vert\text{jar}_i\vert < R_i \cdot \text{floss}_i`$ (inside quadratic zone) |
| **Elliptic contact** | State is `QUADRATIC` **and** normal force $`f_{i_0} \neq 0`$ |
| **Other contacts** (frictionless, pyramidal) | State is `QUADRATIC` **and** $`f_i \neq 0`$ |

> [!IMPORTANT]
> The post-solver constraint acceleration $\text{jar} = J \dot{v}^* - a_\text{ref}$ must be **recomputed** from the solver's final $\dot{v}^*$, because the `efc_state` stored in `mjData` was computed *before* the solver ran and is therefore stale for limits and friction constraints.

## 6. Cone constraint contributions

When a contact constraint is in the `CONE` state (with nonzero force), its local Hessian $`H_\text{local}`$ is a dense $`\text{dim} \times \text{dim}`$ matrix (stored in `contact->H`). These contribute additional rank-$`\text{dim}`$ terms to *both* the primal Hessian and the cross-Hessian:

**Primal Hessian:**

```math
H \mathrel{+}= J_\text{cone}^T\, H_\text{local}\, J_\text{cone}
```

**Cross-Hessian** ($`J^T D_\text{active} B J`$ term):

```math
J^T D_\text{active} B J \mathrel{+}= J_\text{cone}^T\, H_\text{local}\, \text{diag}(B_\text{cone})\, J_\text{cone}
```

where $`J_\text{cone}`$ are the rows of $J$ for this contact and $`B_\text{cone}`$ are the corresponding damping coefficients.

## 7. The $\dot{J}v$ bilinear correction (connect/weld equalities)

For connect and weld equality constraints, $`a_\text{ref}`$ includes the term $`\dot{J}v`$, which is bilinear in $`v`$. Its full velocity derivative $`\partial(\dot{J}v)/\partial v`$ has two parts.

### 7.1 Translational part

For each DOF $k$ in the ancestor chain of a constrained body, the corrected column of $`\partial(\dot{J}v)/\partial v`$ is:

```math
\left[\frac{\partial (\dot{J}v)}{\partial v}\right]_{\!:,k}
= \underbrace{[\dot{J}]_{:,k}}_{\text{first-order}} + \underbrace{\text{correction}_k}_{\text{bilinear}}
```

The **first-order** $`\dot{J}`$ is computed by `mj_jacDot` (already available). The bilinear correction for DOF $k$ has two sub-terms:

#### (a) `cdof_dot` correction

Compute $`\text{crossMotion}(\xi_k, \, \nu_\text{eff})`$, where $`\xi_k \in \mathbb{R}^6`$ is the DOF's motion axis (`cdof[k]`), and $`\nu_\text{eff} \in \mathbb{R}^6`$ is the cumulative spatial velocity of all DOFs that come **after** $k$ in the ancestor chain:

```math
\nu_\text{eff} = \sum_{j \text{ after } k} \xi_j \, v_j
```

For quaternion DOFs (ball joints, or the rotational part of free joints), the effective velocity also includes lower-index quaternion siblings on the same body that have not yet been processed.

The cross-motion product is:

```math
\text{crossMotion}(\xi, \nu) =
\begin{pmatrix}
\omega_\xi \times \omega_\nu \\
\omega_\xi \times v_\nu + v_\xi \times \omega_\nu
\end{pmatrix}
```

where $`\xi = (\omega_\xi, v_\xi)`$ and $`\nu = (\omega_\nu, v_\nu)`$ use the angular-first spatial convention. The translational Jacobian column correction is then the linear part of this cross-motion, plus a lever-arm term:

```math
\text{corr}^\text{cdof}_k = (\omega_\xi \times v_\nu + v_\xi \times \omega_\nu) + (\omega_\xi \times \omega_\nu) \times p
```

where $p$ is the offset from the subtree center of mass to the anchor point.

#### (b) `pvel` correction

The body's angular velocity $`\omega_\text{body}`$ crossed with the translational Jacobian column at the anchor point:

```math
\text{corr}^\text{pvel}_k = \omega_\text{body} \times J^\text{trans}_{:,k}
```

where $`J^\text{trans}_{:,k} = v_{\xi_k} + \omega_{\xi_k} \times p`$.

#### Total translational correction

```math
\text{correction}_k = \text{corr}^\text{cdof}_k + \text{corr}^\text{pvel}_k
```

The corrected $`\partial(\dot{J}v)/\partial v`$ difference between the two bodies is accumulated into the cross-Hessian term $`J^T D_\text{active} B J`$:

```math
\text{for each row } r: \quad
\forall\, a: \quad [J^T D_\text{active} B J]_{a,k} \mathrel{+}= J_{r,a} \cdot D_r \cdot \left[\frac{\partial(\dot{J}v)}{\partial v}\right]_{r,k}
```

### 7.2 Rotational part (weld only)

Weld constraints have 3 additional rotational rows. The rotational constraint acceleration involves a quaternion triple product:

```math
(\dot{J}v)_\text{rot} = \tfrac{1}{2}\, \tau_s \cdot \text{Im}\bigl(\bar{q}_1 \cdot (0, \Delta\omega) \cdot q_{0r}\bigr)
```

where:
- $q_0, q_1$ are the body quaternions
- $`q_{0r} = q_0 \cdot q_\text{ref}`$ (with reference quaternion from constraint data)
- $`\bar{q}_1 = \text{conj}(q_1)`$
- $`\Delta\omega = \omega_1 - \omega_2`$ (angular velocity difference)
- $`\tau_s`$ is the torquescale parameter

Differentiating this with respect to $`v_k`$ produces **five terms** from the product rule on the three $`v`$-dependent factors ($`\bar{q}_1`$, $`\Delta\omega`$, $`q_{0r}`$):

| Term | Expression | Source |
|------|-----------|--------|
| T1a | $`\frac{\partial \bar{\dot{q}}_1}{\partial v_k} \cdot (0,\Delta\omega) \cdot q_{0r}`$ | $`v_k`$ affects $`\dot{q}_1`$ via $`\omega_2`$ |
| T1b | $`\bar{\dot{q}}_1 \cdot (0, \text{axis}_k) \cdot q_{0r}`$ | $`v_k`$ affects $`\Delta\omega`$ |
| T2  | $`\bar{q}_1 \cdot (0, [\partial(\dot{J}_\text{rot} v)/\partial v]_{:,k}) \cdot q_{0r}`$ | bilinear correction to $`\Delta\dot{\omega}`$ |
| T3a | $`\bar{q}_1 \cdot (0, \text{axis}_k) \cdot \dot{q}_{0r}`$ | $`v_k`$ affects $`\Delta\omega`$ (second occurrence) |
| T3b | $`\bar{q}_1 \cdot (0, \Delta\omega) \cdot \frac{\partial \dot{q}_{0r}}{\partial v_k}`$ | $`v_k`$ affects $`\dot{q}_0`$ via $`\omega_1`$ |

where $`\text{axis}_k = J^\text{rot}_{1,:,k} - J^\text{rot}_{2,:,k}`$ is the rotational Jacobian difference at DOF $k$, and $`\dot{q} = \frac{1}{2}(0, \omega) \cdot q`$ is the quaternion time derivative.

The combined rotational correction per DOF $`k`$ is:

```math
\left[\frac{\partial(\dot{J}v)_\text{rot}}{\partial v}\right]_{\!:,k} = \tfrac{1}{2}\, \tau_s \cdot \text{Im}(T_{1a} + T_{1b} + T_2 + T_{3a} + T_{3b})
```

## 8. Algorithm summary

```
Inputs:  M, J, qDeriv (smooth force Jacobian), efc_D, efc_KBIP, efc_force,
         efc_state, qacc (solver output), efc_aref, contact Hessians

1.  Compute D_smooth: scatter sparse qDeriv into dense n_v × n_v matrix
2.  Recompute jar = J·qacc - aref  (post-solver constraint acceleration)
3.  Determine D_active[i] for each constraint (Section 5)
4.  Compute DB[i] = D_active[i] · B[i]   (B[i] = efc_KBIP[4i+1])
5.  Compute JtDBJ = Jᵀ diag(DB) J  (base cross-Hessian contribution)
6.  Add Jdotv bilinear correction to JtDBJ (Section 7, connect/weld only)
7.  Compute H = M + Jᵀ D_active J  (primal Hessian)
8.  Add cone contributions to both H and JtDBJ (Section 6)
9.  Cholesky-factor H
10. Form cross = D_smooth - JtDBJ
11. Solve H⁻¹ cross column-by-column → store in cross
12. result = M · cross - D_smooth
```

> [!NOTE]
> The result matrix $`\partial(J^T f)/\partial v`$ is **symmetric** (it is a Hessian of a scalar energy-like quantity). This can be used as a sanity check against finite differences.

## 9. Finite-difference verification

The companion function `mjd_constraint_velFD` provides a centered finite-difference approximation for testing:

```
for each DOF i:
    save state
    v[i] += eps;  forward(skip_pos);  plus  = qfrc_constraint
    v[i] -= 2eps; forward(skip_pos);  minus = qfrc_constraint
    result[:, i] = (plus - minus) / (2 eps)
    restore state
```

Typical agreement: relative error < 5% with $`\varepsilon = 10^{-6}`$ and absolute tolerance $10^{-6}$.

yuvaltassa:
PS of course if you don't care about the Jdot term, which is only for equality constraints anyway, the above derivation becomes much simpler.

yuvaltassa:
Ah [found it](https://www.roboticsproceedings.org/rss03/p23.pdf) is was a 2008 paper.

Of course the foundational paper on the subject is [Montana 1988](https://pages.github.berkeley.edu/EECS-106/sp22-site/assets/misc/David%20Montana.1988.pdf).

yuvaltassa:
PPS let me know if you want me to publish this function. I have it in C, publishing an inefficient version would be really easy and an efficient one would be hard. This is because I only have dense and I  recompute H. 

- MuJoCo C doesn't save the Newton H, it gets popped off the mjData stack. Saving it rather than recomputing would be a serious PITA structurally, but definitely possible. 
- Writing this all with sparse math is also nontrivial since we'd need a sparse LU solver (the cross Hessian is not symmetric). 

All doable but I hesitate to make the effort without a clear use case. The somewhat wasteful dense function is however ready to go.

etaoxing:
@yuvaltassa

> CRB (inertia) derivatives

No, inertia derivatives are not approximated, and we do not attempt to differentiate CRB. In reverse mode, we only have contractions of the form ∂_q [y^T M(q) w], so we can do one RNE pass with zero velocity and gravity at the converged acc from forward step.

> Trinkle ball rolling on a plane

Sorry I think my "frozen witness narrowphase" phrasing was unclear. We do not freeze the contact geometry (point / normal / distance), so I think we do have the "correct" result for the [sphere/plane contact](https://github.com/google-deepmind/mujoco_warp/blob/d28130a8798ee80b25de843d529d5b95627df4fa/mujoco_warp/_src/collision_primitive_core.py#L48):

```math
d = n^T(c - p) - r, \quad
x_c = c - n\,(r + d/2)
\;\Rightarrow\;
\frac{\partial x_c}{\partial c} = I - \tfrac12 n n^T,
\quad
\frac{\partial x_c}{\partial R_{ball}} = 0,
```

so the contact point moves sideways one-for-one with the center, orientation has no effect, and the
vertical rate along rolling is zero, like you mentioned.

> PS: Jdot term, which is only for equality constraints anyway

Yes, this PR does not currently support connect/weld, but should be able to add.

> PPS: C version of mjd_constraint_vel()

Yes, this PR also recomputes the Newton H. I think this is ok from a "gradient checkpointing" perspective, where we want to save on memory when backpropagating through multiple timesteps. 

Since this PR is not explicitly computing ∂(qfrc_constraint)/∂v, not sure how we could use mjd_constraint_vel() as a oracle test case against this branch, but would be interesting to see your approach if there's a Mujoco branch to look at!

> math behind mjd_constraint_vel()

Thanks for your note on this! I spent some time at formalizing the math for this "hybrid analytic differentiability" PR https://gist.github.com/etaoxing/af0823e7e930ff7771c752e45cf6f53b, trying to follow the Mujoco docs math / definitions as close as possible. Looking forward to hearing your thoughts and hope it helps clarify some things!

thowell:
@etaoxing what is the status on this work? is this code ready for review? what is the minimum warp version that contains all necessary features? thanks!

etaoxing:
@thowell This is ready for a review now! Large refactor of the original July implementation for human readability, and merged the latest changes from main. It uses the pip installable Warp version specified by MJWarp. 

- added the demo examples into `contrib/diffsim`. there are a lot of demos (some require additional XML scenes like Go1 quadruped in `benchmarks/` which are omitted for now). should we incorporate them into the `benchmarks/` tool instead? I can move some of these demos to Newton also. 
- still need to refactor and trim down the test cases, so those are also omitted for now.
- there are some changes in `collision_primitive_core.py‎` that decompose `capsule_box()`, I'll open a different PR for them.
- `adjoint_recompute.py` duplicates some kernels from the forward pass with `enable_grad=True`. We could probably decompose the relevant forward kernels into `wp.func`s that can be called from the adjoint code. Alternatively, we could set those forward kernels as `enable_backward=True`, but that would require a newer version of Warp containing this [bugfix](https://github.com/NVIDIA/warp/commit/27df5c23def4c3838f094d0fefaa7ca692c35364).

- convex narrowphase adjoint will be handled in a future PR.

Can also share a tech report soon for more details on how gradients are computed. Demo videos and plots are available [here](https://etaoxing.com/reports/diff_mjw/diff_mjw_demos.html).

PR #1537, opened 2026-07-20
Simplify dense elliptic Hessian contraction

adenzler-nvidia:
Dense Newton currently materializes every entry of the elliptic cone curvature matrix and then performs a quadratic contraction over the contact dimensions for each generalized-coordinate pair.

This change uses the curvature's diagonal-plus-rank-one structure directly. It computes two projections and a tangent dot product for each Hessian entry, then evaluates the resulting scalar contraction. Work per entry falls from quadratic to linear in the contact dimension without storing the intermediate cone Hessian. Mixed 3-, 4-, and 6-dimensional contacts continue to dispatch at runtime.

The dense regression test constructs a mixed 3/4/6 contact model and compares the generated Hessian with an independent NumPy construction of `J^T C J`.

Performance was measured on an RTX PRO 6000 Blackwell with 8,192 worlds and 1,000 steps. Values are medians of five interleaved runs after unmeasured warmups, comparing `upstream/main` (`63d25e44`) with this PR (`90191e81`). Both benchmarks force Newton, elliptic cones, and dense Jacobians. G1 uses the prerecorded `shuffle_dance.npz` trajectory; Humanoid uses its standard keyframe rollout. Every run converged all 8,192 worlds with unchanged solver iteration distributions.

| Benchmark | `upstream/main` | This PR | Difference |
|---|---:|---:|---:|
| Humanoid, dense elliptic | 2.271 s | 2.078 s | 8.5% faster |
| G1 flat, dense elliptic, prerecorded | 5.229 s | 4.539 s | 13.2% faster |

The focused mixed-contact dense test passes on CUDA. CI is green, including pre-commit and kernel-analyzer.

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -2737,6 +2737,28 @@ def _update_gradient_JTCJ_compact(
     wp.atomic_add(ctx_h_out[worldid, c_dof1], c_dof2, h)
 
 
+@wp.func
+def _elliptic_hessian_entry_from_projections(
+  # In:
+  dm: float,
+  mu_over_t: float,
+  mu_n_over_ttt: float,
+  tangent_diag: float,
+  z01: float,
+  z02: float,
+  projection1: float,
+  projection2: float,
+  tangent_dot: float,
+) -> float:
+  # Contract the diagonal-plus-rank-one curvature without materializing the cone Hessian.
+  return dm * (
+    z01 * z02
+    - mu_over_t * (z01 * projection2 + z02 * projection1)
+    + mu_n_over_ttt * projection1 * projection2
+    + tangent_diag * tangent_dot
+  )
+
+
 @wp.kernel
 def _update_gradient_JTCJ_dense(
   # Model:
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -2803,85 +2825,37 @@ def _update_gradient_JTCJ_dense(
...
-        hcone *= dm * fri1 * fri2
-
-        if hcone != 0.0:
-          h += hcone * efc_J11 * efc_J22
-
-          if dim1id != dim2id:
-            h += hcone * efc_J12 * efc_J21
+    h = _elliptic_hessian_entry_from_projections(
+      dm,
+      math.safe_div(mu, t),
```
thowell: what if we compute `mu_tinv = math.safe_div(mu, t)` above, then provide `mu_tinv` here, and then below `mu * math.safe_div(n, t)` -> `n * mu_tinv`?

thowell (approved):
nice! thanks @adenzler-nvidia

mujoco_warp/_src/solver.py
adenzler-nvidia: Done in 66df0bdb — hoisted `mu_tinv = math.safe_div(mu, t)` and rewrote the tangent-diagonal term as `mu2 - n * mu_tinv`. The `safe_div(n, ttt)` above stays as-is since it divides by t³.

Issue #1539, opened 2026-07-20
Proposal: replace quadratic island discovery with a GPU-appropriate linear-memory path


teerthsharma:
### The feature and motivation

MuJoCo Warp's current island discovery materializes both `tree_tree[nworld, ntree, ntree]` and `stack_scratch[nworld, ntree * ntree]` as `int32` tensors. Those two scratch tensors alone cost

`8 * nworld * ntree^2` bytes.

At the current `upstream/main` commit (`c50dd837`), the allocations and traversal are in [`island.py`](https://github.com/google-deepmind/mujoco_warp/blob/c50dd837aa5b1abe46d15e6343da382a2ea7711b/mujoco_warp/_src/island.py#L26-L292).

| `nworld` | `ntree` | two current scratch tensors |
|---:|---:|---:|
| 8192 | 64 | 256 MiB |
| 1024 | 256 | 512 MiB |
| 64 | 1024 | 512 MiB |
| 1 | 4096 | 128 MiB |

Native MuJoCo has the same quadratic-memory problem in google-deepmind/mujoco#3388 and an open minimum-root DSU implementation in google-deepmind/mujoco#3396. MuJoCo Warp depends on MuJoCo for model/data bindings but has its own Warp island kernels, so the native change does not remove these GPU allocations. The native DSU is a useful correctness baseline, but I do not think it should be ported blindly: a sequential worker per world can underutilize a GPU when `nworld` is small and `ntree` is large, while a parallel GPU connectivity algorithm adds atomics, convergence, determinism, and CUDA graph-capture tradeoffs.

The exact object is a finite hypergraph: dynamic trees are vertices and each active constraint row induces a hyperedge containing the trees in the nonzero support of that Jacobian row. Island discovery asks only for its connected-component partition, equivalently its `H_0`, with the minimum tree id as a canonical component label. This gives several exact linear-memory candidates without importing unrelated manifold machinery.

### Exact candidates screened

I independently screened three propositions against a bipartite tree/constraint BFS oracle. These are correctness experiments, not performance evidence.

1. **Incidence-star DSU.** Replacing every nonempty hyperedge by a star through any representative preserves connected components. Minimum-root DSU is therefore exact regardless of representative, row order, endpoint order, or duplicates. A direct implementation needs `parent[nworld, ntree]` plus incidence scratch and is the simplest production baseline. It matched 20,010 special, random, and metamorphic cases in an isolated CPU test.

[...]

PR #1541, opened 2026-07-20
Replace quadratic island discovery with a linear-memory GPU DSU

teerthsharma:
This replaces the quadratic `tree_tree[nworld, ntree, ntree]` adjacency and DFS stack with a persistent `int32 parent[nworld, ntree]` workspace and three allocation-free, CUDA-capturable phases: reset, direct EFC incidence unions, and one canonical compression/label pass per world. The latest iteration assigns one 32-lane CUDA block to each world and has lanes stride only the active EFC prefix instead of launching `nworld * njmax` capacity threads.

Repeated fixed-support scalar rows with the same `(efc_type, efc_id)` are quotiented before endpoint lookup. This is exact for connect/weld equality constraints, joint friction and limits, and non-flex contacts because duplicate incidence generates the same H0 equivalence relation. Generic dense or sparse Jacobian rows and negative-geometry flex contacts remain fully scanned. Component labels use the minimum active tree id, and the existing canonical DOF/EFC mapping produces bitwise-stable island inputs independent of row order, endpoint order, and duplicate incidence.

The retained `aloha_clutter` checkpoint has 2,048 worlds, 22 trees, 247,308 active EFC rows, and 193,882 consecutive fixed-support repeats. Across five fresh processes and 10 A-B-B-A blocks, the current discovery path matched the original dense kernels exactly in every block and was 1.513x faster by geometric mean, with a conservative process-bootstrap 99% interval of [1.406x, 1.584x]. Pooled medians were 156.738 microseconds for dense and 101.097 microseconds for the current path. Compared with the previous capacity-wide atomic implementation, the new schedule was 1.355x faster with 99% interval [1.265x, 1.420x]. Reusable scratch falls from 7,929,856 bytes to 180,224 bytes for this configuration, a 97.73% reduction and a change from quadratic to linear growth in `ntree`.

The earlier 582-record synthetic matrix found the linear DSU faster than dense discovery in 91 of 93 valid pairs with a median dense/DSU ratio of 2.40. I also tested CUDA block dimensions 32, 64, 128, and 256; all were exact. A dedicated five-process 32-vs-128 comparison had geometric ratio 1.004 with 99% interval [0.990, 1.029], so I kept the simpler one-warp schedule rather than claiming a wider-block improvement.

The deterministic same-state `fwd_position` boundary matched all topology and downstream mapping arrays exactly across five fresh processes. Its 1.013x geometric speedup had a 99% interval of [0.967x, 1.053x], so it is not a performance claim. Two full 2,751-frame replay pairs measured 1.76-3.54x faster island events and 3.72-3.93x faster direct discovery scopes, but whole-step throughput remained 7-12% lower as the chaotic trajectories diverged in non-island `forward` work. Candidate-vs-itself and dense-vs-itself controls also show that full-replay qacc and EFC ordering are not bitwise repeatable on this workload. This PR therefore claims the exact direct-discovery and memory improvements only, not an end-to-end step speedup.

Validation includes 42 focused island tests plus four subtests on CUDA, 41 passed and one skipped plus four subtests on CPU, 90 CUDA-error-checked non-capture tests, graph capture/replay coverage, pre-commit Ruff/format/kernel-analyzer checks, and `uv run pytest -n 8` with 1,217 passed and 18 skipped. The existing unstaged `mesh_plane_complex` fixture offset remains outside this PR.

Addresses #1539.

### DIFF
--- a/mujoco_warp/_src/island.py
+++ b/mujoco_warp/_src/island.py
@@ -25,8 +25,88 @@ from mujoco_warp._src.warp_util import event_scope
 wp.set_module_options({"default_grid_stride": False})
 
 
+@wp.func
+def _dsu_find(parent: wp.array2d[int], worldid: int, tree: int):
+  """Find a root while shortening its strictly decreasing parent path."""
+  current = tree
+  while True:
+    parent_current = parent[worldid, current]
+    if parent_current == current:
+      return current
+    grandparent = parent[worldid, parent_current]
+    if grandparent < parent_current:
+      wp.atomic_min(parent, worldid, current, grandparent)
+    current = parent_current
+  return current
+
+
+@wp.func
+def _dsu_activate(worldid: int, tree: int, tree_island_out: wp.array2d[int]):
+  if tree >= 0:
+    wp.atomic_max(tree_island_out, worldid, tree, 0)
+
+
+@wp.func
+def _dsu_union(parent: wp.array2d[int], worldid: int, tree0: int, tree1: int, tree_island_out: wp.array2d[int]):
+  """Activate endpoints and atomically hook the higher root to the lower root."""
+  if tree0 < 0:
+    tree0 = tree1
+    tree1 = -1
+  if tree0 < 0:
+    return
+  _dsu_activate(worldid, tree0, tree_island_out)
+  if tree1 < 0:
+    return
+  _dsu_activate(worldid, tree1, tree_island_out)
+
+  while True:
+    root0 = _dsu_find(parent, worldid, tree0)
+    root1 = _dsu_find(parent, worldid, tree1)
+    if root0 == root1:
+      return
+    low_root = wp.min(root0, root1)
+    high_root = wp.max(root0, root1)
+    previous = wp.atomic_cas(parent, worldid, high_root, high_root, low_root)
+    if previous == high_root:
+      return
+
+
 @wp.kernel
-def _tree_edges(
+def _reset_dsu(
+  # Data out:
+  nisland_out: wp.array[int],
+  tree_island_out: wp.array2d[int],
+  # Out:
+  island_parent_out: wp.array2d[int],
+):
+  """Reset identity parents and active/output markers in parallel."""
+  worldid, treeid = wp.tid()
+  island_parent_out[worldid, treeid] = treeid
+  tree_island_out[worldid, treeid] = -1
+  if treeid == 0:
+    nisland_out[worldid] = 0
+
+
+@wp.func
+def _is_repeated_fixed_support_efc(
+  # Data in:
+  efc_type_in: wp.array2d[int],
+  efc_id_in: wp.array2d[int],
+  # In:
+  worldid: int,
+  efcid: int,
+) -> bool:
+  """Return whether the preceding scalar row has the same fixed tree support."""
+  if efcid == 0:
+    return False
+  return (
+    efc_type_in[worldid, efcid - 1] == efc_type_in[worldid, efcid]
+    and efc_id_in[worldid, efcid - 1] == efc_id_in[worldid, efcid]
+  )
+
+
+@wp.kernel
+def _island_dsu(
   # Model:
   nv: int,
   body_treeid: wp.array[int],
@@ -49,134 +129,178 @@ def _tree_edges(
   efc_J_colind_in: wp.array3d[int],
   efc_J_in: wp.array3d[float],
   njmax_in: int,
+  # In:
+  chunk_size: int,
+  # Data out:
+  tree_island_out: wp.array2d[int],
   # Out:
-  tree_tree: wp.array3d[int],  # kernel_analyzer: off
+  island_parent_out: wp.array2d[int],
 ):
-  """Find tree edges."""
-  worldid, efcid = wp.tid()
+  """Process one chunk of one world's active EFC prefix with one strided CUDA warp.
 
-  # skip if beyond active constraints
-  if efcid >= wp.min(njmax_in, nefc_in[worldid]):
-    return
+  Blocks are laid out over (world, chunk) rather than world alone, so resident
+  parallelism tracks total constraint capacity instead of batch size. A single world
+  with a long prefix fills the device the same way a large batch of short ones does.
+  """
+  worldid, chunkid, lane = wp.tid()
+  chunk_beg = chunkid * chunk_size
+  chunk_end = wp.min(chunk_beg + chunk_size, wp.min(njmax_in, nefc_in[worldid]))
+  for efcid in range(chunk_beg + lane, chunk_end, wp.block_dim()):
+    efc_type = efc_type_in[worldid, efcid]
+    efc_id = efc_id_in[worldid, efcid]
+    repeated = _is_repeated_fixed_support_efc(efc_type_in, efc_id_in, worldid, efcid)
+    tree0 = int(-1)
+    tree1 = int(-1)
+    use_generic = int(0)
 
-  efc_type = efc_type_in[worldid, efcid]
-  efc_id = efc_id_in[worldid, efcid]
-
-  tree0 = int(-1)
-  tree1 = int(-1)
-  use_generic = int(0)
-
-  # equality (connect/weld)
-  if efc_type == ConstraintType.EQUALITY:
-    eq_t = eq_type[efc_id]
-
-    if eq_t == EqType.CONNECT or eq_t == EqType.WELD:
-      b1 = eq_obj1id[efc_id]
-      b2 = eq_obj2id[efc_id]
-
-      # site semantics
-      if eq_objtype[efc_id] == ObjType.SITE:
-        b1 = site_bodyid[b1]
-        b2 = site_bodyid[b2]
-
-      tree0 = body_treeid[b1]
-      tree1 = body_treeid[b2]
-    else:
-      # JOINT, TENDON, FLEX
-      use_generic = 1
-
-  # joint friction
-  elif efc_type == ConstraintType.FRICTION_DOF:
-    tree0 = dof_treeid[efc_id]
-
-  # joint limit
-  elif efc_type == ConstraintType.LIMIT_JOINT:
-    tree0 = dof_treeid[jnt_dofadr[efc_id]]
-
-  # contact
-  elif (
-    efc_type == ConstraintType.CONTACT_FRICTIONLESS
-    or efc_type == ConstraintType.CONTACT_PYRAMIDAL
-    or efc_type == ConstraintType.CONTACT_ELLIPTIC
-  ):
-    geom_pair = contact_geom_in[efc_id]
-    g1 = geom_pair[0]
-    g2 = geom_pair[1]
-
-    # flex contacts have negative geom ids
-    if g1 >= 0 and g2 >= 0:
-      tree0 = body_treeid[geom_bodyid[g1]]
-      tree1 = body_treeid[geom_bodyid[g2]]
-    else:
-      use_generic = 1
-
-  # generic
-  else:
-    use_generic = 1
-
-  # handle static bodies
-  if use_generic == 0:
-    # swap so tree0 is non-negative if possible
-    if tree0 < 0 and tree1 >= 0:
-      tree0 = tree1
-      tree1 = -1
-
-    # mark the edge
-    if tree0 >= 0:
-      if tree1 < 0 or tree0 == tree1:
-        # self-edge
-        wp.atomic_max(tree_tree, worldid, tree0, tree0, 1)
+    if efc_type == ConstraintType.EQUALITY:
+      eq_t = eq_type[efc_id]
+      if eq_t == EqType.CONNECT or eq_t == EqType.WELD:
+        if repeated:
+          continue
+        body0 = eq_obj1id[efc_id]
+        body1 = eq_obj2id[efc_id]
+        if eq_objtype[efc_id] == ObjType.SITE:
+          body0 = site_bodyid[body0]
+          body1 = site_bodyid[body1]
+        tree0 = body_treeid[body0]
+        tree1 = body_treeid[body1]
       else:
-        # cross-tree edge
-        t1 = wp.min(tree0, tree1)
-        t2 = wp.max(tree0, tree1)
-        wp.atomic_max(tree_tree, worldid, t1, t2, 1)
-        wp.atomic_max(tree_tree, worldid, t2, t1, 1)
-    return
-
-  # generic: scan Jacobian row
-  first_tree = int(-1)
-  has_cross_edge = int(0)
-
-  count = nv
-  rowadr = 0
-  if is_sparse:
-    count = efc_J_rownnz_in[worldid, efcid]
-    rowadr = efc_J_rowadr_in[worldid, efcid]
-
-  for i in range(count):
-    dof = i
-    if is_sparse:
-      sparseid = rowadr + i
-      dof = efc_J_colind_in[worldid, 0, sparseid]
-    else:
-      J_val = efc_J_in[worldid, efcid, dof]
-      if J_val == 0.0:
+        use_generic = 1
+    elif efc_type == ConstraintType.FRICTION_DOF:
+      if repeated:
         continue
+      tree0 = dof_treeid[efc_id]
+    elif efc_type == ConstraintType.LIMIT_JOINT:
+      if repeated:
+        continue
+      tree0 = dof_treeid[jnt_dofadr[efc_id]]
+    elif (
+      efc_type == ConstraintType.CONTACT_FRICTIONLESS
+      or efc_type == ConstraintType.CONTACT_PYRAMIDAL
+      or efc_type == ConstraintType.CONTACT_ELLIPTIC
+    ):
+      geom_pair = contact_geom_in[efc_id]
+      if geom_pair[0] >= 0 and geom_pair[1] >= 0:
+        if repeated:
+          continue
+        tree0 = body_treeid[geom_bodyid[geom_pair[0]]]
+        tree1 = body_treeid[geom_bodyid[geom_pair[1]]]
+      else:
+        use_generic = 1
+    else:
+      use_generic = 1
 
-    tree = dof_treeid[dof]
-    if tree < 0:
+    if use_generic == 0:
+      _dsu_union(island_parent_out, worldid, tree0, tree1, tree_island_out)
       continue
-    if first_tree == -1:
-      first_tree = tree
-    elif tree != first_tree:
-      t1 = wp.min(first_tree, tree)
-      t2 = wp.max(first_tree, tree)
-      wp.atomic_max(tree_tree, worldid, t1, t2, 1)
-      wp.atomic_max(tree_tree, worldid, t2, t1, 1)
-      has_cross_edge = 1
 
-  if first_tree >= 0 and has_cross_edge == 0:
-    wp.atomic_max(tree_tree, worldid, first_tree, first_tree, 1)
+    first_tree = int(-1)
+    count = nv
+    rowadr = int(0)
+    if is_sparse:
+      count = efc_J_rownnz_in[worldid, efcid]
+      rowadr = efc_J_rowadr_in[worldid, efcid]
+
+    for index in range(count):
+      dof = index
+      if is_sparse:
+        dof = efc_J_colind_in[worldid, 0, rowadr + index]
+      elif efc_J_in[worldid, efcid, dof] == 0.0:
+        continue
+      tree = dof_treeid[dof]
+      if tree < 0:
+        continue
+      if first_tree < 0:
+        first_tree = tree
+        _dsu_union(island_parent_out, worldid, tree, -1, tree_island_out)
+      else:
+        _dsu_union(island_parent_out, worldid, first_tree, tree, tree_island_out)
+
+
+@wp.kernel
+def _compress_roots(
+  # Data in:
+  tree_island_in: wp.array2d[int],
+  # Data out:
+  # Out:
+  island_parent_out: wp.array2d[int],
+):
+  """Point every active tree straight at its component's minimum-tree root."""
+  worldid, treeid = wp.tid()
+  if tree_island_in[worldid, treeid] >= 0:
+    island_parent_out[worldid, treeid] = _dsu_find(island_parent_out, worldid, treeid)
+
+
+@wp.kernel
+def _label_roots(
+  # Model:
+  ntree: int,
+  # Data in:
+  # In:
+  island_parent_in: wp.array2d[int],
+  # Data out:
+  nisland_out: wp.array[int],
+  tree_island_out: wp.array2d[int],
+):
+  """Number the roots in ascending tree order, which is MuJoCo's island order.
+
+  Ranking roots is the one inherently sequential step: island i is the i-th smallest
+  root, so it depends on every preceding tree. It reads two ints per tree and does no
+  pointer chasing, leaving the compress and propagate passes free to run per tree.
+  """
+  worldid = wp.tid()
+  nisland = int(0)
+  for treeid in range(ntree):
+    if tree_island_out[worldid, treeid] >= 0 and island_parent_in[worldid, treeid] == treeid:
+      tree_island_out[worldid, treeid] = nisland
+      nisland += 1
+  nisland_out[worldid] = nisland
+
+
+@wp.kernel
+def _propagate_labels(
+  # Data in:
+  # In:
+  island_parent_in: wp.array2d[int],
+  # Data out:
+  tree_island_out: wp.array2d[int],
+):
+  """Copy each root's label down to the trees that point at it.
+
+  A root reads and writes its own slot with the same value, so the only concurrent
+  writes to any slot store identical data.
+  """
+  worldid, treeid = wp.tid()
+  if tree_island_out[worldid, treeid] >= 0:
+    tree_island_out[worldid, treeid] = tree_island_out[worldid, island_parent_in[worldid, treeid]]
+
+
+# Smallest EFC chunk worth its own block, and the resident-block count worth reaching for.
+_DSU_MIN_CHUNK = 256
+_DSU_TARGET_BLOCKS = 2048
 
 
 @event_scope
-def tree_edges(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
-  """Compute tree-tree adjacency matrix."""
-  tree_tree.zero_()
+def direct_dsu(m: types.Model, d: types.Data, island_parent: wp.array2d[int]):
+  """Discover islands with EFC-parallel atomic minimum-root hooks.
+
+  `island_parent` is the (nworld, ntree) disjoint-set workspace, owned by the caller.
+  """
+  # Discovery blocks are laid out over (world, chunk). A batch large enough to occupy the
+  # device keeps one block per world, since extra chunks would mostly launch past the
+  # active prefix; a small batch with a long prefix is split until there is resident work.
+  max_chunks = max(1, -(-d.njmax // _DSU_MIN_CHUNK))
+  nchunk = min(max(1, -(-_DSU_TARGET_BLOCKS // d.nworld)), max_chunks)
+  chunk_size = -(-d.njmax // nchunk)
   wp.launch(
-    kernel=_tree_edges,
-    dim=(d.nworld, d.njmax),
+    _reset_dsu,
+    dim=(d.nworld, m.ntree),
+    inputs=[d.nisland, d.tree_island, island_parent],
+  )
+  wp.launch_tiled(
+    _island_dsu,
+    dim=(d.nworld, nchunk),
     inputs=[
       m.nv,
       m.body_treeid,
@@ -198,84 +322,26 @@ def tree_edges(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
       d.efc.J_colind,
       d.efc.J,
       d.njmax,
+      chunk_size,
+      d.tree_island,
+      island_parent,
     ],
-    outputs=[tree_tree],
+    block_dim=types.BlockDim.island_dsu,
   )
-
-
-@wp.kernel
-def _flood_fill(
-  # Model:
-  ntree: int,
-  # In:
-  tree_tree_in: wp.array3d[int],
-  labels_in: wp.array2d[int],
-  stack_in: wp.array2d[int],
-  # Data out:
-  nisland_out: wp.array[int],
-  tree_island_out: wp.array2d[int],
-  # Out:
-  stack_out: wp.array2d[int],
-):
-  """DFS flood fill to discover islands using tree_tree matrix."""
-  worldid = wp.tid()
-  nisland = int(0)
-
-  # iterate over trees
-  for i in range(ntree):
-    # already assigned
-    if labels_in[worldid, i] != -1:
-      continue
-
-    # check if tree has any edges
-    has_edge = int(0)
-    for j in range(ntree):
-      if tree_tree_in[worldid, i, j] != 0:
-        has_edge = 1
-        break
-    if has_edge == 0:
-      continue
-
-    # DFS: push i onto stack
-    nstack = int(0)
-    stack_out[worldid, nstack] = i
-    nstack = nstack + 1
-
-    while nstack > 0:
-      # pop v from stack
-      nstack = nstack - 1
-      v = stack_in[worldid, nstack]
-
-      # already assigned
-      if labels_in[worldid, v] != -1:
-        continue
-
-      # assign to current island
-      tree_island_out[worldid, v] = nisland
-
-      # push neighbors
-      for neighbor in range(ntree):
-        if tree_tree_in[worldid, v, neighbor] != 0:
-          if labels_in[worldid, neighbor] == -1:
-            stack_out[worldid, nstack] = neighbor
-            nstack = nstack + 1
-
-    # island filled
-    nisland = nisland + 1
-
-  nisland_out[worldid] = nisland
-
-
-@event_scope
-def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
-  d.tree_island.fill_(-1)
-  stack_scratch = wp.empty((d.nworld, m.ntree * m.ntree), dtype=int)
-
   wp.launch(
-    _flood_fill,
+    _compress_roots,
+    dim=(d.nworld, m.ntree),
+    inputs=[d.tree_island, island_parent],
+  )
+  wp.launch(
+    _label_roots,
     dim=d.nworld,
-    inputs=[m.ntree, tree_tree, d.tree_island, stack_scratch],
-    outputs=[d.nisland, d.tree_island, stack_scratch],
+    inputs=[m.ntree, island_parent, d.nisland, d.tree_island],
+  )
+  wp.launch(
+    _propagate_labels,
+    dim=(d.nworld, m.ntree),
+    inputs=[island_parent, d.tree_island],
   )
 
 
@@ -286,18 +352,16 @@ def island(m: types.Model, d: types.Data):
     d.nisland.zero_()
     return
 
-  # Step 1: Find tree edges
-  tree_tree = wp.zeros((d.nworld, m.ntree, m.ntree), dtype=int)
-  tree_edges(m, d, tree_tree)
-
-  # Step 2: DFS flood fill
-  flood_fill(m, d, tree_tree)
+  direct_dsu(m, d, wp.empty((d.nworld, m.ntree), dtype=int))
 
 
 @wp.kernel
 def _island_count_dofs(
+  # Model:
   dof_treeid: wp.array[int],
+  # Data in:
   tree_island_in: wp.array2d[int],
+  # Data out:
   dof_island_out: wp.array2d[int],
   island_nv_out: wp.array2d[int],
 ):
@@ -311,12 +375,14 @@ def _island_count_dofs(
 
 @wp.kernel
 def _island_scan_sizes(
+  # Data in:
   nisland_in: wp.array[int],
-  island_idofadr_out: wp.array2d[int],
-  island_nv_inout: wp.array2d[int],
-  island_nefc_inout: wp.array2d[int],
-  island_iefcadr_out: wp.array2d[int],
+  island_nv_in: wp.array2d[int],
+  # Data out:
   nidof_out: wp.array[int],
+  island_idofadr_out: wp.array2d[int],
+  island_nefc_out: wp.array2d[int],
+  island_iefcadr_out: wp.array2d[int],
 ):
   worldid = wp.tid()
 
@@ -329,44 +395,58 @@ def _island_scan_sizes(
   island_idofadr_out[worldid, 0] = 0
   island_iefcadr_out[worldid, 0] = 0
   for i in range(1, nisland):
-    island_idofadr_out[worldid, i] = island_idofadr_out[worldid, i - 1] + island_nv_inout[worldid, i - 1]
-    island_iefcadr_out[worldid, i] = island_iefcadr_out[worldid, i - 1] + island_nefc_inout[worldid, i - 1]
+    island_idofadr_out[worldid, i] = island_idofadr_out[worldid, i - 1] + island_nv_in[worldid, i - 1]
+    island_iefcadr_out[worldid, i] = island_iefcadr_out[worldid, i - 1] + island_nefc_out[worldid, i - 1]
 
-  nidof = island_idofadr_out[worldid, nisland - 1] + island_nv_inout[worldid, nisland - 1]
+  nidof = island_idofadr_out[worldid, nisland - 1] + island_nv_in[worldid, nisland - 1]
   nidof_out[worldid] = nidof
 
-  # Reset for recount
+  # Reset for recount ( island_nv is final after _island_count_dofs and stays put )
   for i in range(nisland):
-    island_nv_inout[worldid, i] = 0
-    island_nefc_inout[worldid, i] = 0
+    island_nefc_out[worldid, i] = 0
 
 
 @wp.kernel
 def _island_map_dofs(
-  nv: int,
-  dof_island_in: wp.array2d[int],
-  island_idofadr_in: wp.array2d[int],
+  # Model:
+  dof_treeid: wp.array[int],
+  tree_dofadr: wp.array[int],
+  tree_dofnum: wp.array[int],
+  # Data in:
   nidof_in: wp.array[int],
-  island_nv_inout: wp.array2d[int],
+  tree_island_in: wp.array2d[int],
+  island_idofadr_in: wp.array2d[int],
+  # Data out:
   island_dofadr_out: wp.array2d[int],
   map_dof2idof_out: wp.array2d[int],
   map_idof2dof_out: wp.array2d[int],
+  # Out:
   idof_islandid_out: wp.array2d[int],
-  unconstrained_cnt_inout: wp.array2d[int],
 ):
+  """Map DOFs in ascending global order for deterministic solver inputs.
+
+  A DOF's rank inside its island is fixed by the tree layout: DOFs are contiguous per
+  tree and trees are ordered, so the rank is the DOF count of every preceding tree
+  carrying the same label. That closed form makes the ascending order reproducible
+  without a serial per-world scan.
+  """
   worldid, dofid = wp.tid()
 
-  nidof = nidof_in[worldid]
-  island_id = dof_island_in[worldid, dofid]
+  tree = dof_treeid[dofid]
+  island_id = tree_island_in[worldid, tree]
+
+  local_idx = dofid - tree_dofadr[tree]
+  for t in range(tree):
+    if tree_island_in[worldid, t] == island_id:
+      local_idx += tree_dofnum[t]
 
   if island_id >= 0:
-    local_idx = wp.atomic_add(island_nv_inout, worldid, island_id, 1)
     idof = island_idofadr_in[worldid, island_id] + local_idx
     idof_islandid_out[worldid, idof] = island_id
-    wp.atomic_min(island_dofadr_out, worldid, island_id, dofid)
+    if local_idx == 0:
+      island_dofadr_out[worldid, island_id] = dofid
   else:
-    cnt = wp.atomic_add(unconstrained_cnt_inout, worldid, 0, 1)
-    idof = nidof + cnt
+    idof = nidof_in[worldid] + local_idx
 
   map_dof2idof_out[worldid, dofid] = idof
   map_idof2dof_out[worldid, idof] = dofid
@@ -374,11 +454,14 @@ def _island_map_dofs(
 
 @wp.kernel
 def _island_count_constraints(
+  # Data in:
   nefc_in: wp.array[int],
-  njmax_in: int,
-  efc_tree_in: wp.array2d[int],
-  tree_island_in: wp.array2d[int],
   efc_type_in: wp.array2d[int],
+  tree_island_in: wp.array2d[int],
+  njmax_in: int,
+  # In:
+  efc_tree_in: wp.array2d[int],
+  # Data out:
   efc_island_out: wp.array2d[int],
   island_nefc_out: wp.array2d[int],
   island_ne_out: wp.array2d[int],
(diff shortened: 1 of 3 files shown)


teerthsharma:
@thowell I have pushed the first complete implementation and made the evidence explicit in the PR description. The linear-memory and exact same-state correctness goals are met, including canonical DOF/EFC mappings and CUDA capture, but the controlled Aloha pair is not yet an end-to-end improvement: DSU was 4.2% lower in steps/s and 10.2% higher in island event time than the deterministic dense control, with allocator-rounded total memory unchanged.

Would you prefer that I continue optimizing the direct DSU schedule here, or split out the deterministic mapping fix and keep the discovery replacement experimental until it wins on `aloha_clutter`? I can proceed either way.

thowell:
please feel free to continue iterating on the implementation in this pr. thanks!

teerthsharma:
Thanks — I continued iterating in this PR and pushed `5b83f51`.

The new path assigns one 32-lane block per world, strides the active EFC prefix, and collapses only repeated fixed-support scalar rows. On the retained 2,048-world Aloha checkpoint it matched dense discovery exactly in 10/10 A-B-B-A blocks and was 1.513x faster by geometric mean, with a conservative 99% interval of [1.406x, 1.584x]. Scratch drops from 7,929,856 to 180,224 bytes (97.73%). It is also 1.355x faster than the previous capacity-wide atomic schedule.

I tested 32/64/128/256-lane shapes and kept 32 because the dedicated 32-vs-128 result was statistically tied. The same-state `fwd_position` boundary is exact but its timing interval crosses 1, and two full chaotic replay pairs still do not show a whole-step win, so I am deliberately limiting the claim to direct discovery and memory rather than overstating end-to-end performance.

The full local gate is green: 1,217 passed, 18 skipped, with CPU, CUDA error checking, graph capture/replay, Ruff, formatting, and kernel-analyzer coverage. The PR description now contains the complete benchmark breakdown. This iteration is ready for review, and I will keep any further performance work in this PR as requested.

thowell:
@teerthsharma thank you for the update, please post the aloha clutter benchmark outputs before and after the changes proposed in this pr. thanks!

teerthsharma:
@thowell Here are fresh `aloha_clutter` outputs from the exact parent (`081c74e`) and current PR head (`5b83f51`) on the same RTX 4060 Laptop GPU. Both runs used the same 2,751-frame replay, 2,048 worlds, capacities, sleep settings, and `--clear_warp_cache=true`; only the source worktree changed.

Before this PR:

```text
{"jit_duration": 148.21315740000136, "run_time": 182.5835132000284, "steps_per_second": 30857.375352547024, "converged_worlds": 2048, "step": 32322.237580900884, "step.forward.fwd_position.island": 101.8095239443292, "step.forward.fwd_position.island.tree_edges": 49.0014087697439, "step.forward.fwd_position.island.flood_fill": 45.10839236968976}
```

After this PR:

```text
{"jit_duration": 166.96295850000024, "run_time": 188.48736509993614, "steps_per_second": 29890.85234977328, "converged_worlds": 2048, "step": 33374.431660186914, "step.forward.fwd_position.island": 52.00119271267024, "step.forward.fwd_position.island.direct_dsu": 47.033806115583786}
```

The full-step after/before ratio is `0.96868x` (`-3.13%`), so I am not claiming an end-to-end Aloha speedup. In the same outputs, the island event is `1.958x` faster (`101.810 -> 52.001 us`) and the discovery kernels are `2.001x` faster (`tree_edges + flood_fill = 94.110 us -> direct_dsu = 47.034 us`). Both runs report all 2,048 worlds converged.

The command was:

```text
uv run --directory <source> mjwarp-testspeed C:\tmp\mjwarp-issue1539-assets\aloha_clutter\scene_clutter.xml --function=step --nworld=2048 --nconmax=512 --nccdmax=64 --njmax=1024 --nvmax=64 --override=opt.enableflags=SLEEP --init_asleep=true --replay=C:\tmp\mjwarp-issue1539-assets\aloha_clutter\pick_clutter.npz --format=json --event_trace=true --memory=true --measure_solver=true --measure_alloc=true --overflow_behavior=error --clear_warp_cache=true
```

I can also upload the complete raw JSON outputs if that would be useful.

thowell:
@teerthsharma do we understand why there is a performance regression for `step` even though `island` performance is improved?

teerthsharma:
> @teerthsharma do we understand why there is a performance regression for `step` even though `island` performance is improved?

There must be some kind of overhead remaining I am starting to run some tests to be sure .
It's deterministic same-state fwd_position boundary matched all topology and downstream mapping arrays exactly across five fresh processes which means All left is fine tuning such overheads , I wanted to show my work and ask for any guidance in continuing it rather then fix it myself . Shall I continue making it ? Along with any guidance for moving forward?

erikfrey:
@teerthsharma the full benchmark output should be able to show you which part of the physics step got slower.  That might give us some meaningful info.

Just so I understand - are you saying your new island discovery mechanism produces the exact same islands as the old approach?  I ask because solver performance is sensitive to island topology.

teerthsharma:
@erikfrey The full trace does localize it, and it localizes to two separate things: a measurement artifact that accounts for almost all of the reported `step` gap, and one real regression of about 0.2% of `step` that the earlier posts missed. On the island question: yes, exactly the same islands, and that is now a test rather than an assertion.thanks for advice

### Where the reported step gap came from

Back-to-back on the same RTX 4060 Laptop GPU (driver 595.79), parent `081c74e` first and PR head `5b83f51` second, full 2,751-frame `aloha_clutter` replay at 2,048 worlds:

| scope | 081c74e | 5b83f51 | delta |
|---|---|---|---|
| `step` | 22112.355 | 25390.516 | +3278.161 |
| `…fwd_position.collision[0]` | 12226.744 | 14755.977 | +2529.233 |
| ↳ `convex_narrowphase[0]` | 7575.158 | 9222.991 | +1647.833 |
| ↳ `nxn_broadphase[0]` | 4557.346 | 5457.625 | +900.279 |
| `forward.solve` | 6731.916 | 7663.037 | +931.121 |
| `…fwd_position.island` | 124.121 | 44.654 | −79.467 |

77% of the delta lands in `collision`, which no code path in this PR reaches. The work controls from those same two runs are flat: `ncon_mean` 10.220243 → 10.217885, `nefc_mean` 120.71683 → 120.99709, `solver_niter_mean` 2.6538375 → 2.6529517. Identical work, everything uniformly 15-21% slower. Sampling `nvidia-smi` across the pair at 5 s intervals shows why: SM clock 2685 MHz → 1890 MHz, 84-87 °C, `clocks_throttle_reasons.active` set by the end. The baseline ran on a cold GPU and the candidate on a throttled one, so run order and not the diff carried the result.

Locking the SM clock at 1500 MHz (`nvidia-smi -lgc 1500,1500`, below the observed 1890 MHz throttle floor) and running A-B-B-A blocks of the same replay, `5b83f51` against `081c74e` gives 0.99946x steps/s and 1.00054x on `step`. `collision` comes back at 1.00074x, consistent with noise.

### The regression that is real

The same locked-clock runs isolate `compute_island_mapping` at 1.29391x (244.294 → 316.094), reproducible to 0.05% within each arm. Temporary event scopes around its two mapping launches split it as `_island_map_dofs` 47.30 → 69.97 and `_island_map_constraints` 42.70 → 90.59, which sum to the entire 251.17 → 322.80 scope delta.

The cause is `63274a1`, not the DSU. Making the mappings deterministic collapsed both kernels from per-item parallel to one thread per world, each walking `nv` DOFs or `nefc` constraint rows in a serial loop, leaving only `nworld` threads resident.

`a938314` recovers the ascending order from structure instead of from a serial scan. A DOF's rank inside its island is the DOF count of every preceding tree carrying the same label, since DOFs are contiguous per tree and trees are ordered, so `_island_map_dofs` runs at `(nworld, nv)` with no counters. `_island_map_constraints` gives one thread each island and holds its three category counters in registers, so the scan stays serial per island while islands and worlds advance in parallel; that also drops three `(nworld, ntree)` scratch allocations and the `island_nv` reset in `_island_scan_sizes`.

A-B-B-A at 1500 MHz against `081c74e`, before and after:

| scope | before `a938314` | after `a938314` |
|---|---|---|
| `solve.compute_island_mapping` | 1.29391x | 0.97766x |
| `fwd_position.island` | 0.49114x | 0.50394x |
| `step` | 1.00054x | 0.99912x |
| steps/s | 0.99946x | 1.00100x |

Work controls across every block after the fix: `ncon_mean` 1.0003x, `nefc_mean` 1.0004x, `solver_niter_mean` 0.9999x.

### Island topology equivalence

The labels are identical by construction and by test. `_dsu_union` hooks the higher root to the lower, so a component's root is its minimum tree id, and `_compress_and_label` walks trees ascending and numbers roots in that order. The dense path reaches the same order because `_flood_fill` starts each DFS from the lowest unlabeled tree that carries an edge, so both number components by their minimum member.

Two tests cover it. `test_direct_dsu_labels_match_dense_flood_fill_over_a_rollout` compares `tree_island` and `nisland` from the production path against `tree_edges` plus `flood_fill` on every state of a 24-step mixed-contact rollout across 4 worlds, and fails when `_compress_and_label` is mutated to number roots in descending order. `test_dof_mapping_interleaves_islands_and_unconstrained_trees` checks `map_dof2idof`, `map_idof2dof` and `island_dofadr` against MuJoCo on a model whose two islands are separated by unconstrained trees; the two pre-existing canonical-order tests use a single-island chain and pass unchanged when island-local DOF ranking is broken, so that case was previously uncovered.

Full gate on `a938314`: 1,219 passed, 18 skipped, 4 subtests passed, plus Ruff, ruff-format and kernel-analyzer.

### Limits

Every number above comes from one RTX 4060 Laptop GPU and one workload, so the clock-lock result establishes that the reported gap was thermal on this machine rather than that it would be absent on a datacenter part. The unlocked pair is a single A-B ordering, not a randomized one; the locked runs are A-B-B-A, which cancels linear drift but not a nonlinear one. `steps_per_second` at 1.00100x after the fix sits inside the 0.03% same-build spread between the two base arms, so the claim is that whole-step throughput is unchanged, not improved. The rewritten mapping kernels trade total work for parallelism: `_island_map_dofs` is `O(ntree)` per DOF and `_island_map_constraints` scans the active EFC prefix once per island, both of which favour the modest `ntree` of this workload (22) over models with many small islands.

teerthsharma:
# Scaling in ntree

RTX 4060 Laptop GPU, driver 595.79, SM clock locked at 1500 MHz with `nvidia-smi -lgc 1500,1500` to remove thermal drift. Median of 30 graph-captured launches. The model is `ntree` single-DOF bodies welded into `ntree/2` pairs with a sparse Jacobian and contacts disabled, so island count is `ntree/2` by construction. Both paths are asserted to produce identical `tree_island` and `nisland` at matched state before either is timed; the dense column is `tree_edges` plus flood_fill on the same Data. I apologise for my lack of hardware 

**nworld = 1**

| ntree | dense bytes | direct bytes | memory | dense us | direct us | speedup |
| :--- | ---: | ---: | ---: | ---: | ---: | ---: |
| 64 | 32,768 | 256 | 128.0x | 406.40 | 34.10 | 11.92x |
| 256 | 524,288 | 1,024 | 512.0x | 5,896.75 | 60.40 | 97.63x |
| 1,024 | 8,388,608 | 4,096 | 2,048.0x | 92,023.15 | 162.70 | 565.60x |
| 4,096 | 134,217,728 | 16,384 | 8,192.0x | 1,758,156.20 | 584.50 | 3,007.97x |

**nworld = 64**

| ntree | dense bytes | direct bytes | memory | dense us | direct us | speedup |
| :--- | ---: | ---: | ---: | ---: | ---: | ---: |
| 64 | 2,097,152 | 16,384 | 128.0x | 665.55 | 41.55 | 16.02x |
| 256 | 33,554,432 | 65,536 | 512.0x | 10,299.40 | 109.95 | 93.67x |
| 1,024 | 536,870,912 | 262,144 | 2,048.0x | 190,839.60 | 309.45 | 616.71x |

Dense scratch is `2 * nworld * ntree^2 * 4` bytes against `nworld * ntree * 4` for the direct path, so the memory column is exactly `2 * ntree` and the formula reproduces every measured size. The dense path cannot be allocated at all at `nworld = 64, ntree = 4096`, where it requires 8.6 GB against this device's 8 GiB.

## Where the speedup comes from

Three changes compound, each measured against the preceding commit at the same locked clock.

Ranking roots is the only inherently sequential step in labeling, because island i is the i-th smallest root. Splitting the single per-world labeling kernel so that only that rank pass stays serial, while compression and propagation run per tree, is worth a factor of roughly 2.1 at every size: `74.85 to 34.15 us` at `ntree = 64`, `188.55 to 86.85` at `256`, `617.00 to 290.45` at `1,024`, and `2,412.45 to 1,142.50` at `4,096`. Propagation is safe per tree because a root reads and writes its own slot with the same value, so every concurrent write to a slot stores identical data.

Laying blocks out over `(world, chunk)` rather than world alone makes resident parallelism track constraint capacity instead of batch size. Previously a single world strided its whole active prefix with one 32-lane block on one SM. `_dsu_chunking` trades batch size against a resident-block target: a batch that already occupies the device keeps one block per world, and a small batch with a long prefix is split until there is work to schedule. This moves discovery against dense from `317.27x to 565.60x` at `ntree = 1024` and from `1,557.88x to 3,007.97x` at `ntree = 4096`, both at `nworld = 1`.

A third change is a repair rather than a gain. Making the DOF and constraint mappings deterministic had collapsed `_island_map_dofs` and `_island_map_constraints` to one thread per world walking `nv` or `nefc` serially, costing `compute_island_mapping` `1.294x` on aloha_clutter. Both now recover the ascending order from structure: a DOF's rank inside its island is the DOF count of every preceding tree carrying the same label, since DOFs are contiguous per tree and trees are ordered, and each island gets one thread holding its three category counters in registers. That returns `compute_island_mapping` to `0.971x` and drops three `(nworld, ntree)` scratch allocations.

## Batched workload

The 2,048-world 2,751-frame aloha_clutter replay has `ntree = 22`, where quadratic growth is not yet visible and the workload is dominated by collision. A-B-B-A blocks at the locked clock against `081c74e` give `0.4342x` on the island event, `0.9710x` on `compute_island_mapping`, and `1.0009x` on whole-step throughput, with the two same-build base arms `0.03%` apart. Work controls are flat across every block: `ncon_mean 1.0008x`, `nefc_mean 0.9965x`, `solver_niter_mean 0.9999x`.

thowell:
running the aloha clutter benchmark

before (081c74e2)
```
jit_duration: 125.85205416800454
run_time: 26.16317226132378
steps_per_second: 215342.69406346575
converged_worlds: 2048
model_memory: 47799902
data_memory: 1512445968
total_memory: 3919577088
ncon_mean: 8.592442059421574
ncon_p95: 21.23046875
nefc_mean: 115.76444929116686
nefc_p95: 222.0
solver_niter_mean: 2.6153362555661577
solver_niter_p95: 6.0
step: 4633.432970545111
step.forward: 4473.200842244899
step.forward.sensor_acc: 8.486732241666676
step.forward.fwd_actuation: 5.916939467336824
step.forward.update_sleep: 7.346419498584604
step.forward.solve: 1742.1332231579909
step.forward.solve.update_active_dofs: 10.12486367857351
step.forward.solve.compute_island_mapping: 46.393130452706274
step.forward.solve.solve_compact: 1682.8356917980707
step.forward.solve.solve_compact._compact_gather: 429.45074389305444
step.forward.solve.solve_compact._compact_scatter: 10.337876997349303
step.forward.solve.solve_compact._compact_scatter.mul_m: 6.006361454056089
step.forward.fwd_velocity: 82.73834571737785
step.forward.fwd_velocity.com_vel: 24.843636340919037
step.forward.fwd_velocity.tendon_bias: 0.6993820411277234
step.forward.fwd_velocity.passive: 30.061250260766776
step.forward.fwd_velocity.rne: 22.091784943628113
step.forward.fwd_position: 2504.986002790748
step.forward.fwd_position.transmission: 3.8291530455938334
step.forward.fwd_position.wake_equality: 2.8578699225080713
step.forward.fwd_position.tendon_armature: 0.6953834951545399
step.forward.fwd_position.island: 42.30988782443163
step.forward.fwd_position.island.flood_fill: 29.93293404769697
step.forward.fwd_position.island.tree_edges: 8.809876066590737
step.forward.fwd_position.update_sleep[0]: 7.560705243096239
step.forward.fwd_position.update_sleep[1]: 7.113958595603887
step.forward.fwd_position.wake_collision: 3.0367140048809906
step.forward.fwd_position.crb: 22.274820784759218
step.forward.fwd_position.make_constraint: 69.30970542418878
step.forward.fwd_position.fwd_kinematics: 96.13304297138116
step.forward.fwd_position.fwd_kinematics.kinematics: 49.71823750045115
step.forward.fwd_position.fwd_kinematics.tendon: 0.7071973809844001
step.forward.fwd_position.fwd_kinematics.com_pos: 35.617958250522335
step.forward.fwd_position.fwd_kinematics.flex: 0.6953834951545399
step.forward.fwd_position.fwd_kinematics.camlight: 5.217557199248682
step.forward.fwd_position.collision[0]: 2050.248458064918
step.forward.fwd_position.collision[1]: 189.15412604936034
step.forward.fwd_position.collision.nxn_broadphase[0]: 804.3585945125933
step.forward.fwd_position.collision.nxn_broadphase[1]: 112.69611036376851
step.forward.fwd_position.collision.primitive_narrowphase[0]: 14.148576628438127
step.forward.fwd_position.collision.primitive_narrowphase[1]: 10.657579032151991
step.forward.fwd_position.collision.convex_narrowphase[0]: 1228.1245009914126
step.forward.fwd_position.collision.convex_narrowphase[1]: 62.51035978446139
step.forward.fwd_acceleration: 103.08106151478933
step.forward.fwd_acceleration.xfrc_accumulate: 7.776933296856001
step.forward.fwd_acceleration.update_active_dofs: 9.767902587563595
step.forward.fwd_acceleration.smooth_solve_compact: 80.24191196413926
step.forward.sensor_vel: 0.6935659742576383
step.forward.sensor_pos: 0.6932024700782581
step.forward.wake: 8.522804327612251
step.euler: 157.99527457832812
step.euler.fwd_velocity: 84.06815817928747
step.euler.fwd_velocity.com_vel: 25.222464002762717
step.euler.fwd_velocity.tendon_bias: 0.7010178099349348
step.euler.fwd_velocity.passive: 30.03398743206409
step.euler.fwd_velocity.rne: 22.818611853732566
step.euler.update_sleep: 7.350599792287623
step.euler.sleep: 32.76517649628863
```

after (this pr)
```
jit_duration: 129.531484473031
run_time: 25.910115628503263
steps_per_second: 217445.88410103746
converged_worlds: 2048
model_memory: 47799902
data_memory: 1512626192
total_memory: 3919577088
ncon_mean: 8.585279713626862
ncon_p95: 21.16796875
nefc_mean: 115.93747728098873
nefc_p95: 222.0
solver_niter_mean: 2.614443824404762
solver_niter_p95: 6.0
step: 4589.873061626094
step.forward: 4429.947596801
step.forward.fwd_actuation: 5.8940387036639335
step.forward.fwd_acceleration: 103.476917407568
step.forward.fwd_acceleration.xfrc_accumulate: 7.934932709831601
step.forward.fwd_acceleration.update_active_dofs: 9.742275542442039
step.forward.fwd_acceleration.smooth_solve_compact: 80.4900035810923
step.forward.sensor_vel: 0.6952017430648497
step.forward.solve: 1741.170302096562
step.forward.solve.update_active_dofs: 10.157033814379702
step.forward.solve.solve_compact: 1684.1572184518172
step.forward.solve.solve_compact._compact_scatter: 9.990912141447847
step.forward.solve.solve_compact._compact_scatter.mul_m: 5.927980886417125
step.forward.solve.solve_compact._compact_gather: 432.16539577612133
step.forward.solve.compute_island_mapping: 44.06179575388058
step.forward.sensor_acc: 8.480552667187181
step.forward.wake: 8.394805397934944
step.forward.sensor_pos: 0.7017448182936954
step.forward.fwd_position: 2462.664485470805
step.forward.fwd_position.island: 12.392221064922325
step.forward.fwd_position.island.direct_dsu: 10.992729914881846
step.forward.fwd_position.crb: 22.352235841782175
step.forward.fwd_position.fwd_kinematics: 96.2422758280787
step.forward.fwd_position.fwd_kinematics.camlight: 5.203562283900033
step.forward.fwd_position.fwd_kinematics.kinematics: 49.35586928061456
step.forward.fwd_position.fwd_kinematics.tendon: 0.6992002890380332
step.forward.fwd_position.fwd_kinematics.com_pos: 36.09414887601853
step.forward.fwd_position.fwd_kinematics.flex: 0.6910214450019762
step.forward.fwd_position.collision[0]: 2039.9934575229108
step.forward.fwd_position.collision[1]: 186.92166515448724
step.forward.fwd_position.collision.nxn_broadphase[0]: 798.6733935768633
step.forward.fwd_position.collision.nxn_broadphase[1]: 110.84641899004158
step.forward.fwd_position.collision.convex_narrowphase[0]: 1223.5443465039662
step.forward.fwd_position.collision.convex_narrowphase[1]: 62.03925844567671
step.forward.fwd_position.collision.primitive_narrowphase[0]: 14.168137711531758
step.forward.fwd_position.collision.primitive_narrowphase[1]: 10.746455834322836
step.forward.fwd_position.transmission: 3.5610687431999533
step.forward.fwd_position.tendon_armature: 0.6955652472442301
step.forward.fwd_position.make_constraint: 69.37659014724065
step.forward.fwd_position.wake_equality: 2.816066927952594
step.forward.fwd_position.wake_collision: 3.0229406048849516
step.forward.fwd_position.update_sleep[0]: 7.509632892607111
step.forward.fwd_position.update_sleep[1]: 7.339694679882465
step.forward.fwd_velocity: 82.59030865000575
step.forward.fwd_velocity.passive: 30.12540858847651
step.forward.fwd_velocity.rne: 21.789349428033496
step.forward.fwd_velocity.com_vel: 24.690294336087145
step.forward.fwd_velocity.tendon_bias: 0.7033805871009068
step.forward.update_sleep: 7.274445671645861
step.euler: 157.6935658055337
step.euler.fwd_velocity: 83.94474843788201
step.euler.fwd_velocity.passive: 30.16793852915433
step.euler.fwd_velocity.rne: 22.656125439388827
step.euler.fwd_velocity.com_vel: 25.119955810374705
step.euler.fwd_velocity.tendon_bias: 0.6961105035133005
step.euler.sleep: 32.208651477936996
step.euler.update_sleep: 7.236095977559827
```

erikfrey:
@thowell seems pretty good right?  Should we properly review this?

If we do, my one suggestion would be let's see if we can completely remove the old island discovery mechanism (e.g. `flood_fill` and other references).  Not sure it's worth having both around.

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
 @event_scope
 def island(m: types.Model, d: types.Data):
   """Discover constraint islands."""
   if m.ntree == 0:
     d.nisland.zero_()
     return
 
-  # Step 1: Find tree edges
-  tree_tree = wp.zeros((d.nworld, m.ntree, m.ntree), dtype=int)
-  tree_edges(m, d, tree_tree)
```
thowell, line 291: lets remove `tree_edges`

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
-
-  # Step 2: DFS flood fill
-  flood_fill(m, d, tree_tree)
```
thowell, line 294: lets remove `flood_fill`

mujoco_warp/_src/types.py
```diff
@@ -2153,6 +2153,7 @@ class Data:
     cfrc_ext: com-based external force on body                  (nworld, nbody, 6)
     contact: contact data
     efc: constraint data
+    island_parent: compressed minimum-root identity pointer      (nworld, ntree)
```
thowell: if this field is only utilized in `island` lets remove it from `Data` and instead allocate it inline in `island`

mujoco_warp/_src/io_test.py
```diff
@@ -63,6 +63,31 @@ def _allocate_worlds(
   return assignment
 
 
+class IslandArrayAllocationTest(absltest.TestCase):
```
thowell: lets remove this test

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
+
+
+@wp.kernel
+def _reset_dsu(
+  island_parent_out: wp.array2d[int],
+  tree_island_out: wp.array2d[int],
+  nisland_out: wp.array[int],
+):
+  """Reset identity parents and active/output markers in parallel."""
+  worldid, tree = wp.tid()
```
thowell: `tree` -> `treeid`

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
+
+  Blocks are laid out over (world, chunk) rather than world alone, so resident
+  parallelism tracks total constraint capacity instead of batch size. A single world
+  with a long prefix fills the device the same way a large batch of short ones does.
+  """
+  worldid, chunkid, lane = wp.tid()
+  chunk_beg = chunkid * chunk_size
+  chunk_end = wp.min(chunk_beg + chunk_size, wp.min(njmax_in, nefc_in[worldid]))
+  for efcid in range(chunk_beg + lane, chunk_end, wp.block_dim()):
+    _process_island_efc(
```
thowell: since this is only called once, lets inline it

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
+    )
+
+
+@wp.kernel
+def _compress_roots(
+  island_parent_inout: wp.array2d[int],
+  tree_island_in: wp.array2d[int],
+):
+  """Point every active tree straight at its component's minimum-tree root."""
+  worldid, tree = wp.tid()
```
thowell: `tree` -> `treeid`

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
+def _propagate_labels(
+  island_parent_in: wp.array2d[int],
+  tree_island_inout: wp.array2d[int],
+):
+  """Copy each root's label down to the trees that point at it.
+
+  A root reads and writes its own slot with the same value, so the only concurrent
+  writes to any slot store identical data.
+  """
+  worldid, tree = wp.tid()
```
thowell: `tree` -> `treeid`

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
+  """
+  max_chunks = max(1, -(-njmax // _DSU_MIN_CHUNK))
+  nchunk = min(max(1, -(-_DSU_TARGET_BLOCKS // nworld)), max_chunks)
+  return nchunk, -(-njmax // nchunk)
+
+
+@event_scope
+def direct_dsu(m: types.Model, d: types.Data):
+  """Discover islands with EFC-parallel atomic minimum-root hooks."""
+  device = d.island_parent.device
```
thowell: think we can remove `device` here and in the launches

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
+  nchunk, chunk_size = _dsu_chunking(d.nworld, d.njmax)
```
thowell: lets inline the function

mujoco_warp/_src/island.py
```diff
@@ -277,19 +277,385 @@ def flood_fill(m: types.Model, d: types.Data, tree_tree: wp.array3d[int]):
...
+      d.efc.J_rowadr,
+      d.efc.J_colind,
+      d.efc.J,
+      d.njmax,
+      chunk_size,
+      d.island_parent,
+      d.tree_island,
+    ],
+    device=device,
+    block_dim=32,
```
thowell: lets update `types.BlockDim` with this setting

teerthsharma:
@thowell I have tried to fix all the reviewed changes but there are some minor caveats !!! should I change my approach if so do let me know

mujoco_warp/_src/io.py
```diff
@@ -1529,6 +1529,7 @@ def _allocate_island_arrays(
   njmax_size = njmax if enabled else 0
 
   d.nisland = wp.array(np.full(nworld, mjd.nisland), dtype=int)
+  d.island_parent = wp.empty((nworld, ntree_size), dtype=int)
```
thowell: lets remove `island_parent` from `Data` and construct the field inline in `island`

mujoco_warp/_src/island.py
```diff
@@ -284,14 +342,14 @@ def island(m: types.Model, d: types.Data):
...
-  # Step 1: Find tree edges
-  tree_tree = wp.zeros((d.nworld, m.ntree, m.ntree), dtype=int)
-  tree_edges(m, d, tree_tree)
-
-  # Step 2: DFS flood fill
-  flood_fill(m, d, tree_tree)
+  direct_dsu(m, d)
 
 
+# kernel_analyzer: off
```
thowell: lets removal all of these `kernel_analyzer` settings from the file

mujoco_warp/_src/island_test.py
```diff
@@ -24,6 +28,276 @@
 from mujoco_warp._src import island
 from mujoco_warp._src import types
 
+
+# Dense reference island discovery: the ntree x ntree adjacency matrix and DFS flood
```
thowell: lets remove any kernels, function, and tests for the dfs routine.

teerthsharma:
@thowell Can we move for merge!!

mujoco_warp/_src/island.py
```diff
@@ -407,78 +488,62 @@ def _island_count_constraints(
...
 
-  total_gathered_efc = island_iefcadr_in[worldid, nisland - 1] + island_nefc_in[worldid, nisland - 1]
+  iefcadr = island_iefcadr_in[worldid, islandid]
+  fadr = iefcadr + island_ne_in[worldid, islandid]
+  oadr = fadr + island_nf_in[worldid, islandid]
+  ne_mapped = int(0)
+  nf_mapped = int(0)
+  nother_mapped = int(0)
+
+  for efcid in range(wp.min(njmax_in, nefc_in[worldid])):
```
thowell: lets read `nefc_in[worldid]` because the for loop range call

mujoco_warp/_src/island.py
```diff
@@ -49,134 +127,178 @@ def _tree_edges(
...
+  """
+  worldid, treeid = wp.tid()
+  if tree_island_out[worldid, treeid] >= 0:
+    tree_island_out[worldid, treeid] = tree_island_out[worldid, island_parent_in[worldid, treeid]]
+
 
-  if first_tree >= 0 and has_cross_edge == 0:
-    wp.atomic_max(tree_tree, worldid, first_tree, first_tree, 1)
+# Smallest EFC chunk worth its own block, and the resident-block count worth reaching for.
+_DSU_MIN_CHUNK = 256
```
thowell, line 280: are there any limitations for these constant settings, for example, are there any issues with large `ntree` values?

mujoco_warp/_src/island_test.py
thowell, line 1: how closely do these test changes match the mujoco tests?

mujoco_warp/_src/island.py
teerthsharma: Neither constant depends on `ntree` — they size the EFC chunk split from `nworld` and `njmax` only, so the discovery grid is identical at `ntree = 22` and at 4,096, and a wrong value costs occupancy rather than correctness (`test_dsu_chunking_splits_only_when_the_batch_leaves_the_device_idle` pins `nchunk * chunk_size >= njmax` across 45 shapes).

Their real limit is resident parallelism at small `nworld`: `_DSU_MIN_CHUNK = 256` caps a single world at `ceil(njmax / 256)` blocks, and `_DSU_TARGET_BLOCKS = 2048` binds only below `nworld = 2048`; both were tuned on a 24-SM RTX 4060 Laptop GPU, so a larger part wants the block target raised.

Large `ntree` binds in `_label_roots` instead, the serial rank pass at `dim=d.nworld`: on PR head 5daf934 with the SM clock locked at 1500 MHz it is linear at 0.131 us per tree, rising from 67.5% of `direct_dsu` at `ntree = 256` to 96.7% at 4,096 (34.3 -> 536.2 us), with `nisland` exact at every size and the workspace flat at `nworld * ntree * 4` bytes.

mujoco_warp/_src/island_test.py
teerthsharma: These kernels are the GPU form of `mj_dsuRoot`, `mj_dsuMerge` and `mj_dsuAssign` from google-deepmind/mujoco#3396, so `test/engine/engine_island_test.cc` is the reference; 17 of its 30 tests have direct counterparts among the 35 here, including the minimum-root, idempotent-merge, ascending-label and inactive-input cases.

The five `FloodFill*` tests went with the DFS path as requested, and the two `DsuMergeFastPath*` tests cover a `parent`-equality shortcut that `_dsu_union` cannot have under concurrent hooking, so neither group has anything to port.

thowell:
@teerthsharma thanks for the responses and update!

the implementation looks good to me. it would be great if we could consolidate the relevant island discovery tests and ideally match the mujoco tests.

teerthsharma:
> @teerthsharma thanks for the responses and update!
> 
> the implementation looks good to me. it would be great if we could consolidate the relevant island discovery tests and ideally match the mujoco tests.

commit, 2c0d6d2, which carries the `nefc_in[worldid]` hoist. `uv run pytest -n 8` gives 1,258 passed, 19 skipped and 6 subtests passed; the only two failures, `test_collision20` and `test_factor_solve_mixed_blocks`, reproduce unchanged on a clean c822833 checkout and are outside this PR.

Consolidation will drop the `inspect.getsource` assertions in `test_direct_dsu_owns_no_hidden_allocation`, `test_direct_dsu_blocks_cover_constraint_capacity_not_batch_size` and `test_direct_dsu_collapses_only_repeated_fixed_support_rows`, which pin source text rather than behaviour, and fold what they actually guard into the behavioural cases.

Matching `engine_island_test.cc` will port the four cases it covers that this file does not: the 2,000-trial randomized differential against an independent traversal, the 4,096-tree adversarial chain and long-connected boundary case, the static-static diagnostic, and the flex equality rescan. all of it can be computationally heavy (for testing perosnally ) and hence I haven't implemented if this plans sounds good to you I will implement this

mujoco_warp/_src/island_test.py
```diff
@@ -43,365 +68,691 @@
...
-          <weld body1="body1" body2="body2"/>
-        </equality>
-      </mujoco>
-      """
-    )
 
-    mjwarp.fwd_position(m, d)
+# Mixed-constraint pile: static-plane contacts, body-body contacts that merge and split
+# as the pile settles, a weld pair, and a hinge tree carrying both a limit and friction.
+_MIXED_CONTACT_XML = """
```
thowell: is this used? if not, please remove

mujoco_warp/_src/island_test.py
```diff
@@ -43,365 +68,691 @@
...
-    tt = treetree.numpy()
-    self.assertEqual(tt[0, 1, 2], 1)
-    self.assertEqual(tt[0, 2, 1], 1)
+    island.island(m, d)
+    np.testing.assert_array_equal(d.nisland.numpy(), np.array([1], dtype=np.int32))
+    np.testing.assert_array_equal(d.tree_island.numpy(), np.array([[0]], dtype=np.int32))
 
-  def test_worldbody_dofs_ignored(self):
-    """Constraints involving worldbody (tree < 0) should not cause spurious edges."""
+  def test_generic_equality_jacobian_unions_all_active_trees(self):
```
thowell: can we delete `test_generic_equality_jacobian_unions_all_active_trees` and just have `test_generic_equality_matches_in_explicit_dense_and_sparse_modes`?

mujoco_warp/_src/island_test.py
```diff
@@ -898,6 +1249,56 @@ def test_mujoco_parity_mapping(self):
       mjd.map_iefc2efc[:nefc],
     )
 
+  def test_dof_mapping_is_canonical_across_worlds(self):
```
thowell: lets consolidate `test_dof_mapping_is_canonical_across_worlds` and `test_efc_mapping_is_canonical_across_worlds`

mujoco_warp/_src/island_test.py
```diff
@@ -43,365 +68,691 @@
...
-    self.assertEqual(np.sum(tt[1]), 0)
+    expected = np.full((d.nworld, trees), -1, dtype=np.int32)
+    expected[:, 0] = 0
+    expected[:, 1] = 0
+    np.testing.assert_array_equal(d.nisland.numpy(), np.ones(d.nworld, dtype=np.int32))
+    np.testing.assert_array_equal(d.tree_island.numpy(), expected)
 
-  def test_mixed_equality_and_contact(self):
-    """Both equality and contact constraints should contribute to edges."""
+  def test_repeated_island_reset_is_bitwise_stable(self):
```
thowell: lets consolidate `test_repeated_island_reset_is_bitwise_stable` and `test_high_contention_chain_is_bitwise_stable`

mujoco_warp/_src/island_test.py
thowell, line 1: lets remove `test_two_trees_one_constraint_one_island`, `test_three_trees_chain_one_island`, `test_two_disconnected_pairs_two_islands`, `test_multiple_worlds`, `test_three_trees_star_hub_at_end`

thowell:
hi @teerthsharma,

The core implementation in `island.py` looks great: the linear-memory GPU DSU is algorithmically sound, eliminates the quadratic memory wall, handles static bodies properly, and the benchmark validation on Aloha Clutter is solid.

For the tests in `island_test.py`, we want **excellent coverage with the simplest possible tests**.

Below are the key recommendations and a consolidated, ready-to-use refactoring of `island_test.py`.

---

### Key Recommendations

1. **Inline XMLs in `test_data.fixture`**
   - Inline MJCF XML strings directly into each `test_data.fixture(xml="""...""")` invocation so that each test case is self-contained and immediately readable without jumping to shared variables at the top of the file.

2. **Poison Output Fields Before Calling MuJoCo Warp Functions**
   - Follow the repository standard of poisoning destination fields before calling routines under test (`d.nisland.fill_(-1)`, `d.tree_island.fill_(-1)`, and `-1` for mapping indices/counts, or `wp.inf` for floating-point fields). This guarantees that tests catch unwritten or stale entries.

3. **Use MuJoCo C (`mjd`) as the Sole Ground Truth Oracle**
   - Drop `_reference_islands` (the 35-line Python BFS reimplementation) and the random edge pool generator. MuJoCo C (`mujoco.mj_forward`) is already available; asserting directly against `mjd.tree_island` and `mjd.nisland` tests against the true gold standard and removes ~90 lines of ad-hoc test logic.

4. **Eliminate `unittest.mock` Entirely**
   - Remove `_launch_chunking` and `test_warmed_island_does_not_sync_with_host` monkey-patching. Warp's native `wp.ScopedCapture()` (`test_cuda_graph_capture_and_replay`) already proves CUDA graph capturability and detects illegal host syncs without brittle argument-index inspections.

5. **Scale Down Massive Synthetic Models**
   - Replace 1,024-body and 900-body loop-generated XMLs with compact 6-to-8 body fixtures. Path compression, CAS contention, inactive tree handling, and chunk boundaries are exercised identically without the XML compilation latency.

6. **Test the Public Simulation Contract**
   - Call `island.island(m, d)` and assert on `d.tree_island` and `d.nisland` rather than inspecting internal `island_parent` pointers and buffer byte capacities.

---

### Proposed Consolidated Test Suite

Here is a streamlined, idiomatic structure for `mujoco_warp/_src/island_test.py`:

```python
"""Tests for island discovery."""

import mujoco
import numpy as np
import warp as wp
from absl.testing import absltest

import mujoco_warp as mjwarp
from mujoco_warp._src import island


class IslandDiscoveryTopologyTest(absltest.TestCase):
  """Tests island discovery across fundamental graph topologies."""

  def test_connected_chain_collapses_to_single_island(self):
    """A welded chain of trees compresses to a single island rooted at tree 0."""
    count = 8
    bodies = "".join(f'<body name="b{i}" pos="{i * 0.3} 0 0"><freejoint/><geom size=".1"/></body>' for i in range(count))
    welds = "".join(f'<weld body1="b{i}" body2="b{i + 1}"/>' for i in range(count - 1))

    mjm, mjd, m, d = test_data.fixture(
      xml=f"<mujoco><worldbody>{bodies}</worldbody><equality>{welds}</equality></mujoco>",
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_disconnected_pairs_form_independent_islands(self):
    """Disconnected pairs form separate islands with canonical ascending IDs."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body name="a1"><freejoint/><geom size=".1"/></body>
          <body name="a2" pos="1 0 0"><freejoint/><geom size=".1"/></body>
          <body name="b1" pos="5 0 0"><freejoint/><geom size=".1"/></body>
          <body name="b2" pos="6 0 0"><freejoint/><geom size=".1"/></body>
        </worldbody>
        <equality>
          <weld body1="a1" body2="a2"/>
          <weld body1="b1" body2="b2"/>
        </equality>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_unconstrained_trees_remain_unlabeled(self):
    """Unconstrained trees remain labeled -1 and do not create islands."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body name="c1"><freejoint/><geom size=".1"/></body>
          <body name="c2" pos="1 0 0"><freejoint/><geom size=".1"/></body>
          <body name="free1" pos="3 0 0"><freejoint/><geom size=".1"/></body>
          <body name="free2" pos="4 0 0"><freejoint/><geom size=".1"/></body>
        </worldbody>
        <equality><weld body1="c1" body2="c2"/></equality>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_worldbody_constraint_activates_single_tree(self):
    """A constraint between worldbody (tree < 0) and a body activates only that tree."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body name="floating"><freejoint/><geom size=".1"/></body>
        </worldbody>
        <equality><weld body1="world" body2="floating"/></equality>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_no_constraints_no_islands(self):
    """Free bodies with no constraints produce zero islands."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body><freejoint/><geom size=".1"/></body>
        </worldbody>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.zeros(d.nworld, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.full((d.nworld, mjm.ntree), -1, dtype=np.int32))


class IslandDiscoveryConstraintsTest(absltest.TestCase):
  """Tests island discovery across all MuJoCo constraint types."""

  def test_site_connect_and_weld(self):
    """Site-based connect and weld equalities resolve to owning body trees."""
    for kind in ("connect", "weld"):
      with self.subTest(kind=kind):
        mjm, mjd, m, d = test_data.fixture(
          xml=f"""
          <mujoco>
            <worldbody>
              <body name="l" pos="-1 0 0"><freejoint/><site name="s1"/></body>
              <body name="r" pos="1 0 0"><freejoint/><site name="s2"/></body>
            </worldbody>
            <equality><{kind} site1="s1" site2="s2"/></equality>
          </mujoco>
          """,
          nworld=2,
        )
        mjwarp.fwd_position(m, d)

        d.nisland.fill_(-1)
        d.tree_island.fill_(-1)
        island.island(m, d)

        np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
        np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_generic_equality_dense_and_sparse(self):
    """Joint equalities take the generic Jacobian scan in both dense and sparse modes."""
    for jac in (mujoco.mjtJacobian.mjJAC_DENSE, mujoco.mjtJacobian.mjJAC_SPARSE):
      with self.subTest(jacobian=int(jac)):
        mjm, mjd, m, d = test_data.fixture(
          xml="""
          <mujoco>
            <worldbody>
              <body><joint name="j0" type="hinge"/></body>
              <body pos="1 0 0"><joint name="j1" type="hinge"/></body>
            </worldbody>
            <equality><joint joint1="j0" joint2="j1"/></equality>
          </mujoco>
          """,
          nworld=2,
          overrides={"opt.jacobian": jac},
        )
        mjwarp.fwd_position(m, d)

        d.nisland.fill_(-1)
        d.tree_island.fill_(-1)
        island.island(m, d)

        np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
        np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_flex_equality_rescans_all_rows(self):
    """Flex equality rows sharing an ID carry differing tree incidence and are not quotiented."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <option jacobian="sparse"><flag contact="disable" gravity="disable"/></option>
        <worldbody>
          <flexcomp name="f" type="grid" dim="1" count="3 1 1" spacing=".05 .05 .05" radius=".01" mass="1">
            <edge equality="true"/>
            <contact internal="false" selfcollide="none"/>
          </flexcomp>
        </worldbody>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_joint_friction_and_limits(self):
    """Joint limits and frictionloss activate their owning DOF tree."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body><joint type="hinge" limited="true" range="-1 1" frictionloss="1"/></body>
        </worldbody>
      </mujoco>
      """,
      nworld=2,
    )
    d.qpos.fill_(2.0)
    mjd.qpos[:] = 2.0
    mujoco.mj_forward(mjm, mjd)
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_contacts_pyramidal_and_elliptic(self):
    """Colliding geoms activate both trees under pyramidal and elliptic cones."""
    for cone in ("pyramidal", "elliptic"):
      with self.subTest(cone=cone):
        mjm, mjd, m, d = test_data.fixture(
          xml=f"""
          <mujoco>
            <option cone="{cone}"/>
            <worldbody>
              <body pos="-.05 0 0"><freejoint/><geom type="sphere" size=".1"/></body>
              <body pos=".05 0 0"><freejoint/><geom type="sphere" size=".1"/></body>
            </worldbody>
          </mujoco>
          """,
          nworld=2,
        )
        mjwarp.fwd_position(m, d)

        d.nisland.fill_(-1)
        d.tree_island.fill_(-1)
        island.island(m, d)

        np.testing.assert_array_equal(d.nisland.numpy(), np.full(d.nworld, mjd.nisland, dtype=np.int32))
        np.testing.assert_array_equal(d.tree_island.numpy(), np.tile(mjd.tree_island[: mjm.ntree], (d.nworld, 1)))

  def test_static_static_constraint_creates_no_island(self):
    """Constraints between two static bodies create no island, matching MuJoCo."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <option><flag contact="disable"/></option>
        <worldbody>
          <body name="s1"><geom size=".05"/></body>
          <body name="s2" pos="1 0 0"><geom size=".05"/></body>
        </worldbody>
        <equality><weld body1="s1" body2="s2"/></equality>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    self.assertEqual(mjd.nisland, 0)
    np.testing.assert_array_equal(d.nisland.numpy(), np.zeros(d.nworld, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy(), np.full((d.nworld, mjm.ntree), -1, dtype=np.int32))


class IslandDiscoveryExecutionTest(absltest.TestCase):
  """Tests multi-world execution, idempotency, and CUDA graph capturability."""

  def test_inactive_world_prefix_ignored(self):
    """A world with nefc=0 leaves all trees inactive even if njmax > 0."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body name="b1"><freejoint/><geom size=".1"/></body>
          <body name="b2" pos="1 0 0"><freejoint/><geom size=".1"/></body>
        </worldbody>
        <equality><weld body1="b1" body2="b2"/></equality>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    nefc = d.nefc.numpy()
    nefc[1] = 0
    wp.copy(d.nefc, wp.array(nefc, dtype=wp.int32, device=d.nefc.device))

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)

    np.testing.assert_array_equal(d.nisland.numpy(), np.array([1, 0], dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy()[0], np.zeros(mjm.ntree, dtype=np.int32))
    np.testing.assert_array_equal(d.tree_island.numpy()[1], np.full(mjm.ntree, -1, dtype=np.int32))

  def test_repeated_discovery_is_idempotent(self):
    """Poisoning output arrays before each run reproduces bitwise-identical results."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body name="b1"><freejoint/><geom size=".1"/></body>
          <body name="b2" pos="1 0 0"><freejoint/><geom size=".1"/></body>
        </worldbody>
        <equality><weld body1="b1" body2="b2"/></equality>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)
    expected_labels = d.tree_island.numpy().copy()
    expected_nisland = d.nisland.numpy().copy()

    for _ in range(3):
      d.tree_island.fill_(-1)
      d.nisland.fill_(-1)
      island.island(m, d)
      np.testing.assert_array_equal(d.tree_island.numpy(), expected_labels)
      np.testing.assert_array_equal(d.nisland.numpy(), expected_nisland)

  @absltest.skipIf(not wp.get_device().is_cuda, "CUDA graph capture requires a CUDA device.")
  def test_cuda_graph_capture_and_replay(self):
    """Direct DSU is capturable in CUDA graphs with zero host synchronizations."""
    mjm, mjd, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <body name="b1"><freejoint/><geom size=".1"/></body>
          <body name="b2" pos="1 0 0"><freejoint/><geom size=".1"/></body>
        </worldbody>
        <equality><weld body1="b1" body2="b2"/></equality>
      </mujoco>
      """,
      nworld=2,
    )
    mjwarp.fwd_position(m, d)

    d.nisland.fill_(-1)
    d.tree_island.fill_(-1)
    island.island(m, d)
    expected_labels = d.tree_island.numpy().copy()
    expected_nisland = d.nisland.numpy().copy()

    d.tree_island.fill_(-1)
    d.nisland.fill_(-1)
    workspace = wp.empty((d.nworld, m.ntree), dtype=int)
    with wp.ScopedCapture() as capture:
      island.direct_dsu(m, d, workspace)
    wp.capture_launch(capture.graph)

    np.testing.assert_array_equal(d.tree_island.numpy(), expected_labels)
    np.testing.assert_array_equal(d.nisland.numpy(), expected_nisland)


class IslandMappingTest(absltest.TestCase):
  """Tests downstream DOF and constraint mapping parity against MuJoCo C."""

  def test_mapping_parity_with_interleaved_unconstrained_trees(self):
    """Disjoint islands separated by unconstrained trees match MuJoCo mapping parity."""
    bodies = "".join(f'<body name="b{i}" pos="{i} 0 0"><freejoint/><geom size=".1"/></body>' for i in range(6))
    mjm, mjd, m, d = test_data.fixture(
      xml=f"""
      <mujoco>
        <worldbody>{bodies}</worldbody>
        <equality>
          <weld body1="b0" body2="b2"/>
          <weld body1="b3" body2="b5"/>
        </equality>
      </mujoco>
      """,
      nworld=4,
    )
    m.opt.disableflags &= ~types.DisableBit.ISLAND
    mjwarp.fwd_position(m, d)

    # Poison all destination fields before computing mapping
    d.dof_island.fill_(-1)
    d.efc.island.fill_(-1)
    d.island_nv.fill_(-1)
    d.island_nefc.fill_(-1)
    d.island_idofadr.fill_(-1)
    d.island_dofadr.fill_(-1)
    d.island_iefcadr.fill_(-1)
    d.map_dof2idof.fill_(-1)
    d.map_idof2dof.fill_(-1)
    d.map_efc2iefc.fill_(-1)
    d.map_iefc2efc.fill_(-1)

    island.compute_island_mapping(m, d)

    for name, got, want in (
      ("dof_island", d.dof_island.numpy()[:, : mjm.nv], mjd.dof_island[: mjm.nv]),
      ("map_dof2idof", d.map_dof2idof.numpy()[:, : mjm.nv], mjd.map_dof2idof[: mjm.nv]),
      ("map_idof2dof", d.map_idof2dof.numpy()[:, : mjm.nv], mjd.map_idof2dof[: mjm.nv]),
      ("island_dofadr", d.island_dofadr.numpy()[:, : mjd.nisland], mjd.island_dofadr[: mjd.nisland]),
      ("efc_island", d.efc.island.numpy()[:, : mjd.nefc], mjd.efc_island[: mjd.nefc]),
      ("map_efc2iefc", d.map_efc2iefc.numpy()[:, : mjd.nefc], mjd.map_efc2iefc[: mjd.nefc]),
      ("map_iefc2efc", d.map_iefc2efc.numpy()[:, : mjd.nefc], mjd.map_iefc2efc[: mjd.nefc]),
    ):
      np.testing.assert_array_equal(got, np.tile(want, (d.nworld, 1)), err_msg=name)
```

teerthsharma:
> hi @teerthsharma,
> 
> The core implementation in `island.py` looks great: the linear-memory GPU DSU is algorithmically sound, eliminates the quadratic memory wall, handles static bodies properly, and the benchmark validation on Aloha Clutter is solid.
> 
> For the tests in `island_test.py`, we want **excellent coverage with the simplest possible tests**.
> 
> Below are the key recommendations and a consolidated, ready-to-use refactoring of `island_test.py`.
> 
> ### Key Recommendations
> 1. **Inline XMLs in `test_data.fixture`**
>    
>    * Inline MJCF XML strings directly into each `test_data.fixture(xml="""...""")` invocation so that each test case is self-contained and immediately readable without jumping to shared variables at the top of the file.
> 2. **Poison Output Fields Before Calling MuJoCo Warp Functions**
>    
>    * Follow the repository standard of poisoning destination fields before calling routines under test (`d.nisland.fill_(-1)`, `d.tree_island.fill_(-1)`, and `-1` for mapping indices/counts, or `wp.inf` for floating-point fields). This guarantees that tests catch unwritten or stale entries.
> 3. **Use MuJoCo C (`mjd`) as the Sole Ground Truth Oracle**
>    
>    * Drop `_reference_islands` (the 35-line Python BFS reimplementation) and the random edge pool generator. MuJoCo C (`mujoco.mj_forward`) is already available; asserting directly against `mjd.tree_island` and `mjd.nisland` tests against the true gold standard and removes ~90 lines of ad-hoc test logic.
> 4. **Eliminate `unittest.mock` Entirely**
>    
>    * Remove `_launch_chunking` and `test_warmed_island_does_not_sync_with_host` monkey-patching. Warp's native `wp.ScopedCapture()` (`test_cuda_graph_capture_and_replay`) already proves CUDA graph capturability and detects illegal host syncs without brittle argument-index inspections.
> 5. **Scale Down Massive Synthetic Models**
>    
>    * Replace 1,024-body and 900-body loop-generated XMLs with compact 6-to-8 body fixtures. Path compression, CAS contention, inactive tree handling, and chunk boundaries are exercised identically without the XML compilation latency.
> 6. **Test the Public Simulation Contract**
>    
>    * Call `island.island(m, d)` and assert on `d.tree_island` and `d.nisland` rather than inspecting internal `island_parent` pointers and buffer byte capacities.

Applied as proposed — 15 tests over 386 lines, 15/15 passing with 6/6 subtests.

One change was needed to make it compile: test_site_connect_and_weld, test_generic_equality_dense_and_sparse and test_joint_friction_and_limits each gave a moving body a joint and no geom, which MuJoCo rejects with "mass and inertia of moving bodies must be larger than mjMINVAL". Each got a <geom size=".1"/>; nothing else differs from what you posted.

thowell (approved):

Issue #1543, opened 2026-07-21
CUDA 700 (illegal memory access) in crb/factor_m CSR path on sm_89 when num_envs >= 128


kevinzakka:
An mjlab user reported (https://github.com/mujocolab/mjlab/issues/1108) that they were getting CUDA 700 illegal memory access errors when using the new mjlab version (which upgraded mjwarp to 3.10.0.2, from 3.8.1). It fails on a 4090 (sm_89) once num_envs >= 128, but is fine under 64. With CUDA_LAUNCH_BLOCKING=1 the fault lands in _M inside smooth.crb.

We first suspected warp, but ruled it out: warp-lang 1.15.0 works with mjwarp 3.8.1 and fails with 3.10.0.x. So it points to the 3.8 to 3.10 mass-matrix rewrite, where crb moved from qM (_qM_sparse/_qM_dense) to CSR M.

It only reproduces on Ada. On sm_120 we build cleanly at 128/512/1024/4096 envs and compute-sanitizer comes back clean. _M is unchanged from 3.10.0.2 to main and its indexing looks in bounds, so our hunch is the shared-memory tile Cholesky rather than _M itself.

cc @adenzler-nvidia @thowell

adenzler-nvidia:
Thanks for reporting - I will investigate.

thowell:
@adenzler-nvidia found an issue with `set_const` please take a look at https://github.com/google-deepmind/mujoco_warp/pull/1547

thowell:
@kevinzakka this should be resolved (#1547 is merged)

kevinzakka:
Thank you Taylor! Super helpful.

PR #1547, opened 2026-07-22
update `set_const`

thowell:
investigating #1543 looks like there is a bug with the `_accumulate_subtreemass` launch where the accumulation is for `nworld` even if the field has batch dimension 1.

additionally, update `set_const` launches to use the fields sizes instead of `nworld`


adenzler-nvidia (approved):
Thanks, @thowell — I reproduced #1543 on sm_89 and confirmed that this change fixes it when applied directly to v3.10.0.2. The regression test fails without the fix because `body_subtreemass` is accumulated once per data world into the shared model row, and passes with this change. I agree that sizing the `set_const` launches from their output model fields is the right fix; `_M` is only where the earlier corruption becomes visible.

PR #1549, opened 2026-07-22
Use lean launches for fixed-grid kernels

adenzler-nvidia:
Warp generates grid-stride loops by default, adding loop-control and stride bookkeeping to kernels even when MuJoCo Warp launches the complete logical grid. This change opts profiled kernels into Warp 1.15's lean launch using module options and explicit settings for unique-module kernels.

Kernels and modules that regressed during profiling retain grid-stride launches. The change does not modify Warp's global configuration. It raises the minimum Warp version to 1.15 because the per-kernel `grid_stride` option was introduced in that release.

Benchmarks were run on an NVIDIA RTX PRO 6000 Blackwell and report medians of three interleaved runs:

| Benchmark | Worlds | End-to-end result |
|---|---:|---:|
| G1 | 8,192 | ~2.2% faster |
| Three humanoids | 8,192 | ~1.2% faster |
| Aloha clutter | 2,048 | 9.7% faster |

The G1 result combines a 2.0% solver improvement with another 0.2% from the remaining profiled modules. The solver-only three-humanoid comparison remained within run-to-run variance, while the remaining modules improved it by 1.2%. The direct Aloha clutter A/B improved throughput from 625,063 to 685,992 steps/s over 1,000 steps.

Individual Aloha clutter event scopes show where the larger end-to-end gain comes from:

| Event scope | Upstream | This change | Speedup |
|---|---:|---:|---:|
| GPU step | 1562.04 ns/world-step | 1420.18 ns/world-step | 1.100x |
| GPU solve | 552.06 ns/world-step | 526.09 ns/world-step | 1.049x |
| Collision | 538.77 ns/world-step | 426.19 ns/world-step | 1.264x |
| NxN broadphase | 493.46 ns/world-step | 381.37 ns/world-step | 1.294x |
| Compact gather | 308.05 ns/world-step | 224.35 ns/world-step | 1.373x |


mujoco_warp/_src/solver.py
```diff
@@ -3751,7 +3751,7 @@ def _use_incremental(m: types.Model) -> bool:
   return m.opt.solver == types.SolverType.NEWTON and m.opt.cone != types.ConeType.ELLIPTIC
 
 
-@wp.kernel
+@wp.kernel(grid_stride=True)
```
thowell, line 3754: we have a duplicate `_zero_change_counters`. lets remove one while we are here. thanks!

erikfrey: Removed the duplicate `_zero_change_counters` kernel definition. Thanks!

erikfrey:
Resolved merge conflicts with `main` and addressed review comments:

- Resolved dependency version merge conflict in `pyproject.toml`.
- Removed duplicate `_zero_change_counters` kernel in `mujoco_warp/_src/solver.py` as requested by @thowell.

thowell (approved):

thowell:
thanks @adenzler-nvidia! thanks @erikfrey!

PR #1550, opened 2026-07-22
passive adhesion

thowell:
add passive adhesion

https://github.com/google-deepmind/mujoco/commit/a264d0bc8bfa12be9d69f33905be208914ccce6c

---

```
mjwarp-viewer ../mujoco/model/adhesion/fridge_door.xml --nconmax=256 --njmax=1024
```

https://github.com/user-attachments/assets/d48c8599-11b1-49e9-968e-e72300579354

```
mjwarp-viewer ../mujoco/model/adhesion/sand_castle.xml --nconmax=4096 --njmax=8192
```

https://github.com/user-attachments/assets/5fe61a11-8431-4ee0-9fc9-3cee6c06c577


mujoco_warp/_src/passive.py
```diff
@@ -559,34 +560,109 @@ def _fluid(m: Model, d: Data):
...
+  g2 = geoms[1]
+  body1 = body_weldid[geom_bodyid[g1]] if g1 >= 0 else -1
+  body2 = body_weldid[geom_bodyid[g2]] if g2 >= 0 else -1
+
+  pos = contact_pos_in[cid]
+  frame = contact_frame_in[cid]
+  normal = wp.vec3(frame[0, 0], frame[0, 1], frame[0, 2])
+  force = adhesion * normal
+
+  if body1 > 0:
```
yuvaltassa, line 605: Should the signs on `body1` and `body2` be reversed here? (should be answered with a test)

In MuJoCo, the contact frame normal $n$ points from geom 2 to geom 1. To pull the two bodies together in mutual attraction:
- `body1` should experience force along $-n$ (pulling it towards `body2`)
- `body2` should experience force along $+n$ (pulling it towards `body1`)

Currently, `body1` gets `+force` ($+n \cdot \text{adhesion}$) and `body2` gets `-force` ($-n \cdot \text{adhesion}$), which results in repulsion rather than attraction when `body1` has dynamic DOFs (compare with `engine_passive.c:1033-1038`). 

Should this be:
```python
if body1 > 0:
  ...
  wp.atomic_add(qfrc_adhesion_out, worldid, dofid, wp.dot(jacp, -force))

if body2 > 0:
  ...
  wp.atomic_add(qfrc_adhesion_out, worldid, dofid, wp.dot(jacp, force))

File: mujoco_warp/_src/constraint.py [hunk 2]
```diff
@@ -4322,11 +4331,20 @@ def kernel(
       efc_frictionloss_out,
     )
 
+    if wp.static(flg_adhesion):
```
yuvaltassa, line 4334: In pyramidal mode (`not IS_ELLIPTIC`), should the scaling condition check `condim > 1` rather than `dimid > 0`?
For a pyramidal cone, all rows `dimid = 0, ..., 2*(condim-1) - 1` represent pyramid facets that share the normal decomposition equally (as in `mj_adhesionRef`). With `dimid > 0`, the first facet (`dimid = 0`) receives the unscaled `adhesion` while the remaining facets receive `adhesion / (2*(condim - 1))`.
Should this be:

```python
if not wp.static(IS_ELLIPTIC) and condim > 1:
  adhesion = adhesion / float(2 * (condim - 1))
```

mujoco_warp/_src/constraint.py
```diff
@@ -4743,6 +4762,15 @@ def kernel(
...
```
yuvaltassa, line 4765: Could this result in a division by zero when `condim == 1`?
When `not IS_ELLIPTIC` and `condim == 1` (e.g. frictionless flex contacts or in-gap tether contacts where `dim` is clamped to 1), `2 * (condim - 1)` evaluates to 0. 
Should we guard this with `condim > 1` here as well?

```python
if not wp.static(IS_ELLIPTIC) and condim > 1:
  adhesion = adhesion / float(2 * (condim - 1))
```

yuvaltassa (commented):
In `support.py`, should `contact_force_fn` also subtract `contact_adhesion_in[contact_id]` from `force[0]` to report the net interface force?
In C MuJoCo (`mj_contactForce` in `engine_core_util.c:1094`), `result[0] -= con->adhesion` is applied so that `contact_force` reports the net interaction force (which can become negative when adhesive contacts are in tension).

yuvaltassa:
Should we remove the early return `if dsbl_spring and dsbl_damper:` in `passive()`? (or otherwise do it differently)
Currently, if both `DisableBit.SPRING` and `DisableBit.DAMPER` are active, the function returns before running `_qfrc_adhesion`, `_fluid`, or gravcomp.

This is preexisting.

mujoco_warp/_src/passive.py
thowell: there was a related issue with the normal direction that has been resolved and this has been changed to match mujoco

mujoco_warp/_src/constraint.py
thowell: fixed

yuvaltassa (approved):

PR #1554, opened 2026-07-24
Drop the tendon-armature all-sparse forcing

adenzler-nvidia:
Since M became CSR-only with per-block factorization (#1424), every factor path gathers its input from the CSR M at factor time, after the per-step tendon-armature accumulation has written into it — so the `_m_allow_dense` gate that forced armature models onto the all-sparse LDL path has never guarded a real hazard: it landed in the same commit as the factor-time gather that makes it moot. Tendon armature also only ever produces ancestor-pair terms (the accumulation walks dof_parentid), so its contributions land in existing CSR slots inside the diagonal blocks, matching MuJoCo's own projection on the CPU — no per-block representation can miss them. Removing the gate lets models with tendon armature classify their blocks by the normal size rules instead of paying the sparse LDL path at every size. The existing tendon-armature tests cover parity through the newly-permitted dense path.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -191,12 +191,6 @@ def _m_blocks(mjm: mujoco.MjModel):
   return [(int(adr), int(num)) for adr, num in zip(mjm.tree_dofadr, mjm.tree_dofnum) if num > 0]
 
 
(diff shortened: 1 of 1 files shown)


thowell (approved):
thanks @adenzler-nvidia!

PR #1557, opened 2026-07-27
multicontact warn

thowell:
add warning to `io.put_model` for ccd pairs that do not have multicontact support

### DIFF
--- a/mujoco_warp/_src/io_test.py
+++ b/mujoco_warp/_src/io_test.py
@@ -16,6 +16,7 @@
 """Tests for io functions."""
 
 import dataclasses
+import warnings
 from unittest import mock
 
 import mujoco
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/io_test.py
```diff
@@ -3172,6 +3173,59 @@ def test_flex_interp_negative_success(self):
     m = mjwarp.put_model(mjm)
     self.assertEqual(m.has_3d_flex, True)
 
+  # TODO(team): remove after implementing multicontact support for CCD pairs.
+  @parameterized.parameters(
+    ("cylinder", "box"),
+    ("cylinder", "cylinder"),
+    ("cylinder", "mesh"),
+    ("capsule", "cylinder"),
+    ("capsule", "mesh"),
```
kbayes, line 3182: You can drop warnings for spheres and ellipsoids since those only ever generate one contact.

thowell: done

kbayes (approved):

PR #1559, opened 2026-07-27
Enable `refit=True` argument for `render`

hartikainen:
See https://github.com/google-deepmind/mujoco/issues/3435 and https://github.com/google-deepmind/mujoco/pull/3436.

### DIFF
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -19,6 +19,7 @@ from typing import Tuple
 import warp as wp
 
 from mujoco_warp._src import math
+from mujoco_warp._src.bvh import refit_bvh
 from mujoco_warp._src.ray import ray_box
 from mujoco_warp._src.ray import ray_capsule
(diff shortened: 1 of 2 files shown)


erikfrey:
Hi @hartikainen - ouch, I can see this having resulted in a nasty bug.  Hope it didn't bite you guys for very long.

I have heard that `AbstractToken` is used to ensure proper chaining of FFI calls:

https://github.com/jax-ml/jax/blob/75295224bb2d1ff4c48e476f8077622e6b7eac66/jax/_src/core.py#L3095-L3099

Example here:

https://github.com/jax-ml/jax/issues/16303

Maybe there's a simple, generic way in the JAX wrapper code so that users can supply abstract tokens to indicate serial dependencies?  cc @btaba

thowell:
resolved with https://github.com/google-deepmind/mujoco/commit/5e3464f475cc5f376607fa36795fc4a327126a70

PR #1564, opened 2026-07-28
skip unnecessary solver allocations

thowell:
reduce solver memory utilization by skipping unnecessary allocations

### DIFF
--- a/mujoco_warp/_src/solver.py
+++ b/mujoco_warp/_src/solver.py
@@ -79,6 +79,7 @@ def _create_solver_context(m: types.Model, d: types.Data) -> SolverContext:
   alloc_h = m.opt.solver == types.SolverType.NEWTON
   alloc_hfactor = alloc_h and nv > _BLOCK_CHOLESKY_DIM
   alloc_mgrad = m.opt.solver == types.SolverType.CG
+  alloc_incremental = _use_incremental(m)
 
   return SolverContext(
     Jaref=wp.empty((nworld, njmax), dtype=float),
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/solver.py
```diff
@@ -4330,32 +4331,21 @@ def solve_compact(m: types.Model, d: types.Data):
...
+      inputs=[d.ncdof],
+      outputs=[d.cM],
+    )
+    wp.launch(
+      _gather_M_sparse,
+      dim=(d.nworld, m.nv),
+      inputs=[m.M_rownnz, m.M_rowadr, m.M_colind, d.M, d.dof_cdof],
+      outputs=[d.cM],
     )
-  elif not m.is_sparse:
```
erikfrey, line 4358: it's not clear to me why you can delete this `elif` - was this a typo?

erikfrey (commented):
One question

mujoco_warp/_src/solver.py
thowell: this should be ok. the `elif` -> `if` and we dont need to gather `cJ` or `cM` here if `m.is_sparse == True` since these fields are directly read from in the solver now.

mujoco_warp/_src/solver.py
```diff
@@ -97,15 +98,15 @@ def _create_solver_context(m: types.Model, d: types.Data) -> SolverContext:
     improvement=wp.empty((nworld,), dtype=float),
     ls_exhausted=wp.zeros((nworld,), dtype=bool),
     search_unchanged=wp.empty((nworld,), dtype=bool),
-    prev_grad=wp.empty((nworld, nv), dtype=float),
-    prev_Mgrad=wp.empty((nworld, nv), dtype=float),
-    beta=wp.empty((nworld,), dtype=float),
-    beta_den=wp.empty((nworld,), dtype=float),
+    prev_grad=wp.empty((nworld, nv), dtype=float) if alloc_mgrad else wp.empty((nworld, 0), dtype=float),
```
erikfrey, line 101: in the future let's consider something like:

```
SolverContext
SolverCGContext
SolverIncrementalContext
```

something like that might be cleaner in the long run.  no action for now.

mujoco_warp/_src/solver.py
erikfrey: ah i get it now, thanks

erikfrey (approved):
Looks good!

PR #1565, opened 2026-07-31
Limit max convex hull vertices to 64 in Aloha clutter scene

erikfrey:
This change sets `<mesh maxhullvert="64"/>` in default class of `scene_clutter.xml`. Restricting the maximum number of convex hull vertices per mesh significantly reduces convex narrowphase collision overhead in dense clutter scenes with complex mesh geometries.

The scene was reviewed manually with no regressions in motion.

Benchmarks were run for 2,048 parallel rollouts over 2,751 steps:

| Metric / Scope | Upstream (Baseline) | This Change (Optimized) | Speedup / Improvement |
|---|---|---|---|
| Total Simulation Time | 21.89 s | 17.42 s | **1.257x faster** (-20.4% time) |
| Time per Step | 3,884.72 ns | 3,092.61 ns | **1.256x faster** |
| Convex Narrowphase | 1,215.35 ns | 585.27 ns | **2.077x faster** (-51.8% collision time) |
| Rollout Throughput | 257,419 steps/s | 323,351 steps/s | **+25.6% throughput** |

### DIFF
--- a/benchmarks/aloha/scene_clutter.xml
+++ b/benchmarks/aloha/scene_clutter.xml
@@ -10,6 +10,7 @@
   <statistic meansize="0.050000000000000003" extent="0.59999999999999998" center="0 -0.10000000000000001 0.20000000000000001" />
   <default>
     <default class="/">
+      <mesh maxhullvert="64"/>
       <default class="frame">
         <geom type="mesh" group="1" material="black" />
       </default>
--- a/benchmarks/aloha/scene_clutter.xml
+++ b/benchmarks/aloha/scene_clutter.xml
@@ -9,8 +9,8 @@
   </visual>
   <statistic meansize="0.050000000000000003" extent="0.59999999999999998" center="0 -0.10000000000000001 0.20000000000000001" />
   <default>
+    <mesh maxhullvert="64"/>
     <default class="/">
-      <mesh maxhullvert="64"/>
       <default class="frame">
         <geom type="mesh" group="1" material="black" />
       </default>


thowell:
@erikfrey nice!

it would be great if we could improve the collision pipeline performance for high resolution meshes with algorithmic changes so that its not necessary to specify `maxhullvert`.

things we might consider: some form of contact caching between steps / contact state or improved ccd initialization, for example: 1) precompute n directions and their corresponding nearest vertices 2) at run time evaluate the precomputed directions to determine an initialization.

@kbayes what do you think?

kbayes:
@thowell I've been considering over the last few weeks doing a Dobkin-Kirkpatrick hierarchy approach in the compiler, which is kinda what you're alluding to.

erikfrey:
@thowell sounds great - is that a blocker for this PR?

kbayes:
@erikfrey What's the purpose of this PR? Is the benchmark too slow in general for practical use, because maxhullvert just simplifies the scene so no real speedup is occurring? The improvements suggested are things I have been opening discussing over the last few months, but there's some controversies to them, so they're nonblockers.

erikfrey:
Benchmarks serve a few purposes:

1) they give us signal for hill climbing the engine + regression testing
2) they communicate to a user what to expect if they try our engine with a certain workload
3) they also communicate to the user current best practices for optimizing their own scenes (which for now includes paying attention to `maxhullvert`)

so the purpose of this PR is to improve 3)

kbayes:
@erikfrey I see. In that case, it's a good idea to set maxhullvert as small as possible where there's no contact behavior differences. 64 might be too low, idk. There's some good stress tests for narrowphase by itself in MuJoCo already, maybe they need ported over.

erikfrey:
@kbayes yes agreed - I ran the unroll that is part of this benchmark with the lower `maxhullvert` and the motion was the same.

kbayes:
@erikfrey Nice! LGTM from me.

kbayes (approved):

thowell (approved):

PR #1569, opened 2026-08-04
array update for types 

thowell:
updates `Model` and `Data` in types.py to use `array(...)` instead of `wp.array`/`wp.array2d`/etc in order to correctly set `_is_batched` in `_mark_batched` in io.py. additionally, the kernel analyzer is updated to check that all `Model`, `Data`, `Option`, and `Statistic` fields that are arrays use `array(...)`.

- additional docstring alignment changes in types.py


btaba:
Nice catch thanks Taylor!!

contrib/kernel_analyzer/kernel_analyzer/ast_analyzer.py
```diff
@@ -228,6 +237,13 @@ def analyze(source: str, filename: str, type_source: str) -> List[Issue]:
     return []
 
   issues: List[Issue] = []
+  for cls_name in ("Model", "Data", "Option", "Statistic"):
+    if cls_name in type_classes:
+      for attr_name, attr_type in type_classes[cls_name]:
+        if attr_type.startswith("wp.array"):
+          dummy_node = ast.AST()
```
btaba, line 244: just curious what does the error look like with this dummy node business?

mujoco_warp/_src/io.py
```diff
@@ -1240,8 +1245,52 @@ def _check_margin(name, t1, t2, margin):
   else:
     m.flex_face = np.zeros((0, 9), dtype=np.int32)
 
-  # place m on device
   sizes = {f.name: getattr(m, f.name) for f in dataclasses.fields(types.Model) if f.type is int}
+  sizes.update(
+    {
+      "nbody_branches": len(m.body_branches),
```
btaba, line 1251: why is this a manual process, how do folks know to update this?

contrib/kernel_analyzer/kernel_analyzer/ast_analyzer.py
thowell:
```
mujoco_warp/_src/types.py:1836:"qLD_block_adr" in class "Model" must use array(...) annotation instead of direct wp.array.
mujoco_warp/_src/types.py:1844:"body_tree" in class "Model" must use array(...) annotation instead of direct wp.array.
mujoco_warp/_src/types.py:1845:"body_branches" in class "Model" must use array(...) annotation instead of direct wp.array.
mujoco_warp/_src/types.py:1846:"body_branch_start" in class "Model" must use array(...) annotation instead of direct wp.array.
```

mujoco_warp/_src/io.py
thowell: these are sizes that are not specified in `mjModel` so they need to be computed first in order to be used by `_create_array` that is expected a size specified in types.py. if the size is missing from `sizes` in `put_model` a `KeyError` will be raised.

btaba (approved):

PR #1573, opened 2026-08-04
constraint updates in io and types

thowell:
update `Constraint` to define `J` and `J_colind` with `array(...)`. update the kernel analyzer to check `Constraint` for `array(...)` instead of `wp.array`/`wp.array2d`/etc in order to correctly set `_is_batched`.

additionally:
- remove unused argument `njmax_nnz` from `io._create_constraint`
- remove unused `efc_kwargs`


contrib/kernel_analyzer/kernel_analyzer/ast_analyzer.py
```diff
@@ -237,7 +237,7 @@ def analyze(source: str, filename: str, type_source: str) -> List[Issue]:
     return []
 
   issues: List[Issue] = []
-  for cls_name in ("Model", "Data", "Option", "Statistic"):
+  for cls_name in ("Model", "Data", "Option", "Statistic", "Constraint"):
```
btaba, line 240: not blocking: but maybe this could be a constant up top, and then have a test that the kernel analyzer constant is checking all of the classes that meet some acceptance criteria

btaba (approved):

erikfrey (approved):
Nice, I like it.

PR #1576, opened 2026-08-10
revert flex preconditioner

thowell:
update: reverting the flex preconditioner and restoring the mujoco inertia matrix inverse preconditioner

fix bug where scenes with flex and non-flex dofs the cg solver would not use the tree-block inertia matrix inverse for non-flex dofs. adds `dof_isflex` field to `Model` to filter objects that are simple bodies but not part of a flex.


quagla (changes requested):
Have you re-run cloth and aloha-cloth to check the performance?

thowell:
there are additional issues that affect flex, specifically the aloha cloth scene, including silent overflows. planning to address in a follow-up pr.

quagla (approved):

PR #1578, opened 2026-08-12
Export OverflowType in the public API.

PhysicistJohn:
Export the existing `OverflowType` enum from the public `mujoco_warp` package. `Data.overflow` is a programmatic bitmask, but consumers currently have to import `_src.types` or hardcode its bits. This does not change the enum, its values, or overflow behavior.

The public import/identity probe, `mujoco_warp._src.types_test`, and pre-commit pass. The full suite reports 1,236 passed and 23 skipped, with the same unrelated `IOTest.test_put_data_nefc_zero_dense` failure reproduced on unmodified current main.


thowell:
thanks @PhysicistJohn!

thowell (approved):

PhysicistJohn:
Thanks, Taylor. MuJoCo Warp is becoming central to my interest in humanoid AI. I’m hoping to make a few more contributions over the coming weeks.

PR #1580, opened 2026-08-13
Restore Cassie in the kitchen scene generator.

PhysicistJohn:
Restore Cassie to the kitchen scene generator now that magnetometer sensors are implemented. The generator also assumed every robot's root body was named `pelvis`; using the spec's first top-level body preserves G1 behavior and supports Cassie's `cassie-pelvis` root. The pinned Cassie scene generates and completes MuJoCo and MuJoCo Warp forward passes, and G1's compiled model and keyframe-forward state remain unchanged.


thowell:
@erikfrey do we want to update the kitchen benchmark, potentially adding a specific robot version to the nightlies, or should we remove this benchmark?

PR #1581, opened 2026-08-13
Exit the passive viewer loop when the window closes

anxkhn:
## Summary

Use the passive viewer's `is_running()` state as the loop condition.

The loop previously ran forever after the native window closed, continuing to step and synchronize until the process was killed. It now exits cleanly when the viewer reports that it is no longer running.

## Tests

- Added a host-only mocked viewer lifecycle test
- Verified one step and sync occur before shutdown


anxkhn:
hey @thowell, could you please review this pr when you have a chance?

thowell:
@anxkhn thank you for your contribution to mujoco warp!

lets remove viewer_test.py. otherwise lgtm. thanks!

anxkhn:
removed viewer_test.py as requested. thanks for the review!

thowell (approved):

PR #1582, opened 2026-08-13
Respect trajectory timestamps during replay

anxkhn:
## Summary

Sample replay controls on the model timestep using zero-order hold over the full timestamp sequence.

The previous loader calculated one repeat factor from the first interval and applied it to every control. Non-uniform trajectories therefore replayed at incorrect times. The loader now supports per-control timestamps or explicit interval boundaries, validates malformed timelines, and handles floating-point step boundaries consistently. Intervals shorter than one physics step may be skipped by model-timestep sampling.

## Tests

- Covered both timestamp conventions and irregular intervals
- Added validation cases for malformed inputs
- Added a real NPZ round-trip with a minimal MuJoCo model
- Verified the tracked shuffle trajectory

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -4033,12 +4033,16 @@ def make_trajectory(model: mujoco.MjModel, keys: list[int]) -> np.ndarray:
 def load_trajectory(npz_path: str, mjm: mujoco.MjModel, mjd: mujoco.MjData) -> np.ndarray:
   """Load ctrl sequence from NPZ and interpolate to model timestep.
 
-  If the trajectory dt differs from mjm.opt.timestep, each ctrl value is held
-  constant (zero-order hold) for the appropriate number of physics steps.
+  Controls are sampled on the model timestep using zero-order hold. ``times``
+  may contain one timestamp per control or interval boundaries with one extra
+  final timestamp. For per-control timestamps, the final interval repeats the
+  preceding interval duration, or one model timestep for a single control.
+  Sub-timestep control intervals may be skipped when no model-timestep sample
+  falls within them.
 
   The NPZ file should contain:
     - 'ctrl': array of shape (nstep, nu) with ctrl values
-    - 'times': array of shape (nstep,) with timestamps
+    - 'times': array of shape (nstep,) or (nstep + 1,) with timestamps
     - 'qpos' (optional): array of shape (1, nq) - initial state
     - 'qvel' (optional): array of shape (1, nv) - initial state
   """
@@ -4046,8 +4050,18 @@ def load_trajectory(npz_path: str, mjm: mujoco.MjModel, mjd: mujoco.MjData) -> n
   ctrl = data["ctrl"]
   times = data["times"]
 
+  if ctrl.ndim != 2:
+    raise ValueError(f"ctrl must have shape (nstep, nu), got {ctrl.shape}")
   if ctrl.shape[1] != mjm.nu:
     raise ValueError(f"ctrl shape {ctrl.shape} does not match model nu={mjm.nu}")
+  if times.ndim != 1 or len(times) not in (len(ctrl), len(ctrl) + 1) or not len(ctrl):
+    raise ValueError(f"times shape {times.shape} must contain {len(ctrl)} or {len(ctrl) + 1} timestamps")
+  if not np.all(np.isfinite(times)):
+    raise ValueError("times must be finite")
+
+  intervals = np.diff(times)
+  if np.any(intervals <= 0):
+    raise ValueError("times must be strictly increasing")
 
   # set initial state from first frame if available
   if "qpos" in data and data["qpos"].shape[1] == mjm.nq:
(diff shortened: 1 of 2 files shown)


anxkhn:
hey @thowell, could you please review this pr when you have a chance?

mujoco_warp/_src/io.py
```diff
@@ -3883,34 +3883,53 @@ def make_trajectory(model: mujoco.MjModel, keys: list[int]) -> np.ndarray:
...
   if "qvel" in data and data["qvel"].shape[1] == mjm.nv:
     mjd.qvel[:] = data["qvel"][0]
 
-  # determine decimation from timing
-  ctrl_dt = (times[1] - times[0]) if len(times) > 1 else mjm.opt.timestep
-  decimation = max(1, round(ctrl_dt / mjm.opt.timestep))
-
-  # expand: each ctrl held constant for decimation physics steps
-  return np.repeat(ctrl, decimation, axis=0)
+  boundaries = times
```
erikfrey: this is using quite a bit of advanced numpy. can we simplify this a bit to make it easier to follow?  something like this:

```py
  if len(times) == len(ctrl):
    final_dt = intervals[-1] if len(intervals) else mjm.opt.timestep
    times = np.append(times, times[-1] + final_dt)

  n_steps = int(np.round((times[-1] - times[0]) / mjm.opt.timestep))
  sample_times = times[0] + (np.arange(n_steps) + 1e-7) * mjm.opt.timestep
  ctrl_indices = np.searchsorted(times, sample_times, side="right") - 1
  return ctrl[ctrl_indices]
```

mujoco_warp/_src/io.py
```diff
@@ -3883,34 +3883,53 @@ def make_trajectory(model: mujoco.MjModel, keys: list[int]) -> np.ndarray:
...
   """
   data = np.load(npz_path)
   ctrl = data["ctrl"]
   times = data["times"]
 
+  if ctrl.ndim != 2:
+    raise ValueError(f"ctrl must have shape (nstep, nu), got {ctrl.shape}")
   if ctrl.shape[1] != mjm.nu:
     raise ValueError(f"ctrl shape {ctrl.shape} does not match model nu={mjm.nu}")
+  if times.ndim != 1 or len(times) not in (len(ctrl), len(ctrl) + 1) or not len(ctrl):
```
erikfrey: Having `not len(ctrl)` bundled inside the `times` shape check causes an empty `ctrl` array to fail with `ValueError: times shape (0,) must contain 0 or 1 timestamps`.

Could we validate `ctrl` emptiness upfront instead?
```python
if ctrl.ndim != 2 or len(ctrl) == 0:
  raise ValueError(f"ctrl must have shape (nstep, nu) with nstep > 0, got {ctrl.shape}")
if times.ndim != 1 or len(times) not in (len(ctrl), len(ctrl) + 1):
  raise ValueError(f"times shape {times.shape} must contain {len(ctrl)} or {len(ctrl) + 1} timestamps")
```

mujoco_warp/_src/io_test.py
```diff
@@ -16,7 +16,10 @@
 """Tests for io functions."""
 
 import dataclasses
+import pathlib
+import tempfile
 import warnings
+from types import SimpleNamespace
```
erikfrey: Could we avoid `types.SimpleNamespace` and use a minimal `mujoco.MjModel.from_xml_string(...)` / `mujoco.MjData(model)` across these tests instead?

Everywhere else in `mujoco_warp` we construct real models/data (as done in `test_load_trajectory_npz_roundtrip`), which keeps test patterns consistent and tests against actual MuJoCo data structures rather than duck-typed mocks.

mujoco_warp/_src/io_test.py
```diff
@@ -458,6 +461,95 @@ def _assert_eq(a, b, name):
...
+      "ctrl": np.arange(3)[:, None],
+      "times": times,
+    }
+
+    with mock.patch.object(np, "load", return_value=trajectory):
+      ctrl = io.load_trajectory("trajectory.npz", model, data)
+
+    np.testing.assert_array_equal(ctrl, trajectory["ctrl"][expected_indices])
+
+  def test_load_trajectory_npz_roundtrip(self):
```
erikfrey: `load_trajectory` is pretty simple - let's see if we can converge on a single unit test to exercise it.  I think this one (`test_load_trajectory_npz_roundtrip`) is sufficient to start from.

using `mock`  and `SimpleNamespace` feels a bit out of place, and although we do use `mock` elsewhere in the tests I'd prefer to use it sparingly.

Similarly, let's remove the tests traversing `__file__.parents[2]` looking for `benchmarks/` in the git repo root. If tests are executed against an installed wheel/package (where `benchmarks/` isn't included in `site-packages`), this test will fail. Also, `pathlib` isn't used elsewhere in the codebase (we use `etils.epath`).

Let's focus on making `test_load_trajectory_npz_roundtrip` have sufficient coverage of this load trajectory behavior.  Perhaps you can combine it with the parameterized stuff above?

erikfrey (approved):
This is a nice change, thank you.  Just a few nits.

PR #1584, opened 2026-08-14
set island fields in put_data

thowell:
`io.put_data` currently will silently skip setting island fields. this pr sets these island fields and tests this functionality.


kbayes (approved):

thowell:
updated the tests with inline xmls and nworld parameterization

PR #1586, opened 2026-08-15
Stabilize Windows collision and mixed-block tests

ArrebolBlack:
On Windows with Python 3.14, two otherwise valid tests fail for platform-specific fixture reasons.

The complex mesh-plane fixture places one tip exactly tangent to the plane. MuJoCo reports an approximately zero-depth numerical contact while Warp omits it; lifting the body by 1 mm preserves the intended penetrating contact and makes contact generation unambiguous.

The mixed-block solve fixture builds 70 nested bodies, which can overflow the Windows XML parser stack before the solver runs. Using `M_BLOCK_DENSE_MAX + 1` (65 DOFs currently) remains the minimal coverage of the sparse LDL branch without exercising that parser limit.

This changes only test fixtures; collision and solver behavior are unchanged. Validated on Windows with Python 3.14.6: the two targeted tests passed, the full suite reported 1262 passed and 19 skipped, and `pre-commit run -a` passed all hooks.

### DIFF
--- a/mujoco_warp/_src/collision_driver_test.py
+++ b/mujoco_warp/_src/collision_driver_test.py
@@ -433,7 +433,8 @@ class CollisionTest(parameterized.TestCase):
           </asset>
           <worldbody>
             <geom size="40 40 40" type="plane"/>
-            <body pos="0.0 2.0 0.0" euler="90 90 0">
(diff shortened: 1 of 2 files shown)


thowell:
@ArrebolBlack thank you for contributing to mujoco warp! please post a printout of the test failure(s) you are seeing on windows and a uv lock file. thanks!

PR #1593, opened 2026-08-19
Fix numpy array types to avoid expensive object arrays

camevor:
Small fix to avoid unintentional creation of numpy object arrays with warp datatypes.

This improves startup time and memory use, e.g. 12% end-to-end speedup on the G1_flat scene (single sim step), from 2.38 s to 2.11 s.

```bash
uv run mjwarp-testspeed /tmp/benchmark_assets/unitree_g1_flat/scene_flat.xml --nworld=8192 --nconmax=48 --njmax=192 --nstep=1

```


thowell:
nice, thanks @camevor!

thowell (approved):

PR #1594, opened 2026-08-19
gjk restoration for fallback routine

thowell:
`gjk_phase`: spheres and capsules are temporarily shrunk to points and line segments to accelerate GJK simplex convergence. When `dist <= tolerance`, the algorithm falls back to rerunning GJK and EPA with un-shrunk geometries. Previously, this recovery path attempted to reconstruct dimensions and margins arithmetically (`geom.margin = full_margin - size`, `geom.size[0] = size`), which caused two bugs:

- For heterogeneous pairs involving a sphere or capsule against a cylinder, ellipsoid, or box, `size` for the second geom remained 0.0, overwriting `geom.size[0]` with 0.0 and collapsing cylinders into 1D line segments, ellipsoids into 2D ellipses, and boxes into 2D rectangles during the rerun.
- `full_margin` was defined as `size + 0.5 * margin`, so evaluating `full_margin - size` halved the original margin on spheres/capsules and completely wiped the margin on non-rounded shapes, corrupting EPA Minkowski support expansion and contact distances.

This change caches `orig_size` and `orig_margin` at kernel entry and directly restores them for the fallback routine, matching the reference MuJoCo C implementation. Regression tests have been added for capsule-cylinder, sphere-mesh with margin, and capsule-ellipsoid scenarios that utilize the fallback routine.


mujoco_warp/_src/collision_driver_test.py
```diff
@@ -1243,6 +1243,124 @@ def test_ccd_margin_dist(self):
     self.assertEqual(mjd.nefc, 0, "Classic MuJoCo should have no active constraints")
     self.assertEqual(d.nefc.numpy()[0], 0, "MuJoCo Warp should have no active constraints")
 
+  def test_capsule_cylinder_fallback(self):
```
kbayes: Delete this and just had the following test to collision_gjk_test.py. Confirmed locally that it repo's the bug.

  def test_cylinder_capsule(self):
    """Test deep penetration between an cylinder and a capsule to trigger EPA."""
    _, _, m, d = test_data.fixture(
    xml="""
    <mujoco>
      <worldbody>
        <geom type="cylinder" size="2 4" pos="0 0 0"/>
        <geom type="capsule" size="1 1" pos="0 0 5"/>
      </worldbody>
    </mujoco>"""
    )
    dist, _, _, _ = _geom_dist(m, d, 0, 1)
    self.assertAlmostEqual(-1.0, dist)

kbayes:
@thowell good catch!

thowell:
updated the tests

mujoco_warp/_src/collision_gjk_test.py
```diff
@@ -695,6 +695,21 @@ def test_cylinder_box(self):
...
+      <mujoco>
+        <worldbody>
+          <geom type="cylinder" size="2 4" pos="0 0 0"/>
+          <geom type="capsule" size="1 1" pos="0 0 5"/>
+        </worldbody>
+      </mujoco>
+      """
+    )
+    dist, _, _, _ = _geom_dist(m, d, 0, 1)
+    self.assertAlmostEqual(-1.0, dist)
```
kbayes: Looks like you need to loosen the tolerance.

kbayes (approved):

PR #1597, opened 2026-08-20
Size the box-box edge-axis tie-break to float32 rounding

jsw7460:
Fix box_box ghost contacts on level resting surfaces under FP32

When two boxes rest flat against each other, four of the nine edge-cross  separating axes are parallel to the face normal, creating an exact tie in separation depth. In FP32, the legacy relative margin `1.0 - 1e-12` rounds to 1.0, allowing FP32 cancellation noise in cross-product evaluation to  erroneously pick an edge axis over the face axis. The subsequent edge-edge routine degenerates on face contacts and generates phantom contact points hundreds of millimeters deep.

This patch introduces an absolute `edge_slack` proportional to the sum of the box half-extents (`1e-6 * sum(sizes)`), ensuring face axes are stably preferred on near-ties regardless of near-zero resting separation. Adds a regression test covering the degenerate resting pose.

Simply increasing the relative margin does not work: the separations are sums of box extents times rotation entries, so their FP32 error is absolute (~5e-8 m for supports of order 0.4 m) against a resting separation of ~1e-4 m — a relative error near 5e-4. Both 1e-6 and 1e-4 leave every failing pose failing.

(Note: Mujoco's C has since rewritten this collider on main using Sutherland-Hodgman face clipping and precision-aware constants split on mjUSESINGLE. Porting that is likely the better long-term direction)

### DIFF
--- a/mujoco_warp/_src/collision_primitive_core.py
+++ b/mujoco_warp/_src/collision_primitive_core.py
@@ -638,6 +638,15 @@ def box_box(
   separation = wp.float32(margin + s_sum_3[0] + s_sum_3[1] + s_sum_3[2])
   axis_code = wp.int32(-1)
 
+  # An edge axis has to beat the best face axis by more than the rounding error of the
+  # separations themselves. Those are sums of box extents times rotation entries, so their
+  # float32 error is absolute and proportional to the extents, while the separation of a
+  # resting contact is near zero -- a relative margin cannot express that. It matters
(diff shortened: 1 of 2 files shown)


thowell:
@jsw7460 thank you for contributing to mujoco warp!

@kbayes please take a look. thanks!

thowell:
@jsw7460 thank you for contributing to mujoco warp! closing this issue in favor a complete refactor #1614

PR #1598, opened 2026-08-20
Extract set_const functions and tests into dedicated set_const module

erikfrey:
io.py has grown large and contains multiple distinct responsibilities. This change extracts set_const, set_const_0, set_const_fixed, set_const_spring, set_length_range, and their 22 associated GPU kernels from io.py into a new mujoco_warp/_src/set_const.py module. This mirrors the C MuJoCo layout (engine_setconst.c vs engine_io.c) by cleanly separating derived constant physics precomputations from host/device memory transfer. The corresponding unit tests are moved from io_test.py into set_const_test.py. All functions continue to be re-exported from io.py and __init__.py for backward compatibility.


mujoco_warp/_src/set_const.py
```diff
@@ -0,0 +1,981 @@
+# Copyright 2025 The Newton Developers
```
thowell: nit: 2026?

mujoco_warp/_src/set_const.py
```diff
@@ -0,0 +1,981 @@
...
+from mujoco_warp._src import types
+from mujoco_warp._src.types import MJ_MINVAL
+from mujoco_warp._src.types import BiasType
+from mujoco_warp._src.types import TrnType
+from mujoco_warp._src.types import vec10
+
+wp.set_module_options({"default_grid_stride": False})
+
+
+# kernel_analyzer: off
```
thowell, line 33: lets add a todo to refactor this file to enable the kernel_analyzer?

(think the challenge is that many of these kernel arguments are `Model` fields and they are usually inputs but here are outputs)

mujoco_warp/_src/set_const_test.py
```diff
@@ -0,0 +1,1048 @@
+# Copyright 2025 The Newton Developers
```
thowell: nit: 2026

thowell (approved):
thanks @erikfrey!

mujoco_warp/_src/set_const.py
erikfrey: Updated to 2026.

mujoco_warp/_src/set_const.py
erikfrey: Added a TODO comment above `# kernel_analyzer: off`.

mujoco_warp/_src/set_const_test.py
erikfrey: Updated to 2026.

erikfrey:
Applied review suggestions in 231a2c3e:
- Updated copyright year to 2026 in `set_const.py` and `set_const_test.py`.
- Added a TODO comment above `# kernel_analyzer: off` in `set_const.py` to refactor the kernels to enable the kernel analyzer.

erikfrey:
Updated the TODO in `set_const.py` (9d5b9a04) to reflect that the kernel analyzer should be updated to support Model fields as kernel outputs.

PR #1600, opened 2026-08-21
fix `eval_ccd_write_contact`

thowell:


kbayes:
@thowell is it possible to provide a unit test?

thowell:
@kbayes added a test that fails before the change to collision_convex.py and now passes with the fix

kbayes (approved):

PR #1602, opened 2026-08-24
Improve renderer.

kevinzakka:
This PR fixes several rendering correctness issues in the raytracer. In every image below: native MuJoCo (OpenGL) on the left, mjwarp @ main in the middle, and mjwarp @ this PR on the right.

**Meshes were clipped.** The BVH stored half-extents of `0.5 * (pmax - pmin)` about the mesh origin. For meshes whose vertices aren't centered on the origin, this bounding box does not contain the full mesh, causing chunks of geometry to disappear at some view angles.

mesh clipping

*Look at the white bottle's cap: main clips its top rim away, leaving a gray gap between cap and collar.*

**Coincident surfaces flickered.** Which geom won depended on BVH traversal order. Ties now resolve to the higher geom index: the later-declared geom wins, matching native MuJoCo, whose depth test leaves the last draw on top.

coincident geoms

*Two exactly coincident boxes, red declared before green. Main shows whichever the BVH happens to visit first; this PR shows the later-declared green, like native MuJoCo.*

**Backfaces rendered black** when backface culling was disabled. Shading is now two-sided.

two sided shading

*The bin's rim and interior wall face away from the camera: main leaves them black and flat, this PR shades them.*

**Smooth meshes rendered faceted.** Authored vertex normals were ignored. The renderer now interpolates vertex normals and rejects MuJoCo's [0, 0, 1] sentinel normals against the face normal, matching MuJoCo's mesh-normal handling. This is gated by `enable_vertex_normals`.

vertex normals

*Main shades each face flat, banding the ribbed bottle; this PR matches the reference.*

vertex normals and shadows

*The same fix from a wrist camera; both bottles lose their faceting.*

Two smaller changes are included:
* `shadow_light_fraction` exposes the previously hardcoded 0.3 in-shadow light fraction.
* The megakernel is built once per context instead of on every `render()` call.

There is no measurable performance regression on `aloha_clutter_vision` (2048 worlds, 4 cameras @ 64x64, RGB + depth + segmentation, shadows; RTX 5090):

| | main | this PR |
|---|---|---|
| steps/s | 7,921 | 7,936 |
| render, ms/step | 126.18 | 125.97 |

### DIFF
--- a/mujoco_warp/_src/ray.py
+++ b/mujoco_warp/_src/ray.py
@@ -28,6 +28,10 @@ from mujoco_warp._src.types import vec6
 
 wp.set_module_options({"enable_backward": False, "default_grid_stride": False})
 
+# Widen ray prune bounds so a coincident hit survives to the closest-hit tie-break.
+RAY_TOL_REL = 1.0e-6
+RAY_TOL_ABS = 1.0e-9
+
 
 @wp.func
 def _ray_map(pos: wp.vec3, mat: wp.mat33, pnt: wp.vec3, vec: wp.vec3) -> Tuple[wp.vec3, wp.vec3]:
@@ -714,19 +718,20 @@ def ray_mesh_with_bvh(
   f = int(-1)
 
   lpnt, lvec = _ray_map(pos, mat, pnt, vec)
+  mesh_bvh = mesh_bvh_id[mesh_geom_id]
 
   # Culling drops back-facing triangles, not the mesh: like a rasterizer, march
   # past culled hits to the next surface (bounded to keep the kernel in check).
   offset = float(0.0)
   for _ in range(8):
-    hit = wp.mesh_query_ray(mesh_bvh_id[mesh_geom_id], lpnt + lvec * offset, lvec, max_t - offset, t, u, v, sign, n, f)
+    hit = wp.mesh_query_ray(mesh_bvh, lpnt + lvec * offset, lvec, max_t - offset, t, u, v, sign, n, f)
     if not hit:
       return -1.0, wp.vec3(0.0, 0.0, 0.0), 0.0, 0.0, -1, -1
     if (not cull_backfaces) or wp.dot(lvec, n) < 0.0:
       normal = mat @ n
       normal = wp.normalize(normal)
       return offset + t, normal, u, v, f, mesh_geom_id
-    offset += t * (1.0 + 1.0e-6) + 1.0e-9
+    offset += t * (1.0 + RAY_TOL_REL) + RAY_TOL_ABS
 
   return -1.0, wp.vec3(0.0, 0.0, 0.0), 0.0, 0.0, -1, -1
 
--- a/mujoco_warp/_src/ray.py
+++ b/mujoco_warp/_src/ray.py
@@ -714,17 +714,19 @@ def ray_mesh_with_bvh(
   f = int(-1)
 
   lpnt, lvec = _ray_map(pos, mat, pnt, vec)
-  hit = wp.mesh_query_ray(mesh_bvh_id[mesh_geom_id], lpnt, lvec, max_t, t, u, v, sign, n, f)
 
-  if not hit:
-    return -1.0, wp.vec3(0.0, 0.0, 0.0), 0.0, 0.0, -1, -1
-
-  # Front-face hit, or back-face hit when cull is disabled: rotate the
-  # local-space normal into world space and return the hit.
-  if (not cull_backfaces) or wp.dot(lvec, n) < 0.0:
-    normal = mat @ n
-    normal = wp.normalize(normal)
-    return t, normal, u, v, f, mesh_geom_id
+  # Culling drops back-facing triangles, not the mesh: like a rasterizer, march
+  # past culled hits to the next surface (bounded to keep the kernel in check).
+  offset = float(0.0)
+  for _ in range(8):
+    hit = wp.mesh_query_ray(mesh_bvh_id[mesh_geom_id], lpnt + lvec * offset, lvec, max_t - offset, t, u, v, sign, n, f)
+    if not hit:
+      return -1.0, wp.vec3(0.0, 0.0, 0.0), 0.0, 0.0, -1, -1
+    if (not cull_backfaces) or wp.dot(lvec, n) < 0.0:
+      normal = mat @ n
+      normal = wp.normalize(normal)
+      return offset + t, normal, u, v, f, mesh_geom_id
+    offset += t * (1.0 + 1.0e-6) + 1.0e-9
 
   return -1.0, wp.vec3(0.0, 0.0, 0.0), 0.0, 0.0, -1, -1
 
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -356,11 +356,12 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
     bounds_nr = int(0)
     ngeom = bvh_ngeom + flex_bvh_ngeom
 
-    # A coincident geom sits exactly at dist; widen the prune so it survives to
-    # reach the tie-break below.
+    # max_t is exclusive: widen so coincident hits reach the tie-break below.
     while wp.bvh_query_next(query, bounds_nr, dist * (1.0 + 1.0e-6) + 1.0e-9):
       gi_global = bounds_nr
       local_id = gi_global - (worldid * ngeom)
+
+      query_dist = dist * (1.0 + 1.0e-6) + 1.0e-9
 
       d = float(-1.0)
       hit_mesh_id = int(-1)
@@ -397,7 +398,7 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
             geom_xmat_in[worldid, gi],
             ray_origin_world,
             ray_dir_world,
-            dist,
+            query_dist,
             cull_backfaces,
           )
       if wp.static(int(GeomType.SPHERE) in geom_ray_types):
@@ -465,7 +466,7 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
               geom_xmat_in[worldid, gi],
               ray_origin_world,
               ray_dir_world,
-              dist * (1.0 + 1.0e-6) + 1.0e-9,  # max_t is exclusive
+              query_dist,
               cull_backfaces,
             )
       if wp.static(int(GeomType.FLEX) in geom_ray_types):
@@ -502,7 +503,7 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
               d = 0.0 if hit else -1.0
             else:
               flex_gr = flex_group_root[worldid, flexid]
-              d, n, u, v, f = ray_flex_with_bvh(flex_bvh_id, flexid, flex_gr, ray_origin_world, ray_dir_world, dist)
+              d, n, u, v, f = ray_flex_with_bvh(flex_bvh_id, flexid, flex_gr, ray_origin_world, ray_dir_world, query_dist)
               if d >= 0.0:
                 hit_mesh_id = flexid
 
@@ -517,11 +518,9 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
         if d >= 0.0 and d < dist:
           return hit_geom_id, d, n, u, v, f, hit_mesh_id
       else:
-        # MuJoCo's GL depth test (GL_LEQUAL) leaves the last-drawn geom on top, so
-        # coincident surfaces resolve to the higher geom index.
-        closer = d < dist
-        if wp.abs(d - dist) <= 1.0e-6 * wp.max(dist, 1.0) and geom_id >= 0 and hit_geom_id > geom_id:
-          closer = True
+        # Ties go to the higher geom index, like MuJoCo's GL depth test.
+        tol = 1.0e-6 * wp.max(dist, 1.0)
+        closer = (d < dist - tol) or (wp.abs(d - dist) <= tol and (geom_id < 0 or hit_geom_id > geom_id))
         if d >= 0.0 and closer:
           dist = d
           normal = n
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -848,6 +848,11 @@ def render(m: Model, d: Data, rc: RenderContext):
       wp.static(rc_static["enable_backface_culling"]),
     )
 
+    if wp.static(not rc_static["enable_backface_culling"]):
+      # Two-sided shading: light a back-facing hit as if it faced the viewer.
+      if geom_id >= 0 and wp.dot(normal, ray_dir_world) > 0.0:
+        normal = -normal
+
     splat_color = wp.vec3(0.0)
     splat_transmittance = float(1.0)
     splat_depth = float(-1.0)
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -348,7 +348,9 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
     bounds_nr = int(0)
     ngeom = bvh_ngeom + flex_bvh_ngeom
 
-    while wp.bvh_query_next(query, bounds_nr, dist):
+    # A coincident geom sits exactly at dist; widen the prune so it survives to
+    # reach the tie-break below.
+    while wp.bvh_query_next(query, bounds_nr, dist * (1.0 + 1.0e-6) + 1.0e-9):
       gi_global = bounds_nr
       local_id = gi_global - (worldid * ngeom)
 
@@ -455,7 +457,7 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
               geom_xmat_in[worldid, gi],
               ray_origin_world,
               ray_dir_world,
-              dist,
+              dist * (1.0 + 1.0e-6) + 1.0e-9,  # max_t is exclusive
               cull_backfaces,
             )
       if wp.static(int(GeomType.FLEX) in geom_ray_types):
@@ -507,7 +509,12 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
         if d >= 0.0 and d < dist:
           return hit_geom_id, d, n, u, v, f, hit_mesh_id
       else:
-        if d >= 0.0 and d < dist:
+        # MuJoCo's GL depth test (GL_LEQUAL) leaves the last-drawn geom on top, so
+        # coincident surfaces resolve to the higher geom index.
+        closer = d < dist
+        if wp.abs(d - dist) <= 1.0e-6 * wp.max(dist, 1.0) and geom_id >= 0 and hit_geom_id > geom_id:
+          closer = True
+        if d >= 0.0 and closer:
           dist = d
           normal = n
           geom_id = hit_geom_id
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -20,6 +20,8 @@ import warp as wp
 
 from mujoco_warp._src import math
 from mujoco_warp._src.bvh import SPLAT_MIN_RESPONSE
+from mujoco_warp._src.ray import RAY_TOL_ABS
+from mujoco_warp._src.ray import RAY_TOL_REL
 from mujoco_warp._src.ray import ray_box
 from mujoco_warp._src.ray import ray_capsule
 from mujoco_warp._src.ray import ray_cylinder
@@ -357,11 +359,11 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
     ngeom = bvh_ngeom + flex_bvh_ngeom
 
     # max_t is exclusive: widen so coincident hits reach the tie-break below.
-    while wp.bvh_query_next(query, bounds_nr, dist * (1.0 + 1.0e-6) + 1.0e-9):
+    while wp.bvh_query_next(query, bounds_nr, dist * (1.0 + RAY_TOL_REL) + RAY_TOL_ABS):
       gi_global = bounds_nr
       local_id = gi_global - (worldid * ngeom)
 
-      query_dist = dist * (1.0 + 1.0e-6) + 1.0e-9
+      query_dist = dist * (1.0 + RAY_TOL_REL) + RAY_TOL_ABS
 
       d = float(-1.0)
       hit_mesh_id = int(-1)
@@ -519,7 +521,7 @@ def _make_cast_ray(geom_ray_types: Tuple[int], first_hit: bool = False) -> wp.Fu
           return hit_geom_id, d, n, u, v, f, hit_mesh_id
       else:
         # Ties go to the higher geom index, like MuJoCo's GL depth test.
-        tol = 1.0e-6 * wp.max(dist, 1.0)
+        tol = RAY_TOL_REL * wp.max(dist, 1.0)
         closer = (d < dist - tol) or (wp.abs(d - dist) <= tol and (geom_id < 0 or hit_geom_id > geom_id))
         if d >= 0.0 and closer:
           dist = d
@@ -860,14 +862,12 @@ def _build_megakernel(m: Model, rc: RenderContext):
       face = wp.transpose(mat) @ normal
       tri = mesh_facenormal[mesh_faceadr[mesh_id] + f]
       adr = mesh_normaladr[mesh_id]
-      normal = wp.normalize(
-        mat
-        @ (
-          vertex_normal(mesh_normal[adr + tri[0]], face) * u
-          + vertex_normal(mesh_normal[adr + tri[1]], face) * v
-          + vertex_normal(mesh_normal[adr + tri[2]], face) * (1.0 - u - v)
-        )
+      vec = (
+        vertex_normal(mesh_normal[adr + tri[0]], face) * u
+        + vertex_normal(mesh_normal[adr + tri[1]], face) * v
+        + vertex_normal(mesh_normal[adr + tri[2]], face) * (1.0 - u - v)
       )
+      normal = wp.normalize(mat @ vec)
 
     if wp.static(not rc_static["enable_backface_culling"]):
       # Two-sided shading: light a back-facing hit as if it faced the viewer.
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -674,18 +674,8 @@ def _make_compute_lighting(cast_ray_first_hit: wp.Function) -> wp.Function:
   return compute_lighting
 
 
-@event_scope
-def render(m: Model, d: Data, rc: RenderContext):
-  """Render the current frame.
-
-  Outputs are stored in buffers within the render context.
-
-  Args:
-    m: The model on device.
-    d: The data on device.
-    rc: The render context on device.
-  """
-  rc.seg_data.fill_(wp.vec2i(-1, -1))
+def _build_megakernel(m: Model, rc: RenderContext):
+  """Construct the specialised megakernel for this context."""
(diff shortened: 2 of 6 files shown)


thowell:
thanks @kevinzakka! will you post some before and after images that highlight the changes? additionally we can run the aloha clutter vision benchmark to check performance before and after: `uv run python3 benchmarks/run.py -f aloha_clutter_vision`.

kevinzakka:
@thowell This is still a draft PR, will review closely with @StafaH once his schedule frees up.

kevinzakka:
FYI, if you're wondering why the white bottle's cap still looks off in the mesh clipping image it's because that asset's under-rim triangles are wound inward, so the default `enable_backface_culling=True` culls them and we render the bottle body underneath. Native MuJoCo doesn't cull. If we set `enable_backface_culling=False` in mjwarp after my fixes, the result more closely aligns with MuJoCo:

culling_comparison

So question to @StafaH and @thowell: now that backfaces shade correctly, should the default just be `False` to match native MuJoCo? Culling is slightly faster but shows holes on any mesh with imperfect winding.

havess:
Drive-by 2 cents: render two-sided by default, especially since AFAICT culling backfaces is a visual choice here and not performance related. Inverted polygon authoring is generally a mistake/bad form and rasterizers cull backfaces for perf reasons.

thowell:
awesome work @kevinzakka!

some things we should consider updating:

| Item | Location | Issue | Recommendation |
| :--- | :--- | :--- | :--- |
| **1. Coincident Tie-Break** | [render.py:523-528](file:///usr/local/google/home/taylorhowell/mujoco_warp/mujoco_warp/_src/render.py#L523) | Asymmetric check allows lower geom ID to win if numerical roundoff yields $d_0 < d_1$ within tolerance. | Use a symmetric window: `closer = (d < dist - tol) or (abs(d - dist) <= tol and (geom_id < 0 or hit_geom_id > geom_id))`. |
| **2. BVH Query Distance** | [render.py:401,507](file:///usr/local/google/home/taylorhowell/mujoco_warp/mujoco_warp/_src/render.py#L401) | `HFIELD` (line 401) and `FLEX` (line 507) pass un-widened `dist`, risking premature pruning by `mesh_query_ray`. | Define `query_dist = dist * (1.0 + 1e-6) + 1e-9` once at line 362 and pass to all BVH queries. |
| **3. Match MuJoCo Defaults** | [render_util.py:324](file:///usr/local/google/home/taylorhowell/mujoco_warp/mujoco_warp/_src/render_util.py#L324) | `enable_backface_culling=True` culls single-sided geometry (e.g. bottle caps, open bins). | Default `enable_backface_culling=False` to match native MuJoCo's OpenGL renderer. |
| **4. Megakernel Typing** | [types.py:2657](file:///usr/local/google/home/taylorhowell/mujoco_warp/mujoco_warp/_src/types.py#L2657) | `_megakernel` is assigned dynamically as an undeclared attribute on `RenderContext`. | Declare `_megakernel: wp.Kernel | None = None` on `RenderContext` matching `types.py` style. |
| **5. Test Suite Additions** | [render_test.py](file:///usr/local/google/home/taylorhowell/mujoco_warp/mujoco_warp/_src/render_test.py) | Missing tests for disabled vertex normals, cache persistence, and $\Delta d$ tie-breaking. | Add 3 unit tests with inlined XMLs in `test_data.fixture`, tested fields pre-filled (`wp.inf` for floats, `-1` for counts/IDs), and visible benefit explanations above each test. |

---

### 1. Fix Asymmetric Tie-Breaking in `cast_ray`

If two coincident surfaces compute slightly different distances due to numerical precision ($|d_0 - d_1| \le \text{tol}$), evaluating `closer = d < dist` allows a lower-index geom visited second to overwrite an earlier higher-index geom. Partitioning the distance check makes the tie-breaker deterministic:

```diff
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -522,9 +522,10 @@ def cast_ray(
       else:
         # MuJoCo's GL depth test (GL_LEQUAL) leaves the last-drawn geom on top, so
         # coincident surfaces resolve to the higher geom index.
-        closer = d < dist
-        if wp.abs(d - dist) <= 1.0e-6 * wp.max(dist, 1.0) and geom_id >= 0 and hit_geom_id > geom_id:
-          closer = True
+        tol = 1.0e-6 * wp.max(dist, 1.0)
+        closer = (d < dist - tol) or (
+          wp.abs(d - dist) <= tol and (geom_id < 0 or hit_geom_id > geom_id)
+        )
         if d >= 0.0 and closer:
           dist = d
           normal = n
```

---

### 2. Apply Widened `query_dist` Consistently Across All BVH Queries

`dist * (1.0 + 1.0e-6) + 1.0e-9` prevents premature pruning by `wp.mesh_query_ray` due to exclusive `max_t`. Defining `query_dist` once ensures `HFIELD` and `FLEX` receive the same expansion:

```diff
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -360,7 +360,8 @@ def cast_ray(
     # A coincident geom sits exactly at dist; widen the prune so it survives to
     # reach the tie-break below.
-    while wp.bvh_query_next(query, bounds_nr, dist * (1.0 + 1.0e-6) + 1.0e-9):
+    query_dist = dist * (1.0 + 1.0e-6) + 1.0e-9
+    while wp.bvh_query_next(query, bounds_nr, query_dist):
       gi_global = bounds_nr
       local_id = gi_global - (worldid * ngeom)
@@ -400,7 +401,7 @@ def cast_ray(
             geom_xmat_in[worldid, gi],
             ray_origin_world,
             ray_dir_world,
-            dist,
+            query_dist,
             cull_backfaces,
           )
@@ -466,7 +467,7 @@ def cast_ray(
               geom_xmat_in[worldid, gi],
               ray_origin_world,
               ray_dir_world,
-              dist * (1.0 + 1.0e-6) + 1.0e-9,  # max_t is exclusive
+              query_dist,
               cull_backfaces,
             )
@@ -506,7 +507,7 @@ def cast_ray(
             else:
               flex_gr = flex_group_root[worldid, flexid]
-              d, n, u, v, f = ray_flex_with_bvh(flex_bvh_id, flexid, flex_gr, ray_origin_world, ray_dir_world, dist)
+              d, n, u, v, f = ray_flex_with_bvh(flex_bvh_id, flexid, flex_gr, ray_origin_world, ray_dir_world, query_dist)
               if d >= 0.0:
                 hit_mesh_id = flexid
```

---

### 3. Match Native MuJoCo Defaults: `enable_backface_culling = False`

Native MuJoCo disables backface culling (`glDisable(GL_CULL_FACE)`). Now that two-sided shading properly illuminates back-facing hits, defaulting `enable_backface_culling=False` provides 1:1 parity and prevents unintended holes in open meshes:

```diff
--- a/mujoco_warp/_src/render_util.py
+++ b/mujoco_warp/_src/render_util.py
@@ -324,7 +324,7 @@ def create_render_context(
   use_ambient_lighting: bool = True,
   use_precomputed_rays: bool = True,
   render_skybox: bool = False,
-  enable_backface_culling: bool = True,
+  enable_backface_culling: bool = False,
   shadow_light_fraction: float = 0.3,
   enable_vertex_normals: bool = True,
   enable_specular: bool = True,
```

```diff
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -2548,7 +2548,7 @@ class RenderContext:
     enable_backface_culling: drop primitive ray hits whose normal faces away
       from the ray (i.e. the ray origin is inside the geom). Matches MuJoCo's
-      mesh-ray rule. When False, the renderer reports inner-surface hits, which
-      are shaded two-sided.
+      mesh-ray rule. Default False to match native MuJoCo's two-sided rendering.
```

---

### 4. Declare `_megakernel` Concrete Field on `RenderContext` (Option B)

Declaring `_megakernel: wp.Kernel | None = None` matches the codebase's pattern for optional/cached dataclass fields (e.g. `compact_m_full: Optional["Model"] = None`, `passive: Callable | None = None`) without introducing `dataclasses.field`. It provides explicit static typing and prevents dynamic attribute warnings:

```diff
--- a/mujoco_warp/_src/types.py
+++ b/mujoco_warp/_src/types.py
@@ -2654,6 +2654,7 @@ class RenderContext:
   enable_specular: bool
   enable_emission: bool
   enable_per_light_ambient: bool
+  _megakernel: wp.Kernel | None = None
```

---

### 5. Suggested Test Additions in `render_test.py`

**Test: `test_vertex_normals_can_be_disabled`**  
*Benefit:* Tests compiler gating for `enable_vertex_normals=False`, confirming that disabling vertex normals falls back to faceted face normals instead of smooth interpolation on curved surfaces. Pre-filling output buffers ensures the kernel actively writes fresh outputs.

```python
  def test_vertex_normals_can_be_disabled(self):
    n, r = 64, 0.3
    i = np.arange(n) + 0.5
    phi = np.arccos(1.0 - 2.0 * i / n)
    theta = np.pi * (1.0 + 5.0**0.5) * i
    pts = r * np.stack([np.cos(theta) * np.sin(phi), np.sin(theta) * np.sin(phi), np.cos(phi)], axis=-1)
    vert_str = " ".join(f"{x:.5f}" for x in pts.ravel())
    mjm, _, m, d = test_data.fixture(
      xml=f"""
      <mujoco>
        <asset><mesh name="sphere" vertex="{vert_str}"/></asset>
        <worldbody>
          <light pos="0 -2 1" dir="0 0.9 -0.45" directional="true" diffuse="0.55 0.55 0.55"/>
          <camera pos="0 -1.6 0" xyaxes="1 0 0 0 0 1" fovy="30"/>
          <geom type="mesh" mesh="sphere" rgba="0.8 0.8 0.8 1"/>
        </worldbody>
      </mujoco>
      """
    )
    rc_smooth = mjw.create_render_context(
      mjm, cam_res=(64, 64), render_rgb=True, render_depth=True, render_seg=True, enable_vertex_normals=True, use_ambient_lighting=False
    )
    rc_flat = mjw.create_render_context(
      mjm, cam_res=(64, 64), render_rgb=True, render_depth=True, render_seg=True, enable_vertex_normals=False, use_ambient_lighting=False
    )
    rc_smooth.rgb_data.fill_(0)
    rc_smooth.depth_data.fill_(wp.inf)
    rc_smooth.seg_data.fill_(wp.vec2i(-1, -1))
    mjw.render(m, d, rc_smooth)

    rc_flat.rgb_data.fill_(0)
    rc_flat.depth_data.fill_(wp.inf)
    rc_flat.seg_data.fill_(wp.vec2i(-1, -1))
    mjw.render(m, d, rc_flat)

    red_smooth = _unpack_rgb(rc_smooth.rgb_data.numpy()[0])[..., 0]
    red_flat = _unpack_rgb(rc_flat.rgb_data.numpy()[0])[..., 0]
    seg = rc_smooth.seg_data.numpy()[0].reshape(-1, 2)[..., 0]

    unique_smooth = np.unique(red_smooth[seg == 0])
    unique_flat = np.unique(red_flat[seg == 0])
    self.assertGreater(len(unique_smooth), len(unique_flat) * 2)
```

**Test: `test_coincident_geoms_with_delta_d_resolves_to_higher_index`**  
*Benefit:* Validates tie-break symmetry when coincident surfaces have slight numerical distance variations ($10^{-7}$), ensuring the higher geom ID wins deterministically regardless of BVH traversal order. Inlining XML and filling depth with `wp.inf` and `seg_data` with `-1` confirms both float and count/ID fields are actively updated by the raytracer.

```python
  def test_coincident_geoms_with_delta_d_resolves_to_higher_index(self):
    mjm, _, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <light pos="0 -1 1"/>
          <camera pos="0 -2 0" xyaxes="1 0 0 0 0 1"/>
          <geom type="box" size="0.2 0.2 0.2" rgba="1 0 0 1"/>
          <geom type="box" pos="0 1e-7 0" size="0.2 0.2 0.2" rgba="0 1 0 1"/>
        </worldbody>
      </mujoco>
      """
    )
    rc = mjw.create_render_context(mjm, cam_res=(32, 32), render_depth=True, render_seg=True)
    rc.depth_data.fill_(wp.inf)
    rc.seg_data.fill_(wp.vec2i(-1, -1))
    mjw.render(m, d, rc)
    seg = rc.seg_data.numpy()[0].reshape(-1, 2)[..., 0]
    depth = rc.depth_data.numpy()[0].reshape(-1)

    self.assertGreater(np.count_nonzero(seg == 1), 0)
    self.assertEqual(np.count_nonzero(seg == 0), 0)
    self.assertTrue(np.all(np.isfinite(depth[seg == 1])))
```

**Test: `test_megakernel_cached_across_renders`**  
*Benefit:* Ensures `rc._megakernel` is created on the initial render and preserved on subsequent invocations, protecting against module leaks and unnecessary recompilations across simulation loops and JAX FFI calls. Pre-filling float and count buffers confirms execution across multiple frames.

```python
  def test_megakernel_cached_across_renders(self):
    mjm, _, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <worldbody>
          <light pos="0 0 2"/>
          <camera pos="0 -2 0" xyaxes="1 0 0 0 0 1"/>
          <geom type="sphere" size="0.2" rgba="1 0 0 1"/>
        </worldbody>
      </mujoco>
      """
    )
    rc = mjw.create_render_context(mjm, cam_res=(32, 32), render_rgb=True, render_depth=True, render_seg=True)
    self.assertIsNone(getattr(rc, "_megakernel", None))

    rc.rgb_data.fill_(0)
    rc.depth_data.fill_(wp.inf)
    rc.seg_data.fill_(wp.vec2i(-1, -1))
    mjw.render(m, d, rc)
    kernel_first = rc._megakernel
    self.assertIsNotNone(kernel_first)

    rc.rgb_data.fill_(0)
    rc.depth_data.fill_(wp.inf)
    rc.seg_data.fill_(wp.vec2i(-1, -1))
    mjw.render(m, d, rc)
    self.assertIs(rc._megakernel, kernel_first)
```

kevinzakka:
Thanks for the review @thowell! I applied everything except the culling default. A few notes:

- The suggested tie-break test actually passes on the old code (the BVH happens to visit the pair in the harmless order on CPU), so the committed test adds extra geoms to shuffle traversal order. It fails on GPU without the fix.
- `query_dist` is recomputed per candidate instead of hoisted, otherwise the prune bound never tightens as hits land.
- In the culling default: native MuJoCo actually culls by default so I kept `True`. The real bug was that `ray_mesh_with_bvh` dropped the whole mesh at the first backface hit, whereas GL culls the triangle and draws whatever is behind it. It now marches past culled hits, which fixes the bottle cap at default settings.

mujoco_warp/_src/ray.py
```diff
@@ -714,17 +714,19 @@ def ray_mesh_with_bvh(
...
-  # local-space normal into world space and return the hit.
-  if (not cull_backfaces) or wp.dot(lvec, n) < 0.0:
-    normal = mat @ n
-    normal = wp.normalize(normal)
-    return t, normal, u, v, f, mesh_geom_id
+  # Culling drops back-facing triangles, not the mesh: like a rasterizer, march
+  # past culled hits to the next surface (bounded to keep the kernel in check).
+  offset = float(0.0)
+  for _ in range(8):
+    hit = wp.mesh_query_ray(mesh_bvh_id[mesh_geom_id], lpnt + lvec * offset, lvec, max_t - offset, t, u, v, sign, n, f)
```
thowell: can we read `mesh_bvh_id[mesh_geom_id]` once before the loop?

mujoco_warp/_src/ray.py
```diff
@@ -714,17 +714,19 @@ def ray_mesh_with_bvh(
...
+    if not hit:
+      return -1.0, wp.vec3(0.0, 0.0, 0.0), 0.0, 0.0, -1, -1
+    if (not cull_backfaces) or wp.dot(lvec, n) < 0.0:
+      normal = mat @ n
+      normal = wp.normalize(normal)
+      return offset + t, normal, u, v, f, mesh_geom_id
+    offset += t * (1.0 + 1.0e-6) + 1.0e-9
```
thowell: we have these tolerances `1.0e-6` and `1.0e-9` in ray.py and render.py. can we make these constants in one of the files and then use the constants throughout?

mujoco_warp/_src/render.py
```diff
@@ -348,10 +356,13 @@ def cast_ray(
     bounds_nr = int(0)
     ngeom = bvh_ngeom + flex_bvh_ngeom
 
-    while wp.bvh_query_next(query, bounds_nr, dist):
+    # max_t is exclusive: widen so coincident hits reach the tie-break below.
+    while wp.bvh_query_next(query, bounds_nr, dist * (1.0 + 1.0e-6) + 1.0e-9):
```
thowell: see comment about constants

mujoco_warp/_src/render.py
```diff
@@ -348,10 +356,13 @@ def cast_ray(
...
       gi_global = bounds_nr
       local_id = gi_global - (worldid * ngeom)
 
+      query_dist = dist * (1.0 + 1.0e-6) + 1.0e-9
```
thowell: see comment about constants

mujoco_warp/_src/render.py
```diff
@@ -841,6 +849,31 @@ def _render_megakernel(
...
+      and mesh_id >= 0
+      and f >= 0
+      and geom_type[geom_id] == int(GeomType.MESH.value)
+    ):
+      mat = geom_xmat_in[worldid, geom_id]
+      face = wp.transpose(mat) @ normal
+      tri = mesh_facenormal[mesh_faceadr[mesh_id] + f]
+      adr = mesh_normaladr[mesh_id]
+      normal = wp.normalize(
+        mat
```
thowell: nit: can we modify the formatting here? maybe sum the vectors so that we have something like:
```
vec = ... + ... + ....
normal = wp.normalize(mat @ vec)
```

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
     # Verify that the two worlds rendered different skybox backgrounds
     self.assertFalse(np.array_equal(rgb_w0, rgb_w1))
 
+  def test_mesh_bounds_contain_offcentre_mesh(self):
+    # Recentred on its centroid the pyramid spans -h/4 to 3h/4, but a half-extent
+    # of 0.5 * (pmax - pmin) reaches only h/2, cutting the apex off. The camera
+    # sits on that cut looking along the horizon, so rays above the centre row
+    # only travel upwards: they reach the apex only if it is inside the bounds.
+    b, h = 0.1, 0.4
+    xml = f"""
```
thowell: lets inline the xml in test_data.fixture and also instead of an f-string lets just set the constants in the xml directly

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+  @parameterized.parameters("box", "mesh")
+  def test_coincident_geoms_resolve_to_higher_index(self, geom_type: str):
+    # Two geoms in the same place: MuJoCo's GL depth test (GL_LEQUAL) leaves the
+    # last-drawn geom on top, so the higher index must win every pixel, not
+    # whichever the BVH reaches first.
+    s = 0.2
+    vert = " ".join(f"{x} {y} {z}" for x in (-s, s) for y in (-s, s) for z in (-s, s))
+    asset = f'<asset><mesh name="m" vertex="{vert}"/></asset>' if geom_type == "mesh" else ""
+    geom = 'type="mesh" mesh="m"' if geom_type == "mesh" else f'type="box" size="{s} {s} {s}"'
+    xml = f"""
```
thowell: lets inline the xml in test_data.fixture

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+  @parameterized.parameters(False, True)
+  def test_coincident_geoms_with_roundoff_resolve_to_higher_index(self, extra_geoms: bool):
+    # The lower-index box sits nearer by less than the tie tolerance: roundoff
+    # must not let it win. The extra geoms reshuffle the BVH traversal order so
+    # both visit orders of the pair are exercised.
+    ring = "".join(
+      f'<geom type="sphere" size="0.05" pos="{0.7 * np.cos(t):.3f} {0.5 * np.sin(t):.3f} 0" rgba="0.5 0.5 0.9 1"/>'
+      for t in np.linspace(0.0, 2.0 * np.pi, 9, endpoint=False)
+    )
+    xml = f"""
```
thowell: lets inline the xml in test_data.fixture

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+    seg = rc.seg_data.numpy()[0].reshape(-1, 2)[..., 0]
+    nbox = mjm.ngeom - 2, mjm.ngeom - 1
+
+    self.assertGreater(np.count_nonzero(seg == nbox[1]), 0)
+    self.assertEqual(np.count_nonzero(seg == nbox[0]), 0)
+
+  def test_backface_cull_skips_culled_triangles(self):
+    # An inside-out tetra: like GL, culling must drop its near faces yet still
+    # draw the far face behind them, not drop the whole mesh.
+    xml = """
```
thowell: lets inline the xml in test_data.fixture

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+    rc = mjw.create_render_context(mjm, cam_res=(32, 32), render_seg=True, enable_backface_culling=True)
+    mjw.render(m, d, rc)
+    seg = rc.seg_data.numpy()[0].reshape(-1, 2)[..., 0]
+
+    self.assertGreater(np.count_nonzero(seg == 0), 0)
+
+  def test_backfaces_are_shaded_two_sided(self):
+    # Camera inside the box: every visible face points away from it. Shading
+    # those hits with an unflipped normal leaves the interior at zero diffuse.
+    xml = """
```
thowell: lets inline the xml in test_data.fixture

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+    self.assertGreater(int(rgb.min()), 0)
+
+  def test_unusable_vertex_normals_fall_back_to_the_face(self):
+    # A bare-vertex hull has no authored normals: at a cube corner MuJoCo's face
+    # contributions cancel and it stores the [0, 0, 1] sentinel. Rejecting those
+    # against the face normal must render the hull exactly like the primitive.
+    s = 0.3
+    vert = " ".join(f"{x} {y} {z}" for x in (-s, s) for y in (-s, s) for z in (-s, s))
+
+    def render(geom):
```
thowell: dont think we need this function 'render', lets just run the code directly in the test scope and provide the output from `_unpack_rgb` to the test

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+
+    self.assertGreater(np.count_nonzero(seg[:32] >= 0), 100)
+
+  @parameterized.parameters("box", "mesh")
+  def test_coincident_geoms_resolve_to_higher_index(self, geom_type: str):
+    # Two geoms in the same place: MuJoCo's GL depth test (GL_LEQUAL) leaves the
+    # last-drawn geom on top, so the higher index must win every pixel, not
+    # whichever the BVH reaches first.
+    s = 0.2
+    vert = " ".join(f"{x} {y} {z}" for x in (-s, s) for y in (-s, s) for z in (-s, s))
```
thowell: lets add both assets to the xml and provide the values directly -- think this will make the test a bit easier to read and understand the scene setup

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+    rgb = _unpack_rgb(rc.rgb_data.numpy()[0])
+
+    self.assertGreater(int(rgb.min()), 0)
+
+  def test_unusable_vertex_normals_fall_back_to_the_face(self):
+    # A bare-vertex hull has no authored normals: at a cube corner MuJoCo's face
+    # contributions cancel and it stores the [0, 0, 1] sentinel. Rejecting those
+    # against the face normal must render the hull exactly like the primitive.
+    s = 0.3
+    vert = " ".join(f"{x} {y} {z}" for x in (-s, s) for y in (-s, s) for z in (-s, s))
```
thowell: lets directly add the actual values to the xml to make the scene a bit easier to ready and understand

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+
+    def render(geom):
+      xml = f"""
```
thowell: lets inline the xml in test_data.fixture

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+
+  def test_authored_normals_shade_smoothly(self):
+    # On a sphere hull neighbouring faces stay inside the tolerance, so MuJoCo
+    # keeps real vertex normals: interpolating them must leave no flat facet.
+    n, r = 64, 0.3
+    i = np.arange(n) + 0.5
+    phi = np.arccos(1.0 - 2.0 * i / n)
+    theta = np.pi * (1.0 + 5.0**0.5) * i
+    pts = r * np.stack([np.cos(theta) * np.sin(phi), np.sin(theta) * np.sin(phi), np.cos(phi)], axis=-1)
+    xml = f"""
```
thowell: lets inline the xml in test_data.fixture

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+    rc = mjw.create_render_context(mjm, cam_res=(64, 64), render_rgb=True, render_seg=True, use_ambient_lighting=False)
+    mjw.render(m, d, rc)
+    red = _unpack_rgb(rc.rgb_data.numpy()[0])[..., 0]
+    seg = rc.seg_data.numpy()[0].reshape(-1, 2)[..., 0]
+    value, count = np.unique(red[seg == 0], return_counts=True)
+
+    self.assertGreater(len(value), 120)
+    self.assertLess(int(count.max()), 40)
+
+  def test_vertex_normals_can_be_disabled(self):
```
thowell: can this test and the one above be combined via some sort of parameterization?

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+    self.assertIsNone(rc._megakernel)
+    mjw.render(m, d, rc)
+    kernel = rc._megakernel
+    self.assertIsNotNone(kernel)
+    mjw.render(m, d, rc)
+    self.assertIs(rc._megakernel, kernel)
+
+  def test_shadow_light_fraction_scales_shadowed_light(self):
+    # How much of a light survives its own shadow: 0 is black, 1 is no shadow.
+    xml = """
```
thowell: lets inline this xml in test_data.fixture

mujoco_warp/_src/render_test.py
```diff
@@ -687,6 +687,273 @@ def test_per_world_skybox_textures(self):
...
+      <worldbody>
+        <light pos="0 0 3" dir="0 0 -1" directional="true" diffuse="1 1 1" castshadow="true"/>
+        <camera pos="0 -1.2 1.2" xyaxes="1 0 0 0 0.7 0.7" fovy="60"/>
+        <geom type="plane" size="2 2 0.1" rgba="0.8 0.8 0.8 1"/>
+        <geom type="box" pos="0 0 0.5" size="0.25 0.25 0.02" rgba="0.2 0.2 0.9 1"/>
+      </worldbody>
+    </mujoco>
+    """
+    shadowed = []
+    for fraction in (0.0, 0.3, 0.6, 1.0):
```
thowell: lets parameterize this test by `fraction`

thowell:
@kevinzakka left some comments that are mostly about code style. otherwise lgtm. thanks!

kevinzakka:
@thowell Applied everything as suggested.

thowell (approved):

PR #1604, opened 2026-08-24
Add ray offset for orthographic cameras

sibocw:
Currently, `compute_ray` (called upon rendering context initialization) only returns the ray direction. For orthographic cameras, this is a constant (0, 0, -1), which is fine.

However, for orthographic cameras, we also need the ray's offset from the camera center. If I understand it correctly, I think this is currently missing. The result is that all pixels are looking at the same location.

To verify this, I set up this test with two boxes and a small green sphere at the center. As expected, all pixels rendered from the orthographic camera seem to be looking at the same green center pixel:

```python
import mujoco
import numpy as np
from PIL import Image, ImageFont, ImageDraw

import mujoco_warp as mjw

XML = """
<mujoco>
  <worldbody>
    <camera name="perspective_camera" pos="0 -10 0" xyaxes="1 0 0 0 0 1"/>
    <camera name="orthographic_camera" pos="0 -10 0" xyaxes="1 0 0 0 0 1" projection="orthographic" fovy="10"/>
    <geom name="left_box" type="box" size="1 2 1" pos="-2 0 0" rgba="1 0 0 1"/>
    <geom name="right_box" type="box" size="1 2 1" pos="2 0 0" rgba="0 0 1 1"/>
    <geom name="center_sphere" type="sphere" size="0.1" pos="0 0 0" rgba="0 1 0 1"/>
  </worldbody>
</mujoco>
"""


def render_camera(mjm, mjd, cam_name, width=256, height=256):
  m = mjw.put_model(mjm)
  d = mjw.put_data(mjm, mjd)
  rc = mjw.create_render_context(mjm, cam_res=(width, height), render_rgb=True, cam_active=[cam_name])
  mjw.render(m, d, rc)

  packed = rc.rgb_data.numpy()[0]
  r = (packed >> 16) & 0xFF
  g = (packed >> 8) & 0xFF
  b = packed & 0xFF
  return np.stack([r, g, b], axis=-1).reshape(height, width, 3).astype(np.uint8)


if __name__ == "__main__":
  mjm = mujoco.MjModel.from_xml_string(XML)
  mjd = mujoco.MjData(mjm)
  mujoco.mj_forward(mjm, mjd)

  img_persp = render_camera(mjm, mjd, "perspective_camera")
  img_ortho = render_camera(mjm, mjd, "orthographic_camera")

  FONT = ImageFont.load_default(size=12)
  FILL = (255, 255, 255)
  OFFSET = (10, 10)

  panels = []
  for img, label in [(img_persp, "Perspective"), (img_ortho, "Orthographic")]:
    img_pil = Image.fromarray(img)
    draw = ImageDraw.Draw(img_pil)
    draw.text(OFFSET, label, font=FONT, fill=FILL)
    panels.append(np.array(img_pil))

  gap = np.zeros((panels[0].shape[0], 10, 3), dtype=np.uint8)
  out = np.hstack([panels[0], gap, panels[1]])
  Image.fromarray(out).save("demo_ortho_bug.png")
```

demo_ortho_bug_before

I propose adding a `ray_offset` attribute to the rendering context and modifying the `render` function accordingly. The output of the same script is now like this:

demo_ortho_bug_after

### DIFF
--- a/mujoco_warp/_src/render_test.py
+++ b/mujoco_warp/_src/render_test.py
@@ -409,8 +409,7 @@ class RenderTest(parameterized.TestCase):
   @parameterized.named_parameters(("precomputed_rays", True), ("dynamic_rays", False))
   def test_segmentation_orthographic_matches_mujoco(self, use_precomputed_rays: bool):
     """Orthographic segmentation should match native MuJoCo, including vertical orientation."""
-    mjm, mjd, _m, d = test_data.fixture(xml=self._ORTHOGRAPHIC_SCENE)
-    m = mjw.put_model(mjm)
+    mjm, mjd, m, d = test_data.fixture(xml=self._ORTHOGRAPHIC_SCENE)
     cam_w, cam_h = 64, 64
 
     rc = mjw.create_render_context(
--- a/mujoco_warp/_src/render_test.py
+++ b/mujoco_warp/_src/render_test.py
@@ -377,6 +377,56 @@ class RenderTest(parameterized.TestCase):
 
     np.testing.assert_array_equal(warp_seg_np, mj_seg)
 
+  def test_render_segmentation_orthographic(self):
+    """Orthographic camera's rays must shift across pixels, depending on its offset."""
+    xml = """
+      <mujoco>
+        <worldbody>
+          <camera name="cam" pos="0 -10 0" xyaxes="1 0 0 0 0 1" projection="orthographic" fovy="10"/>
+          <geom name="left_box" type="box" size="1 2 1" pos="-2 0 0" rgba="1 0 0 1"/>
+          <geom name="right_box" type="box" size="1 2 1" pos="2 0 0" rgba="0 0 1 1"/>
+        </worldbody>
+      </mujoco>
+    """
+    mjm, mjd, m, d = test_data.fixture(xml=xml)
+
+    rc = mjw.create_render_context(mjm, nworld=1, cam_res=(64, 64), render_rgb=False, render_depth=False, render_seg=True)
+    mjw.render(m, d, rc)
+
+    seg = rc.seg_data.numpy().reshape(1, 64, 64, 2)[0]
+    left_ids = set(np.unique(seg[:, :32, 0])) - {-1}
+    right_ids = set(np.unique(seg[:, 32:, 0])) - {-1}
+    self.assertTrue(left_ids, "left half of the frame should hit the left sphere")
+    self.assertTrue(right_ids, "right half of the frame should hit the right sphere")
+    self.assertFalse(left_ids & right_ids, "the two spheres are distinct geoms")
+
+  @absltest.skipIf(not _HAS_RENDERER, "MuJoCo rendering requires OpenGL")
+  def test_segmentation_orthographic_matches_mujoco(self):
+    """Orthographic segmentation should match native MuJoCo."""
+    xml = """
+      <mujoco>
+        <worldbody>
+          <camera name="cam" pos="0 -10 0" xyaxes="1 0 0 0 0 1" projection="orthographic" fovy="10"/>
+          <geom name="left_box" type="box" size="1 2 1" pos="-2 0 0" rgba="1 0 0 1"/>
+          <geom name="right_box" type="box" size="1 2 1" pos="2 0 0" rgba="0 0 1 1"/>
+        </worldbody>
+      </mujoco>
+    """
+    mjm, mjd, _m, d = test_data.fixture(xml=xml)
+    m = mjw.put_model(mjm)
+    cam_w, cam_h = 64, 64
+
+    rc = mjw.create_render_context(mjm, nworld=1, cam_res=(cam_w, cam_h), render_seg=[True])
+    mjw.render(m, d, rc)
+    warp_seg_np = rc.seg_data.numpy()[0].reshape(-1, 2)
+
+    with mujoco.Renderer(mjm, height=cam_h, width=cam_w) as renderer:
+      renderer.update_scene(mjd, camera="cam")
+      renderer.enable_segmentation_rendering()
+      mj_seg = renderer.render().reshape(-1, 2)
+
+    np.testing.assert_array_equal(warp_seg_np, mj_seg)
+
   @absltest.skipIf(not _HAS_RENDERER, "MuJoCo rendering requires OpenGL")
   def test_depth_matches_mujoco(self):
     """Depth values should match native MuJoCo (planar depth, not Euclidean)."""
--- a/mujoco_warp/_src/render_test.py
+++ b/mujoco_warp/_src/render_test.py
@@ -377,46 +377,49 @@ class RenderTest(parameterized.TestCase):
 
     np.testing.assert_array_equal(warp_seg_np, mj_seg)
 
+  # The two boxes sit in diagonally opposite quadrants (bottom-left and top-right, in
+  # image space) so that a flip along either the horizontal or the vertical ray axis
+  # is caught by a single scene, unlike a left/right or top/bottom split alone, each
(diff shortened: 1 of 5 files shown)


thowell:
@sibocw thanks for the contribution to mujoco warp!

here are a few items we can address:

### 1. Fix Y-Axis Inversion (Vertical Flip)
In `compute_ray` (`mujoco_warp/_src/render_util.py`), `y` is currently mapped starting from `bottom`:
```python
bottom = -0.5 * sensor_h
y = bottom + sensor_h * v
```
For `py = 0` (top of the image, `v ≈ 0`), this sets `y = -0.5 * sensor_h` (bottom of camera view), which flips the rendered image upside-down.
**Fix:** Match the perspective camera convention and native MuJoCo's `mju_camPixelRay`:
```python
top = 0.5 * sensor_h
bottom = -0.5 * sensor_h
y = top + (bottom - top) * v  # or: y = (0.5 - v) * sensor_h
```
---
### 2. FOV Handling with `sensorsize`
In native MuJoCo (`engine_vis_visualize.c` and `engine_support.c`), orthographic cameras ignore `sensorsize` and define vertical extent strictly using `fovy` (in length units). If an XML defines both `sensorsize` and `fovy`, checking `sensorsize` causes a mismatch with MuJoCo.
**Fix:** Simplify the orthographic branch to always use `fovy`:
```python
aspect = float(img_w) / float(img_h)
sensor_h = fovy
sensor_w = sensor_h * aspect
```
---
### 3. Optimize Perspective-Only Scenes in `_render_megakernel`
In `_render_megakernel` (`mujoco_warp/_src/render.py`), every ray currently performs a global memory load from `ray_offset`, a 3x3 matrix multiplication, and a vector addition even for perspective-only scenes (where `ray_offset` is zero).
**Suggestion:** Statically guard the offset calculation using `rc_static`:
```python
# In create_render_context:
rc_static["has_orthographic_camera"] = any(
    cam_projection[cid] == ProjectionType.ORTHOGRAPHIC for cid in cam_id_map
)
# In _render_megakernel:
if wp.static(rc_static["has_orthographic_camera"]):
  ray_origin_world = cam_xpos_in[worldid, mujoco_cam_id] + cam_mat_world @ ray_offset_local_cam
else:
  ray_origin_world = cam_xpos_in[worldid, mujoco_cam_id]
```
---
### 4. Test Improvements
- **Asymmetric Test Scene:** Update `test_segmentation_orthographic_matches_mujoco` to place objects vertically separated along +/-Z (e.g. `top_box` at `0 0 2` and `bottom_box` at `0 0 -2`) so vertical orientation is verified against `mujoco.Renderer`.
- **Add Depth and RGB Parity Tests:** Add tests verifying orthographic depth (`test_depth_orthographic_matches_mujoco`) and RGB matching native MuJoCo.
- **Dynamic Rays (`use_precomputed_rays=False`):** Add a test case verifying orthographic cameras when precomputed rays are disabled.
- **Typo in test comments:** In `test_render_segmentation_orthographic`, update the assertion messages referencing "spheres" to "boxes".
---
### 5. Typo
- `render_util.py:171`: `# always poiting forward` -> `# always pointing forward`.

sibocw:
@thowell  Sounds good, about point 2: the [XML reference for `body/camera`](https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-camera) reads

> sensorsize: real(2), “0 0”
> Size of the camera sensor in length units. When specified, all intrinsic attributes become active and fovy is ignored. The field-of-view is then computed automatically from the focal length and sensor size.

Do you want to follow this or native MuJoCo's implementation?

sibocw:
@thowell I think these issues are addressed now. I felt that a pixel-level parity test with the native MuJoCo could be too fragile, so I added tests against expected analytical results instead. What do you think?

thowell:
@sibocw thanks for the fast update!

- comment 1: lets follow the mujoco implementation
- comment 2: lets go with analytical comparisons instead of exact pixel matching as suggested

sibocw:
Great, then the code is ready as is for review. @thowell

mujoco_warp/_src/render_test.py
```diff
@@ -377,6 +377,100 @@ def test_segmentation_matches_mujoco(self):
...
+    self.assertTrue(bottom_left_ids, "bottom-left quadrant should hit the bottom-left box")
+    self.assertTrue(top_right_ids, "top-right quadrant should hit the top-right box")
+    self.assertFalse(bottom_left_ids & top_right_ids, "the two boxes are distinct geoms")
+
+  @absltest.skipIf(not _HAS_RENDERER, "MuJoCo rendering requires OpenGL")
+  @parameterized.named_parameters(("precomputed_rays", True), ("dynamic_rays", False))
+  def test_segmentation_orthographic_matches_mujoco(self, use_precomputed_rays: bool):
+    """Orthographic segmentation should match native MuJoCo, including vertical orientation."""
+    mjm, mjd, _m, d = test_data.fixture(xml=self._ORTHOGRAPHIC_SCENE)
+    m = mjw.put_model(mjm)
```
thowell: we can remove this and use the `Model` object returned above and lets rename the above `_m` to `m`?

mujoco_warp/_src/render_util.py
```diff
@@ -79,54 +79,79 @@ def compute_ray(
...
-    half_width = half_height * aspect
-    left = -half_width
-    right = half_width
-    top = half_height
-    bottom = -half_height
+    # Compute ray direction
+    direction = wp.vec3(0.0, 0.0, -1.0)  # always pointing forward
+
+    # Compute ray offset from center
+    aspect = float(img_w) / float(img_h)
```
thowell: lets compute `1.0 / float(img_h)` once as `inv_img_h` and use here and below

mujoco_warp/_src/render_util.py
```diff
@@ -79,54 +79,79 @@ def compute_ray(
...
+    sensor_h = fovy
+    sensor_w = sensor_h * aspect
+    left = -0.5 * sensor_w
+    top = 0.5 * sensor_h
+    bottom = -0.5 * sensor_h
```
thowell: `-top`?

mujoco_warp/_src/render_util.py
```diff
@@ -79,54 +79,79 @@ def compute_ray(
...
+    u = (float(px) + 0.5) / float(img_w)
+    v = (float(py) + 0.5) / float(img_h)
+    x = left + sensor_w * u
+    y = top + (bottom - top) * v
+    offset = wp.vec3(x, y, 0.0)
+
+  else:  # projection == ProjectionType.PERSPECTIVE:
+    # Compute ray direction
+    aspect = float(img_w) / float(img_h)
```
thowell: same thing here lets precompute `1.0 / float(img_h)` and reuse

mujoco_warp/_src/render_util.py
```diff
@@ -79,54 +79,79 @@ def compute_ray(
...
+      target_aspect = float(img_w) / float(img_h)
+      sensor_aspect = sensor_w / sensor_h
+      if target_aspect > sensor_aspect:
+        sensor_h = sensor_w / target_aspect
+      elif target_aspect < sensor_aspect:
+        sensor_w = sensor_h * target_aspect
+
+      inv_fx_znear = znear / fx
+      inv_fy_znear = znear / fy
+      left = -inv_fx_znear * (sensor_w * 0.5 - cx)
```
thowell: we could compute `0.5 * sensor_w` and `0.5 * sensor_h` once each and then reuse

thowell:
left a few more comments, otherwise lgtm. great work @sibocw!

sibocw:
Thanks! How about now? @thowell

thowell (approved):

PR #1607, opened 2026-08-24
Skip tactile preprocessing when no tactile taxels exist

WT-MM:
Skip tactile-contact scratch allocations and preprocessing in `sensor_acc` when `m.nsensortaxel` is zero.

Currently, models without tactile taxels still allocate `weld_geom_count` and `weld_geom_list` and launch `_preprocess_tactile_contacts` over `d.naconmax` contact slots on each `sensor_acc` call, followed by a zero-taxel tactile launch. With this change, tactile preprocessing and evaluation run only when taxels exist. Models with taxels are unchanged, and other acceleration-sensor processing is untouched.

### DIFF
--- a/mujoco_warp/_src/sensor.py
+++ b/mujoco_warp/_src/sensor.py
@@ -2542,65 +2542,66 @@ def sensor_acc(m: Model, d: Data):
     ],
   )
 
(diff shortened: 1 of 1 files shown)


thowell:
nice fix, thanks!

thowell (approved):

PR #1612, opened 2026-08-27
update tactile sensor

thowell:
This change addresses several correctness bugs, missing features, and GPU performance bottlenecks in MuJoCo Warp's tactile sensor implementation compared to the MuJoCo C reference.

Key improvements:
- Accumulate taxel force channels in thread-local registers rather than global atomic writes, eliminating memory race conditions.
- Zero sensor output channels upon thread entry when contacts break or when no contacts are present.
- Rotate tangent friction vectors into the world frame (`sensor_mat @ tang_geom`) to align with relative spatial velocity coordinates.
- Clamp penetration depth and tangent velocities by `sensor_cutoff`.
- Support SDF plugin mesh transforms (`mesh_quat` and `mesh_pos`), safely skip non-SDF mesh geoms lacking octrees, and exclude self-geom sensing during self-collision.
- Vectorize `Model.body_has_tactile` initialization in `io.py` and skip contact preprocessing for bodies without tactile sensors.
- Add bounded contact deduplication in preprocessing and kernel scans to prevent buffer saturation by duplicate contacts.
- Guard overflow warnings statically behind `warn_overflow` and record `OverflowType.TACTILE` in `d.overflow` when contact capacity is exceeded.
- Safely ignore flex contacts (`geom < 0`) with a TODO for future flex body support.
- Keep temporary contact tracking buffers allocated inline in `sensor_acc` to avoid modifying `Data`.
- Add parameterized multi-world unit tests in `sensor_test.py` (`nworld=1` and `nworld=2`) verifying deduplication, cutoff clamping, zeroing on contact break, overflow flag propagation, and non-octree mesh handling.

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -881,6 +881,12 @@ def put_model(mjm: mujoco.MjModel, batch_sizes: dict[str, int] | None = None) ->
   # Note: body_has_tactile is indexed by weld root body ID (body_weldid)
   m.body_has_tactile = np.zeros(mjm.nbody, dtype=bool)
   m.body_has_tactile[tactile_weldid] = True
+  unique_tactile_welds = np.unique(tactile_weldid)
+  m.ntactileweld = int(len(unique_tactile_welds))
+  weld_tactile_id = np.full(mjm.nbody, -1, dtype=np.int32)
+  for idx, weld in enumerate(unique_tactile_welds):
+    weld_tactile_id[weld] = idx
+  m.weld_tactile_id = weld_tactile_id
 
   # Per-block scalar/tile/sparse layout (see m_block_layout).
   _lay = m_block_layout(mjm)
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -876,6 +876,10 @@ def put_model(mjm: mujoco.MjModel, batch_sizes: dict[str, int] | None = None) ->
     if mjm.sensor_type[i] == mujoco.mjtSensor.mjSENS_TACTILE
     for j in range(mjm.mesh_vertnum[mjm.sensor_objid[i]])
   ]
+  tactile_geomid = mjm.sensor_refid[mjm.sensor_type == mujoco.mjtSensor.mjSENS_TACTILE]
+  tactile_weldid = mjm.body_weldid[mjm.geom_bodyid[tactile_geomid]]
+  m.body_has_tactile = np.zeros(mjm.nbody, dtype=bool)
+  m.body_has_tactile[tactile_weldid] = True
 
   # Per-block scalar/tile/sparse layout (see m_block_layout).
   _lay = m_block_layout(mjm)
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -878,6 +878,7 @@ def put_model(mjm: mujoco.MjModel, batch_sizes: dict[str, int] | None = None) ->
   ]
   tactile_geomid = mjm.sensor_refid[mjm.sensor_type == mujoco.mjtSensor.mjSENS_TACTILE]
   tactile_weldid = mjm.body_weldid[mjm.geom_bodyid[tactile_geomid]]
+  # Note: body_has_tactile is indexed by weld root body ID (body_weldid)
   m.body_has_tactile = np.zeros(mjm.nbody, dtype=bool)
   m.body_has_tactile[tactile_weldid] = True
 
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -878,9 +878,6 @@ def put_model(mjm: mujoco.MjModel, batch_sizes: dict[str, int] | None = None) ->
   ]
   tactile_geomid = mjm.sensor_refid[mjm.sensor_type == mujoco.mjtSensor.mjSENS_TACTILE]
   tactile_weldid = mjm.body_weldid[mjm.geom_bodyid[tactile_geomid]]
-  # Note: body_has_tactile is indexed by weld root body ID (body_weldid)
-  m.body_has_tactile = np.zeros(mjm.nbody, dtype=bool)
-  m.body_has_tactile[tactile_weldid] = True
   unique_tactile_welds = np.unique(tactile_weldid)
   m.ntactileweld = int(len(unique_tactile_welds))
   weld_tactile_id = np.full(mjm.nbody, -1, dtype=np.int32)
--- a/mujoco_warp/_src/sensor.py
+++ b/mujoco_warp/_src/sensor.py
@@ -2151,7 +2151,7 @@ def _preprocess_tactile_contacts_kernel(warn_overflow: int):
     # Model:
     body_weldid: wp.array[int],
     geom_bodyid: wp.array[int],
-    body_has_tactile: wp.array[bool],
+    weld_tactile_id: wp.array[int],
     # Data in:
     contact_geom_in: wp.array[wp.vec2i],
     contact_worldid_in: wp.array[int],
@@ -2186,18 +2186,19 @@ def _preprocess_tactile_contacts_kernel(warn_overflow: int):
       geom = geom2 if side == 0 else geom1
 
       # Skip bodies without tactile sensors
-      if not body_has_tactile[weld]:
+      tactile_idx = weld_tactile_id[weld]
+      if tactile_idx < 0:
         continue
 
       # Exact atomic bitset test-and-set guarantees one slot reservation per distinct geom.
       # Duplicate contacts for the same geom cannot consume buffer slots or displace distinct geoms.
       word_idx = geom // 32
       bit_mask = wp.uint32(1) << wp.uint32(geom % 32)
-      old_seen = wp.atomic_or(weld_geom_seen_out[worldid, weld], word_idx, bit_mask)
+      old_seen = wp.atomic_or(weld_geom_seen_out[worldid, tactile_idx], word_idx, bit_mask)
       if (old_seen & bit_mask) == wp.uint32(0):
-        idx = wp.atomic_add(weld_geom_count_out[worldid], weld, 1)
+        idx = wp.atomic_add(weld_geom_count_out[worldid], tactile_idx, 1)
         if idx < MJ_MAXCONPAIR:
-          weld_geom_list_out[worldid, weld, idx] = geom
+          weld_geom_list_out[worldid, tactile_idx, idx] = geom
         else:
           if wp.static(bool(warn_overflow & OverflowType.TACTILE)):
             if idx == MJ_MAXCONPAIR:
@@ -2240,6 +2241,7 @@ def _sensor_tactile(
   plugin: wp.array[int],
   plugin_attr: wp.array[vec_pluginattr],
   geom_plugin_index: wp.array[int],
+  weld_tactile_id: wp.array[int],
   taxel_vertadr: wp.array[int],
   taxel_sensorid: wp.array[int],
   # Data in:
@@ -2260,6 +2262,7 @@ def _sensor_tactile(
   geom_id = sensor_refid[sensor_id]
   parent_body = geom_bodyid[geom_id]
   parent_weld = body_weldid[parent_body]
+  tactile_idx = weld_tactile_id[parent_weld]
 
   ncon = mesh_vertnum[mesh_id]
   nchannel = sensor_dim[sensor_id] // ncon
@@ -2269,7 +2272,10 @@ def _sensor_tactile(
   for c in range(nchannel):
     sensordata_out[worldid, sensor_adr[sensor_id] + c * ncon + vertid] = 0.0
 
-  geom_count = weld_geom_count_in[worldid, parent_weld]
+  if tactile_idx < 0:
+    return
+
+  geom_count = weld_geom_count_in[worldid, tactile_idx]
   if geom_count == 0:
     return
 
@@ -2296,21 +2302,11 @@ def _sensor_tactile(
 
   max_geoms = wp.min(geom_count, MJ_MAXCONPAIR)
   for g in range(max_geoms):
-    geom = weld_geom_list_in[worldid, parent_weld, g]
+    geom = weld_geom_list_in[worldid, tactile_idx, g]
     # Unlike MuJoCo C which excludes self-geom during preprocessing, weld_geom_list is shared
     # across all sensors on the same weld body. Skipping geom == geom_id here ensures each
     # sensor on the weld body filters its own geom without affecting others.
     if geom < 0 or geom == geom_id:
-      continue
-
-    # Load-bearing deduplication check: ensures geoms are processed at most once even if
-    # concurrent preprocessing inserted duplicates into weld_geom_list.
-    is_dup = int(0)
-    for j in range(g):
-      if weld_geom_list_in[worldid, parent_weld, j] == geom:
-        is_dup = int(1)
-        break
-    if is_dup == int(1):
       continue
 
     contact_type = geom_type[geom]
@@ -2615,17 +2611,17 @@ def sensor_acc(m: Model, d: Data):
   )
 
   if m.nsensortaxel:
-    weld_geom_count = wp.zeros((d.nworld, m.nbody), dtype=int)
-    weld_geom_list = wp.full((d.nworld, m.nbody, MJ_MAXCONPAIR), -1, dtype=int)
+    weld_geom_count = wp.zeros((d.nworld, m.ntactileweld), dtype=int)
+    weld_geom_list = wp.full((d.nworld, m.ntactileweld, MJ_MAXCONPAIR), -1, dtype=int)
     nwords = (m.ngeom + 31) // 32
-    weld_geom_seen = wp.zeros((d.nworld, m.nbody, nwords), dtype=wp.uint32)
+    weld_geom_seen = wp.zeros((d.nworld, m.ntactileweld, nwords), dtype=wp.uint32)
     wp.launch(
       _preprocess_tactile_contacts_kernel(int(m.opt.warn_overflow)),
       dim=d.naconmax,
       inputs=[
         m.body_weldid,
         m.geom_bodyid,
-        m.body_has_tactile,
+        m.weld_tactile_id,
         d.contact.geom,
         d.contact.worldid,
         d.nacon,
@@ -2668,6 +2664,7 @@ def sensor_acc(m: Model, d: Data):
         m.plugin,
         m.plugin_attr,
         m.geom_plugin_index,
+        m.weld_tactile_id,
         m.taxel_vertadr,
         m.taxel_sensorid,
         d.geom_xpos,
--- a/mujoco_warp/_src/sensor.py
+++ b/mujoco_warp/_src/sensor.py
@@ -33,6 +33,7 @@ from mujoco_warp._src.types import ContactType
 from mujoco_warp._src.types import Data
 from mujoco_warp._src.types import DataType
 from mujoco_warp._src.types import DisableBit
+from mujoco_warp._src.types import GeomType
 from mujoco_warp._src.types import JointType
 from mujoco_warp._src.types import Model
 from mujoco_warp._src.types import ObjType
@@ -2143,41 +2144,72 @@ def _transform_spatial(vec: wp.spatial_vector, dif: wp.vec3) -> wp.vec3:
   return wp.spatial_bottom(vec) - wp.cross(dif, wp.spatial_top(vec))
 
 
-@wp.kernel(grid_stride=True)
-def _preprocess_tactile_contacts(
-  # Model:
-  body_weldid: wp.array[int],
-  geom_bodyid: wp.array[int],
-  # Data in:
-  contact_geom_in: wp.array[wp.vec2i],
-  contact_worldid_in: wp.array[int],
-  nacon_in: wp.array[int],
-  # Out:
-  weld_geom_count_out: wp.array2d[int],
-  weld_geom_list_out: wp.array3d[int],
-):
-  conid = wp.tid()
-  ncon = nacon_in[0]
-  if conid >= ncon:
-    return
-  worldid = contact_worldid_in[conid]
-  contact_geom = contact_geom_in[conid]
-  weld1 = body_weldid[geom_bodyid[contact_geom[0]]]
-  weld2 = body_weldid[geom_bodyid[contact_geom[1]]]
-  geom1 = contact_geom[0]
-  geom2 = contact_geom[1]
+@cache_kernel
+def _preprocess_tactile_contacts_kernel(warn_overflow: bool):
+  @wp.kernel(module="unique", enable_backward=False, grid_stride=True)
+  def _preprocess_tactile_contacts(
+    # Model:
+    body_weldid: wp.array[int],
+    geom_bodyid: wp.array[int],
+    body_has_tactile: wp.array[bool],
+    # Data in:
+    contact_geom_in: wp.array[wp.vec2i],
+    contact_worldid_in: wp.array[int],
+    nacon_in: wp.array[int],
+    # Data out:
+    overflow_out: wp.array[int],
+    # Out:
+    weld_geom_count_out: wp.array2d[int],
+    weld_geom_list_out: wp.array3d[int],
+  ):
+    conid = wp.tid()
+    ncon = nacon_in[0]
+    if conid >= ncon:
+      return
 
-  for side in range(2):
-    if side == 0:
-      weld = weld1
-      geom = geom2
-    else:
-      weld = weld2
-      geom = geom1
+    contact_geom = contact_geom_in[conid]
+    geom1 = contact_geom[0]
+    geom2 = contact_geom[1]
 
-    idx = wp.atomic_add(weld_geom_count_out[worldid], weld, 1)
-    if idx < MJ_MAXCONPAIR:
-      weld_geom_list_out[worldid, weld, idx] = geom
+    # TODO(team): support flex body contacts (MuJoCo C resolves via mj_flexBody).
+    # Flex body contacts have negative geom indices; skip for now.
+    if geom1 < 0 or geom2 < 0:
+      return
+
+    worldid = contact_worldid_in[conid]
+    weld1 = body_weldid[geom_bodyid[geom1]]
+    weld2 = body_weldid[geom_bodyid[geom2]]
+
+    for side in range(2):
+      weld = weld1 if side == 0 else weld2
+      geom = geom2 if side == 0 else geom1
+
+      # Skip bodies without tactile sensors
+      if not body_has_tactile[weld]:
+        continue
+
+      cur_count = weld_geom_count_out[worldid, weld]
+      already_present = int(0)
+      limit = wp.min(cur_count, MJ_MAXCONPAIR)
+      for k in range(limit):
+        if weld_geom_list_out[worldid, weld, k] == geom:
+          already_present = int(1)
+          break
+
+      if already_present == int(0):
+        idx = wp.atomic_add(weld_geom_count_out[worldid], weld, 1)
+        if idx < MJ_MAXCONPAIR:
+          weld_geom_list_out[worldid, weld, idx] = geom
+        else:
+          if wp.static(warn_overflow):
+            if idx == MJ_MAXCONPAIR:
+              wp.printf(
+                "tactile sensor collision pair overflow: number of contacting geoms >= %u\n",
+                MJ_MAXCONPAIR,
+              )
+          wp.atomic_or(overflow_out, worldid, int(OverflowType.TACTILE))
+
+  return _preprocess_tactile_contacts
 
 
 @wp.kernel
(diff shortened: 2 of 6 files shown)


mujoco_warp/_src/sensor.py
```diff
@@ -2228,35 +2262,49 @@ def _sensor_tactile(
...
   tang2 = wp.vec3(0.0, 0.0, 0.0)
   if has_frame:
-    tang1 = math.rot_vec_quat(mesh_normal[offset + 1], quat)
-    tang2 = math.rot_vec_quat(mesh_normal[offset + 2], quat)
+    offset = mesh_normaladr[mesh_id] + 3 * vertid
+    quat = mesh_quat[mesh_id]
+    tang1_geom = math.rot_vec_quat(mesh_normal[offset + 1], quat)
+    tang2_geom = math.rot_vec_quat(mesh_normal[offset + 2], quat)
+    tang1 = sensor_mat @ tang1_geom
+    tang2 = sensor_mat @ tang2_geom
```
smallquail: This applies `mesh_quat` on top of the rotation into the world frame. `engine_sensor.c` no longer applies `mesh_quat` to the tangents at all: google-deepmind/mujoco@94cc2b147 deleted those `mju_rotVecQuat` calls and rotates the velocity into the geom frame instead.

```c
mju_mulMatTVec3(vel_rel_geom, geom_mat, vel_rel);
mjtNum tang1[3] = {mesh_normal[normal_stride*j + 3], ...};   // raw, no mesh_quat
sensordata[1*ncon + j] += mju_abs(mju_dot3(vel_rel_geom, tang1));
```

So this computes `dot(v, R_geom * Q_mesh * t)` where C computes `dot(v, R_geom * t)`. The two agree only when the sensor mesh's `mesh_quat` is identity.

Dropping the `mesh_quat` rotation is also what makes the code self-consistent with the taxel position a few lines up, which is `sensor_mat @ pos + geom_xpos` with no `mesh_quat`. `mjCMesh::Rotate` (`src/user/user_mesh.cc`) applies the same matrix to vertices and normals, and `Process` then stores that transform as `mesh_pos`/`mesh_quat`, so `mesh_normal` and `mesh_vert` are in the same frame and need the same treatment. Rotating a stored normal by `mesh_quat` takes it back toward the original mesh frame, which is the frame the SDF branch below deliberately converts into for the plugin.

With `mesh_quat` dropped, rotating `vel_rel` by `transpose(sensor_mat)` once is equivalent to rotating both tangents, and saves a matvec.

mujoco_warp/_src/sensor.py
```diff
@@ -2285,30 +2342,35 @@ def _sensor_tactile(
...
+    if -depth > max_depth:
+      max_depth = -depth
 
     if has_frame:
-      forceT[1] = wp.abs(wp.dot(vel_rel, tang1))
-      forceT[2] = wp.abs(wp.dot(vel_rel, tang2))
+      vel_other = _transform_spatial(
+        cvel_in[worldid, body],
+        geom_xpos_in[worldid, geom] - subtree_com_in[worldid, body_rootid[body]],
+      )
```
smallquail, line 2353: The reference point changed in the same C commit: the other body's spatial velocity is now evaluated at the taxel rather than at the other geom's origin.

```c
mju_transformSpatial(vel_other, d->cvel + 6*body, 0, xpos, ...);   // was d->geom_xpos + 3*geom
```

These differ whenever the contacting body has angular velocity. `vel_sensor` above is already evaluated at `xpos`, so as written the two velocities are taken at different material points. Should be `xpos - subtree_com_in[worldid, body_rootid[body]]`.

mujoco_warp/_src/sensor_test.py
```diff
@@ -927,18 +929,147 @@ def test_tactile_sensor_geom_deduplication(self):
...
     )
 
-    d.sensordata.zero_()
+    d.sensordata.fill_(wp.inf)
     mjw.sensor_acc(m, d)
 
-    warp_sensordata = d.sensordata.numpy()[0]
+    warp_sensordata = d.sensordata.numpy()
+    for w in range(nworld):
+      _assert_eq(warp_sensordata[w], mjd.sensordata, f"tactile_sensordata_world_{w}")
```
smallquail, line 941: Every fixture added in this PR uses `qpos="0 0 1 1 0 0 0"` and none passes `qvel_noise`, so `cvel` is zero and channels 1 and 2 are identically zero in both implementations. The comparison against MuJoCo therefore covers only the depth channel, which leaves the tangent rotation and the `vel_other` change untested.

`fixture(qvel_noise=...)` would cover both: it also runs `mj_step` 10 times, which gives the non-identity rotations needed to distinguish the tangent frames.

mujoco_warp/_src/sensor.py
```diff
@@ -2557,16 +2619,18 @@ def sensor_acc(m: Model, d: Data):
     weld_geom_count = wp.zeros((d.nworld, m.nbody), dtype=int)
     weld_geom_list = wp.full((d.nworld, m.nbody, MJ_MAXCONPAIR), -1, dtype=int)
     wp.launch(
-      _preprocess_tactile_contacts,
+      _preprocess_tactile_contacts_kernel(bool(m.opt.warn_overflow)),
```
smallquail: This gates on the whole bitmask rather than the `TACTILE` bit, so `m.opt.warn_overflow &= ~mjw.OverflowType.TACTILE` will not silence the print. Elsewhere the factory takes `int(m.opt.warn_overflow)` and gates per bit, e.g. `collision_convex.py:436` and `collision_flex.py:658`:

```python
if wp.static(bool(warn_overflow & OverflowType.HFIELD)):
```

The warning text is also missing the `"To disable the print warning: m.opt.warn_overflow &= ~mjw.OverflowType.TACTILE (or = 0 for all)\n"` line that the other overflow messages carry.

mujoco_warp/_src/sensor.py
```diff
@@ -2143,41 +2144,72 @@ def _transform_spatial(vec: wp.spatial_vector, dif: wp.vec3) -> wp.vec3:
...
+      weld = weld1 if side == 0 else weld2
+      geom = geom2 if side == 0 else geom1
+
+      # Skip bodies without tactile sensors
+      if not body_has_tactile[weld]:
+        continue
+
+      cur_count = weld_geom_count_out[worldid, weld]
+      already_present = int(0)
+      limit = wp.min(cur_count, MJ_MAXCONPAIR)
```
smallquail: This dedup is best-effort rather than exact: `cur_count` is a plain read, and a slot is written only after the `atomic_add` that reserves it. Two threads carrying the same geom can both miss and both append, and a scanning thread can read a reserved-but-unwritten `-1`. Duplicates can therefore still fill `weld_geom_list` and raise `OverflowType.TACTILE` spuriously.

Results stay correct only because `_sensor_tactile` keeps its own `is_dup` scan, so that scan is load-bearing and worth a comment saying so. Given that the downstream scan is what actually guarantees correctness, it is worth measuring whether the extra O(MJ_MAXCONPAIR) global-memory read per contact per side here pays for itself.

Worth noting that `test_tactile_sensor_geom_deduplication` passes on the downstream scan alone, so it does not exercise this path.

mujoco_warp/_src/sensor.py
```diff
@@ -2228,35 +2262,49 @@ def _sensor_tactile(
...
-      break
+  max_depth = float(0.0)
+  tang1_acc = float(0.0)
+  tang2_acc = float(0.0)
 
+  max_geoms = wp.min(geom_count, MJ_MAXCONPAIR)
+  for g in range(max_geoms):
     geom = weld_geom_list_in[worldid, parent_weld, g]
-    if geom < 0:
+    if geom < 0 or geom == geom_id:
```
smallquail, line 2309: C excludes the sensor's own geom during dedup (`g2 >= 0 && g2 != geom_id && !seen_geom[g2]`), so it never occupies a slot. Skipping it here instead means it consumes one of the `MJ_MAXCONPAIR` entries under self-collision.

That looks unavoidable given `weld_geom_list` is per-weld and shared across sensors on the same weld, so a comment noting the deliberate difference would help.

mujoco_warp/_src/sensor.py
```diff
@@ -2285,30 +2342,35 @@ def _sensor_tactile(
...
+      vel_other = _transform_spatial(
+        cvel_in[worldid, body],
+        geom_xpos_in[worldid, geom] - subtree_com_in[worldid, body_rootid[body]],
+      )
+      vel_rel = vel_sensor - vel_other
+      tang1_acc += wp.abs(wp.dot(vel_rel, tang1))
+      tang2_acc += wp.abs(wp.dot(vel_rel, tang2))
+
+  cutoff = sensor_cutoff[sensor_id]
+  if cutoff > 0.0:
```
smallquail, line 2360: This applies an unconditional `wp.min` rather than going through the datatype-aware `_write_scalar`/`_write_vector` path the other sensors use. Tactile is `mjDATATYPE_REAL`, so C clips to `[-cutoff, cutoff]`; same result here since the values are non-negative, but it diverges from the convention in this file.

Also, given the TODO in the test noting that XML does not expose `cutoff` on `<tactile>`, worth confirming that a cutoff for this sensor is intended before adding the code path.

mujoco_warp/_src/io.py
```diff
@@ -876,6 +876,10 @@ def _check_margin(name, t1, t2, margin):
     if mjm.sensor_type[i] == mujoco.mjtSensor.mjSENS_TACTILE
     for j in range(mjm.mesh_vertnum[mjm.sensor_objid[i]])
   ]
+  tactile_geomid = mjm.sensor_refid[mjm.sensor_type == mujoco.mjtSensor.mjSENS_TACTILE]
+  tactile_weldid = mjm.body_weldid[mjm.geom_bodyid[tactile_geomid]]
+  m.body_has_tactile = np.zeros(mjm.nbody, dtype=bool)
+  m.body_has_tactile[tactile_weldid] = True
```
smallquail: This is indexed by weld id, and the kernel reads it as `body_has_tactile[weld]`. The name and the `types.py` docstring ("body has tactile sensor") both read as body-indexed; `weld_has_tactile`, or a note in the docstring, would avoid a misuse later.

smallquail (commented):
Most of this matches the C reference and is a clear improvement: unconditional per-taxel zeroing, register accumulation instead of atomics, the non-octree mesh skip, the SDF `mesh_pos`/`mesh_quat` transform, `body_has_tactile` gating, and the `OverflowType.TACTILE` plumbing all look right.

Two things look blocking. The tangential-channel frame handling and the `vel_other` reference point no longer agree with `engine_sensor.c` as of google-deepmind/mujoco@94cc2b147, which rewrote `tactile_taxel_batch` for the same two bugs with a different resolution. Neither divergence is reachable by the tests added here, so green CI does not cover them.

The rest are smaller consistency points.

quagla (commented):
I left some comments with my other account.

thowell:
@quagla thanks for the review! all items should be addressed now.

---

Applied updates addressing all review comments:

- Updated tangent channel evaluation to rotate relative velocity into the geom frame once via `wp.transpose(sensor_mat) @ vel_rel` and dropped the `mesh_quat` rotation on `mesh_normal`, matching MuJoCo C commit google-deepmind/mujoco@94cc2b147.
- Evaluated `vel_other` spatial velocity at the taxel contact point `xpos` via `xpos - subtree_com_in[worldid, body_rootid[body]]` rather than at the colliding geom center.
- Added `test_tactile_sensor_dynamic_parity` with non-zero velocities and a wedge mesh with normal and tangent frames, validating all three channels (penetration depth, slip U, slip V) against MuJoCo C across `nworld=1` and `nworld=2`.
- Gated the overflow warning print specifically on `OverflowType.TACTILE` and added the standard disable hint (`m.opt.warn_overflow &= ~mjw.OverflowType.TACTILE`).
- Documented concurrency semantics in kernel comments, noting that preprocessing deduplication is best-effort to reduce buffer usage while the downstream `is_dup` check in `_sensor_tactile` is load-bearing for exactness.
- Added an inline comment explaining that `weld_geom_list` is shared across all sensors on the same weld body, which is why self-geom filtering occurs per-sensor in `_sensor_tactile`.
- Switched cutoff clamping to symmetric `wp.clamp(val, -cutoff, cutoff)` according to `mjDATATYPE_REAL` semantics.
- Clarified in `types.py` and `io.py` that `body_has_tactile` is indexed by weld root body ID (`body_weldid`).

sylvesterkaczmarek (commented):
There is still a correctness hole in the best-effort preprocessing dedup once the fixed `MJ_MAXCONPAIR` buffer is under pressure.

Two threads for the same geom can both miss the plain-read duplicate scan, reserve separate slots with `atomic_add`, and store duplicates. The downstream `is_dup` loop prevents double-counting those duplicates, but it cannot recover a *different* geom that was never stored because the duplicate reservations consumed the finite prefix first. For example, enough concurrent contacts with geom A can occupy the available slots; a later geom B then gets `idx >= MJ_MAXCONPAIR` and is dropped. `_sensor_tactile` will deduplicate A, but B's penetration/slip contribution is gone entirely.

That means the downstream scan guarantees no duplicate contribution, not exactness. MuJoCo C avoids this by exact `seen_geom` dedup before building the contact-geom list.

Could admission guarantee one slot per distinct geom before overflow, or otherwise ensure distinct geoms cannot be displaced by duplicate reservations? A useful regression would create duplicate pressure from geom A plus a distinct geom B whose contribution changes the expected tactile output, then assert B is still represented.

thowell:
@sylvesterkaczmarek thanks for the review!

Updated contact preprocessing admission to guarantee that duplicate contacts cannot displace distinct geoms:

- Replaced the best-effort linear scan with an exact atomic test-and-set using `wp.atomic_or` on a per-weld-body bitset (`weld_geom_seen_out` with shape `(d.nworld, m.nbody, (m.ngeom + 31) // 32)`). Because `wp.atomic_or` returns the prior mask value, exactly one thread per distinct geom receives `(old_seen & bit_mask) == 0` and proceeds to reserve a slot via `atomic_add`.
- Duplicate contacts for already-seen geoms immediately exit without calling `atomic_add` or consuming slots in `weld_geom_list_out`. This guarantees that duplicate contacts cannot exhaust the 50-slot buffer (`MJ_MAXCONPAIR`) or cause subsequent distinct geoms to be dropped prior to real overflow.
- Added regression test `test_tactile_sensor_duplicate_pressure_retention` in `sensor_test.py` parameterized across `nworld=1` and `nworld=2`. It applies 60 duplicate contacts for geom A alongside a single contact for distinct geom B, asserting that `OverflowType.TACTILE` is not triggered, both geoms are retained with non-zero sensor readings, and values match MuJoCo C parity.

mujoco_warp/_src/sensor_test.py
```diff
@@ -927,18 +929,278 @@ def test_tactile_sensor_geom_deduplication(self):
...
+    # into the geom frame was corrected).
+    mjm, mjd, m, d = test_data.fixture(
+      xml="""
+      <mujoco>
+        <option>
+          <flag multiccd="enable" nativeccd="disable"/>
+        </option>
+        <asset>
+          <mesh name="sensor_mesh" builtin="wedge" params="3 3 45 45 0" scale=".2 .2 .2"
+                refquat="0.5 0.5 0.5 -0.5"/>
```
smallquail, line 1092: This note is accurate about *why*, and that is the problem: compiling this model shows all three of the corrected quantities are degenerate here.

```
mesh_quat[0]      = [1. 0. 0. 0.]      exactly identity
geom_xmat[sensor] = identity
cvel[slider] ang  = [0. 0. 0.]         pure translation
```

So `wp.transpose(sensor_mat) @ vel_rel` is a no-op, the dropped `mesh_quat` rotation was a no-op, and `_transform_spatial` is independent of its reference point when the angular part is zero. The test confirms the slip channels are populated and the layout is right, which is real value, but it cannot distinguish this PR's arithmetic from the pre-fix arithmetic.

Checking which variations would discriminate, using the max component gap between `R*t` and `R*Q_mesh*t`:

| variant | tangent gap | contacting-body angular velocity |
|---|---|---|
| as submitted | 0.000000 | `[0 0 0]` |
| drop `refquat` | **1.366** | `[0 0 0]` |
| rotate sensor body | 0.000000 | `[0 0 0]` |
| drop `refquat` + hinge joint | **1.366** | `[0 2 0]` |

Rotating the sensor body does not help: `dot(v, R*t) == dot(R^T*v, t)`, so only a non-identity `mesh_quat` separates the two formulations. Dropping `refquat` yields `mesh_quat = [0.5, 0.5, -0.5, 0.5]` and covers the tangent fix; swapping the slide joint for a hinge covers the velocity fix.

The blocker is the reference, not the scene. Both CI jobs run pre-fix MuJoCo: `uv.lock` pins the nightly `3.12.1.dev968306640`, whose dev number sits below the fix's `PiperOrigin-RevId: 972096410`, and the stable job's `pip install .` resolves PyPI `mujoco 3.12.0` (uploaded 2026-08-20, before the fix). So a discriminating test would fail today in both jobs.

Suggestion: keep this test as the always-on parity check, and add a second one, with no `refquat` and a hinge joint on the contacting body, under `absltest.skipIf` on the installed MuJoCo lacking the fix (nightly below that CL, or stable below the first release containing it), matching the existing `skipIf` idiom in this file. Bumping `uv.lock` past the fix makes the locked job exercise it now; the stable job picks it up when the next release lands. Either way the note above should say coverage is gated on the reference rather than describe the test as exact parity validation.

smallquail (commented):
Summary of this round, in plain terms:

1. **The new parity test can't detect the fix it's meant to cover** (inline on `sensor_test.py`). Its scene has no mesh rotation and no spinning contact body, so pre-fix and post-fix code produce identical output there. Ask: add a second test with a rotated mesh (drop `refquat`) and a hinge on the contacting body, skipped when the installed MuJoCo predates the C fix, which both CI jobs currently do. **Blocking.**

2. **The duplicate check inside `_sensor_tactile` is now dead code** (inline on `sensor.py`). The preprocessing dedup became exact with the bitset change, so the backup loop can never trigger, yet it still runs for every taxel. Ask: delete it, along with its "load-bearing" comment, which came from my earlier review. **Should fix.**

3. **Scratch buffers are sized by `nbody` but only tactile bodies ever use them** (inline on `sensor.py`). In large batches that's hundreds of MB allocated and zeroed every step. Ask: size them by tactile-weld count. **Follow-up work, not blocking.**

For the record: my earlier claim that the downstream `is_dup` scan made results exact was wrong. @sylvesterkaczmarek's example, duplicate contacts displacing a distinct geom out of the buffer, is correct, and the bitset fixes it properly.

mujoco_warp/_src/sensor.py
```diff
@@ -2228,37 +2261,50 @@ def _sensor_tactile(
...
+    # Unlike MuJoCo C which excludes self-geom during preprocessing, weld_geom_list is shared
+    # across all sensors on the same weld body. Skipping geom == geom_id here ensures each
+    # sensor on the weld body filters its own geom without affecting others.
+    if geom < 0 or geom == geom_id:
       continue
 
+    # Load-bearing deduplication check: ensures geoms are processed at most once even if
+    # concurrent preprocessing inserted duplicates into weld_geom_list.
     is_dup = int(0)
     for j in range(g):
```
smallquail: This comment is now false, and the wording is mine from the previous round: it described the racy design and should retire with it. With the `atomic_or` test-and-set, one slot is reserved per distinct geom per `(world, weld)`, so `weld_geom_list` provably holds no duplicates and this scan can never fire.

It costs O(g^2) global loads per taxel. For a pad in the 1-5k taxel range touching ~10 geoms that is roughly 50k-250k redundant loads per sensor per step.

Suggest deleting it. If it stays as defence in depth, the comment should say so rather than call it load-bearing.

mujoco_warp/_src/sensor.py
```diff
@@ -2556,17 +2617,22 @@ def sensor_acc(m: Model, d: Data):
   if m.nsensortaxel:
     weld_geom_count = wp.zeros((d.nworld, m.nbody), dtype=int)
     weld_geom_list = wp.full((d.nworld, m.nbody, MJ_MAXCONPAIR), -1, dtype=int)
+    nwords = (m.ngeom + 31) // 32
+    weld_geom_seen = wp.zeros((d.nworld, m.nbody, nwords), dtype=wp.uint32)
```
smallquail: Follow-up material, not a blocker for this PR; noting it here so it does not get lost.

The bitset is correct, but it lands in a family of per-step allocations dimensioned over `m.nbody` while only tactile welds are ever written (every insertion is already behind `if not body_has_tactile[weld]`).

At `nworld=8192, nbody=100, ngeom=1000`, allocated and zeroed on every `sensor_acc` call:

| buffer | size |
|---|---|
| `weld_geom_list` | ~164 MB |
| `weld_geom_seen` | ~105 MB |
| `weld_geom_count` | ~3 MB |

`m.body_has_tactile` already marks the rows that matter, so a prefix-sum to a dense tactile-weld index would shrink all three by `nbody / ntactileweld`, usually one or two welds against tens or hundreds of bodies.

The `weld_geom_list` allocation predates this PR, so this belongs in a follow-up rather than in this change.

quagla:
Adde some more comments.

thowell:
@quagla

Summary of the latest changes addressing the recent review feedback:

* Dynamic parity test guarding (discussion_r3914467210): Clarified the comments on test_tactile_sensor_dynamic_parity to explain that the test scene uses pure translation and identity orientations, which gives exact parity with MuJoCo C 3.12.1. Guarded both test_tactile_sensor_dynamic_parity and test_tactile_sensor_duplicate_pressure_retention with @absltest.skipIf(not util_pkg.check_version("mujoco>=3.12.1"), "Requires mujoco>=3.12.1").

* Deduplication cleanup in taxel kernel (discussion_r3914467221): Removed the redundant O(G^2) is_dup linear scan from _sensor_tactile. The exact atomic bitset test-and-set in _preprocess_tactile_contacts already ensures each contacting geom is inserted into weld_geom_list at most once per world and weld body, eliminating duplicate checking inside the taxel threads.

* Dense tactile weld buffer indexing (discussion_r3914467229): In sensor_acc, dimensioned weld_geom_count, weld_geom_list, and weld_geom_seen by (d.nworld, m.ntactileweld, ...) instead of m.nbody. Added ntactileweld: int and weld_tactile_id: array("nbody", int) to Model and put_model (mapping tactile weld roots to 0..ntactileweld-1 and non-tactile bodies to -1), updated kernels to index buffers by tactile_idx, and added test_tactile_sensor_dense_weld_indexing covering multi-body scenes where ntactileweld << nbody and models without tactile sensors.

mujoco_warp/_src/sensor_test.py
```diff
@@ -927,18 +930,369 @@ def test_tactile_sensor_geom_deduplication(self):
...
+
+    d.sensordata.fill_(wp.inf)
+    mjw.forward(m, d)
+    np.testing.assert_allclose(d.sensordata.numpy(), 0.0, atol=1e-6)
+
+  @parameterized.parameters(1, 2)
+  @absltest.skipIf(
+    not util_pkg.check_version("mujoco>=3.12.1"),
+    "Requires mujoco>=3.12.1",
+  )
```
smallquail: This predicate is false on every CI configuration: the repo's `_parse_version` ranks `3.12.1.devNNN` below `3.12.1` by design, and stable is 3.12.0. Checked against the repo's own parser:

```
3.12.1.dev968306640   (locked nightly)      satisfies mujoco>=3.12.1 : False
3.12.1.dev972096410   (post-fix nightly)    satisfies mujoco>=3.12.1 : False
3.12.0                (PyPI stable)         satisfies mujoco>=3.12.1 : False
3.12.1                (does not exist yet)  satisfies mujoco>=3.12.1 : True
```

Net effect is that this test, `duplicate_pressure_retention` and `dense_weld_indexing` are all skipped (CI: 1343 passed / 38 skipped on the previous head, 1339 / 44 on this one), and a `uv.lock` bump would not change that.

None of the three depends on the C fix: this one passes on any MuJoCo precisely because its scene is degenerate, the dedup test compares against a C path that has always deduped (`mj_addContact` is available in 3.12.0), and the indexing test is internal. Suggest dropping all three guards.

For the still-missing discriminating test (no `refquat`, hinge on the contacting body), a guard that works: parse `mujoco.__version__`; if it has a `.dev` suffix, require the dev number >= 972096410, otherwise require >= 3.12.1. `check_version` cannot express that. A version-agnostic alternative is a small behavioral probe (compile a scene with non-identity `mesh_quat`, run `mj_sensorAcc`, compare to the post-fix formula in numpy), if relying on the dev number as the CL is unwelcome.

mujoco_warp/_src/io.py
```diff
@@ -876,6 +876,17 @@ def _check_margin(name, t1, t2, margin):
     if mjm.sensor_type[i] == mujoco.mjtSensor.mjSENS_TACTILE
     for j in range(mjm.mesh_vertnum[mjm.sensor_objid[i]])
   ]
+  tactile_geomid = mjm.sensor_refid[mjm.sensor_type == mujoco.mjtSensor.mjSENS_TACTILE]
+  tactile_weldid = mjm.body_weldid[mjm.geom_bodyid[tactile_geomid]]
+  # Note: body_has_tactile is indexed by weld root body ID (body_weldid)
+  m.body_has_tactile = np.zeros(mjm.nbody, dtype=bool)
+  m.body_has_tactile[tactile_weldid] = True
```
smallquail: `body_has_tactile` is no longer read by any kernel now that `weld_tactile_id` exists (`< 0` is the same test). Suggest removing it rather than carrying two fields for one fact.

smallquail (commented):
One regression to flag from this round, and two small cleanups.

1. **The new `skipIf` guards turn off three tests everywhere.** `check_version("mujoco>=3.12.1")` is false for every nightly (the parser ranks `.dev` builds below the stable) and for 3.12.0, so `dynamic_parity`, `duplicate_pressure_retention` and `dense_weld_indexing` now run on no CI job; the run went from 38 to 44 skipped. None of the three depends on the C fix, so the guards should simply be removed. Details inline.
2. **The discriminating test still isn't there.** The ask was a *second* scene, with no `refquat` and a hinge on the contacting body, gated on the fix being present. Gating needs a predicate that admits nightlies at or above the fix; a note inline on how.
3. `body_has_tactile` is now unread; `weld_tactile_id` supersedes it.

quagla:
One more.

thowell:
Addressed review feedback from @smallquail:

- Removed the method-level `@absltest.skipIf` guards from `test_tactile_sensor_dynamic_parity`, `test_tactile_sensor_duplicate_pressure_retention`, and `test_tactile_sensor_dense_weld_indexing`. None of these three tests depends on the C reference bugfix, so they now run unconditionally across all CI configurations without skips.
- Added `test_tactile_sensor_dynamic_discriminating` with no `refquat` on the wedge mesh (giving non-identity `mesh_quat`) and a hinge joint on the contacting body (introducing rotational velocity). The Warp pipeline execution and non-empty channel checks run unconditionally on all CI jobs, while the reference parity check against MuJoCo C is gated inline using `util_pkg.check_version("mujoco>=3.12.1.dev972096410")` to admit post-fix nightlies and future stable releases >= 3.12.1.
- Removed deprecated and unread `body_has_tactile` from `Model` (docstring and dataclass) and `io.put_model`, as `weld_tactile_id` (< 0) fully supersedes it.

smallquail (approved):
LGTM. Verified this round rather than read it:

- Guards removed; CI is back to `1347 passed, 38 skipped` on all four jobs, with the six un-gated cases restored and the two new discriminating cases running everywhere.
- `test_tactile_sensor_dynamic_discriminating` genuinely discriminates. Compiling its scene in MuJoCo 3.12.0 gives `mesh_quat = [0.5, 0.5, -0.5, 0.5]`, a non-identity sensor `geom_xmat`, and rotator angular velocity `[0, 2, 0]`. A numpy re-implementation of the pre-fix formula reproduces 3.12.0's output exactly, and the post-fix formula differs from it by 0.40 on the slip channels.
- Nice property of the scene: the rotator's geom origin sits on its hinge axis, so pre-fix C gives exactly zero slip while the fixed contact-point velocity does not. The unconditional `slip U non-empty` assertion therefore already exercises the `vel_other` fix on every CI job today.
- `mujoco>=3.12.1.dev972096410` evaluates correctly under the repo parser: false for the locked nightly and 3.12.0, true for post-fix nightlies and any stable >= 3.12.1.
- `body_has_tactile` is fully removed.

Two non-blocking notes for after merge:

1. The gated parity check is the only thing verifying the `mesh_quat` half of the fix, and `uv.lock` still pins `dev968306640`, below the fix. Bumping the pin is what switches that check on; worth doing as the next change.
2. `_parse_version` compares non-numeric components as strings, so a ten-digit dev number (`dev1000000000`) sorts below `dev972096410` and would silently re-disable the check. Not imminent, but worth a note in `util_pkg` or a numeric-aware compare next time someone touches it.

quagla (approved):
Left two nits but LGTM otherwise, thanks!

thowell:
@quagla

1. update the uv.lock so that the dev version of mujoco is installed and runs this test with the ci
2. will leave this as future work for now

thanks!

Issue #1546, opened 2026-07-21
Mesh-mesh GJK/EPA contact is not invariant to common world translation


mmichelis:
Follow-up to #1527 after #1530. For the five-vertex convex mesh in the repro below, MuJoCo Warp creates a valid constraint at `y=0` but misses it after both bodies are translated to `y=1`. A common translation should not change the result.

Tested with MuJoCo Warp 3.10.0.2 at `081c74e276cfc7f2484e98b80b1377aa4b3c64bd`, MuJoCo 3.10.0, Warp 1.16.0.dev20260716, and CUDA 12.9 on an RTX 5090.

```text
y=0.0: native_contacts=3, native_distance=-0.0816492254
y=0.0: mjwarp_contacts=1, constraint=True,  distance=-0.08164963871, addresses=[[0, 1, 2, -1]]
y=1.0: native_contacts=3, native_distance=-0.0816492254
y=1.0: mjwarp_contacts=1, constraint=False, distance=1.19209238e-08, addresses=[[-1, -1, -1, -1]]
```

## Root cause

Both meshes receive the same translation, so their relative transform is unchanged: `(A + t) - (B + t) = A - B`. MJWarp instead transforms each support point to absolute float32 world coordinates before subtracting the pair, effectively computing `float32(A + t) - float32(B + t)`.

Near `y=1`, float32 spacing is about `1.19e-7`, so relevant simplex components of a few `1e-9` round away. GJK consequently produces a three-point triangle instead of a four-point tetrahedron containing the origin. EPA cannot initialize a full-dimensional polytope and leaves the tiny positive GJK distance unchanged.

The zero contact margin requires a negative distance to create a constraint. The per-geom `gap=0.01` values sum to `0.02`, so MJWarp retains the positive-distance candidate as a raw contact but leaves its constraint addresses at `-1`. The gap exposes the failure; it does not cause it.

The five-vertex mesh is a minimized, near-degenerate trigger for the precision loss; none of its four-vertex subsets reproduces the same failure.

## Proposed fix

Run rigid non-heightfield GJK, EPA, and multicontact in a pair-local frame, then restore the origin when writing the contact position:

```python
origin = geom1.pos
geom1.pos = wp.vec3(0.0)
geom2.pos -= origin

# GJK, EPA, and multicontact

contact_pos = 0.5 * (witness1 + witness2) + origin
```

[...]

PR #1603, opened 2026-08-24
Add support for variable distance multicontact

kbayes:
See:

https://github.com/google-deepmind/mujoco/commit/ec506dbde38d82894bfd6f39e6acef26b75acaab
https://github.com/google-deepmind/mujoco/commit/17e6d44a221cd97bc719ccb112f2bdc0ae147806

Fixes #1510

PR #1613, opened 2026-08-27
Run rigid CCD GJK/EPA in a pair-local frame to fix common-translation precision loss

rootkiller6788:
Mesh-mesh GJK/EPA formed support points in absolute float32 world coordinates, so a common translation of both bodies did not cancel exactly and the small simplex components GJK needs to certify a tetrahedron containing the origin rounded away. A near-degenerate five-vertex mesh pair correctly produced a constraint at y=0 but silently dropped it at y=1, so the scene depended on its position in the world (issue #1546).

This change runs the rigid non-heightfield CCD narrowphase (GJK, EPA, and multicontact) in a pair-local frame anchored at geom1, so the common translation cancels before support-point rounding instead of after, and restores the world origin on the written contact position. Contact distance and frame are translation-invariant and are unaffected.

A regression test builds the five-vertex mesh pair at y=0 and y=1, runs the full forward pass, and asserts both produce an identical penetrating contact.

Verified with the issue repro (constraint restored at y=1 with distance -0.0816496, matching y=0) and the existing collision suites: collision_gjk_test, collision_driver_test, and collision_primitive_core_test pass with this change. The one failure in this environment, collision_driver_test test_collision20 (mesh_plane_complex), also fails on main without this change.


mujoco_warp/_src/collision_convex.py
```diff
@@ -816,6 +816,17 @@ def eval_ccd_write_contact(
       cutoff = 1.0e32
     else:
       cutoff = gap
+    # Run GJK/EPA in a pair-local frame anchored at geom1 so that the support
+    # points subtracted during the Minkowski-difference iterations stay close to
```
kbayes: Remove long comment for something like

"Shift origin to geom1 position for translation invariance with numerical precision."

mujoco_warp/_src/collision_driver_test.py
```diff
@@ -608,6 +608,62 @@ def test_collision(self, fixture):
     else:
       self.assertEqual(d.nacon.numpy()[0], mjd.ncon)
 
+  def test_mesh_mesh_common_translation(self):
+    """A common world translation must not change the CCD contact result (#1546).
```
kbayes: Remove long doc string. Just say

"test collision against translation"

kbayes:
@rootkiller6788, I pulled this in and didn't see any regressions. A couple of nits for you to resolve and we can merge this in. Thank you.

rootkiller6788:
Thank you for the review and for confirming you didn't see any regressions. I've resolved both nits and also merged in the latest main to clear the merge conflict.

- collision_convex.py: replaced the long comment with a one-liner, "Shift origin to geom1 position for translation invariance with numerical precision."
- collision_driver_test.py: trimmed the docstring down to "Test collision against translation."

The branch had drifted from main (25 commits, including the variable-distance multicontact change in #1603 that introduced per-contact dists). I merged main in; the one real conflict was at the write_contact call in ccd_kernel_builder, where I kept main's dists[i] per-contact distance together with the + origin restore of the contact position in world coordinates.

Re-verified on the merged tree (CPU device): the new test_mesh_mesh_common_translation, collision_gjk_test.py, and the CCD/convex subsets of collision_driver_test.py (ccd_margin_dist, convex_contact_frame_parity, plane_meshtet) all pass. Could you please take another look?

kbayes (approved):

PR #1616, opened 2026-08-28
Add optional supersampling to the renderer

kevinzakka:
This PR adds anti-aliasing via supersampling. `create_render_context(samples_per_pixel=n)` casts n^2 rays per pixel instead of one, spread evenly across the pixel, and averages the result. The default of 1 leaves the renderer unchanged.

Only RGB is averaged: depth and segmentation keep a single sample, since averaging object IDs is meaningless. Supersampling requires precomputed rays (the default).

supersampling

*A slanted silhouette at 1, 2, and 4 samples per pixel.*

supersampling closeup

*The same edge rendered at 96x96 and magnified: 1 spp is a binary staircase, 4 spp blends by coverage.*

aa on a wrist camera

*1 vs 2 spp from a robot wrist camera. Look at the table's back edge and the shadow boundaries.*

Cost is the expected n^2 renders, slightly sublinear thanks to fixed per-launch overheads (clutter scene, 2048 worlds, 4 cameras @ 64x64, RGB + depth + segmentation, shadows; RTX 5090):

| samples_per_pixel | ms/step | vs 1 spp |
|---|---|---|
| 1 | 111.0 | 1.0x |
| 2 | 409.1 | 3.7x |
| 3 | 921.5 | 8.3x |

At the default `samples_per_pixel=1` the `aloha_clutter_vision` benchmark is unchanged from main (7,921 steps/s on both).

### DIFF
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -704,6 +704,12 @@ def _build_megakernel(m: Model, rc: RenderContext):
   # Static parameters extracted for JAX FFI closure.
   rc_static = {f.name: getattr(rc, f.name) for f in dataclasses.fields(rc) if f.type in (int, wp.uint32, bool, float, wp.vec3)}
   rc_static["enable_specular_or_emission"] = rc.enable_specular or rc.enable_emission
+  bg = int(rc.background_color)
+  rc_static["background_color_vec3"] = wp.vec3(
+    float((bg >> 16) & 0xFF) / 255.0,
+    float((bg >> 8) & 0xFF) / 255.0,
+    float(bg & 0xFF) / 255.0,
+  )
   M_NLIGHT = m.nlight
 
   aa = rc.samples_per_pixel * rc.samples_per_pixel > 1
@@ -926,7 +932,8 @@ def _build_megakernel(m: Model, rc: RenderContext):
         dist,
       )
 
-    if render_seg[camid] and geom_id != -1:
+    # Depth and seg are single-sample outputs: only the first pass writes them.
+    if ray_base == 0 and render_seg[camid] and geom_id != -1:
       if geom_id == -2:
         seg_out[worldid, seg_adr[camid] + rayid_local] = wp.vec2i(mesh_id, int(ObjType.FLEX))
       else:
@@ -934,7 +941,7 @@ def _build_megakernel(m: Model, rc: RenderContext):
 
     # Early Out
     if geom_id == -1:
-      if render_depth[camid]:
+      if ray_base == 0 and render_depth[camid]:
         depth = 0.0
         if wp.static(has_splats):
           if splat_depth > 0.0:
@@ -952,25 +959,13 @@ def _build_megakernel(m: Model, rc: RenderContext):
         store_pixel(worldid, rgb_adr[camid] + rayid_local, skybox_color, rgb_out, aa_accum_out)
       elif render_rgb[camid]:
         if wp.static(has_splats):
-          packed = wp.static(rc_static["background_color"])
-          background = wp.vec3(
-            float((packed >> wp.uint32(16)) & wp.uint32(0xFF)),
-            float((packed >> wp.uint32(8)) & wp.uint32(0xFF)),
-            float(packed & wp.uint32(0xFF)),
-          ) * wp.static(1.0 / 255.0)
-          pixel_color = splat_color + background * splat_transmittance
+          pixel_color = splat_color + wp.static(rc_static["background_color_vec3"]) * splat_transmittance
           store_pixel(worldid, rgb_adr[camid] + rayid_local, pixel_color, rgb_out, aa_accum_out)
         elif wp.static(aa):
-          packed = wp.static(rc_static["background_color"])
           store_pixel(
             worldid,
             rgb_adr[camid] + rayid_local,
-            wp.vec3(
-              float((packed >> wp.uint32(16)) & wp.uint32(0xFF)),
-              float((packed >> wp.uint32(8)) & wp.uint32(0xFF)),
-              float(packed & wp.uint32(0xFF)),
-            )
-            * wp.static(1.0 / 255.0),
+            wp.static(rc_static["background_color_vec3"]),
             rgb_out,
             aa_accum_out,
           )
@@ -978,7 +973,7 @@ def _build_megakernel(m: Model, rc: RenderContext):
           rgb_out[worldid, rgb_adr[camid] + rayid_local] = wp.static(rc_static["background_color"])
       return
 
-    if render_depth[camid]:
+    if ray_base == 0 and render_depth[camid]:
       # Planar depth: project Euclidean distance onto the camera's optical axis.
       # In camera-local coordinates, the optical axis is -Z. The Z-component of the
       # normalized ray direction is negative, so -ray_dir_local_cam[2] gives cos(θ)
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -675,6 +675,40 @@ def _make_compute_lighting(cast_ray_first_hit: wp.Function) -> wp.Function:
   return compute_lighting
 
 
+@wp.kernel
+def _aa_accumulate(
+  # In:
+  rgb: wp.array2d[wp.uint32],
+  # Out:
+  accum: wp.array2d[wp.vec3],
+):
+  worldid, i = wp.tid()
+  c = rgb[worldid, i]
+  accum[worldid, i] += wp.vec3(
+    float(c & wp.uint32(0xFF)),
+    float((c >> wp.uint32(8)) & wp.uint32(0xFF)),
+    float((c >> wp.uint32(16)) & wp.uint32(0xFF)),
+  )
+
+
+@wp.kernel
+def _aa_resolve(
+  # In:
+  accum: wp.array2d[wp.vec3],
+  inv_n: float,
+  # Out:
+  rgb: wp.array2d[wp.uint32],
+):
+  worldid, i = wp.tid()
+  c = accum[worldid, i] * inv_n
+  rgb[worldid, i] = (
+    wp.uint32(wp.clamp(c[0], 0.0, 255.0))
+    | (wp.uint32(wp.clamp(c[1], 0.0, 255.0)) << wp.uint32(8))
+    | (wp.uint32(wp.clamp(c[2], 0.0, 255.0)) << wp.uint32(16))
+    | (wp.uint32(255) << wp.uint32(24))
+  )
+
+
 def _build_megakernel(m: Model, rc: RenderContext):
   """Construct the specialised megakernel for this context."""
   has_splats = rc.splat_count > 0
@@ -741,6 +775,7 @@ def _build_megakernel(m: Model, rc: RenderContext):
     cam_id_map: wp.array[int],
     ray: wp.array[wp.vec3],
     ray_offset: wp.array[wp.vec3],
+    ray_base: int,
     rgb_adr: wp.array[int],
     depth_adr: wp.array[int],
     seg_adr: wp.array[int],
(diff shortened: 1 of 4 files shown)


mujoco_warp/_src/render.py
```diff
@@ -1094,93 +1173,123 @@ def _render_megakernel(
...
+  _render_megakernel = getattr(rc, "_megakernel", None)
+  if _render_megakernel is None:
+    _render_megakernel = _build_megakernel(m, rc)
+    object.__setattr__(rc, "_megakernel", _render_megakernel)
+
+  nsamples = rc.samples_per_pixel * rc.samples_per_pixel
+  if nsamples > 1:
+    rc.aa_accum.zero_()
+
+  for sample in range(nsamples):
```
havess, line 1197: OOC did you try pushing this loop into render to avoid the kernel launch?

The other arguably more common/performant thing I've seen is to push the number of samples to the end of the grid dim. So Width x Height x N_Samples and just accumulate straight into the pixel value.

kevinzakka:
Regarding @havess's comment, I benchmarked all three on the clutter scene with 2048 worlds x 4 cams @ 64x64 on a RTX 5090:

| ms/step | per-sample launches (current) | loop inside kernel | samples in grid dim + atomics |
|---|---|---|---|
| spp 1 | 104.6 | 104.8 | 104.6 |
| spp 2 | 407.0 | 407.6 | 403.7 |
| spp 3 | 923.3 | 925.5 | 910.1 |

Doing one launch instead of n^2 (the in-kernel loop) is not faster so the launches cost nothing in this case. The grid layout wins ~1% by letting samples fill each other's idle tails, which I don't think is worth the atomics.

thowell:
@kevinzakka

some suggestions

### 1. Guard `depth_out` and `seg_out` writes to pass 0 (`ray_base == 0`)

> [!IMPORTANT]
> Currently, multi-pass rendering causes object silhouettes in `seg_data` to dilate and leaves `depth_data` biased toward the last subpixel corner.

#### Issue:
The PR description notes that *"depth and segmentation keep a single sample, since averaging object IDs is meaningless."* However, currently all $n^2$ passes execute the depth and segmentation write logic:
1. **Segmentation (`seg_out`):** Because misses (`geom_id == -1`) do not write to `seg_out`, `seg_out` retains hits from previous passes. Across $n^2$ jittered passes, `seg_out` accumulates a union (OR) of all sub-pixel hits, causing object silhouettes in segmentation masks to dilate.
2. **Depth (`depth_out`):** Because every pass overwrites `depth_out`, the final buffer contains the depth of only the *last* sample (`sample = nsamples - 1`, the bottom-right corner sub-pixel), rather than the first sample or pixel center.
3. **Memory bandwidth:** Writing depth and segmentation $n^2$ times incurs unnecessary global memory stores on passes $1 \dots n^2 - 1$.

#### How to resolve:
Guard both writes with `if ray_base == 0:`. Since `ray_base` is uniform across the entire grid for each launch, this incurs zero warp divergence and eliminates all redundant memory traffic on subsequent passes.

#### Suggested change in `mujoco_warp/_src/render.py`:
```python
    # Inside _render_megakernel:
    if ray_base == 0 and render_seg[camid] and geom_id != -1:
      if geom_id == -2:
        seg_out[worldid, seg_adr[camid] + rayid_local] = wp.vec2i(mesh_id, int(ObjType.FLEX))
      else:
        seg_out[worldid, seg_adr[camid] + rayid_local] = wp.vec2i(geom_id, int(ObjType.GEOM))

    # Early Out
    if geom_id == -1:
      if ray_base == 0 and render_depth[camid]:
        depth = 0.0
        if wp.static(has_splats):
          if splat_depth > 0.0:
            depth = splat_depth * -ray_dir_local_cam[2]
        depth_out[worldid, depth_adr[camid] + rayid_local] = depth
      if wp.static(rc_static["render_skybox"]) and render_rgb[camid]:
        ...
      return

    if ray_base == 0 and render_depth[camid]:
      depth = dist
      if wp.static(has_splats):
        if splat_depth > 0.0:
          depth = splat_depth
      depth_out[worldid, depth_adr[camid] + rayid_local] = depth * -ray_dir_local_cam[2]
```

---

### 2. Allocate `aa_accum` based on RGB pixels (`ri`) instead of `total`

#### Issue:
In `create_render_context`:
```python
aa_accum = wp.zeros((nworld, int(total) if nsamples > 1 else 1), dtype=wp.vec3)
```
`total` includes pixels from all cameras, including depth-only and segmentation-only cameras. However, `store_pixel` indexes `aa_accum` via `adr = rgb_adr[camid] + rayid_local`, which only spans $[0, ri - 1]$. Similarly, `_aa_resolve` launches over `dim = rc.rgb_data.shape` ($(nworld, ri)$). If depth-only cameras exist, elements from $ri$ to $total - 1$ are never accessed, wasting 12 bytes per pixel per world of VRAM and unnecessary zeroing overhead.

#### Suggested change in `mujoco_warp/_src/render_util.py`:
```python
  aa_accum = wp.zeros((nworld, ri if nsamples > 1 else 1), dtype=wp.vec3)
```

---

### 3. Validate `render_rgb` when `samples_per_pixel > 1`

#### Issue:
If a user creates a render context with `samples_per_pixel > 1` but `render_rgb=False` across all cameras (e.g. depth-only pipeline), $ri = 0$. `render()` will execute $n^2$ launches doing no RGB work, repeatedly overwriting depth/seg, and then launch `_aa_resolve` with a 0-width dimension.

#### Suggested change in `mujoco_warp/_src/render_util.py`:
```python
  if samples_per_pixel > 1 and not use_precomputed_rays:
    raise ValueError("samples_per_pixel > 1 requires use_precomputed_rays=True: dynamic rays carry no sub-pixel jitter.")
  if samples_per_pixel > 1 and ri == 0:
    raise ValueError("samples_per_pixel > 1 requires at least one camera with render_rgb=True.")
```

---

### 4. Precompute background color `wp.vec3` in `rc_static`

> [!TIP]
> Precomputing the unpacked background color at context creation avoids runtime bit-shifting and redundant normalization/scaling in the megakernel.

#### Issue:
In `_render_megakernel`, missed rays with `aa=True` unpack the 32-bit packed background color on every thread via bitwise shifts, convert to float, scale by `1.0 / 255.0`, and then `store_pixel` immediately multiplies by `255.0`. Since `rc.background_color` is known when building the megakernel, this can be precomputed once.

#### Suggested change in `mujoco_warp/_src/render.py`:
```python
  # In _build_megakernel:
  bg_packed = rc.background_color
  rc_static["background_color_vec3"] = wp.vec3(
    float((bg_packed >> 16) & 0xFF) / 255.0,
    float((bg_packed >> 8) & 0xFF) / 255.0,
    float(bg_packed & 0xFF) / 255.0,
  )

  # Inside _render_megakernel miss branch:
  elif wp.static(aa):
    store_pixel(
      worldid,
      rgb_adr[camid] + rayid_local,
      wp.static(rc_static["background_color_vec3"]),
      rgb_out,
      aa_accum_out,
    )
```

---

### 5. Additional Test Coverage

While `test_supersampling_antialiases_edges` nicely verifies monotonic sub-pixel edge blending for RGB silhouettes, adding unit tests for the following edge cases and multi-buffer interactions will prevent regressions:

#### A. Depth & Segmentation Invariance with Supersampling
Verifies that `seg_data` does not suffer from boundary dilation and that `depth_data` remains consistent when supersampling is enabled:
```python
def test_supersampling_preserves_segmentation_and_depth(self):
  mjm, _, m, d = test_data.fixture(
    xml=(
      '<mujoco><worldbody><light pos="0 -2 2" dir="0 1 -1" directional="true"/>'
      '<camera pos="0 -2 0" xyaxes="1 0 0 0 0 1" fovy="30"/>'
      '<geom type="box" size="0.3 0.3 0.3" rgba="0.9 0.9 0.9 1"/></worldbody></mujoco>'
    )
  )
  rc_1spp = mjw.create_render_context(
    mjm, cam_res=(64, 64), render_rgb=True, render_depth=True, render_seg=True, samples_per_pixel=1
  )
  mjw.render(m, d, rc_1spp)
  seg_1spp = rc_1spp.seg_data.numpy()[0]
  depth_1spp = rc_1spp.depth_data.numpy()[0]

  rc_2spp = mjw.create_render_context(
    mjm, cam_res=(64, 64), render_rgb=True, render_depth=True, render_seg=True, samples_per_pixel=2
  )
  mjw.render(m, d, rc_2spp)
  seg_2spp = rc_2spp.seg_data.numpy()[0]
  depth_2spp = rc_2spp.depth_data.numpy()[0]

  # Hits should match pass 0 and not dilate to all 4 sub-pixel rays
  hit_1spp = seg_1spp[:, 0] >= 0
  hit_2spp = seg_2spp[:, 0] >= 0
  self.assertLess(abs(int(np.sum(hit_2spp)) - int(np.sum(hit_1spp))), 10)
  # Background depth remains zero
  self.assertTrue(np.all(depth_2spp[~hit_2spp] == 0.0))
```

#### B. Argument Validation
Ensures invalid parameter combinations fail fast with descriptive error messages:
```python
def test_supersampling_validation(self):
  mjm, _, _, _ = test_data.fixture(xml="<mujoco/>")
  with self.assertRaisesRegex(ValueError, "samples_per_pixel must be at least 1"):
    mjw.create_render_context(mjm, samples_per_pixel=0)
  with self.assertRaisesRegex(ValueError, "samples_per_pixel > 1 requires use_precomputed_rays=True"):
    mjw.create_render_context(mjm, samples_per_pixel=2, use_precomputed_rays=False)
  with self.assertRaisesRegex(ValueError, "requires at least one camera with render_rgb=True"):
    mjw.create_render_context(mjm, samples_per_pixel=2, render_rgb=False)
```

#### C. Multi-Camera with Mixed Outputs
Verifies that camera addressing offsets, ray indices, and buffer shapes remain correct when some cameras are RGB and others are depth-only:
```python
def test_supersampling_multi_camera_mixed_outputs(self):
  mjm, _, m, d = test_data.fixture(
    xml=(
      '<mujoco><worldbody><camera name="cam_rgb" pos="0 -2 0" xyaxes="1 0 0 0 0 1" fovy="30"/>'
      '<camera name="cam_depth" pos="0 0 2" xyaxes="1 0 0 0 1 0" fovy="30"/>'
      '<geom type="box" size="0.2 0.2 0.2" rgba="1 0 0 1"/></worldbody></mujoco>'
    )
  )
  rc = mjw.create_render_context(
    mjm,
    cam_res=[(32, 32), (32, 32)],
    render_rgb=[True, False],
    render_depth=[False, True],
    samples_per_pixel=2,
  )
  mjw.render(m, d, rc)

  # rgb_data and aa_accum should match active RGB camera pixels only (1024)
  self.assertEqual(rc.rgb_data.shape, (1, 1024))
  self.assertEqual(rc.aa_accum.shape, (1, 1024))
  # depth_data should match depth camera pixels only (1024)
  self.assertEqual(rc.depth_data.shape, (1, 1024))
  self.assertGreater(np.count_nonzero(rc.depth_data.numpy()[0]), 0)
```

#### D. Batched Worlds (`nworld > 1`)
Ensures accumulation and resolution work across multiple environments without cross-world corruption:
```python
def test_supersampling_batched_worlds(self):
  nworld = 4
  mjm, _, m, d = test_data.fixture(
    xml=(
      '<mujoco><worldbody><light pos="0 -2 2" dir="0 1 -1" directional="true"/>'
      '<camera pos="0 -2 0" xyaxes="1 0 0 0 0 1" fovy="30"/>'
      '<geom type="box" size="0.3 0.3 0.3" euler="0 22 0" rgba="0.8 0.2 0.2 1"/></worldbody></mujoco>'
    ),
    nworld=nworld,
  )
  rc = mjw.create_render_context(mjm, nworld=nworld, cam_res=(32, 32), render_rgb=True, samples_per_pixel=2)
  mjw.render(m, d, rc)

  self.assertEqual(rc.rgb_data.shape, (nworld, 32 * 32))
  for w in range(nworld):
    rgb = _unpack_rgb(rc.rgb_data.numpy()[w])
    unique_vals = len(np.unique(rgb[..., 0]))
    self.assertGreater(unique_vals, 2)  # antialiased edge produces intermediate blends
```

kevinzakka:
Thanks @thowell, applied all fixes!

thowell (approved):

thowell:
lgtm, thanks @kevinzakka!

PR #1619, opened 2026-08-29
Honor use_textures in create_render_context

shailensobhee:
`create_render_context` takes a `use_textures` parameter (default `True`) and its docstring documents it as "Whether to use textures", but the implementation ignores it: the texture-creation loop always calls `create_warp_texture` (which allocates a `wp.Texture2D`) for every texture in the model, even when the caller passed `use_textures=False`. Raycast-only callers that never sample a texture still pay the full texture-materialization cost, and on backends without texture-object support this fails hard. Concretely, `wp.Texture2D` maps to a hipArray texture object on ROCm/HIP, which is not supported there, so `create_render_context` raises "operation not supported" (CUDA error 801) even for a pure raycast context that requested no textures.

This change guards the texture-creation loop with the existing `use_textures` flag, so `use_textures=False` skips texture allocation entirely. When `use_textures=True` the behavior is unchanged: `textures_registry` is populated exactly as before, and `textures` stays a `wp.array(..., dtype=wp.Texture2D)` (empty when nothing is created, which already happened for models with `ntex == 0`). The added test asserts the registry is empty when `use_textures=False` and fully populated when `use_textures=True` on a model that contains a texture.

One note on testing: I develop on AMD (MI300X/ROCm) via the ROCm Warp fork, which is currently based on Warp 1.13. mujoco_warp `main` requires `warp-lang>=1.15` (it uses `wp.kernel(grid_stride=...)`, unavailable in 1.13), so I could not execute the GPU test suite on my hardware. I verified `ruff check` and `ruff format` pass and the changed modules compile, and confirmed the change against the surrounding code by inspection. I would appreciate a maintainer running the render tests on CUDA to confirm.

### DIFF
--- a/mujoco_warp/_src/render_util.py
+++ b/mujoco_warp/_src/render_util.py
@@ -496,10 +496,7 @@ def create_render_context(
       flex_group_root[:, f] = group_root.numpy()
 
   textures_registry = []
-  # Only materialize GPU textures when the caller actually needs them. The
-  # loop below calls wp.Texture2D, which requires backend texture-object
-  # support (hipArray textures are unsupported on some non-CUDA backends);
-  # raycast-only contexts pass use_textures=False and never sample textures.
+  # Only materialize GPU textures when the caller actually needs them.
   if use_textures:
     for i in range(mjm.ntex):
       textures_registry.append(create_warp_texture(mjm, i))
--- a/mujoco_warp/_src/render_util.py
+++ b/mujoco_warp/_src/render_util.py
@@ -496,8 +496,13 @@ def create_render_context(
       flex_group_root[:, f] = group_root.numpy()
 
   textures_registry = []
-  for i in range(mjm.ntex):
-    textures_registry.append(create_warp_texture(mjm, i))
+  # Only materialize GPU textures when the caller actually needs them. The
+  # loop below calls wp.Texture2D, which requires backend texture-object
+  # support (hipArray textures are unsupported on some non-CUDA backends);
+  # raycast-only contexts pass use_textures=False and never sample textures.
+  if use_textures:
+    for i in range(mjm.ntex):
+      textures_registry.append(create_warp_texture(mjm, i))
   textures = wp.array(textures_registry, dtype=wp.Texture2D)
 
   # Locate skybox texture
--- a/mujoco_warp/_src/render_util_test.py
+++ b/mujoco_warp/_src/render_util_test.py
@@ -179,6 +179,32 @@ class RenderUtilTest(parameterized.TestCase):
     group_np = rc.group.numpy()
     _assert_eq(group_np, np.repeat(np.arange(nworld), rc.bvh_ngeom), "render context group values")
 
+  def test_use_textures_flag(self):
+    """create_render_context honors use_textures: skip texture materialization when False."""
+    textured_xml = """
+    <mujoco>
+      <asset>
+        <texture name="tex" type="2d" builtin="flat" rgb1="1 0 0" width="4" height="4"/>
+        <material name="mat" texture="tex"/>
+      </asset>
+      <worldbody>
+        <camera name="cam" pos="0 -3 2" xyaxes="1 0 0 0 0.6 0.8" resolution="16 16" output="rgb"/>
+        <geom type="plane" size="5 5 0.1" material="mat"/>
+        <geom type="sphere" size="0.5" pos="0 0 1" material="mat"/>
+      </worldbody>
+    </mujoco>
+    """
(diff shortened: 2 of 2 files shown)


thowell:
@shailensobhee thank you for contributing to mujoco warp. please sign the cla. thanks!

shailensobhee:
> @shailensobhee thank you for contributing to mujoco warp. please sign the cla. thanks!

CLA signed, linked to this Github account.

mujoco_warp/_src/render_util.py
```diff
@@ -496,8 +496,13 @@ def create_render_context(
       flex_group_root[:, f] = group_root.numpy()
 
   textures_registry = []
-  for i in range(mjm.ntex):
-    textures_registry.append(create_warp_texture(mjm, i))
+  # Only materialize GPU textures when the caller actually needs them. The
```
thowell: lets simplify this comment to the first sentence

mujoco_warp/_src/render_util_test.py
```diff
@@ -179,6 +179,32 @@ def test_bvh_creation(self, nworld):
...
+        <material name="mat" texture="tex"/>
+      </asset>
+      <worldbody>
+        <camera name="cam" pos="0 -3 2" xyaxes="1 0 0 0 0.6 0.8" resolution="16 16" output="rgb"/>
+        <geom type="plane" size="5 5 0.1" material="mat"/>
+        <geom type="sphere" size="0.5" pos="0 0 1" material="mat"/>
+      </worldbody>
+    </mujoco>
+    """
+    mjm, mjd, m, d = test_data.fixture(xml=textured_xml)
```
thowell: lets inline the xml in `test_data.fixture`

mujoco_warp/_src/render_util_test.py
```diff
@@ -179,6 +179,32 @@ def test_bvh_creation(self, nworld):
...
+    self.assertGreater(mjm.ntex, 0, "test model must contain at least one texture")
+
+    rc_off = mjw.create_render_context(mjm, cam_res=(16, 16), use_textures=False)
+    self.assertEqual(len(rc_off.textures_registry), 0, "no textures should be created when use_textures=False")
+    self.assertFalse(rc_off.use_textures)
+
+    rc_on = mjw.create_render_context(mjm, cam_res=(16, 16), render_rgb=True, use_textures=True)
```
thowell: do we need `render_rgb=True` here? if not, lets remove

thowell:
left some minor comments, otherwise lgtm. thanks @shailensobhee!

mujoco_warp/_src/render_util.py
shailensobhee: Done in c81a005: comment trimmed to just the first sentence.

mujoco_warp/_src/render_util_test.py
shailensobhee: Done in c81a005: the XML is now inlined directly into the test_data.fixture(xml=...) call.

mujoco_warp/_src/render_util_test.py
shailensobhee: Good catch, removed in c81a005. render_rgb is not needed here (defaults to the model camera output flags) and the test only asserts on textures_registry and use_textures.

shailensobhee:
Thanks @thowell! Addressed all three in c81a005: trimmed the comment to the first sentence, inlined the XML into the `test_data.fixture` call, and removed the unnecessary `render_rgb=True`. `ruff check` and `ruff format` still pass.

shailensobhee:
Update: I was able to run this after all, so please disregard the testing caveat in the PR description.

I ran the full `render_util_test.py` suite (including the new `test_use_textures_flag`) against current `main` with the change applied, on `warp-lang 1.17.0` using the CPU device on my host. All 14 tests pass. Both branches of the new test are genuinely exercised: for a model with `ntex=1`, `use_textures=False` yields `len(textures_registry)==0` and `use_textures=True` yields `len(textures_registry)==1`. `wp.Texture2D` materializes fine on the CPU device, and the pre-existing `test_create_warp_texture` / `test_render_context_with_textures` also pass, so the `use_textures=True` path is covered.

```
14 passed, 3 warnings in 6.41s
```

The one thing I still have not exercised is a CUDA GPU device specifically (no NVIDIA hardware on my side), but the CPU and CUDA paths share the same `Texture2D` Python API and all texture tests pass, so CI should confirm the GPU path. The original ROCm motivation stands: `wp.Texture2D` maps to a hipArray texture object that is unsupported on the ROCm/HIP backend, which is exactly what `use_textures=False` now lets raycast-only callers avoid.

thowell (approved):

PR #1620, opened 2026-08-29
optimize flex performance

thowell:
This PR introduces targeted performance optimizations for flexible body simulation in MuJoCo Warp, focusing on broadphase collision filtering, farthest point contact sampling, constraint solver linesearch, and conjugate gradient search direction updates.

During the sweep-and-prune broadphase, 2D flex elements are prefiltered using a separating axis test between planar triangle bounding disks, eliminating non-colliding element pairs before narrowphase dispatch. Farthest point contact sampling is parallelized using block-cooperative reductions with a strict total-order tie-breaker and early termination when candidate counts are within capacity, matching MuJoCo C contact filtering semantics. In the constraint solver, quadratic costs for invariant equality constraint rows are pre-reduced prior to the linesearch loop, and the Fletcher-Reeves beta calculation is fused into a single vectorized reduction kernel. In data initialization, eager full-size tensor allocation is skipped for compacted solver arrays that are subsequently populated by compact workspace setup.

Benchmark results below isolate code speedups on identical configurations with zero buffer overflows using dense Jacobians:

Cloth (32 worlds, nconmax=14000, nccdmax=14000, njmax=3600, dense Jacobian, zero overflow):
- Simulation speed: 433 steps/s -> 4,194 steps/s (9.68x speedup)
- Flex collision latency: 2,181 us -> 119 us (18.39x faster)
- CG solve latency: 80 us -> 74 us (1.08x faster)
- Total step latency: 2,305 us -> 238 us (9.70x faster)

Aloha Cloth (32 worlds, nconmax=8000, njmax=20000, opt.iterations=1000, dense Jacobian, zero overflow):
- Simulation speed: 1,576 steps/s -> 2,044 steps/s (1.30x speedup)
- Flex collision latency: 413 us -> 95 us (4.34x faster)
- CG solve latency: 165 us -> 326 us
- Total step latency: 634 us -> 488 us (1.30x faster)

Standalone performance of the PR commit with updated benchmark suite settings (256 worlds, zero overflow):
- Cloth (256 worlds, nconmax=3500, nccdmax=3400, njmax=3600): 7,996 steps/s, total step latency 125 us (flex collision 66 us, CG solve 31 us)
- Aloha Cloth (256 worlds, nconmax=6400, nccdmax=3400, njmax=20000): 3,856 steps/s, total step latency 259 us (flex collision 34 us, CG solve 177 us)


mujoco_warp/_src/collision_flex.py
```diff
@@ -2065,12 +2255,14 @@ def _compute_filter_key(
...
-  # Deduplication compares exact cand_elem values, so any wrap-around (> 65535 elements)
-  # affects sort order only without compromising contact deduplication correctness.
-  elem_key = ((wp.int64(elem1 + 1) & wp.int64(0xFFFF)) << wp.int64(16)) | (wp.int64(elem2 + 1) & wp.int64(0xFFFF))
 
-  key_out[i] = (group_key << wp.int64(32)) | elem_key
+  # Spatial projection key: project position onto diagonal vector u = (1, 1, 1) / sqrt(3).
+  # Clamped monotonic 1D integer mapping with 1 um resolution.
+  p = cand_pos[i]
+  s = (p[0] + p[1] + p[2]) * wp.static(1.0 / math.sqrt(3.0))
+  spatial_key = wp.clamp(wp.int64((s + 100.0) * 1000000.0), wp.int64(0), wp.int64(2147483647))
```
smallquail: This is float32 arithmetic on a value near 1e8 (`s ~ +/-1 m` gives `(s+100)*1e6 ~ 1.01e8`), where the ulp is 8. So the key is quantized to ~8 um, not 1 um, and the neighbour scan below breaks when keys differ by >= 1001. Two candidates 999 um apart along the projection can land 1016 apart in key space and never get compared, so a dominated duplicate survives. Mild consequence (an extra candidate, not a missing contact), but live with the shipped constants.

Suggest doing the offset in integer space: `wp.int64(s * 1e6) + wp.int64(100000000)`. The float product then stays under 2e6 (ulp 0.25) and the +100 m clamp range is unchanged.

mujoco_warp/_src/collision_flex.py
```diff
@@ -2131,6 +2323,8 @@ def _filter_flex_candidates_sorted(
   i = sort_val[si]
   my_key = sort_key[si]
   my_group = my_key >> wp.int64(32)
+  my_spatial = my_key & wp.int64(0x7FFFFFFF)
+  eps_key = wp.int64(epsilon * 1000000.0) + wp.int64(1)
```
smallquail: Exact today only because `1e-3 * 1e6` is an integer. With a fractional `epsilon * 1e6` the truncation makes the window one unit too small (e.g. 5.5 gives keys 11 and 5 for points 5.4 apart). `ceil(...) + 1` makes it safe for any epsilon.

mujoco_warp/_src/collision_flex.py
```diff
@@ -1266,6 +1443,11 @@ def kernel(
 
       worldelemid += nsweep_in
 
+      if wp.static(is_self):
+        word = flex_selfcollide_mask[elem1, elem2 >> 5]
```
smallquail: The mask encodes exactly the truth table of `_exclude_self_collision` (same vertex, or same body with `b >= 0`), which still runs a few lines below. So for self pairs the runtime check can never return True after this point (it is dead), and the mask has bought one 4-byte load in exchange for the table.

Either delete the runtime check on the self path (then the mask does something, and it should show up in a measurement) or delete the mask. Also, on "maintaining dynamic weld exclusions" in the description: there is no weld logic in this path; `_exclude_self_collision` does not read `body_weldid`.

mujoco_warp/_src/io.py
```diff
@@ -1044,6 +1047,55 @@ def _check_margin(name, t1, t2, margin):
   m.flex_vertflexid = flex_vertflexid
   m.flex_shelladr = flex_shelladr
 
+  # Precompute flex_selfcollide_mask: topologically adjacent elements sharing vertices/bodies
```
smallquail: This table is `nflexelem^2 / 8` bytes of model memory. On the benchmark cloth (1682 elements) it is 0.36 MB and builds in 0.06 s, so it is invisible there, but it is 0.8 GB at 200x200 and 4 GB at 300x300, for a per-element neighbour set of ~12. A per-element neighbour list would be a few hundred bytes per element. Worth an ablation of SAT-only vs SAT+mask on the benchmarks before committing to the quadratic form.

mujoco_warp/_src/io.py
```diff
@@ -1785,6 +1837,22 @@ def make_data(
     "map_iefc2efc": None,
     "dof_islandid": None,
     "efc_islandid": None,
+    # compact arrays
```
smallquail: What this does: keep the generic field loop below from eagerly allocating full `(nworld, nvmax_pad, nvmax_pad)` arrays for `cM`, `cqLD`, `cJ`, ... that `_allocate_compact_arrays` immediately replaces. That is a real fix for transient peak memory (gigabytes at 8192 worlds), but the commit message and PR bullet describe something else ("dynamically based on active DOFs"). Suggest saying what it does, and deriving the skip list from `_allocate_compact_arrays` instead of two hand-maintained lists of fifteen; a new `c*` field silently regresses otherwise.

mujoco_warp/_src/collision_flex.py
```diff
@@ -2647,13 +3061,29 @@ def _allocate_flex_workspace(m: Model, d: Data) -> FlexWorkspace:
     flex_group_start_indices = wp.full(nmax_groups, -1, dtype=int)
     flex_fps_min_dist = wp.empty(d.naconmax, dtype=float)
     flex_num_groups = wp.zeros(1, dtype=int)
+    fps_scratch_dist = wp.empty((nmax_groups, 64), dtype=float)
```
smallquail: `64` is hard-coded here, in the three `dim=(nmax_groups, 64)` launches, and as the stride and reduce width inside five kernels. One named constant.

mujoco_warp/_src/collision_flex.py
```diff
@@ -2462,113 +2668,313 @@ def _tie_break_fps(
...
     if cand_active_sorted[si] == 1:
-      total_active += 1
+      local_active += 1
+      c_idx = sort_val[si]
+      d_val = cand_dist[c_idx]
+      if d_val < min_d:
+        min_d = d_val
+        sel_cidx = c_idx
+      elif d_val == min_d:
+        if _tie_break_fps(c_idx, sel_cidx, cand_elem):
```
smallquail, line 2719: `_tie_break_fps` orders on `(elem1, elem2)` only, so candidates from the same element pair compare equal and the winner depends on scan order, which now differs from the serial kernel. Adding `curr_idx < sel_idx` as the final key makes it a strict total order and the parallel result bit-identical to the serial one regardless of partitioning.

mujoco_warp/_src/flex_test.py
```diff
@@ -1732,6 +1732,33 @@ def test_plane_cloth_no_fps_limit(self, nworld):
...
+      </mujoco>
+      """,
+      nworld=nworld,
+      nconmax=1500,
+    )
+    d.nacon.fill_(-1)
+    mjw.kinematics(m, d)
+    mjw.collision(m, d)
+    nacon = int(d.nacon.numpy()[0])
+    self.assertEqual(nacon, types.MJ_MAXCONPAIR * nworld)
```
smallquail, line 1762: This checks the cap, not the selection. Two cheap, strong tests would cover the new code properly: (a) on a self-colliding cloth, the contact set with the SAT filter equals the set without it (it is a pure prefilter, so any difference is a bug); (b) parallel FPS selects the same candidates as a serial/numpy reference.

benchmarks/cloth/__init__.py
```diff
@@ -2,17 +2,19 @@
   {
     "name": "cloth",
     "mjcf": "scene.xml",
-    "nworld": 32,
-    "nconmax": 2200,
-    "njmax": 8000,
+    "nworld": 256,
```
smallquail, line 5: The "Before" column ran with buffer overflow (dropped contacts mean less work *and* different physics), and this PR changes `nworld`, `nconmax`, `njmax`, and adds `nccdmax` in the same change. The 6-12x therefore mixes a code change, a batch-size change, and an overflow/no-overflow comparison. For a number that isolates the code: baseline with buffers raised until overflow is zero, same `nworld` as the optimized run, Jacobian setting stated.

smallquail (commented):
Summary in plain terms. The solver changes (equality-row pre-reduction, fused beta) are correct: I checked the algebra, the row ordering including islands, and the existing `constraints.xml` parity test covers them. The SAT prefilter and the parallel FPS are the right ideas and match the serial semantics. Three things to fix before merge, then nits:

1. **The new spatial sort key loses precision in float32** and can make the dedup scan stop early. Inline at the key computation.
2. **The adjacency mask is N^2 memory and the runtime check it replaces is still there**, so nothing is saved. Pick one. Inline at the sweep and at the build.
3. **The benchmark table doesn't isolate the code change**: the baseline ran with overflow and the configs changed in the same PR. Inline on the benchmark config.

Description accuracy: the "compact workspace" commit skips a wasted eager allocation rather than allocating "dynamically", and there are no weld exclusions in the self-collision path. Details inline.

quagla:
I posted some comments, nice speedups!

thowell:
the pr is updated

---

Below is a summary of the changes responding to each review comment.
 
### 1. Spatial Sort Key Precision
> "This is float32 arithmetic on a value near 1e8 (`s ~ +/-1 m` gives `(s+100)*1e6 ~ 1.01e8`), where the ulp is 8. So the key is quantized to ~8 um, not 1 um, and the neighbour scan below breaks when keys differ by >= 1001. Two candidates 999 um apart along the projection can land 1016 apart in key space and never get compared, so a dominated duplicate survives. Mild consequence (an extra candidate, not a missing contact), but live with the shipped constants.
>
> Suggest doing the offset in integer space: `wp.int64(s * 1e6) + wp.int64(100000000)`. The float product then stays under 2e6 (ulp 0.25) and the +100 m clamp range is unchanged."

Applied. In `mujoco_warp/_src/collision_flex.py`, moved the +100 m offset addition into integer space: `wp.clamp(wp.int64(s * 1000000.0) + wp.int64(100000000), wp.int64(0), wp.int64(2147483647))`. The float product remains below 2e6 where float32 ULP is < 0.25 um, eliminating the 8 um quantization.

### 2. Fractional Epsilon Safety in Neighbor Search
> "Exact today only because `1e-3 * 1e6` is an integer. With a fractional `epsilon * 1e6` the truncation makes the window one unit too small (e.g. 5.5 gives keys 11 and 5 for points 5.4 apart). `ceil(...) + 1` makes it safe for any epsilon."

Applied. In `mujoco_warp/_src/collision_flex.py`, updated the window threshold to `eps_key = wp.int64(wp.ceil(epsilon * 1000000.0)) + wp.int64(1)`.

### 3 & 4. Adjacency Bitmask Redundancy & Quadratic Memory Footprint
> "The mask encodes exactly the truth table of `_exclude_self_collision` (same vertex, or same body with `b >= 0`), which still runs a few lines below. So for self pairs the runtime check can never return True after this point (it is dead), and the mask has bought one 4-byte load in exchange for the table.
>
> Either delete the runtime check on the self path (then the mask does something, and it should show up in a measurement) or delete the mask. Also, on 'maintaining dynamic weld exclusions' in the description: there is no weld logic in this path; `_exclude_self_collision` does not read `body_weldid`."
>
> "This table is `nflexelem^2 / 8` bytes of model memory. On the benchmark cloth (1682 elements) it is 0.36 MB and builds in 0.06 s, so it is invisible there, but it is 0.8 GB at 200x200 and 4 GB at 300x300, for a per-element neighbour set of ~12. A per-element neighbour list would be a few hundred bytes per element. Worth an ablation of SAT-only vs SAT+mask on the benchmarks before committing to the quadratic form."

Removed the quadratic bitmask entirely. We ran an ablation measuring SAT-only vs. SAT+mask:
- Disabling the mask accelerated flex collision latency by 2.8% at 256 worlds (65.2 us -> 63.4 us) and 8.6% at 32 worlds (42.1 us -> 38.5 us), as avoiding 2D global memory loads across warps outweighed the 9-iteration register check in `_exclude_self_collision`.
- Model loading time dropped by 45% (0.16 s -> 0.09 s).
- Contact sets matched bit-for-bit with zero differences across all rollouts.
- Removing the table eliminates the O(N^2) memory cliff (saving ~4 GB at 300x300 resolution).

We cleaned up `Model` fields (`flex_selfcollide_mask`, `nflexelem_words`), removed the precomputation loop from `io.py`, kept `_exclude_self_collision` in the sweep.

### 5. Compact Solver Workspace Eager Allocation
> "What this does: keep the generic field loop below from eagerly allocating full `(nworld, nvmax_pad, nvmax_pad)` arrays for `cM`, `cqLD`, `cJ`, ... that `_allocate_compact_arrays` immediately replaces. That is a real fix for transient peak memory (gigabytes at 8192 worlds), but the commit message and PR bullet describe something else ('dynamically based on active DOFs'). Suggest saying what it does, and deriving the skip list from `_allocate_compact_arrays` instead of two hand-maintained lists of fifteen; a new `c*` field silently regresses otherwise."

Applied. Defined a single canonical `_COMPACT_DATA_FIELDS: tuple[str, ...]` tuple in `mujoco_warp/_src/io.py` adjacent to `_allocate_compact_arrays`. Both `make_data` and `put_data` now populate their initialization placeholders from this tuple (`**{name: None for name in _COMPACT_DATA_FIELDS}`), ensuring any future compact fields cannot silently regress into eager allocation.

### 6. FPS Block Size Named Constant
> "`64` is hard-coded here, in the three `dim=(nmax_groups, 64)` launches, and as the stride and reduce width inside five kernels. One named constant."

Applied. Defined `_FPS_BLOCK_SIZE: int = 64` at module scope in `mujoco_warp/_src/collision_flex.py` and replaced all 12 hard-coded occurrences across scratch buffer allocations, launch dimensions, and kernel loop strides / reductions.

### 7. Strict Total Order for FPS Tie-Breaking
> "`_tie_break_fps` orders on `(elem1, elem2)` only, so candidates from the same element pair compare equal and the winner depends on scan order, which now differs from the serial kernel. Adding `curr_idx < sel_idx` as the final key makes it a strict total order and the parallel result bit-identical to the serial one regardless of partitioning."

Applied. Appended `return curr_idx < sel_idx` as the final key in `_tie_break_fps` in `mujoco_warp/_src/collision_flex.py`. This enforces a strict total ordering `(distance, elem1, elem2, candidate_index)`, making the parallel reduction result bit-identical to serial reference scanning regardless of thread block partitioning.

### 8. Unit Tests for SAT Prefilter and Parallel FPS Parity
> "This checks the cap, not the selection. Two cheap, strong tests would cover the new code properly: (a) on a self-colliding cloth, the contact set with the SAT filter equals the set without it (it is a pure prefilter, so any difference is a bug); (b) parallel FPS selects the same candidates as a serial/numpy reference."

Applied. In `mujoco_warp/_src/flex_test.py`:
- Added `d.contact.dist.fill_(wp.inf)` alongside `d.nacon.fill_(-1)` in `test_flex_fps_capping`.
- Added `test_flex_sat_prefilter_conservative`: runs collision on a self-colliding cloth with SAT enabled vs. disabled (`enable_sat=True/False`). Asserts that SAT culls candidate pairs (`ncollision` is strictly reduced) while producing an exact bit-identical contact set (`pos`, `dist`) after sorting.
- Added `test_parallel_fps_numpy_parity`: runs the parallel FPS collision pipeline on overlapping cloths and compares selections against a pure NumPy reference implementing greedy farthest point sampling with the strict `(dist, elem1, elem2, idx)` tie-breaker. Asserts 100% bit-identical candidate selection across all groups.

mujoco_warp/_src/flex_test.py
```diff
@@ -1732,6 +1733,191 @@ def test_plane_cloth_no_fps_limit(self, nworld):
...
+        selected.append(best_cand)
+        new_pos = cand_pos[best_cand]
+        for c_idx in group_cands:
+          d2 = np.sum((cand_pos[c_idx] - new_pos) ** 2)
+          if d2 < min_dist[c_idx]:
+            min_dist[c_idx] = d2
+
+      warp_selected = sorted([c_idx for c_idx in group_cands if cand_active[c_idx] == 1])
+      np_selected = sorted(selected)
+      self.assertEqual(warp_selected, np_selected)
```
smallquail, line 1929: The reference does not compute what the kernel computes, so exact equality cannot hold on this scene:

- The reference ranks by **squared** distance (`np.sum((a-b)**2)`, float32, numpy summation order); the kernels rank by `wp.length(...)` (float32 `sqrt` of `x*x+y*y+z*z`). Same ordering in exact arithmetic, different rounding on ties.
- Ties are decided by exact float equality in both, and this scene, an 8x8 lattice stacked exactly 1 cm over an identical lattice, has many candidates at geometrically identical distances. One tie breaks differently (312 vs 484 in the CI log) and the greedy chain diverges from there.
- The reference also has no `max_d <= 0.0` stop, which the kernel has in `_parallel_fps_resolve_max`; on coincident positions the two would select different counts.

Suggest: perturb the fixture so exact ties cannot occur (`qpos_noise=` on `test_data.fixture`), mirror the kernel's `sqrt` and stop rule in the reference, and keep the exact-match assertion. With a strict total order on the tie-break and no exact ties, bit-identical selection is a sound invariant.

mujoco_warp/_src/collision_driver.py
```diff
@@ -886,6 +886,7 @@ def collision(
   m: Model,
   d: Data,
   awake_prev: Optional[wp.array] = None,
+  enable_sat: bool = True,
```
smallquail: This keyword is on the public `collision()` signature only so `test_flex_sat_prefilter_conservative` can toggle the filter. A module-level flag in `collision_flex.py` that the test patches would keep it off the API surface.

smallquail (commented):
Thanks: all seven code items land cleanly, and settling the mask question with an ablation was exactly the right call.

1. **`test_parallel_fps_numpy_parity` is failing on CI, and the test is what is wrong**, not the kernel. Details inline on the assertion.
2. **PR description is stale.** Bullet 4 still describes the removed adjacency bitmask and "dynamic weld exclusions".
3. **Benchmark table is still not apples-to-apples** (unchanged from the earlier comment on `benchmarks/cloth/__init__.py`): the baseline ran through overflow, and `nworld`, `nconmax`, `njmax` and `nccdmax` changed in the same PR. One added row, with the baseline buffers raised until overflow is zero, the same `nworld` as the optimized run, and the Jacobian setting stated, would make the headline number defensible.
4. Nit inline on `collision()`: the new `enable_sat` keyword exists only for the test.

quagla:
Left some minor comments.

smallquail:
**Request: mirror the contact-selection changes in MuJoCo C, so the two implementations keep producing the same contacts.**

The motivation is behavior preservation. MuJoCo C and mjWarp should return the same flex contact set from the same state, and today the place where they can differ is the per-pair contact cap (`filterFlexContacts` in `engine_collision_driver.c` on the C side, the FPS stage in `collision_flex.py` here). This PR makes the warp side deterministic; without a matching change in C the two selections diverge on ties and stay diverged.

Three items:

1. **Tie-break order.** `_tie_break_fps` now imposes a strict total order `(elem1, elem2, candidate index)`. C's `filterFlexContacts` uses strict `>` in both the deepest-seed scan and the farthest-point scan, so ties resolve to whichever contact was emitted first by the sweep, an ordering that is neither semantic nor stable. Adopting the same `(elem1, elem2, index)` key in C would make the two selections agree on lattice and symmetric-drape scenes, where exact ties are common.

2. **Proximity dedup before the cap.** Warp merges candidates within 1 mm (`_filter_flex_candidates_sorted`, `epsilon = 1e-3`) before selecting; C has no cross-pair dedup (only the per-pair witness dedup in `mjc_penetration`), so the cap sees different inputs on the two sides and keeps different survivors. Either side can move; the ask is a decision, recorded in both places, rather than the current silent difference.

3. **Record the SAT decision in C.** The SAT prefilter added here was evaluated in C at the `mjc_ConvexElem` call in `mj_collideElems`: contact sets identical (0 rejected-but-contacting pairs over 15.8 M), step time unchanged (8718 vs 8781 us on the cloth benchmark), because GJK's reject path already costs ~0.16 us/pair, the same as the SAT. Not worth porting, but worth a two-line comment at that spot in C saying so, with the numbers, so the structural difference is deliberate and nobody re-measures it.

Items 1 and 2 are the ones that affect physics; item 3 is bookkeeping.

quagla:
I left one request regarding C mujoco, I think we should make sure we don't deviate in behavior between the two.

quagla (approved):

thowell:
updated the pr description including updated benchmark results comparing the before and after commits with modified settings to avoid overflow in order to directly compare performance. results with the updated settings for the latest pr commit are also provided.

PR #1628, opened 2026-09-01
Add `Option.run_rne_postconstraint` and decouple rne_postconstraint from sensor stage

### LINKED ISSUE #1625: Decouple rne_postconstraint from the disabled sensor stage for non-sensor consumers
### Problem

MuJoCo Warp currently calls `smooth.rne_postconstraint()` only inside `sensor_acc()`, after the `mjDSBL_SENSOR` early return:

```text
solver.solve
sensor_acc
  if mjDSBL_SENSOR: return
  ...
  rne_postconstraint
integration
```

Sources: [`forward.py`](https://github.com/google-deepmind/mujoco_warp/blob/7ec1b57f48ce5bec0b1b2982c2ca48d5186044a3/mujoco_warp/_src/forward.py#L1401-L1423), [`sensor.py`](https://github.com/google-deepmind/mujoco_warp/blob/7ec1b57f48ce5bec0b1b2982c2ca48d5186044a3/mujoco_warp/_src/sensor.py#L2522-L2675), [`smooth.py`](https://github.com/google-deepmind/mujoco_warp/blob/7ec1b57f48ce5bec0b1b2982c2ca48d5186044a3/mujoco_warp/_src/smooth.py#L1743-L1825).

This is sufficient when only enabled MuJoCo sensors consume `cacc`, `cfrc_ext`, and `cfrc_int`. It does not support non-sensor consumers. Newton, for example, requests the same RNE outputs for its extended state attributes `body_qdd` and `body_parent_f`. It currently sets `Model.sensor_rne_postconstraint`, but when `mjDSBL_SENSOR` is set the call remains unreachable, so Newton converts stale arrays every step. The downstream silent-stale-state bug is tracked in [newton-physics/newton#4109](https://github.com/newton-physics/newton/issues/4109).

This blocks an otherwise valid separation in deterministic Isaac Lab runs: MuJoCo's internal sensors need to remain disabled because their kernels use atomics that fall outside the deterministic atomic-record budget, while Isaac Lab's IMU, PVA, and joint-wrench sensors are separate consumers that read Newton state — which is itself filled by RNE — rather than MuJoCo `sensordata`. They could keep working if Newton received fresh RNE-derived state.

### Potential fix

Make `rne_postconstraint` a general, optional post-constraint computation rather than scheduling it exclusively as sensor work. Provide a static/CUDA-Graph-safe request for non-sensor consumers and run RNE exactly once after `solver.solve()` and before integration when any active consumer needs it:

```text
solver.solve
optional general rne_postconstraint
sensor_acc
integration
```

The request/scheduling contract should ensure:

- `mjDSBL_SENSOR` still skips all sensor measurement work.
- An external consumer can request fresh `cacc`, `cfrc_ext`, and `cfrc_int` while sensors are disabled.
- Sensors enabled and requiring RNE continue to work without a duplicate RNE launch.
- Sensors disabled with no external request do not trigger otherwise-unused RNE work.
- The new path remains CUDA Graph compatible and is tested under the intended deterministic mode.

Simply moving the current `m.sensor_rne_postconstraint` call above the disable check would not fully satisfy this contract: that flag may be true only because the model contains an RNE-dependent sensor, even though that sensor is disabled and no external consumer needs the result.

With a general request path, Newton can publish fresh extended state while MuJoCo sensors remain disabled, and downstream Isaac Lab sensors can operate in deterministic runs without the current incompatibility guard in [isaac-sim/IsaacLab#7334](https://github.com/isaac-sim/IsaacLab/pull/7334).

thowell:
### Add `Option.run_rne_postconstraint` and decouple RNE post-constraint from sensor stage

Fixes #1625

## Problem

In MuJoCo, `mj_rnePostConstraint` evaluates body spatial accelerations (`d.cacc`), internal forces (`d.cfrc_int`), and external forces (`d.cfrc_ext`). In MuJoCo Warp, this was previously called exclusively inside `sensor.sensor_acc(m, d)`.

However, if sensors are disabled (`m.opt.disableflags & DisableBit.SENSOR`) or if a model contains no acceleration-dependent sensors (`not m.sensor_rne_postconstraint`), `smooth.rne_postconstraint(m, d)` was completely skipped. Downstream consumers (such as Newton / Isaac Lab in Newton issue #4109) rely on `cacc`, `cfrc_ext`, and `cfrc_int` for spatial acceleration and contact force computations in deterministic simulations where sensors are intentionally disabled. Without running `rne_postconstraint`, these fields remain unpopulated or stale.

## Solution

1. **Add `Option.run_rne_postconstraint: bool`**:
   - Added `run_rne_postconstraint: bool = False` to `mjw.Option` and initialized it in `put_model()`.
   - Added `"opt.run_rne_postconstraint"` to `mjw_only_fields` in `io.override_model` so it can be configured via model overrides.

2. **Decouple Post-Constraint RNE in `forward` & `step2`**:
   - In `forward.py`, evaluate `smooth.rne_postconstraint(m, d)` immediately following `solver.solve(m, d)` when:
     ```python
     if m.opt.run_rne_postconstraint or (not (m.opt.disableflags & DisableBit.SENSOR) and m.sensor_rne_postconstraint):
       smooth.rne_postconstraint(m, d)
     ```
   - Then call `sensor.sensor_acc(m, d, skip_rne_postconstraint=True)`.

3. **Update `sensor_acc` API with `skip_rne_postconstraint`**:
   - Added `skip_rne_postconstraint: bool = False` parameter to `sensor.sensor_acc(m, d)`.
   - When `skip_rne_postconstraint=True`, `sensor_acc` skips internal evaluation of `smooth.rne_postconstraint`.
   - Standalone and inverse dynamics invocations default to `False`, preserving backwards-compatible behavior.

## Testing

- Added `test_sensor_acc_skip_rne_postconstraint` in `mujoco_warp/_src/sensor_test.py` to verify unit-level behavior of `skip_rne_postconstraint` and ensure `m.opt.run_rne_postconstraint` defaults to `False`.
- Added `test_run_rne_postconstraint` in `mujoco_warp/_src/forward_test.py` to verify that when `DisableBit.SENSOR` is active and `run_rne_postconstraint=True`, `cacc`, `cfrc_int`, and `cfrc_ext` match upstream MuJoCo across `forward()` and `step2()` while `sensordata` remains empty, and that setting `run_rne_postconstraint=False` bypasses computation.

### DIFF
--- a/mujoco_warp/_src/forward.py
+++ b/mujoco_warp/_src/forward.py
@@ -1405,7 +1405,9 @@ def forward(m: Model, d: Data):
   fwd_acceleration(m, d, factorize=True)
 
   solver.solve(m, d)
-  sensor.sensor_acc(m, d)
+  if m.opt.run_rne_postconstraint or (not (m.opt.disableflags & DisableBit.SENSOR) and m.sensor_rne_postconstraint):
+    smooth.rne_postconstraint(m, d)
+  sensor.sensor_acc(m, d, skip_rne_postconstraint=True)
 
 
 @event_scope
@@ -1448,7 +1450,9 @@ def step2(m: Model, d: Data):
   fwd_actuation(m, d)
   fwd_acceleration(m, d)
   solver.solve(m, d)
-  sensor.sensor_acc(m, d)
+  if m.opt.run_rne_postconstraint or (not (m.opt.disableflags & DisableBit.SENSOR) and m.sensor_rne_postconstraint):
+    smooth.rne_postconstraint(m, d)
+  sensor.sensor_acc(m, d, skip_rne_postconstraint=True)
 
   # integrate with Euler or implicitfast
   if m.opt.integrator in (IntegratorType.IMPLICITFAST, IntegratorType.IMPLICIT):
--- a/mujoco_warp/_src/forward_test.py
+++ b/mujoco_warp/_src/forward_test.py
@@ -2053,6 +2053,53 @@ class DCMotorTest(parameterized.TestCase):
 
     np.testing.assert_allclose(d.qvel.numpy()[0], mjd.qvel, atol=1e-3, rtol=1e-3)
 
+  def test_run_rne_postconstraint(self):
+    """Tests run_rne_postconstraint option in forward and step2."""
+    mjm, mjd, m, d = test_data.fixture(
+      xml="""
+      <mujoco>
+        <worldbody>
+          <body name="body0" pos="0 0 1">
+            <freejoint/>
+            <geom type="sphere" size=".1" mass="1"/>
+            <site name="site0"/>
+          </body>
+        </worldbody>
+        <sensor>
+          <accelerometer site="site0"/>
+        </sensor>
+      </mujoco>
+      """,
+      overrides={"opt.disableflags": DisableBit.SENSOR, "opt.run_rne_postconstraint": True},
+    )
+    self.assertTrue(m.opt.run_rne_postconstraint)
+    mujoco.mj_forward(mjm, mjd)
+    mujoco.mj_rnePostConstraint(mjm, mjd)
+
+    # 1. run_rne_postconstraint=True: RNE runs in forward and step2, sensors skipped
+    d.cacc.fill_(wp.inf)
+    d.cfrc_int.fill_(wp.inf)
+    d.cfrc_ext.fill_(wp.inf)
+    mjw.forward(m, d)
+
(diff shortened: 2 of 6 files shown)


sylvesterkaczmarek (commented):
The split looks consistent. forward()/step2() now run RNE exactly when the explicit option requests it or enabled sensors need it, and skip_rne_postconstraint=True prevents the old sensor path from running it twice. Standalone sensor_acc() keeps the old default behavior. The disabled-sensor regression covers the new use case directly. I don't see a blocker.

yuvaltassa (approved):

PR #1629, opened 2026-09-01
updates for delays and history

thowell:
This PR refines delay and history handling to support batched delay fields per world and align history buffer initialization and reset behavior with MuJoCo C.

Previously, delay and history arrays assumed a single model representation across worlds. Kernels now use 2D modulo indexing (`worldid % shape[0]`) across actuator and sensor delay, interval, and history buffers so that per-world variations can be specified with `batch_sizes`. In addition, `init_sensor_history` previously defaulted `phase` to zeros when omitted, which clobbered the user slot storing `time_prev` for interval sensors and broke subsequent interval scheduling; it now preserves the existing user slot when `phase is None` and accepts either a scalar float or a 1D Warp array. `init_ctrl_history` and `init_sensor_history` now also validate IDs, buffer allocations, and array shapes on the host before kernel launch, and `reset_history` is exported in the top-level package.

* Support per-world batched delay, interval, and history buffer arrays.
* Preserve interval sensor user slot (`time_prev`) in `init_sensor_history` when `phase is None`, and support scalar float phase broadcasting.
* Add host-side ID, buffer allocation, and array shape validation to `init_ctrl_history` and `init_sensor_history`.
* Export `reset_history` in `mujoco_warp`.
* Expand test coverage for batched delay resets, selective reset masks, interval sensor user slot preservation, and input validation against MuJoCo C.


yuvaltassa (approved):
Fixed a bunch of latent bug too 🙂 
Nice one, LGTM

PR #1630, opened 2026-09-01
Fix sparse joint-limit capacity estimate

camevor:
Issue: The equality operator of the `mujoco._enums.mjtJoint` enum evaluates to `False` for numpy types. This can lead to underestimation of the maximum number of constraint nonzeros in `_default_njmax_nnz`, as the following check evaluates to `False` since the enum's equality operator is used:

```python
     elif jnt_type in (mujoco.mjtJoint.mjJNT_SLIDE, mujoco.mjtJoint.mjJNT_HINGE):
```

This change converts `jnt_type` to native int before the membership test.

Example with MuJoCo 3.12.1:
```python
import mujoco

model = mujoco.MjModel.from_xml_string(
    """
    <mujoco>
      <worldbody>
        <body>
          <joint type="hinge"/>
          <geom size="0.1"/>
        </body>
      </worldbody>
    </mujoco>
    """
)

array_value = model.jnt_type[0]
enum_value = mujoco.mjtJoint.mjJNT_HINGE

print(type(array_value), type(enum_value))
# <class 'numpy.int32'> <class 'mujoco._enums.mjtJoint'>
print(array_value == enum_value)
# True
print(enum_value == array_value)
# False
print(array_value in (enum_value,))
# False
print(int(array_value) in (enum_value,))
# True
```

### DIFF
--- a/mujoco_warp/_src/io_test.py
+++ b/mujoco_warp/_src/io_test.py
@@ -512,8 +512,8 @@ class IOTest(parameterized.TestCase):
         <option jacobian="sparse"/>
         <worldbody>
           <body>
-            <joint type="slide" limited="true" range="-1 1" frictionloss="1"/>
-            <joint type="hinge" limited="true" range="-1 1" frictionloss="1"/>
+            <joint type="slide" limited="true" range="-1 1"/>
+            <joint type="hinge" limited="true" range="-1 1"/>
             <geom size="0.1" contype="0" conaffinity="0"/>
           </body>
         </worldbody>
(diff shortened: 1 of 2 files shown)


mujoco_warp/_src/io_test.py
```diff
@@ -504,6 +504,25 @@ def test_augmented_cholesky_padding(self, nv, expected):
...
 
+  def test_default_njmax_nnz_includes_scalar_joint_limits(self):
+    """Include slide and hinge limits in the sparse Jacobian capacity."""
+    mjm = mujoco.MjModel.from_xml_string(
+      """
+      <mujoco>
+        <option jacobian="sparse"/>
+        <worldbody>
+          <body>
+            <joint type="slide" limited="true" range="-1 1" frictionloss="1"/>
```
thowell: can we remove frictionloss in this line and the line below? we would need to update the test to check for 2 instead of 4?

thowell:
thanks @camevor! one comment otherwise lgtm.

mujoco_warp/_src/io_test.py
camevor: Sure! removed in 6a24d588

thowell (approved):

PR #1553, opened 2026-07-23
aloha clutter vision benchmark

thowell:
adds a new benchmark: aloha clutter vision

4 cameras from mujoco menagerie [aloha/scene.xml](https://github.com/google-deepmind/mujoco_menagerie/blob/main/aloha/aloha.xml) added to the scene_clutter.xml. rgb, depth, and segmentation 64x64 images are rendered.

aloha_clutter_vision_rgb_2x2

aloha_clutter_vision_depth_2x2

aloha_clutter_vision_segmentation_2x2

this pr builds on #1375

---

This report presents performance, throughput, memory utilization, and timing breakdown for the `aloha_clutter_vision` benchmark across parallel environment counts (`nworld` = 1, 32, 64, 256, 512, 1024).

## 1. Overview & Setup

- **GPU Device**: NVIDIA RTX 6000 Ada Generation (48 GiB VRAM, Driver 550.163.01)
- **Benchmark Name**: `aloha_clutter_vision`
- **Model File**: `scene_clutter.xml`
- **Functions Benchmarked**: `["step", "refit_bvh", "render"]`
- **Active Cameras (4)**: `overhead_cam`, `worms_eye_cam`, `wrist_cam_left`, `wrist_cam_right`
- **Render Resolution**: 64 x 64 pixels per camera
- **Render Modalities**: RGB (True), Depth (True), Segmentation (True), Shadows (True)
- **Physics Settings**: `nconmax=512`, `nccdmax=64`, `njmax=1024`, `nvmax=64`, `override="opt.enableflags=SLEEP"`, `init_asleep=True`

---

## 2. Benchmark Results Summary

| `nworld` | Steps / Sec (SPS) | Step Time (µs) | Refit BVH (µs) | Render Time (µs) | Total Memory (GiB) | Data Memory (GB) |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| **1** | **411** | 1,654.7 | 14.0 | 748.1 | 0.561 GiB | 0.0008 GB |
| **32** | **3,834** | 75.0 | 0.60 | 184.7 | 0.623 GiB | 0.024 GB |
| **64** | **4,550** | 41.4 | 0.35 | 177.8 | 0.654 GiB | 0.047 GB |
| **256** | **5,265** | 13.6 | 0.11 | 176.1 | 1.029 GiB | 0.189 GB |
| **512** | **5,442** | 8.2 | 0.07 | 175.4 | 1.529 GiB | 0.378 GB |
[...]

PR #1632, opened 2026-09-01
Fix create_render_context rejecting an empty cam_active

kevinzakka:
`create_render_context` infers the format of `cam_active` from its first element, and every branch is guarded on the list being non-empty:

```python
if len(cam_active) > 0 and isinstance(cam_active[0], (bool, np.bool_)):
  ...
elif len(cam_active) > 0 and isinstance(cam_active[0], str):
  ...
elif len(cam_active) > 0 and isinstance(cam_active[0], (int, np.integer)):
  ...
else:
  raise ValueError(f"Invalid cam_active format: {cam_active}")
```

An empty list matches nothing and falls through to the error:

```
ValueError: Invalid cam_active format: []
```

A length-`ncam` boolean mask is empty exactly when the model has no cameras, which is routine for callers that want a render context only for its BVH and never render a frame. Raycast sensors are one such caller: on a camera-free model, `[False] * mjm.ncam` is `[]`. This worked before #1553 added the name and index forms.

This change handles the empty list up front as an empty selection, which makes the `len(cam_active) > 0` guard on the remaining branches redundant. An empty list is now also accepted when the model does have cameras, where the length assert previously rejected it.

`test_cam_active_empty` covers both cases and fails on main with the `ValueError` above. A context with `nrender == 0` still drives `refit_bvh` and BVH-accelerated `rays`, which read only BVH fields and no camera-derived state.

### DIFF
--- a/mujoco_warp/_src/render_util.py
+++ b/mujoco_warp/_src/render_util.py
@@ -358,6 +358,7 @@ def create_render_context(
     enabled_geom_groups: The geom groups to render.
     cam_active: List of booleans, camera names (str), or camera indices (int) indicating
                 which cameras to include in rendering. If None, all cameras are included.
+                An empty list includes no cameras.
     flex_render_smooth: Whether to render flex meshes smoothly.
     use_precomputed_rays: Use precomputed rays instead of computing during rendering.
                           When using domain randomization for camera intrinsics, set to False.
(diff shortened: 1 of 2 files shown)


thowell (approved):
nice! thanks @kevinzakka

PR #1633, opened 2026-09-01
multiccd update

thowell:


kbayes:
@thowell thanks! #1651 fixed the underlying issue which had to do with edge contact. However this PR gives me an idea that we might be able to kill out some multiccd memory buffers by using the contact direction to pick the best face normals, instead of looping through the relevant ones. Feel free to close

Issue #1590, opened 2026-08-18
Deterministic execution blocked: _sensor_tactile mixes atomic_max and atomic_add on one array


NY-WaKeUp:
# Deterministic execution blocked: `_sensor_tactile` mixes `atomic_max` and `atomic_add` on one array

## Summary

No model can be simulated under Warp's deterministic execution mode. `_sensor_tactile`
writes the normal channel with `wp.atomic_max` and the two shear channels with
`wp.atomic_add`, all into `sensordata_out`, which violates Warp's documented
["one reduction family per target"](https://nvidia.github.io/warp/v1.15/user_guide/deterministic_execution.html)
rule, so the module fails to compile.

This is not limited to models that use tactile sensors: the kernel is compiled with the
rest of the module regardless of whether the model instantiates one. Reproduced on a
model with `nsensortaxel == 0`.

## Reproduction

```python
import warp as wp
wp.config.deterministic = wp.DeterministicMode.RUN_TO_RUN  # must precede the import
import mujoco_warp
# build any model and step it
```

```
warp._src.codegen.WarpCodegenError: Error while parsing function "_sensor_tactile"
  wp.atomic_add(sensordata_out, worldid, sensor_adr[sensor_id] + 1 * dim + vertid, forceT[1])
Deterministic mode does not support mixing 'max' and 'add' reductions on array
'sensordata_out' in the same kernel
```

Only reproduces on a cold kernel cache. With a warm cache, modules compiled before the
option was set are reused, so the run appears to succeed while silently not being
deterministic — worth knowing when testing this.

Versions: warp-lang 1.16.0, mujoco-warp 3.11.0, mujoco 3.11.0, CUDA 12.9, driver 13.0,
RTX 4090 (sm_89).

## Cause

`_src/sensor.py`, in `_sensor_tactile`:

```python
dim = sensor_dim[sensor_id] // 3
wp.atomic_max(sensordata_out, worldid, sensor_adr[sensor_id] + 0 * dim + vertid, forceT[0])
wp.atomic_add(sensordata_out, worldid, sensor_adr[sensor_id] + 1 * dim + vertid, forceT[1])
wp.atomic_add(sensordata_out, worldid, sensor_adr[sensor_id] + 2 * dim + vertid, forceT[2])
```

The reduction choices are the right physics — peak normal force wants `max`, shear
components want `add` — but the three channels share one array and Warp analyses
reduction families per target array, not per index range.

## Suggested fix

[...]

PR #1636, opened 2026-09-02
Fix convex narrowphase contact loss under Warp deterministic mode

### LINKED ISSUE #1635: Warp deterministic mode silently drops all convex narrowphase contacts (bodies fall through geometry)
Under Warp's deterministic modes, every geom pair that MuJoCo Warp routes to the convex narrowphase produces zero contacts, silently: `mj_step` returns normally, no warning is printed, and bodies fall straight through geometry. Pairs routed to the primitive narrowphase are unaffected.

Minimal repro (a plain box resting on a plain box, no mesh assets, no sensors):

```python
import warp as wp

wp.config.deterministic = wp.DeterministicMode.RUN_TO_RUN
wp.config.deterministic_max_records = 64

import mujoco
import mujoco_warp as mjw

XML = """
<mujoco>
  <worldbody>
    <geom type="box" pos="0 0 .1" size="1 1 .1"/>
    <body pos="0 0 .29">
      <freejoint/>
      <geom type="box" size=".1 .1 .1"/>
    </body>
  </worldbody>
</mujoco>
"""

mjm = mujoco.MjModel.from_xml_string(XML)
mjd = mujoco.MjData(mjm)
mujoco.mj_forward(mjm, mjd)
m = mjw.put_model(mjm)
d = mjw.put_data(mjm, mjd, nworld=1)
mjw.collision(m, d)
print(int(d.nacon.numpy()[0]))  # 0; the same script without the two wp.config lines prints 4
```

Left to step, the small box sinks through the large one (final z -4.58 after 200 steps, against 0.30 resting in default mode), with exit code 0 throughout. Without `deterministic_max_records` raised, the first symptom is a loud `Deterministic scatter buffer overflow` RuntimeError; raising it past the default converts the failure into the silent form. Reproduced at current main (595bd6f, and 3879591 before it) on two GPU architectures — an RTX 3090 Ti (sm_86) and a GTX 1630 (sm_75), which produce the same final z to the digit — on warp-lang 1.15.0 and 1.16.0, and on the mujoco-warp 3.11.0 wheel.

The dispatch table predicts the failure exactly. In a six-pair sweep (each pair with a default-mode control), every pair `MJ_COLLISION_TABLE` routes to the primitive narrowphase reproduces its default-mode trajectory bit for bit under RUN_TO_RUN, and every pair it routes to the convex narrowphase drops to zero contacts: sphere/plane and sphere/box and mesh/plane pass; mesh/box, mesh/mesh, and box/box all fail. Broadphase is not involved: `d.ncollision` hands the narrowphase the same candidate-pair count in both modes, and `d.overflow` stays clean.

Mechanism, read from Warp's source rather than inferred: Warp runs slot-allocating kernels twice, counting reservations on the first pass and replaying them on the second, and suppresses all array stores during the counting pass (`warp/_src/codegen.py:5288`; the guard is kernel-wide with no per-array opt-out, `warp/_src/deterministic.py` `needs_store_guard`). EPA keeps its polytope in global scratch (`epa_vert` and friends, sized `(naccdmax, 10 + 2 * epa_iterations)`), so the counting pass reads back scratch it was forbidden to write, takes different branches than the replay, and reserves a contact count the replay disagrees with. The gate that matters is the `dist >= gap` early return after the EPA solve in `ccd_kernel`: on the counting pass `dist` is computed from unwritten scratch, the thread leaves the kernel before reserving, and the prefix scan allocates zero slots. `ccd_hfield_kernel` has the same shape and the same measured failure (a box sinks through a heightfield terrain). The silence is Warp's documented capture-time behavior (overflow checks are disabled during graph capture and replay), so nothing on this path can warn.

Related: #562 is the determinism umbrella; #1422/#1425/#1533 build an `opt.deterministic` inside MuJoCo Warp itself and touch none of `collision_convex.py`, so this is a disjoint failure mode that affects anyone who reaches for Warp's own `wp.config.deterministic` today.

We have a fix (a `deterministic` build flag on the two CCD kernel builders that splits contact writing into a second kernel reading staged, previous-launch memory, so counting and replay agree; the default path compiles unchanged) with a regression test; PR to follow shortly.

wagner-austin:
Under Warp's deterministic modes (`wp.config.deterministic = RUN_TO_RUN` or `GPU_TO_GPU`), every geom pair the collision table routes to the convex narrowphase silently produces zero contacts, and bodies fall through geometry with exit code 0. Warp runs slot-allocating kernels twice, counting reservations before replaying them, and suppresses all array stores during the counting pass; EPA keeps its polytope in global scratch, so the counting pass reads back values it was never allowed to write, takes different branches than the replay, and the `dist >= gap` gate after the EPA solve returns before any contact slot is reserved. The heightfield CCD kernel has the same shape and the same measured failure. Full analysis in the linked issue.

This change gives `ccd_kernel_builder` and `ccd_hfield_kernel_builder` a `deterministic` build flag, following the `warn_overflow` precedent. When set, the CCD kernels stop reserving contact slots themselves and instead stage each candidate pair's solved results into buffers indexed by `collisionid` (atomic-free, stable), and a second kernel reserves and writes contacts by reading that staged, previous-launch memory, so the counting and replay passes agree exactly. The contact-writing tail is factored into a shared `write_ccd_contacts` function used by both paths. `convex_narrowphase` reads `wp.config.deterministic` at build time to pick the variant, which is correct because Warp binds a module's determinism mode at first compile.

Tradeoffs and disclosures, so nothing is discovered in review:

* The default path compiles to the same kernels as before and pays nothing; the staging buffers collapse to size one when the flag is off, and are sized by `naconmax` when it is on.
* The new test runs its deterministic half in a subprocess. Warp fixes a module's determinism mode when the module is first compiled, so an in-process before/after comparison silently tests nothing. There is no subprocess precedent in this suite; happy to reshape it if you prefer another mechanism.
* **Using these modes requires raising `wp.config.deterministic_max_records`, and this change alone is not sufficient without it.** The default of 0 means the code-generated lower bound, which the CCD narrowphase exceeds: eagerly, the fixture in this PR's own test raises `Deterministic counter buffer overflow in kernel 'ccd_kernel_builder__locals__ccd_write_kernel'`, which is why the test sets the bound to 64. The under-count is not confined to the convex path — humanoid at nworld=256 reports `ncon_mean` 9.074 at the default against 9.243 non-deterministic.
* **Under CUDA graph capture that overflow is silent, which is the dangerous case.** Inside a capture the host-side readback cannot run, so the condition that raises eagerly raises nothing and contacts simply go missing — the same failure shape this PR fixes, reached by a different route. `wp.config.deterministic_debug = True` exposes it as `Warp deterministic scatter overflow: capacity=1024 dest=...`, but debug is off by default.
* Cost depends almost entirely on whether CUDA graphs are captured, so both regimes are given. Without capture, `RUN_TO_RUN` costs 2-2.7x per step on the box-on-box fixture, and the same factor applies to primitive-only scenes that never touch the new kernel, so it is Warp's two-pass replay cost rather than a cost of this split. With capture — which is what `mjwarp-testspeed` and production use — capture removes per-launch overhead from the baseline but not the replay cost, and the same comparison on humanoid at nworld=256, against a baseline median of 437,881 steps/s over 7 reps, is 64.8x.
* `collision_flex` and `collision_sdf` were both checked and are already sound; neither is modified here, and the two rest on different strengths of evidence. Flex is **inspection only**: its `nacon` reservation sits in `_write_filtered_contacts`, whose `cand_dist` gate is a read-only input produced by an earlier launch, so counting and replay cannot branch differently. I could not measure it — all four in-tree flex fixtures report zero contacts even with determinism off, so they would have agreed under any configuration and prove nothing. For SDF, `gradient_descent` writes to no global array, and `write_contact` derives `active`/`detected` from passed-in scalars and takes its early return before `wp.atomic_add`, so the counting and replay passes cannot branch differently. Measured as well as argued: all eight in-tree SDF fixtures report identical contact counts with determinism off, under `RUN_TO_RUN` with `deterministic_max_records=64`, and under `RUN_TO_RUN` at the default bound (40/1/28/28/34/0/39/39 in every configuration). SDF is unaffected by the record bound because it reserves once per thread rather than behind a data-dependent loop.

Verification, on an RTX 3090 Ti (sm_86) with the box-on-box cells replicated on a GTX 1630 (sm_75):

* `uv run pytest -n 8` passes with one pre-existing failure (`test_collision20`, which fails identically on untouched main).
* The new `test_deterministic_convex_narrowphase` passes with the fix and fails 0 != 4 with the flag forced off (negative control observed, not assumed).
* All seven geom-pair fixtures report identical contact totals and final heights between default and deterministic modes, to the digit; the three convex-routed pairs were at zero contacts before the fix.
* A six-box coupled pile stepped 150 times gives four distinct state digests over four default-mode runs, and one digest across six rollouts in two separate processes under `RUN_TO_RUN`, which is the property the mode exists to provide. An eight-world batch reproduces the same way.
* On the second architecture (GTX 1630), unpatched main under `RUN_TO_RUN` reproduces the defect with the same final height to the digit (box falls to z = -4.58), and this branch under `RUN_TO_RUN` matches the default trajectory exactly (720 contacts over 200 steps, final z 0.299892).
* `GPU_TO_GPU` verified cross-architecture, which the earlier revision of this description listed as untested. One box-on-box pair at offset (10, 20, 0.1), same commit, warp-lang 1.15.0, mujoco 3.12.1 and driver 591.86 on both machines, every float printed at `%.17g`: an RTX 3090 Ti (sm_86) and a GTX 1630 (sm_75) emit byte-identical output — 4 contacts, 16 floats, no divergence in the last bit, compared mechanically rather than by eye. The offset is the point: it is exactly where a missing pair-local origin restore would show.
* Deterministic-mode kernel compilation is expensive on smaller hardware, which is worth stating so it is not mistaken for a hang: on the GTX 1630 the two `ccd_kernel` modules took 173,960 ms and 334,395 ms to compile, against seconds on the 3090 Ti. One-time, then cached under `~/.cache/warp/<version>`.
* Cost figures above are 5 reps eager on the test's own fixture (2.62 / 2.83 / 3.86 / 3.39 / 3.43x) and 7 baseline reps graph-captured on humanoid, each arm in its own process because Warp binds determinism mode at first compile.
* Verified against both warp-lang 1.15.0 and 1.16.0, which measure identically.

Fixes: https://github.com/google-deepmind/mujoco_warp/issues/1635
Related to: https://github.com/google-deepmind/mujoco_warp/issues/562

### DIFF
--- a/mujoco_warp/_src/collision_convex.py
+++ b/mujoco_warp/_src/collision_convex.py
@@ -168,8 +168,16 @@ def ccd_hfield_kernel_builder(
   epa_iterations: int,
   geomgeomid: int,
   warn_overflow: int,
+  deterministic: bool,
 ):
-  """Kernel builder for heightfield CCD collisions (no multiccd args)."""
+  """Kernel builder for heightfield CCD collisions (no multiccd args).
+
+  When ``deterministic`` is set, contact writing is split into a second kernel
+  (``ccd_hfield_write_kernel``) for the same reason as in ``ccd_kernel_builder``: the
+  per-prism CCD results come from EPA scratch that Warp's counting pass cannot read back
+  consistently, so the number of contact slots this kernel would reserve differs between
+  the counting pass and the replay pass.
+  """
 
   # runs convex collision on a set of geom pairs to recover contact info
   @wp.kernel(module="unique", enable_backward=False, grid_stride=False)
@@ -252,6 +260,11 @@ def ccd_hfield_kernel_builder(
     nacon_out: wp.array[int],
     # Data out:
     overflow_out: wp.array[int],
+    # Out:
+    hfield_ncon_out: wp.array[int],
+    hfield_cdist_out: wp.array2d[float],
+    hfield_cpos_out: wp.array2d[wp.vec3],
+    hfield_cnormal_out: wp.array2d[wp.vec3],
   ):
     collisionid = wp.tid()
     if collisionid >= ncollision_in[0]:
@@ -520,40 +533,48 @@ def ccd_hfield_kernel_builder(
           count += 1
 
     # contact 0: minimum distance
-    write_contact(
-      naconmax_in,
-      0,
-      min_dist,
-      min_pos,
-      make_frame(min_normal),
-      margin,
-      gap,
-      condim,
-      friction,
-      solref,
-      solreffriction,
-      solimp,
-      adhesion,
-      geoms,
-      collision_pairid,
-      worldid,
-      contact_dist_out,
-      contact_pos_out,
-      contact_frame_out,
-      contact_includemargin_out,
-      contact_friction_out,
-      contact_solref_out,
-      contact_solreffriction_out,
-      contact_solimp_out,
-      contact_dim_out,
-      contact_geom_out,
-      contact_efc_address_out,
-      contact_worldid_out,
-      contact_type_out,
-      contact_geomcollisionid_out,
-      contact_adhesion_out,
-      nacon_out,
-    )
+    if wp.static(deterministic):
+      # Stage the selected contact for ccd_hfield_write_kernel instead of reserving a
+      # slot here; see the builder docstring.
+      hfield_ncon_out[collisionid] = 1
+      hfield_cdist_out[collisionid, 0] = min_dist
+      hfield_cpos_out[collisionid, 0] = min_pos
+      hfield_cnormal_out[collisionid, 0] = min_normal
+    else:
+      write_contact(
+        naconmax_in,
+        0,
+        min_dist,
+        min_pos,
+        make_frame(min_normal),
+        margin,
+        gap,
+        condim,
+        friction,
+        solref,
+        solreffriction,
+        solimp,
+        adhesion,
+        geoms,
+        collision_pairid,
+        worldid,
+        contact_dist_out,
+        contact_pos_out,
+        contact_frame_out,
+        contact_includemargin_out,
+        contact_friction_out,
+        contact_solref_out,
+        contact_solreffriction_out,
+        contact_solimp_out,
+        contact_dim_out,
+        contact_geom_out,
+        contact_efc_address_out,
+        contact_worldid_out,
+        contact_type_out,
+        contact_geomcollisionid_out,
+        contact_adhesion_out,
+        nacon_out,
+      )
 
     # TODO(team): routine for select subset of contacts
     if wp.static(True):
@@ -579,40 +600,46 @@ def ccd_hfield_kernel_builder(
       pos1 = wp.vec3(hfield_contact_pos[id1, 0], hfield_contact_pos[id1, 1], hfield_contact_pos[id1, 2])
       normal1 = wp.vec3(hfield_contact_normal[id1, 0], hfield_contact_normal[id1, 1], hfield_contact_normal[id1, 2])
 
-      write_contact(
-        naconmax_in,
-        1,
-        hfield_contact_dist[id1],
-        pos1,
-        make_frame(normal1),
-        margin,
-        gap,
-        condim,
-        friction,
-        solref,
-        solreffriction,
-        solimp,
-        adhesion,
-        geoms,
-        collision_pairid,
-        worldid,
-        contact_dist_out,
-        contact_pos_out,
-        contact_frame_out,
-        contact_includemargin_out,
-        contact_friction_out,
-        contact_solref_out,
-        contact_solreffriction_out,
-        contact_solimp_out,
-        contact_dim_out,
-        contact_geom_out,
-        contact_efc_address_out,
-        contact_worldid_out,
-        contact_type_out,
-        contact_geomcollisionid_out,
-        contact_adhesion_out,
-        nacon_out,
-      )
+      if wp.static(deterministic):
+        hfield_ncon_out[collisionid] = 2
+        hfield_cdist_out[collisionid, 1] = hfield_contact_dist[id1]
+        hfield_cpos_out[collisionid, 1] = pos1
+        hfield_cnormal_out[collisionid, 1] = normal1
+      else:
+        write_contact(
+          naconmax_in,
+          1,
+          hfield_contact_dist[id1],
+          pos1,
+          make_frame(normal1),
+          margin,
+          gap,
+          condim,
+          friction,
+          solref,
+          solreffriction,
+          solimp,
+          adhesion,
+          geoms,
+          collision_pairid,
+          worldid,
+          contact_dist_out,
+          contact_pos_out,
+          contact_frame_out,
+          contact_includemargin_out,
+          contact_friction_out,
+          contact_solref_out,
+          contact_solreffriction_out,
+          contact_solimp_out,
+          contact_dim_out,
+          contact_geom_out,
+          contact_efc_address_out,
+          contact_worldid_out,
+          contact_type_out,
+          contact_geomcollisionid_out,
+          contact_adhesion_out,
+          nacon_out,
+        )
 
       # contact 2: point furthest from min_pos - pos1 line
       dist_min1 = wp.cross(min_normal, min_pos - pos1)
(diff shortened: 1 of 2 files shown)


wagner-austin:
@thowell thanks for triaging #1635. One check before this goes stale: I think the title made it read as a determinism feature request, when what it reports is silent wrong results.

Two different tracks here. #1422 / #1533 build `opt.deterministic`, mjwarp's own deterministic kernels. This PR is about `wp.config.deterministic`, Warp's built-in mode — mjwarp doesn't support it, but doesn't refuse it either. It compiles, runs, exits 0, and returns zero contacts for every geom pair the table routes to the convex narrowphase, so bodies fall through geometry. #1590 is the same class one stage earlier, except that one fails loudly.

You asked on #1281 for a benchmark of custom deterministic kernels against Warp's general-purpose determinism. I don't think that ever got measured, so here it is. RTX 3090 Ti, warp 1.15.0, mujoco 3.12.1, CUDA graph capture active in every arm, each track normalised against its own non-deterministic baseline on its own tree (so the base drift between branches can't leak in). Humanoid, nworld=256, baseline median 437,881 steps/s over 7 reps:

| | overhead | contacts |
|---|---|---|
| `opt.deterministic` (#1533) | **3.1x** | correct |
| `wp.config.deterministic`, `deterministic_max_records=64` | **64.8x** | correct |
| `wp.config.deterministic`, default record bound | 12.5x | **wrong** — `ncon` 9.074 vs 9.243 |

On a 12-box pile that routes through the convex narrowphase the same comparison is 10.9x versus ~173x, though I'd treat that row as suggestive rather than settled: #1533's baseline finds `ncon` 5.03 on that scene where this branch finds 17.28, so it is doing less contact work and the ratio may flatter it. The humanoid row is clean — both trees report `ncon` 11.786 identically.

So custom kernels win by roughly an order of magnitude. That is the answer to your question, and it argues for the #1422/#1533 track as the primary one.

Two things it also surfaced that belong on this PR regardless:

1. **Warp's default record bound silently under-counts contacts even with this fix applied.** `wp.config.deterministic_max_records` defaults to 0, meaning the code-generated lower bound, and Warp's own docs say to raise it for kernels with data-dependent loops that static analysis cannot bound — which the CCD narrowphase is. At the default, `wp.config.deterministic_debug = True` prints `Warp deterministic scatter overflow: capacity=1024 dest=...` repeatedly, and contacts go missing. Debug is off by default, so it is silent. My test sets the bound; the PR description does not mention it, and it should — including on primitive-only scenes, which also under-count.
2. **The "2-2.7x" figure in my PR description does not hold** and I am withdrawing it. Across `nworld` 1 to 1024 and both record bounds I could not get the Warp path below ~4.7x, and the cheapest configuration that is also correct is the 64.8x above. The description's structural claim — that primitive-only scenes pay the same factor, so the cost is Warp's replay rather than this change — holds at the default bound (12.5x vs ~20x) but breaks at the correct one (64.8x vs ~173x).

Happy to close this if the answer is that `wp.config.deterministic` should hard-error at init rather than be fixed; I'd send that instead. Otherwise I'll rebase and correct the description — #1613 and #1653 have both landed in these kernels since I opened this.

wagner-austin:
Correction to my comment above: the 2-2.7x in the description is not wrong. I withdrew it too fast.

I had measured only through `mjwarp-testspeed`, which captures a CUDA graph. The original figure was measured eagerly, on this PR's own fixture. That exact scene, eager, `deterministic_max_records=64` (the value the test sets), 5 reps: 2.62x, 2.83x, 3.86x, 3.39x, 3.43x. The figure is right for its regime — the description omits the conditions.

Graph-captured, the same comparison on humanoid at nworld=256, against a baseline median of 437,881 steps/s over 7 reps, is 64.8x. Capture is what `mjwarp-testspeed` and production use, so both numbers belong in the description.

One finding stands on its own. **Under graph capture, a deterministic scatter overflow is silent.** Eager, the default record bound cannot run this PR's own fixture at all — Warp raises `Deterministic counter buffer overflow in kernel 'ccd_kernel_builder__locals__ccd_write_kernel'`. Inside a capture the host-side readback cannot run, so the identical condition raises nothing and contacts go missing instead.

PR #1639, opened 2026-09-02
Fetch pinned benchmark assets without requiring git 2.49

### LINKED ISSUE #1634: benchmarks/run.py needs git >= 2.49 for clone --revision, and a failed clone poisons the asset cache
On 595bd6f, `benchmarks/run.py` cannot fetch its menagerie assets on a machine with a distro git, and the failure leaves the asset cache in a state that stays broken afterwards. Two separate things, but the second is what made the first confusing to debug, so I am reporting them together.

`_assemble_benchmark` clones pinned assets with

```python
_git("clone", repo["source"], repo_dir.as_posix(), "--depth", "1", "--revision", repo["ref"])
```

(`benchmarks/run.py:110`, and the same line at `benchmarks/sweep.py:130`). `git clone --revision` was added in git 2.49.0. On Ubuntu 22.04, whose git is 2.34.1, it fails:

```
$ uv run python3 benchmarks/run.py -f unitree_g1
[...] Discovered 3 benchmarks: [unitree_g1_flat, unitree_g1_hfield, unitree_g1_hfield_render]
error: unknown option `revision'
subprocess.CalledProcessError: Command '('git', 'clone', 'https://github.com/google-deepmind/mujoco_menagerie.git', '/tmp/benchmark_assets/_git/mujoco_menagerie/affef0836947b64cc06c4ab1cbf0152835693374', '--depth', '1', '--revision', 'affef0836947b64cc06c4ab1cbf0152835693374')' returned non-zero exit status 129.
```

Nothing documents a minimum git version, and 2.49 is recent enough (March 2025) that it is not in Ubuntu 22.04 or 24.04, Debian stable, or several CI images.

The second problem is independent of git version. The mkdir happens *before* the clone:

```python
if not repo_dir.exists():
  repo_dir.mkdir(parents=True, exist_ok=True)
  _git("clone", ...)
```

so any clone failure — old git, network, an interrupt — leaves an empty directory at the cache path. On every later run the `if not repo_dir.exists()` guard sees it, skips the clone, and falls through to `shutil.copytree(repo_dir / repo_path, ...)`, which raises `FileNotFoundError`. Upgrading git does not clear it; you have to know to delete the directory by hand. I confirmed the empty directory is left behind after the failure above.

For what it is worth, `git init` plus `git fetch --depth 1 origin <sha>` and a checkout of `FETCH_HEAD` gets the same pinned shallow tree on 2.34.1, so a fallback is possible if you would rather not raise the floor. Cloning into a temporary directory and renaming it into place on success would fix the cache poisoning independently.

Versions: mujoco_warp 595bd6f, warp-lang 1.15.0, mujoco 3.12.1, Python 3.10.20, git 2.34.1, Ubuntu 22.04.

Happy to send a PR for either or both if you have a preference on which way to go.

rakhimovv:
Fixes #1634.

`benchmarks/run.py` and `benchmarks/sweep.py` fetch their pinned menagerie assets with `git clone --revision`, which needs git 2.49 (March 2025). That is newer than the git in Ubuntu 22.04 (2.34.1), Ubuntu 24.04 (2.43.0) and Debian trixie (2.47.3), so on those machines any benchmark with external assets dies right away on ``error: unknown option `revision'``. This swaps the clone for `git init` plus `git fetch --depth 1 <source> <ref>` and a checkout of `FETCH_HEAD`, which gets the same single pinned commit and works back to at least git 2.34.1. I went with the fallback rather than documenting a minimum git version since it costs one extra subprocess call and keeps the scripts runnable on stock LTS images, but happy to switch to a version check and a clear error message if you would rather raise the floor. All four sources the benchmarks pin are on github.com, and I checked that each of their pinned refs fetches this way on 2.34.1.

The second half is the cache poisoning, which is independent of git version. The cache directory was created before the clone ran, so any failed or interrupted fetch left an empty directory at the cache path; on later runs the `if not repo_dir.exists()` guard accepted it and fell through to a `copytree` that raised `FileNotFoundError`. Upgrading git did not help, you had to know to delete the directory. The fetch now lands in a sibling directory that is renamed into place only on success, and the guard keys on `.git` instead of bare existence, which also means a directory left behind by the current code heals itself on the next run rather than needing a manual `rm`. The staging directory is created inside the cache directory's parent so the rename stays on one filesystem.

Two things I deliberately left alone, in case you would rather they were handled. A cache entry that somehow has a `.git` but an incomplete worktree is still trusted; validating it would mean comparing `rev-parse HEAD` against the ref, which would refetch on every run for anyone who pins a branch name instead of a SHA. And two runs sharing an `assets_root` will now have the loser of the rename race fail with `ENOTEMPTY` rather than quietly interleaving two clones into one directory, which is louder than before but still not actually handled.

Both files carry their own copy of the helper, matching how `_git` and `_assemble_benchmark` are already duplicated between them.

Verified on Ubuntu 22.04 with git 2.34.1, warp-lang 1.15.0, mujoco 3.12.1, H100. `uv run python3 benchmarks/run.py -f unitree_g1` runs all three benchmarks to completion starting from the empty poisoned cache directory my earlier failed run had left behind, with no manual cleanup, and reports `converged_worlds: 8192` for each. Fetch by SHA checked against all four pinned sources (mujoco_menagerie, mujoco, aloha_sim, myo_sim), and a deliberately bad ref confirmed to leave nothing at the cache path in both files. `uv run pytest -n 8` passes (1349 passed, 29 skipped), though `norecursedirs` excludes `benchmarks/`, so that is a no-collateral-damage check rather than coverage of this change.

### DIFF
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -66,21 +66,23 @@ def _ensure_pinned_clone(source: str, ref: str, dst: Path):
   """Make dst a shallow checkout of ref from source, reusing it if it already is one."""
   if (dst / ".git").exists():
     return
-  # a dst without .git is a leftover from an interrupted fetch, rebuild it rather than trust it
-  shutil.rmtree(dst, ignore_errors=True)
+  # a dst without .git is a leftover from an interrupted fetch, rebuild it rather than trust it.
+  # rmtree only handles directories, so clear a file or symlink at dst separately or the rename
+  # below fails
+  if dst.is_dir() and not dst.is_symlink():
+    shutil.rmtree(dst, ignore_errors=True)
+  elif dst.exists() or dst.is_symlink():
+    dst.unlink(missing_ok=True)
   # "git clone --revision" does this in one step but needs git >= 2.49, newer than the git in
   # current LTS distros. fetch into a sibling directory and rename it into place so a failed or
   # interrupted fetch cannot leave a partial dst that later runs mistake for a good checkout.
   dst.parent.mkdir(parents=True, exist_ok=True)
-  staging = Path(tempfile.mkdtemp(prefix=f".{dst.name}.", dir=dst.parent))
-  try:
+  with tempfile.TemporaryDirectory(prefix=f".{dst.name}.", dir=dst.parent) as tmp_dir:
+    staging = Path(tmp_dir)
     _git("init", "--quiet", staging.as_posix())
     _git("fetch", "--quiet", "--depth", "1", source, ref, cwd=staging)
     _git("checkout", "--quiet", "FETCH_HEAD", cwd=staging)
-  except BaseException:
-    shutil.rmtree(staging, ignore_errors=True)
-    raise
-  staging.rename(dst)
+    staging.rename(dst)
 
 
 def _uv_run(*args, cwd: Path | None = None):
--- a/benchmarks/run.py
+++ b/benchmarks/run.py
@@ -62,6 +62,27 @@ def _git(*args, cwd: Path | None = None, check: bool = True):
   return subprocess.run(("git",) + args, cwd=cwd, env=env, check=check, capture_output=True, text=True)
 
 
+def _ensure_pinned_clone(source: str, ref: str, dst: Path):
+  """Make dst a shallow checkout of ref from source, reusing it if it already is one."""
+  if (dst / ".git").exists():
+    return
+  # a dst without .git is a leftover from an interrupted fetch, rebuild it rather than trust it
+  shutil.rmtree(dst, ignore_errors=True)
+  # "git clone --revision" does this in one step but needs git >= 2.49, newer than the git in
+  # current LTS distros. fetch into a sibling directory and rename it into place so a failed or
+  # interrupted fetch cannot leave a partial dst that later runs mistake for a good checkout.
+  dst.parent.mkdir(parents=True, exist_ok=True)
+  staging = Path(tempfile.mkdtemp(prefix=f".{dst.name}.", dir=dst.parent))
+  try:
+    _git("init", "--quiet", staging.as_posix())
+    _git("fetch", "--quiet", "--depth", "1", source, ref, cwd=staging)
+    _git("checkout", "--quiet", "FETCH_HEAD", cwd=staging)
+  except BaseException:
+    shutil.rmtree(staging, ignore_errors=True)
+    raise
+  staging.rename(dst)
+
+
 def _uv_run(*args, cwd: Path | None = None):
   """Run a uv command, returning CompletedProcess."""
   log.info("Command: uv run %s", " ".join(args))
@@ -105,9 +126,7 @@ def _assemble_benchmark(bm: dict):
 
     # repo clones are stored in the format: <assets_root>/_git/<repo_source>/<repo_ref>
     repo_dir = Path(_ARGS.assets_root) / "_git" / Path(repo["source"]).stem / repo["ref"]
-    if not repo_dir.exists():
-      repo_dir.mkdir(parents=True, exist_ok=True)
-      _git("clone", repo["source"], repo_dir.as_posix(), "--depth", "1", "--revision", repo["ref"])
+    _ensure_pinned_clone(repo["source"], repo["ref"], repo_dir)
 
     if "*" in repo_path:
       parts = Path(repo_path).parts
--- a/benchmarks/sweep.py
+++ b/benchmarks/sweep.py
@@ -86,21 +86,23 @@ def _ensure_pinned_clone(source: str, ref: str, dst: Path):
   """Make dst a shallow checkout of ref from source, reusing it if it already is one."""
   if (dst / ".git").exists():
     return
-  # a dst without .git is a leftover from an interrupted fetch, rebuild it rather than trust it
-  shutil.rmtree(dst, ignore_errors=True)
+  # a dst without .git is a leftover from an interrupted fetch, rebuild it rather than trust it.
+  # rmtree only handles directories, so clear a file or symlink at dst separately or the rename
+  # below fails
+  if dst.is_dir() and not dst.is_symlink():
+    shutil.rmtree(dst, ignore_errors=True)
+  elif dst.exists() or dst.is_symlink():
+    dst.unlink(missing_ok=True)
   # "git clone --revision" does this in one step but needs git >= 2.49, newer than the git in
   # current LTS distros. fetch into a sibling directory and rename it into place so a failed or
   # interrupted fetch cannot leave a partial dst that later runs mistake for a good checkout.
   dst.parent.mkdir(parents=True, exist_ok=True)
-  staging = Path(tempfile.mkdtemp(prefix=f".{dst.name}.", dir=dst.parent))
-  try:
+  with tempfile.TemporaryDirectory(prefix=f".{dst.name}.", dir=dst.parent) as tmp_dir:
+    staging = Path(tmp_dir)
     _git("init", "--quiet", staging.as_posix())
     _git("fetch", "--quiet", "--depth", "1", source, ref, cwd=staging)
     _git("checkout", "--quiet", "FETCH_HEAD", cwd=staging)
-  except BaseException:
-    shutil.rmtree(staging, ignore_errors=True)
-    raise
-  staging.rename(dst)
+    staging.rename(dst)
 
 
 def _uv_run(*args, cwd: Path | None = None):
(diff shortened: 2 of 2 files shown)


thowell:
Thanks for putting this together, @rakhimovv! This is a great fix that addresses both Git version compatibility on standard LTS distributions (Ubuntu 22.04 / 24.04) and the cache poisoning issue from interrupted fetches.

Here are two quick suggestions to simplify the implementation and make cleanup even cleaner:

---

### 1. Simplify with `tempfile.TemporaryDirectory`

Instead of manually creating a directory with `tempfile.mkdtemp` and catching `BaseException` to call `shutil.rmtree`, you can use Python's standard `tempfile.TemporaryDirectory` context manager:

- **Automatic cleanup:** Automatically cleans up the staging directory on any failure, unhandled exception, or `KeyboardInterrupt` (`Ctrl+C`).
- **No-op on success:** Once `staging.rename(dst)` succeeds, the original staging path no longer exists, so `TemporaryDirectory.__exit__` quietly exits without error.
- **Eliminates boilerplate:** Removes the need for explicit `except BaseException: ... raise` blocks.

---

### 2. Clean up Non-Directory Paths at `dst`

`shutil.rmtree` only handles directory trees. If `dst` happens to exist as a regular file or broken symlink, `shutil.rmtree` raises `NotADirectoryError`, which `ignore_errors=True` silently suppresses—causing the subsequent rename to fail. Checking for a file/symlink before `rmtree` prevents this:

```python
if dst.is_dir() and not dst.is_symlink():
  shutil.rmtree(dst, ignore_errors=True)
elif dst.exists() or dst.is_symlink():
  dst.unlink(missing_ok=True)
```

---

### Proposed Refactor

Here is the updated implementation incorporating both suggestions:

```python
def _ensure_pinned_clone(source: str, ref: str, dst: Path):
  """Make dst a shallow checkout of ref from source, reusing it if it already is one."""
  if (dst / ".git").exists():
    return

  # A dst without .git is a leftover from an interrupted fetch, rebuild it rather than trust it.
  if dst.is_dir() and not dst.is_symlink():
    shutil.rmtree(dst, ignore_errors=True)
  elif dst.exists() or dst.is_symlink():
    dst.unlink(missing_ok=True)

  # "git clone --revision" does this in one step but needs git >= 2.49, newer than the git in
  # current LTS distros. Fetch into a sibling directory and rename it into place so a failed or
  # interrupted fetch cannot leave a partial dst that later runs mistake for a good checkout.
  dst.parent.mkdir(parents=True, exist_ok=True)
  with tempfile.TemporaryDirectory(prefix=f".{dst.name}.", dir=dst.parent) as tmp_dir:
    staging = Path(tmp_dir)
    _git("init", "--quiet", staging.as_posix())
    _git("fetch", "--quiet", "--depth", "1", source, ref, cwd=staging)
    _git("checkout", "--quiet", "FETCH_HEAD", cwd=staging)
    staging.rename(dst)
```

#### Diff:

```diff
-  shutil.rmtree(dst, ignore_errors=True)
+  if dst.is_dir() and not dst.is_symlink():
+    shutil.rmtree(dst, ignore_errors=True)
+  elif dst.exists() or dst.is_symlink():
+    dst.unlink(missing_ok=True)
+
   dst.parent.mkdir(parents=True, exist_ok=True)
-  staging = Path(tempfile.mkdtemp(prefix=f".{dst.name}.", dir=dst.parent))
-  try:
+  with tempfile.TemporaryDirectory(prefix=f".{dst.name}.", dir=dst.parent) as tmp_dir:
+    staging = Path(tmp_dir)
     _git("init", "--quiet", staging.as_posix())
     _git("fetch", "--quiet", "--depth", "1", source, ref, cwd=staging)
     _git("checkout", "--quiet", "FETCH_HEAD", cwd=staging)
-  except BaseException:
-    shutil.rmtree(staging, ignore_errors=True)
-    raise
-  staging.rename(dst)
+    staging.rename(dst)
```

please let us know if you have any questions, thanks!

rakhimovv:
Thank you for the careful read — both suggestions are in as of 2700299.

`TemporaryDirectory` is a clear improvement over the `mkdtemp` plus `except BaseException` pair, and it removes the boilerplate exactly as you describe. I did want to convince myself of the no-op-after-rename property before relying on it, since `shutil.rmtree` moved from `onerror` to `onexc` in 3.12 and this repo supports 3.10 through 3.14, so I ran it on all five: the context manager exits cleanly once the staging path has been renamed away, and it still removes the staging directory when an exception is raised inside the block. The non-directory `dst` case is real too — against the previous implementation, a `dst` left as a regular file or a broken symlink fails at the rename with `NotADirectoryError: [Errno 20]`, and both now succeed.

One small correction to the rationale, though it does not change the fix: `shutil.rmtree` refuses a symlink outright rather than following it, so the old code was never at risk of deleting a symlink's target — it suppressed that `OSError` under `ignore_errors` and then failed at the rename. The change turns a hard failure into a working path rather than preventing data loss. I also checked `dst` as a symlink to a real directory while I was in there: a link to a plain directory is unlinked and rebuilt with the target left intact, and a link to a valid checkout is reused in place.

One pre-existing limitation neither version changes, noting it only so it is written down somewhere: two concurrent `run.py` invocations sharing an `--assets_root` can still collide, since the second `staging.rename(dst)` hits `ENOTEMPTY` once the first has populated `dst`. Benchmarks are assembled serially and the nightly matrix runs on separate runners, so it does not arise in practice today, and the original `git clone --revision` form had the same property.

rakhimovv:
Correction to my last comment: I applied the cleanup change to `run.py` only, which left `sweep.py` behind. `sweep.py` keeps its own verbatim copy of `_ensure_pinned_clone`, and the first commit in this PR had updated both in lockstep, so the second one silently split them and `sweep.py` still failed on a non-directory `dst`. Fixed in d2f90f4 — the two copies are byte-identical again and `sweep.py` passes the same checks.

Worth mentioning since it is what caused the slip: `run.py` and `sweep.py` currently share `_git`, `_ensure_pinned_clone` and `_uv_run` as byte-identical copies, and `_discover_benchmarks` and `_assemble_benchmark` differ only in a local variable name, a type annotation, and `sys.path.insert(0, ...)` versus `sys.path.append(...)`. Any fix to the shared helpers has to be written twice, which is exactly what went wrong here. I have deliberately not touched that in this PR, since restructuring the two entry points is a bigger change than the bug at hand and belongs in its own review rather than buried in this diff. Happy to leave it as an observation for you to weigh.

thowell:
lets introduce a helper file: benchmarks/common.py or benchmarks/utils.py that contains shared utilities in order to reduce code duplication. thanks!

rakhimovv:
Will do. One thing to flag before I send it, since it decides part of the shape: `contrib/systemd/README.md` installs the nightly by copying `sweep.py` on its own to `~/.local/bin/mjwarp-sweep`, and the unit runs `ExecStart=%h/.local/bin/mjwarp-sweep forward`, so today `sweep.py` executes as a lone file outside the repo. The moment it imports `benchmarks/common.py` that copy step breaks with `ModuleNotFoundError`, and it breaks at the next nightly rather than at install time. That constraint is what stopped me proposing the helper file myself.

It is easy to handle, it just has to move together with the extraction, so I will install the scripts into a directory instead of a single file — `~/.local/share/mjwarp-benchmarks/` holding `sweep.py` and `common.py`, with the unit pointed there. That keeps the nightly independent of any repo checkout, which is the property the copy step was giving you. If you would rather it just run out of a checkout, say so and I will do that instead.

I had already opened #1644 with the opposite conclusion — a test that pins the copies together, on the reasoning above — about three minutes after your comment landed, so it crossed with you. Ignore its current contents; I will repurpose that PR for the extraction rather than leave a stale argument sitting there, and keep this one as just the fetch fix.

rakhimovv:
The helper file is up in #1644, repurposed from what was previously sitting there. benchmarks/common.py now holds the five shared helpers plus the PATH bootstrap, the logging setup and clone_if_needed, and the two that used to read the _ARGS global take their arguments explicitly instead.

The systemd install moves with it, as flagged: the scripts go into ~/.local/share/mjwarp-benchmarks/ as a pair and the unit points there. I verified both directions rather than assuming — the two files run fine from an empty directory outside any repo, and sweep.py alone fails with ModuleNotFoundError. There is also a one-time migration for whichever machine currently runs the nightly, spelled out in the README: re-copy the service file and daemon-reload, because otherwise the old unit keeps running the old ~/.local/bin/mjwarp-sweep and the nightly would quietly benchmark stale code rather than fail.

Extracting the helpers had a nice side effect: ensure_pinned_clone is reachable from the test suite now, so the cache-poisoning behaviour this PR fixes has real coverage rather than just my say-so.

Thanks for steering me to the helper file — it is the better shape, and the systemd bit turned out to be easy once it moved alongside. This PR is unchanged and still just the fetch fix.

PR #1637, opened 2026-09-02
Add native Panda threading control-replay benchmark

ooctipus:
Depends on **#1684** for native SDF collision correctness and **#1685** for applying dense/sparse overrides to the prepared binary model. Both fixes should land before this benchmark. Their prerequisite patches remain included in this branch while those PRs are pending, so the replay stays executable.

Add a three-second native threading benchmark with one Panda and an M8 nut and bolt on the NIST board. The nut starts partly threaded; the RJ45 connector, gear and rod are already assembled. During replay, the nut rotates, advances and remains engaged with the bolt.

The package contains the meshes, one initial position/velocity state and 300 actuator targets. Existing replay code holds each target for eight 1.25 ms physics steps, covering all 2,400 steps without policy inference, episode resets or intermediate state restoration.

https://github.com/user-attachments/assets/6d681083-3467-4714-b4a8-8bad442d61bc

With #1638, #1642 and #1643 applied together, sparse solver time on an RTX 5090 is **27.4% shorter at 2,048 worlds, 37.6% shorter at 4,096, and 40.8% shorter at 8,192**. At 4,096 worlds, solver time is 1.3239 ms dense → 0.8264 ms sparse, while whole-step time is 42.8314 → 41.8698 ms (2.24% shorter). Native SDF collision remains the bottleneck: 40.4585 ms, or 96.6% of the sparse whole step.

Both dense and sparse were freshly measured on [the same combined optimization build](https://github.com/ooctipus/mujoco_warp/commit/ebb4ee4aa500003fce8b72aae06d5772ac3b1f15), using identical scene bytes, initial positions and velocities, controls and capacities. The three optimization changes remain in their respective PRs. All timings cover the complete three-second continuous rollout and exclude compilation. The 4,096-world results use the median of three runs per layout; other world counts use one run per layout. Dense and sparse follow their own evolving trajectories. The video above shows the benchmark replay before applying these optimizations.

Dense versus sparse with all three optimizations on the continuous threading replay

[...]

PR #1638, opened 2026-09-02
Optimize sparse contact constraint launches

ooctipus:
This is the first implementation change in the sparse Newton optimization stack. The benchmark-only #1637 is independent and can merge at any time; the required implementation merge order is this PR → #1642 → #1643. The isolated implementation comparison is current main `d9c1c4e` → `3b3f816`.

Sparse contact construction currently launches over capacity-sized contact grids even when each world has only a short active EFC prefix. This change selects one warp per world when a high-world-count sparse batch supplies at least four GPU occupancy waves, and the contact Jacobian and update kernels then traverse active contact EFC rows. Dense, low-world-count, and flex-contact configurations retain their existing dispatch. Passive adhesion behavior is preserved, with forced-dispatch tests covering sparse pyramidal and elliptic contacts against MuJoCo.

On physical GPU0, an RTX 5090, I ran five interleaved warm-cache processes per revision and report medians:

- On the original single-Panda diagnostic with the reviewer-requested `nconmax=32` and `njmax=64`, throughput improved from 7.540M to 7.702M steps/s (+2.15%), step time fell from 122.321 to 121.518 ns/world-step (-0.66%), and `make_constraint` fell from 11.133 to 10.524 ns (-5.47%). Solve time was within run noise (-0.62%).
- On the updated five-Panda sparse benchmark from #1637 with `nconmax=64` and `njmax=192`, throughput was within run spread (3.165M → 3.155M, -0.30%), while step time fell 0.15% and `make_constraint` fell 1.77%.

All 4,096 worlds converged in every run with matching contact, EFC-row, and solver-iteration statistics. Validated with targeted forced-dispatch tests, all pre-commit checks, and `uv run pytest -n 8`: 1,357 passed and 29 skipped.

PR #1646, opened 2026-09-03
pandas benchmark

thowell:
This update upgrades the multi-arm Franka Emika Panda benchmark #1637 from a static gripper contact scene into a dynamic sequential pick-and-lift manipulation task with MuJoCo Menagerie visual assets.

rollout_pandas

Key changes:
- Updated `panda.xml` to include visual mesh geoms matching MuJoCo Menagerie and added the official `home` keyframe.
- Renamed the scene to `scene_pandas.xml` and replaced static contact blocks with a tabletop and five freejoint boxes of varying dimensions, masses, and distinct colors. Object heights are kept below 44 mm to guarantee clearance against the palm's collision capsule.
- Added a 1,700-step (8.5 s) replay trajectory (`pick_objects.npz`) that initializes all arms at the home pose and controls them sequentially from left to right (Arm 0 to Arm 4) to descend, pinch-grip their objects, and raise them back to the home configuration.
- Added camera parameter flags (`--cam_azimuth`, `--cam_elevation`, `--cam_lookat_x`, `--cam_lookat_y`) to `mujoco_warp/record.py` and updated camera initialization to respect scene visual configuration defaults. Re-recorded `rollout_pandas.webp` with an unobstructed front perspective that orbits along with the manipulation sequence.
- Optimized benchmark capacities in `__init__.py` and `README.md` to 8,192 parallel worlds, `nconmax = 48`, and `njmax = 192`. These represent the smallest multiples of 16 that prevent overflow during the rollout (peak contacts reach 44 and peak constraint rows reach 181; `nconmax = 32` and `njmax = 176` both overflow).
- Running with `opt.jacobian = sparse` achieves 1.34M steps/s across 8,192 worlds with 100% solver convergence and 2.78 GB memory usage. Note that `opt.jacobian = dense` is not supported for this benchmark because the 75-DoF system (45 arm DoFs + 30 freejoint DoFs) exceeds the current 60-DoF limit in `put_model`.


ooctipus:
Thanks @thowell — I benchmarked the proposed five-Panda/five-object scene first.

That model has `5 × 9 + 5 × 6 = 75` DoFs. At 75 DoFs, dense is unsupported because the current limit is 60, and the sparse Newton line-search/gradient specialization in #1642 and #1643 does not activate because it currently requires `nv <= 50`. Therefore, the five-station model cannot provide a sparse-versus-dense comparison or exercise the complete intended optimization path.

On an RTX 5090 using physical GPU 0, five order-balanced warm-cache runs of the exact five-station replay at 8,192 worlds, `nconmax=48`, and `njmax=192` measured:

- Base sparse `d9c1c4e`: 2.452M steps/s
- Full stack `b72320e`: 2.441M steps/s
- Change: −0.45% throughput, +0.59% step time, +2.25% solve time, and −7.24% `make_constraint`

All worlds converged, but only the active-row constraint work shows a benefit; there is no end-to-end improvement from the full stack on the 75-DoF model.

I therefore tested a centered three-Panda/three-object variant, retaining the middle three stations and their corresponding controls and initial states from the same 1,700-step replay. This gives `3 × 9 + 3 × 6 = 45` DoFs, so dense is supported and all three optimization PRs are exercised.

Using 8,192 worlds and the smallest safe capacity multiples, `nconmax=32` and `njmax=128`, five order-balanced warm-cache runs produced:

| Configuration | Throughput | Step | Solve | `make_constraint` |
|---|---:|---:|---:|---:|
| Base `d9c1c4e`, dense | 3.975M steps/s | 248.47 ns | 106.21 ns | 32.84 ns |
| Base `d9c1c4e`, sparse | 4.685M steps/s | 210.58 ns | 82.40 ns | 22.45 ns |
| Full stack `b72320e`, sparse | 4.788M steps/s | 206.43 ns | 80.28 ns | 20.10 ns |

Base sparse versus base dense is **+17.87% throughput**, with step time down 15.25% and solve time down 22.42%. The full #1638 → #1642 → #1643 stack versus base sparse is **+2.20% throughput**, with step time down 1.97%, solve time down 2.57%, and `make_constraint` down 10.46%.

All 8,192 worlds converged in every run. The three-station rollout averaged 14.79 contacts and 79.92 EFC rows per world, with p95 values of 17.40 and 99; peak EFC usage was 115 rows. This preserves the dynamic manipulation workload while giving a clean sparse-over-dense signal and exercising the specialized solver path.

thowell:
@ooctipus looking at the isaac lab environments https://isaac-sim.github.io/IsaacLab/main/source/overview/environments.html#contact-rich-manipulation instead of multiple robots + objects per world do we want something like 1 robot arm (~9 dofs) + 1 object (6 dofs) + 1 static object (0 dofs) per world and the interaction between the object and the static object generates a large number of contacts? if yes, what is the order of the expected number of contacts?

ooctipus:
Thanks @thowell — yes, that topology is closer to the IsaacLab Factory task: one articulated arm, one free manipulated object, and one static mating object, with most contacts generated between the two objects.

I checked a four-environment IsaacLab Factory Board rollout containing threaded-nut and gear-mesh assemblies. Once engaged, it generated:

- 151–210 raw collision contacts per environment on average
- 221–326 MJWarp contact records per environment on average after paired-guard expansion
- 241 active EFC rows at the peak in the busiest environment

The contact arrays are shared across worlds, so the first two values are batch averages rather than strict per-world maxima.

More demanding threaded and gear-mesh states can produce substantially more solver work. In a separate gear-mesh run, one world requested 1,928 EFC rows and overflowed the previous `njmax=1500`, causing constraints beyond the capacity to be dropped. That is why the Factory configuration uses `nconmax=600` and `njmax=3000`: they are family-wide capacity ceilings with headroom, not the expected active count in every rollout.

I tested several native MJWarp reductions. The performance runs below used an RTX 5090 on physical GPU 0, 8,192 worlds, and five order-balanced warm-cache repetitions:

| Scene | Workload | Result |
|---|---|---|
| Original single Panda and cube | Approximately 2–5 contacts and 12–24 EFC rows; corrected capacities `32/64` | Full stack: **+3.66%** throughput, **−8.46%** solve, **−5.53%** `make_constraint`. The earlier 37–50% result came from the oversized `600/3000` capacities. |
| Five Pandas with static objects | `nv=45`; averages 27.12 contacts and 119.22 EFC rows | Base sparse was **25.70%** faster than base dense, but the full stack was effectively flat at **−0.35%** versus base sparse. |
| Five Pandas with five free objects from #1646 | `nv=75` | Dense is unsupported and the `nv <= 50` specializations do not activate. Full stack: **−0.45%** throughput. |
| Three Pandas with three free objects | `nv=45`; averages 14.79 contacts and 79.92 EFC rows; peak 115 rows; capacities `32/128` | Base sparse was **17.87%** faster than base dense. Full stack added **2.20%** throughput and reduced `make_constraint` by **10.46%**. |
| Analytic-SDF nut and bolt | Average 6.3 contacts; p95 13; maximum 19 | The native geometry did not generate enough contacts for this benchmark. |
| One Panda, one compound free body, and one static plate | The free body has 25 collision pads. The reference state generates exactly 100 contacts and 401–403 EFC rows; the dynamic rollout averages 66.4 contacts and 276.9 rows; capacities `112/416` | Base sparse was **2.37%** faster than base dense. The full stack added **2.24%** throughput and reduced solve by **7.68%**, but `make_constraint` increased **4.98%**. |

These results suggest that contact count alone does not reproduce the large optimization signal; the DoF count and Jacobian sparsity structure also matter. Simply constructing approximately 100 contacts did not recover the original 37–50% result.

So far, the three-Panda/three-object replay is the best balanced benchmark for the sparse solver work: it has dynamic manipulation, supports a dense comparison, activates the complete optimization stack, and shows improvements both from sparse structure and from the proposed changes. The one-arm compound-contact scene could still be useful as a separate contact-rich microbenchmark.

Would you prefer this two-benchmark split, or should I continue iterating on the one-arm scene before updating #1637?

PR #1647, opened 2026-09-04
Remove deprecated Warp API usage

shi-eric:
Warp has deprecated `warp.jax_experimental` and `wp.config.quiet`, with both scheduled for removal in 1.18. They are still used by the JAX test, examples, command-line tools, and notebooks.

This switches the JAX code to `wp.jax_callable` and `wp.JaxCallableGraphMode`, and replaces `quiet` with the equivalent `log_level` setting. The replacement APIs are already available in Warp 1.15, so the minimum version does not need to change. Graph capture and logging behavior remain the same.

`uv run pytest -n 8` passes (1,371 passed, 29 skipped). The JAX test also passes on CUDA with deprecation warnings treated as errors. Pre-commit and notebook JSON validation pass.


thowell (approved):
thanks @shi-eric!

PR #1648, opened 2026-09-06
Add object-local texture projection and skip unused image uploads

tkelestemur:
Primitives and meshes without UVs currently sample a single texture coordinate, so checkerboard materials render as solid colors. Add object-local box projection for boxes, spheres, ellipsoids, cylinders, capsules, and meshes without valid UVs. The pattern follows object rotation and per-world dimensions, while plane mapping and valid mesh UV interpolation keep their existing behavior. Projection selects the dominant local normal axis and uses one texture lookup per surface hit; the tradeoff is seams at axis transitions and distortion on curved surfaces.

Texture conversion now uploads only the selected image's byte slice instead of the whole texture bank for each image. Render contexts resolve active cameras and their RGB outputs before uploading images, avoiding uploads for depth/segmentation-only contexts and RGB contexts with both surface textures and skyboxes disabled. An enabled skybox still gets its images when surface textures are disabled.

Regression tests cover spatial checker patterns and per-world texture swaps on all six geometry types, explicit mesh UV precedence, translated/rotated objects, local projection-axis selection, per-world primitive sizes, camera defaults and overrides, camera filtering, skybox-only rendering, and image-slice conversion for one-, three-, and four-channel PNGs. Material selection uses the existing `put_model(batch_sizes={"mat_texid": nworld})` API.

Validated on macOS with Warp 1.15.0 CPU and MuJoCo 3.12.1.dev974703000:

- `uv run pytest -n 8 --cpu mujoco_warp/_src/render_test.py mujoco_warp/_src/render_util_test.py -q`: 104 passed, 1 CUDA-only test skipped.
- `uv run pytest -n 8 --cpu --maxfail=8`: 1,389 passed, 34 skipped, 1 failure in the unchanged `IOTest.test_put_data_nefc_zero_dense` (`mjd.nefc` is 1 instead of 0). The same failure reproduces on unmodified upstream commit `917a14b`, where 1,357 tests passed and 34 were skipped.
- Pre-commit checks for the changed files passed, including Ruff, formatting, and kernel analysis.

### DIFF
--- a/mujoco_warp/_src/render.py
+++ b/mujoco_warp/_src/render.py
@@ -182,6 +182,8 @@ def sample_texture(
   tex: wp.Texture2D,
   pos: wp.vec3,
   rot: wp.mat33,
+  size: wp.vec3,
+  normal: wp.vec3,
   mesh_facetexcoord: wp.array[wp.vec3i],
   mesh_texcoord: wp.array[wp.vec2],
   mesh_texcoord_offsets: wp.array[int],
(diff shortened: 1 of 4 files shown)


thowell:
@tkelestemur thanks identifying these differences and opening a pr!

finding after investigating that there are more differences so have opened a pr for the renderer https://github.com/google-deepmind/mujoco_warp/pull/1654. please take a look and let us know if #1654 resolves the items reported in this pr. thanks!

sylvesterkaczmarek (commented):
The fallback projection uses `geom_size` as the normalization extent for `MESH`, but MuJoCo ignores geom `size` for mesh geoms; their extent comes from the mesh asset. A mesh without texcoords can therefore divide by an unrelated/default geom size and produce incorrectly scaled UVs, while the current checker test only verifies that coordinates vary. Could the mesh fallback normalize by the compiled mesh bounds/asset extent, or exclude meshes from this primitive fallback until that extent is available, with a regression using a scaled mesh without texcoords?

thowell:
should be superseded by #1654. thanks!

PR #1651, opened 2026-09-08
Fix issue in multiccd where edge contact can be on the wrong face

kbayes:
See https://github.com/google-deepmind/mujoco/commit/bbb71a28a4d16db0d42f98da84e9e56ff663df5f


thowell (approved):

thowell:
@kbayes should we close https://github.com/google-deepmind/mujoco_warp/pull/1633? additionally, please add a test.

kbayes:
@thowell unfortunately I can't add a test. The mujoco tests depends on the cylinder geom having multiccd support, which I couldn't find a repro with any other geom.

PR #1659, opened 2026-09-10
updates for cylinder multiccd

### LINKED ISSUE #1658: CUDA symbol lookup failure with Warp 1.16 and 1.17
### Description

MuJoCo Warp 3.13.0 can fail when multiple convex collision calls run in the same Python process with Warp 1.16.0 or 1.17.0. The first error is:

```text
Warp CUDA error 500: named symbol not found
Warp CUDA error: Failed to lookup kernel function
ccd_kernel_builder__locals__ccd_kernel_7a3eb851_cuda_kernel_forward in module
```

This is followed by a `KeyError` for the same kernel's shared-memory metadata. The `KeyError` is secondary to the CUDA symbol lookup failure.

The failure depends on execution order, so affected tests may pass when run alone.

### Reproduction

This sequence reproduces the problem on MuJoCo Warp 3.13.0:

```bash
WARP_CACHE_PATH="$(mktemp -d)" uv run --with warp-lang==1.17.0 pytest -n 0 \
  mujoco_warp/_src/collision_driver_test.py::CollisionTest::test_collision1 \
  mujoco_warp/_src/collision_driver_test.py::CollisionTest::test_collision2
```

Results with fresh Warp caches:

```text
Warp 1.15.0: 2 passed
Warp 1.16.0: 1 passed, 1 failed
Warp 1.17.0: 1 passed, 1 failed
```

The MuJoCo Warp trigger was introduced by [cylinder multiccd (#1653)](https://github.com/google-deepmind/mujoco_warp/commit/b32c082e9a47300609b441d0cdeff67ecf0a5b4e). The new cylinder face helper uses deferred `wp.static(values[i])` expressions inside a `@wp.func`.

The underlying Warp bug is tracked in [NVIDIA/warp#1934](https://github.com/NVIDIA/warp/issues/1934), which includes a standalone reproducer without MuJoCo. Warp hashes the `module="unique"` kernel before resolving the deferred static expressions in the helper. After MuJoCo Warp loads the kernel with different block dimensions for the occupancy query and launch, a later occupancy query looks for the updated kernel symbol in the older executable.

### Workarounds for users

Setting the convex CCD block dimension to 256 keeps the occupancy query and kernel launch on the same module variant:

```python
m = mjw.put_model(mjm)
m.block_dim.convex_ccd = 256
```

This must be done before the first call to `collision()`, `forward()`, or `step()`. The two-test reproduction passes with this setting on Warp 1.17.0. The larger block dimension may reduce performance for some workloads.

Pinning `warp-lang==1.15.0` is another temporary option for users who cannot change the model configuration.

### Possible MuJoCo Warp mitigation

MuJoCo Warp could avoid the trigger by removing the deferred `wp.static()` expressions from `_cylinder_face`. Evaluating those constants directly in the generated kernel works with the affected Warp releases and would preserve the existing `convex_ccd=64` launch configuration. Explicitly unrolling the 16 face assignments would also avoid runtime trigonometric operations, though it would make the helper more repetitive.

The root fix belongs in Warp and is tracked in NVIDIA/warp#1934. This issue can alert affected users and track any temporary MuJoCo Warp mitigation until a Warp release containing the fix is available.

thowell:
should fix #1658

### DIFF
--- a/mujoco_warp/_src/collision_gjk.py
+++ b/mujoco_warp/_src/collision_gjk.py
@@ -57,8 +57,8 @@ _FACE_INVALID_BIT = wp.constant(wp.uint32(0x40000000))
 _FACE_INVALID_OR_DELETED_MASK = wp.constant(wp.uint32(0xC0000000))
 
 # Precomputed circle coordinates for 16-gon cylinder face approximation
-_CYLINDER_COS_16 = wp.static(tuple(math.cos(i * math.pi / 8.0) for i in range(16)))
-_CYLINDER_SIN_16 = wp.static(tuple(math.sin(i * math.pi / 8.0) for i in range(16)))
+_CYLINDER_COS_16 = tuple(math.cos(i * math.pi / 8.0) for i in range(16))
+_CYLINDER_SIN_16 = tuple(math.sin(i * math.pi / 8.0) for i in range(16))
 
 
 @wp.struct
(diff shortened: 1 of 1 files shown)


shi-eric (approved):
Looks good, thanks!

PR #1661, opened 2026-09-10
support flex edge passive forces and 1D flex potential energy

thowell:
MuJoCo flex models can specify edge-level stiffness and damping on flex edges, and 1D flexes contribute elastic potential energy from edge deformation. Previously, MuJoCo Warp did not compute flex edge passive forces or 1D flex edge potential energy, and models with rigid flexes were not explicitly rejected.

This change adds support for flex edge spring and damper passive forces and 1D flex edge potential energy, while raising a NotImplementedError for rigid flexes in put_model.

Add flex_edgestiffness and flex_edgedamping arrays to Model.
Add _spring_damper_flexedge_passive kernel in passive.py to compute spring and damper forces along flex edges into qfrc_spring and qfrc_damper.
Add _energy_pos_passive_flex kernel in sensor.py to evaluate potential energy for 1D flexes into d.energy.
Raise NotImplementedError in io.py when flex_rigid is non-zero.
Add parameterized unit tests in flex_test.py validating passive forces, potential energy, disable flags, and rigid flex rejection against MuJoCo.


mujoco_warp/_src/passive.py
```diff
@@ -271,6 +271,74 @@ def _spring_damper_tendon_passive(
...
+
+  damping = float(0.0)
+  if not dsbl_damper:
+    damping = flex_edgedamping[f]
+
+  if stiffness == 0.0 and damping == 0.0:
+    return
+
+  rownnz = flexedge_J_rownnz[edgeid]
+  if rownnz == 0:
```
smallquail, line 315: C skips rigid edges explicitly here — `if (m->flexedge_rigid[e]) continue;` in `engine_passive.c`, and the same guard in the energy loop in `engine_sensor.c`. There is no `flexedge_rigid` array in mjWarp, so `rownnz == 0` is standing in for it. That nulls the force here, but `_energy_pos_passive_flex` in sensor.py has no equivalent guard at all: it relies on `displacement == 0.0` being bit-exact for a welded edge, and `flexedge_length` is recomputed from `flexvert_xpos` every step, so a rigidly moved body gives a rounding-level displacement and a small spurious potential energy. Negligible numerically, but the predicate C uses is not implemented and no test has rigid edges. Exposing `flexedge_rigid` on Model would let both kernels use the same condition as C.

mujoco_warp/_src/passive.py
```diff
@@ -271,6 +271,74 @@ def _spring_damper_tendon_passive(
...
+    frc_damper = -damping * flexedge_velocity_in[worldid, edgeid]
+
+  if frc_spring == 0.0 and frc_damper == 0.0:
+    return
+
+  rowadr = flexedge_J_rowadr[edgeid]
+  for k in range(rownnz):
+    sparseid = rowadr + k
+    colind = flexedge_J_colind[sparseid]
+    J = flexedge_J_in[worldid, sparseid]
```
smallquail, line 333: `flexedge_J` is not written for interpolated flexes: `_flex_edges` in smooth.py returns early for trilinear vertices (`b1 < 0 or b2 < 0`) after writing `flexedge_length` but before filling the Jacobian row. C avoids this by skipping edge processing for such flexes entirely (`if (m->flex_rigid[f] || m->flex_interp[f]) continue;` in `engine_core_smooth.c`). The only guard here is `rownnz`, which comes from Model and stays nonzero. Worth confirming the compiler forces `flex_edgestiffness`/`flex_edgedamping` to zero for trilinear flexes — if it does not, this reads uninitialized memory. A `flex_interp` check would make it independent of that.

mujoco_warp/_src/passive.py
```diff
@@ -271,6 +271,74 @@ def _spring_damper_tendon_passive(
...
+  # In:
+  dsbl_spring: bool,
+  dsbl_damper: bool,
+  # Data out:
+  qfrc_spring_out: wp.array2d[float],
+  qfrc_damper_out: wp.array2d[float],
+):
+  worldid, edgeid = wp.tid()
+
+  for i in range(nflex):
```
smallquail: `f` is used after this loop but is only assigned inside the `if`, unlike `stiffness` and `damping` below which are pre-initialized. Same pattern already exists at smooth.py:370, so not new here, but a `flex_edgeflexid` array mirroring `flex_vertflexid` / `flex_elemflexid` would replace the O(nflex) scan with a lookup and remove the uninitialized read in all three places at once.

mujoco_warp/_src/sensor.py
```diff
@@ -3007,6 +3007,43 @@ def _energy_pos_passive_tendon(
...
+      break
+
+  if flex_dim[f] > 1:
+    return
+
+  stiffness = flex_edgestiffness[f]
+  if stiffness == 0.0:
+    return
+
+  displacement = flexedge_length0[edgeid] - flexedge_length_in[worldid, edgeid]
```
smallquail, line 3037: No `flexedge_rigid` guard here — see the note on `_spring_damper_flexedge_passive` in passive.py. C's energy loop skips rigid edges (`if (!m->flexedge_rigid[e])` in `engine_sensor.c`).

mujoco_warp/_src/flex_test.py
```diff
@@ -3662,6 +3663,88 @@ def test_flex_contact_nnz_runtime_bound(self, xml: str):
...
 
+class FlexEdgeTest(parameterized.TestCase):
+  """Tests for flex edge properties."""
+
+  @parameterized.product(
+    nworld=[1, 2],
+    spring=(0, DisableBit.SPRING),
+    damper=(0, DisableBit.DAMPER),
+  )
+  def test_flex_edge_passive_and_energy(self, nworld, spring, damper):
```
smallquail, line 3674: Coverage is 1D only. A `dim=2` flex with edge stiffness gets passive forces but no potential energy in C — the flexedge spring-damper loop has no dim check, the energy loop has `flex_dim[i] > 1` — and this implementation matches, but nothing pins that asymmetry. A dim=2 case asserting nonzero `qfrc_spring` with no flex contribution to `energy` would lock it in cheaply.

quagla (changes requested):
LGTM, I left some nits

thowell:
@quagla do we want to support rigid flex in mujoco warp?

thowell:
@quagla

Thanks for the thorough review and helpful pointers. Here is a summary of the updates made to address each comment:

> **Comment 1 & 4 (Rigid edges in passive forces and potential energy):**
> "C skips rigid edges explicitly here — `if (m->flexedge_rigid[e]) continue;` in `engine_passive.c`, and the same guard in the energy loop in `engine_sensor.c`. There is no `flexedge_rigid` array in mjWarp, so `rownnz == 0` is standing in for it. That nulls the force here, but `_energy_pos_passive_flex` in sensor.py has no equivalent guard at all: it relies on `displacement == 0.0` being bit-exact for a welded edge, and `flexedge_length` is recomputed from `flexvert_xpos` every step, so a rigidly moved body gives a rounding-level displacement and a small spurious potential energy. Negligible numerically, but the predicate C uses is not implemented and no test has rigid edges. Exposing `flexedge_rigid` on Model would let both kernels use the same condition as C."
> "No `flexedge_rigid` guard here — see the note on `_spring_damper_flexedge_passive` in passive.py. C's energy loop skips rigid edges (`if (!m->flexedge_rigid[e])` in `engine_sensor.c`)."

Exposed `flexedge_rigid: array("nflexedge", bool)` on `Model` (automatically populated from `mjModel.flexedge_rigid` during `put_model`). Added `if flexedge_rigid[edgeid]: return` guards at the top of both `_spring_damper_flexedge_passive` in `passive.py` and `_energy_pos_passive_flex` in `sensor.py`, mirroring C MuJoCo `engine_passive.c` and `engine_sensor.c`. Added `test_flex_edge_rigid` in `flex_test.py` with pinned/welded flex vertices to verify that `flexedge_rigid` is set and potential energy matches MuJoCo.

---

> **Comment 2 (Interpolated flex handling):**
> "`flexedge_J` is not written for interpolated flexes: `_flex_edges` in smooth.py returns early for trilinear vertices (`b1 < 0 or b2 < 0`) after writing `flexedge_length` but before filling the Jacobian row. C avoids this by skipping edge processing for such flexes entirely (`if (m->flex_rigid[f] || m->flex_interp[f]) continue;` in `engine_core_smooth.c`). The only guard here is `rownnz`, which comes from Model and stays nonzero. Worth confirming the compiler forces `flex_edgestiffness`/`flex_edgedamping` to zero for trilinear flexes — if it does not, this reads uninitialized memory. A `flex_interp` check would make it independent of that."

Checked MuJoCo C compiler behavior: in `engine_setconst.c` (`makeFlexSparse`), MuJoCo explicitly skips Jacobian allocation when `m->flex_interp[f]` is true, resulting in `flexedge_J_rownnz[e] = 0`. Since `_spring_damper_flexedge_passive` already guards with `if rownnz == 0: return` before reading `flexedge_J_in`, it safely skips uninitialized Jacobian entries without needing an extra `flex_interp` kernel argument. This also matches C `engine_passive.c`, which does not check `flex_interp`.

---

> **Comment 3 (Edge-to-flex mapping array):**
> "`f` is used after this loop but is only assigned inside the `if`, unlike `stiffness` and `damping` below which are pre-initialized. Same pattern already exists at smooth.py:370, so not new here, but a `flex_edgeflexid` array mirroring `flex_vertflexid` / `flex_elemflexid` would replace the O(nflex) scan with a lookup and remove the uninitialized read in all three places at once."

Added `flex_edgeflexid: array("nflexedge", int)` to `Model`, precomputed in `put_model` in `io.py` alongside `flex_vertflexid` and `flex_elemflexid`. Replaced the $O(n_{\text{flex}})$ linear search loops with direct $O(1)$ indexing `f = flex_edgeflexid[edgeid]` across `_spring_damper_flexedge_passive` (`passive.py`), `_energy_pos_passive_flex` (`sensor.py`), and `_flex_edges` (`smooth.py`).

---

> **Comment 5 (2D flex test coverage):**
> "Coverage is 1D only. A `dim=2` flex with edge stiffness gets passive forces but no potential energy in C — the flexedge spring-damper loop has no dim check, the energy loop has `flex_dim[i] > 1` — and this implementation matches, but nothing pins that asymmetry. A dim=2 case asserting nonzero `qfrc_spring` with no flex contribution to `energy` would lock it in cheaply."

Added `test_flex_2d_edge_passive_and_energy` to `flex_test.py`. It tests a 2D cloth flex with edge damping and stiffness overrides, verifying that non-zero passive spring/damper forces are applied while flex potential energy remains strictly zero, pinning the C MuJoCo asymmetry.

thowell:
Updated the PR with a few follow-up test and kernel cleanups:

- **Multi-world test coverage (`flex_test.py`):** Parameterized `test_flex_2d_edge_passive_and_energy` and `test_flex_edge_rigid` over `nworld=(1, 2)`, and perturbed world 1's state (`qpos` / `qvel`) across all `FlexEdgeTest` cases when `nworld == 2` so each world is validated against its own CPU `MjData` reference and produces distinct outputs.
- **Rigid edge displacement check (`flex_test.py`):** Set a non-zero rest-length displacement (`flexedge_length0[0] = 0.2`) on the pinned rigid edge in `test_flex_edge_rigid` so the test directly exercises the `if flexedge_rigid[edgeid]: return` guard in `_energy_pos_passive_flex`.
- **Kernel signature cleanup (`smooth.py`):** Removed the unused `flexedge_J_colind` parameter from `_flex_edges`.

quagla (approved):

Issue #716, opened 2025-09-18
reset_data of specific worlds


johannespitz:
Specifically for RL training, but also in general, it may be necessary to reset only some of the simulated worlds.
Therefore, as an extension to #525 it would be great to expose such a function. 

Often it is desirable to reset to certain states, for example I want to reset to a stable grasp for simulating in-hand manipulation.
MuJoCo provides `mj_getState` and `mj_setState` for this use case (https://github.com/google-deepmind/mujoco/blob/main/src/engine/engine_support.c#L188).
Is it possible to implement "indexed" versions of these functions, as well? Should I create a separate issue for that?

thowell:
@johannespitz let's open an issue for the mj_getState + mj_setState features. thanks!

kevinzakka:
@johannespitz It should be sufficient to overwrite `qpos`, `qvel` (and `act` if that is part of the state) and clearing `xfrc_applied` and `qfrc_applied` if necessary for your desired world.

johannespitz:
@kevinzakka Thank you.
So I don't have to overwrite [contact states](https://github.com/google-deepmind/mujoco_warp/blob/7999a0d1d2774e19a9d5d1139ebc0d568fe2d443/mujoco_warp/_src/io.py#L1800) for example?
(As long as I `forward` or `step` before reading out other fields, I assume.)

That's interesting because I am already overwriting all those, but the RL agent requires much longer for learning to hold the object than it does with our old simulation (isaacsim). Maybe it's a problem that I have to record the reset states with `noslip_iterations` which are not yet available in mjw? I assumed it must have to do with the resets because when I (accidentally) didn't overwrite the last joint in each finger (they are not actuated on the real hand), the RL agent learns to hold the object quicker than if I overwrite the full `qpos` and `qvel`.

thowell:
#720 adds the per-world reset feature

PR #1665, opened 2026-09-12
fix: compact packed contacts on selective reset_data

dundysm:
## Summary

Selective `reset_data` with a per-world mask corrupted the globally packed `contact[0:nacon]` / `contact.worldid` prefix:

- `reset_contact` cleared selected rows in place and set `worldid` to 0 without compacting
- `reset_nworld` zeroed the single global `d.nacon[0]` only from `worldid == 0` after the mask check

With one contact in each of two worlds, resetting only world 0 left `(nacon, worldids) == (0, [])` (hiding world 1), and resetting only world 1 left `(2, [0, 0])` (relabeling the cleared row as a second world-0 contact). Expected: `(1, [1])` and `(1, [0])`.

Tracking: https://github.com/google-deepmind/mujoco/issues/3572  
Related: #716 / #720 (per-world reset), #586 (all-world reset)

## Fix

- Full-world reset (`reset is None`): still clears active contacts and zeros `nacon` from `reset_nworld`
- Selective reset: new single-thread `compact_contact` stably keeps rows whose world is **not** selected, clears the vacated tail, and writes the retained count to `d.nacon`
- `reset_nworld` no longer touches `nacon` on the selective path

No API-time scratch allocation (graph-capture friendly; same spirit as #720). Serial over the active contact prefix is intentional for correctness and determinism on this infrequent path; a parallel scan/compact can be a follow-up if maintainers want it.

## Test

Adds `test_reset_data_world_packed_contacts` (plane+sphere, 2 worlds, masks `[True, False]` and `[False, True]`), covering packed membership and `get_data_into` host `ncon`. Existing `test_reset_data_world` only checked `qvel`.

## Verification

Contributor machine has **no NVIDIA driver / CUDA** (`nvidia-smi` missing). Verified on Warp **CPU** device:

```bash
uv run pytest mujoco_warp/_src/io_test.py -k "reset_data" --cpu -q
# 15 passed

uv run ruff check mujoco_warp/_src/io.py mujoco_warp/_src/io_test.py
uv run ruff format --check mujoco_warp/_src/io.py mujoco_warp/_src/io_test.py
```

POC expectations on CPU after the fix:

- reset world 0 only -> packed `(1, [1])`
- reset world 1 only -> packed `(1, [0])`

Please confirm on a CUDA CI / GPU runner as well; local CUDA repro was not run here.

cc @thowell @mmossg

### DIFF
--- a/mujoco_warp/_src/io.py
+++ b/mujoco_warp/_src/io.py
@@ -2491,8 +2491,11 @@ def reset_data(m: types.Model, d: types.Data, reset: Optional[wp.array] = None):
         return
 
     solver_niter_out[worldid] = 0
-    if worldid == 0:
-      nacon_out[0] = 0
+    # nacon is shared across worlds. Zero only on all-world reset; selective reset
+    # updates it in compact_contact after retaining unselected-world rows.
+    if wp.static(reset is None):
+      if worldid == 0:
+        nacon_out[0] = 0
     ne_out[worldid] = 0
     nf_out[worldid] = 0
     nl_out[worldid] = 0
@@ -2552,7 +2555,6 @@ def reset_data(m: types.Model, d: types.Data, reset: Optional[wp.array] = None):
     # Data in:
     nacon_in: wp.array[int],
     # In:
-    reset_in: wp.array[bool],
     nefcaddress: int,
     # Data out:
     contact_dist_out: wp.array[float],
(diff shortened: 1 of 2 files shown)


dundysm:
@thowell @mmossg selective reset_data was clearing packed contacts in place and zeroing nacon from world 0 only; this pr compacts the retained prefix instead. tracks google-deepmind/mujoco#3572. local machine has no cuda so i verified on warp cpu (+ ruff); please sanity-check on a gpu runner when you can.

dundysm:
thanks @thowell for grabbing #3572; this pr is the compacting fix when you have a chance to look. cla is green; i only had warp cpu locally so a gpu sanity check would be appreciated

PR #1668, opened 2026-09-12
flex contact deduplication

thowell:
This PR implements efficient sort-based proximity deduplication for flex contact candidates (flex-geom, flex-self, and flex-flex) with $1\text{ mm}$ spatial thresholding, early-exit 1D projection pruning, and strict total order tie-breaking.

Narrowphase collision detection for deformable elements (such as 2D cloth meshes and 3D soft bodies) frequently generates duplicate contacts at shared vertices and adjacent edges. Passing redundant contacts directly to the solver leads to over-constrained contact dynamics, artificial stiffness, and excessive memory allocations (`nconmax`).

> [!NOTE]
> **Context on Recent Reversion**: Proximity deduplication was recently reverted in [PR #1620](https://github.com/google-deepmind/mujoco_warp/pull/1620) (commit `18cafe1e`) in an effort to enforce exact contact count parity with MuJoCo C. In practice, however, **we should have kept it**. Without deduplication, contact-rich scenes like `aloha_cloth` suffered from massive contact bloat due to redundant narrowphase detections across adjacent elements, requiring `nconmax` to be inflated up to 6,400. Re-introducing and refining this pass eliminates redundant contacts within $\epsilon = 10^{-3}\text{ m}$ prior to constraint construction or Farthest Point Sampling (FPS), reducing `nconmax` back to 3,600 and substantially improving throughput.

---

### 1. Spatial Projection & Filter Keys
- Candidate sorting keys pack a 32-bit `group_key` (encoding `worldid`, `geom_id`, `flex_id`, and self-collision status) in the high bits, and a 31-bit `spatial_key` in the low bits.
- The `spatial_key` maps contact positions onto the diagonal unit vector $\mathbf{u} = \frac{1}{\sqrt{3}}(1, 1, 1)^T$ with $1\,\mu\text{m}$ resolution:
  $$s = \mathbf{u} \cdot \mathbf{p} = \frac{p_x + p_y + p_z}{\sqrt{3}}$$
- By Cauchy-Schwarz, the 1D projection distance is a strict lower bound on Euclidean distance:
  $$|s_i - s_j| \le \|\mathbf{p}_i - \mathbf{p}_j\|$$
  Consequently, if $|s_i - s_j| \ge \epsilon$, then $\|\mathbf{p}_i - \mathbf{p}_j\| \ge \epsilon$ is guaranteed.

### 2. Early-Exit Sorted Deduplication Kernel (`_filter_flex_candidates_sorted`)
- Radix sort arranges candidates so that elements belonging to the same collision group are contiguous and monotonically sorted by $s$.
- Each candidate searches backward ($j = si - 1$) and forward ($j = si + 1$) within its group.
- As soon as $|s_i - s_j| \ge \epsilon_{\text{key}}$, the loop terminates immediately (`break`).
- On average, threads inspect only **0 to 2 neighbors**, reducing comparison complexity from $O(k)$ per candidate within a group to $O(1)$.

### 3. Strict Total Order Tie-Breaking (`_compare_candidate`)
- If a neighbor is within $\epsilon$, domination is decided by:
  1. Smaller penetration distance ($d_j < d_i$).
  2. Candidate generation index ($j < i$) when $d_j = d_i$.
- Using candidate indices establishes a strict total order, preventing duplicate leaks when candidates share identical positions and distances, while avoiding uncoalesced global reads of element indices (`cand_elem`) or coordinate comparisons.

### 4. Tail Array Initialization
- Threads with $si \ge \text{ncandlimit}$ explicitly set `cand_active_out[sort_val[si]] = 0`, ensuring deterministic zero-initialization across the entire allocation buffer.

---

## Benchmark & Memory Impact

With duplicate contacts pruned, peak contact counts drop significantly:

| Benchmark | Parameter | Previous | Updated | Observed p95 |
| :--- | :--- | :--- | :--- | :--- |
| `aloha_cloth` | `nconmax` | 6400 | **3600** | $\approx 896$ |
| `aloha_cloth` | `nccdmax` | 3400 | **3000** | — |
| `cloth` | `nconmax` | 3500 | **3600** | $\approx 1746$ |
| `cloth` | `nccdmax` | 3400 | **3600** | — |

- **`aloha_cloth`**: Runs at $\approx 8,110$ SPS with zero contact overflow under the reduced capacity.
- **`cloth`**: Runs at $\approx 8,482$ SPS with zero contact overflow.

---

## Unification of MuJoCo C and MuJoCo Warp

> [!IMPORTANT]
> **Ongoing Alignment**: There is currently a semantic divergence between MuJoCo C and MuJoCo Warp in flex contact filtering:
> - **MuJoCo C** does not perform spatial proximity deduplication. Instead, it relies on Farthest Point Sampling (FPS) strictly capped at `mjMAXCONPAIR` (default 4) for flex-geom pairs, and does not filter out duplicate contacts that occur at shared vertices or edges.
> - **MuJoCo Warp** uses proximity deduplication ($1\text{ mm}$ threshold) for flex-geom collisions, and applies proximity deduplication prior to FPS for self- and flex-flex collisions.
>
> While deduplication was briefly removed to mirror MuJoCo C, doing so highlighted that raw candidate propagation without spatial deduplication does not scale to contact-rich, fine-mesh scenes like **`aloha_cloth`**. Moving forward, we need to find a way to unify MuJoCo C and MuJoCo Warp contact filtering semantics while robustly supporting such scenes. Potential paths forward include:
> 1. Introducing proximity deduplication upstream in MuJoCo C as part of the standard collision pipeline.
> 2. Developing a configurable FPS mechanism in MuJoCo Warp with user-controllable `mj_maxconpair` paired with proximity deduplication across both implementations.


quagla:
Are you going to follow suggestion 1?

thowell:
@quagla not sure what is the best path -- think we need to be able to support more contacts between flexes and geoms in mujoco, but not sure that we should use the 1mm deduplication criteria used here for mujoco warp since ideally we have something scale invariant. any alternative ideas we should consider? @yuvaltassa any recommendations?

quagla (approved):

PR #1670, opened 2026-09-14
Fix benchmarks/run.py --view to forward the same benchmark flags as testspeed

erikfrey:
_view_benchmark only forwarded nconmax, nccdmax, njmax, and replay, while _run_benchmark (used for testspeed) forwarded every field in a benchmark's dict generically. This meant benchmarks relying on other fields, like aloha_clutter's nvmax, override (opt.enableflags=SLEEP), and init_asleep, silently lost those settings under --view and simulated incorrectly (sleep never enabled, wrong DOF sizing).

Factored the flag-building logic shared by both into _bm_flags, which _run_benchmark and _view_benchmark now both call, with --view excluding nworld (intentionally forced to 1 for interactive viewing) and function (a testspeed-only flag mjwarp-viewer doesn't register).

Verified by inspecting the printed viewer command for aloha_clutter, which now includes --nvmax=56 --override=opt.enableflags=SLEEP --init_asleep=True, and by running --view on humanoid, aloha_pot, and aloha_clutter to confirm no regressions.


thowell (approved):
thanks @erikfrey!

PR #1675, opened 2026-09-17
Reuse the CCD launch module for occupancy queries

ooctipus:
Convex CCD compiles two variants of each kernel on a cold cache: `_ccd_grid_size()` loads the unique kernel module at Warp's default block dimension 256 for its occupancy query, then the collision launch loads another variant at the configured 64 threads. Set `module_options={"block_dim": block_dim}` when constructing the unique CCD kernel, so occupancy and execution reuse one compiled module. This API also supports the minimum Warp version, 1.15. The launch configuration and collision calculations stay the same.

The default configuration became affected in [#1472](https://github.com/google-deepmind/mujoco_warp/pull/1472), which changed `convex_ccd` from 256 to 64. [#1464](https://github.com/google-deepmind/mujoco_warp/pull/1464) introduced the configurable launch and occupancy query with a matching 256-thread default; custom block dimensions already exposed the mismatch there.

A standalone two-box benchmark compared base `82e102f` with `fc4935f` on an RTX 5090 / Threadripper PRO 7965WX, using Warp 1.17.0, MuJoCo 3.12.0 and CUDA toolkit 12.9. The scene has two unit boxes: one fixed at the origin and one free at `(0.1, 0.1, 0.95)`, with zero gravity and default 64-thread CCD. Each revision ran in three fresh processes with separate empty Warp/CUDA caches; the middle pair reversed execution order. Times below are medians, with ranges in parentheses.

| Measurement | Before | After |
|---|---:|---:|
| Cold first collision | 26.35 s (26.18–26.52) | 13.50 s (13.49–13.51) |
| CCD module compilation/loading | 25.62 s (25.45–25.78) | 12.78 s (12.77–12.79) |
| CCD variants compiled per run | 2 | 1 |

Cold first-collision latency decreased by **48.8%**. Timing brackets the first `mjw.collision()` with GPU synchronization and excludes imports, initialization and model conversion; it is not overall application startup time. All six runs produced exactly matching contact counts, distances, positions and frames.

The regression checks compiled-module reuse at both 64 and 128 threads. Before the fix, each case compiles two variants and fails the identity check; with the fix, occupancy and launch reuse the same module. All eight selected GPU collision regression/parity checks pass on both Warp 1.15.0 and 1.17.0 using `uv run pytest -n 8`. On Warp 1.15.0, the full CPU collision test file passes with 68 passed and 2 GPU-only cases skipped. Repository pre-commit checks pass.


thowell (approved):
lgtm, thanks @ooctipus!

Issue #1083, opened 2026-01-27
Heightfield collision overflow and inconsistent contact/contact force visualization


louislelay:
I am reporting two issues encountered while using `mjlab` with the [Kangaroo](https://pal-robotics.com/robot/kangaroo/) robot. While the symptoms appear with `mjlab`, the underlying cause seems to stem from the `mujoco-warp` integration.

I cannot share the files publicly, but I have granted access to the repository to @thowell and @kbayes for internal debugging. If needed, I can add other people. See here for the invite: https://github.com/pal-robotics/pal_mjlab/invitations

### **1. Heightfield Collision Overflow**

**Warnings:** `height field collision overflow, number of collisions >= 50 - please adjust resolution: decrease the number of hfield rows/cols or modify size of colliding geom`

This warning triggers consistently with the Kangaroo robot on rough terrain, even with a minimal number of capsule colliders (2 per foot). Interestingly, this issue does not occur with the G1 robot under identical terrain conditions.

**To Reproduce:**
Run the following command on the `main` branch:

```bash
uv run train Mjlab-Velocity-Flat-Pal-Kangaroo --env.scene.num-envs 4096
```

### **2. Inconsistent Contact Force Visualization**

There is a major discrepancy in how contact forces are rendered in the visualizer for Kangaroo compared to other models (like G1/Go1):

* **Kangaroo:** Contacts are often missing from the interface, and contact force arrows almost never appear during normal stance or gait.
* **Specific Behavior:** Force arrows only become visible when the robot tips over significantly (e.g., falling backward).
* **G1/Go1:** Contact forces are displayed correctly and continuously as expected.

This has been tested across multiple setups: **RTX 4060 (Laptop), 3080, and 5090** on **Ubuntu 24.04**.

**To Reproduce Visualization Issue:**
Run with the zero agent to inspect the static/falling robot:

```bash
uv run play Mjlab-Velocity-Flat-Pal-Kangaroo --agent zero
```

| Description | Screenshot |
|-------------|------------|
| Missing some contact and all contact force when both feet touch the ground | ![kangaroo_standing](https://github.com/user-attachments/assets/a15d2635-aafc-49c2-88cd-9056497ecc07) |
| Force showing only when Kangaroo is falling | ![kangaroo_falling](https://github.com/user-attachments/assets/d08d8c4c-04a3-47b0-b8ac-83021773f443) |
[...]

PR #1677, opened 2026-09-18
Fix heightfield narrowphase dropping contacts and repeating its overflow warning

OscarMrZ:
Hello!

This is a proposed fix of the issues discussed here: https://github.com/google-deepmind/mujoco_warp/issues/1083#issuecomment-5729439489

The heightfield kernel guards its MJ_MAXCONPAIR buffer with `if ncontact == 0`, but ccd() returns a contact count of 1 on every path except EPA failure. Separated geoms come back with ncontact == 1 and dist == FLOAT_MAX. Prisms that are not in contact therefore consume the 50 slots, and the furthest-point selection picks entries that write_contact later discards, so real contacts are lost. The non-heightfield kernel already filters on distance rather than the count; this does the same in the heightfield path, using write_contact's own criterion so a slot is only taken by a contact that will actually be written.

Separately, the overflow warning sat inside the per-prism loop and continued rather than warning once, so it re-fired for every remaining prism, per thread, per world, per step. It now counts skipped prisms and warns once per pair after the loops.

Repro: https://github.com/OscarMrZ/mjwarp-hfield-repro/tree/main


thowell:
@OscarMrZ thank you for contributing to mujoco warp!

a couple of recommendations for this pr:

### 1. Break out of nested loops once `MJ_MAXCONPAIR` is reached

Once `count >= MJ_MAXCONPAIR`, the buffer is full and cannot store more contacts. Currently, the loop uses `continue`, which causes the thread to keep spinning through all remaining columns and rows—including pre-initializing vertex heights from global memory (`hfield_data`).

Breaking out of the inner loop and outer loops as soon as `count >= MJ_MAXCONPAIR` avoids these redundant global memory reads and loop iterations. We can also add an early return `if count == 0: return` to skip degenerate `write_contact` calls when no prisms collide.

```python
# add both triangles from this cell
for i in range(2):
  if count >= MJ_MAXCONPAIR:
    noverflow = 1
    break

...

if count == 0:
  return
```

---

### 2. Add a regression test

In `test_hfield_maxconpair`, the box covers the entire heightfield, so 100% of the prisms in the subgrid collide. That test passed with 4 contacts both before and after this change, so it doesn't verify the fix for separated prisms starving active contacts.

Adding a dedicated test in `collision_driver_test.py` ensures we permanently catch regressions of this bug:

```python
def test_hfield_sparse_subgrid_contacts(self):
  """Tests that non-colliding prisms in the subgrid do not starve active contacts."""
  _, _, m, d = test_data.fixture(
    xml="""
    <mujoco>
      <asset>
        <hfield name="terrain" nrow="40" ncol="40" size="1.0 1.0 0.1 0.1"/>
      </asset>
      <worldbody>
        <geom type="hfield" hfield="terrain"/>
        <body pos="0 0 0.095" euler="0 0 45">
          <freejoint/>
          <geom type="box" size="0.9 0.02 0.1"/>
        </body>
      </worldbody>
    </mujoco>
    """
  )

  d.nacon.fill_(-1)
  mjw.kinematics(m, d)
  mjw.collision(m, d)

  # Must recover 4 boundary contacts rather than collapsing to 1
  np.testing.assert_equal(d.nacon.numpy()[0], 4)
  self.assertTrue(d.overflow.numpy()[0] & types.OverflowType.HFIELD)
```

sylvesterkaczmarek (commented):
This prefilter is stricter than write_contact when a collision sensor is attached. write_contact still records a sensor contact with pairid_in[1] >= 0 even when dist >= margin + gap, but this drops the candidate before that path runs. Heightfield collision sensors can therefore lose separated observations. Could this mirror the full write_contact condition and cover a sensor case?

OscarMrZ:
Hi @thowell, @sylvesterkaczmarek, thanks a lot for your suggestions!

Regarding @thowell suggestions, I introduced the following changes:

1. Breaking the nested loop directly instead of iterating over all remaining contacts. I added the necessary checks in the parent loops so they also break.
2. Added two regression tests, `test_hfield_sparse_subgrid_contacts` which checks that the four contacts are correctly returned even in the case of an overflow (now after the fix, of real contacts) and that the overflow bit is set correctly; and `test_hfield_sparse_subgrid_no_overflow` which checks that non contacts do not fill the buffer and consequently overflow is not set, as there are fewer contacts that MJ_MAXCONPAIR.

And regarding @sylvesterkaczmarek comment, I'm not very sure what would be the behavior of the proposed check, as the ccd call in `ccd_hfield_kernel` is [hardcoding](https://github.com/OscarMrZ/mujoco_warp/blob/decd7fee6a4a0a0e39d2be03a441ee6b38d0dc32/mujoco_warp/_src/collision_convex.py#L471-L491) cutoff to 0.0, which means [distance will be FLOAT_MAX](https://github.com/OscarMrZ/mujoco_warp/blob/decd7fee6a4a0a0e39d2be03a441ee6b38d0dc32/mujoco_warp/_src/collision_gjk.py#L715) except if `lower <= 0.0`. Should maybe something like the check of the [non-heightfield kernel](https://github.com/OscarMrZ/mujoco_warp/blob/decd7fee6a4a0a0e39d2be03a441ee6b38d0dc32/mujoco_warp/_src/collision_convex.py#L836-L840) be introduced for this to work properly? I also wonder if hfield dist sensors are supported at all considering that [mujoco rejects](https://github.com/google-deepmind/mujoco/blob/4a9f16818c225672dd299424efde4be3c90b855d/src/user/user_objects.cc#L7824-L7829) compiling them.

All of this to say that I will happily change the line to `if ncontact == 0 or (dist >= margin + gap and collision_pairid[1] == -1):` but this would re-introduce contact starvation when there is a contact sensor, so I'd like to ask for some clarifications in this regard if possible. 

Thanks a lot for your time!

sylvesterkaczmarek (approved):
Rechecked current decd7fee. You're right about the sensor case: collision_pairid[1] is populated for GEOMDIST/GEOMNORMAL/GEOMFROMTO sensors, and upstream MuJoCo rejects heightfields for that sensor family, so my earlier concern does not apply to supported heightfield pairs. The dist >= margin + gap prefilter is therefore appropriate here. The new sparse-contact/no-overflow regressions and early loop exits also cover the starvation and repeated-work paths. My concern is resolved.

thowell:
@OscarMrZ thanks for incorporating the previous feedback! The kernel changes and overflow handling logic in `collision_convex.py` look great.

Before merging, there are a few minor adjustments needed for the two new unit tests in `collision_driver_test.py` to match MuJoCo Warp testing conventions:

---

### 1. Inline XML Directly in `test_data.fixture`

* **Issue**: The tests declare an intermediate variable (`_XML = """..."""`) and pass `xml=_XML`.
* **Recommended Fix**: Inline the multi-line XML string literal directly as the keyword argument inside `test_data.fixture(xml="""...""")` to maintain consistency across the test suite.

---

### 2. Pre-Fill Tested Fields Before `mjw.collision`

* **Issue**: `test_data.fixture` initializes `d` by running CPU `mujoco.mj_forward(mjm, mjd)`. Asserting on `d.nacon` and `d.overflow` without first resetting them means a no-op kernel or early exit could falsely pass.
* **Recommended Fix**: Overwrite tested fields before calling `mjw.collision(m, d)`:
  ```python
  d.nacon.fill_(-1)
  d.overflow.zero_()
  ```

---

### 3. Parameterize for `nworld: 1, 2` with Heterogeneous Worlds

* **Issue**: The tests only exercise `nworld = 1`. Because `overflow_out` is indexed by `worldid` (`wp.atomic_or(overflow_out, worldid, OverflowType.HFIELD)`), testing batched execution ensures overflow state does not cross-talk between worlds.
* **Recommended Fix**: Parameterize the tests with `@parameterized.parameters(1, 2)` and perturb world 1 (e.g., offset the box vertically so it is out of contact). This tests both the active/overflowing subgrid in world 0 and the early-exit `if count == 0: return` path in world 1 simultaneously.

---

### Recommended Code

```python
  @parameterized.parameters(1, 2)
  def test_hfield_sparse_subgrid_contacts(self, nworld):
    """Tests that non-colliding prisms in the subgrid do not starve active contacts."""
    # The box is rotated 45 deg, so its subgrid is a large square while the box itself only
    # occupies the diagonal: 1458 prisms, of which more than MJ_MAXCONPAIR are in contact.
    _, _, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <asset>
          <hfield name="hfield" nrow="40" ncol="40" size="1 1 .1 .1"/>
        </asset>
        <worldbody>
          <body pos="0 0 .095" euler="0 0 45">
            <freejoint/>
            <geom type="box" size=".9 .02 .1"/>
          </body>
          <geom type="hfield" hfield="hfield"/>
        </worldbody>
      </mujoco>
      """,
      nworld=nworld,
    )

    if nworld == 2:
      # Vary world 1 so it does not collide
      qpos = d.qpos.numpy()
      qpos[1, 2] += 1.0
      d.qpos.assign(qpos)
      mjw.kinematics(m, d)

    d.nacon.fill_(-1)
    d.overflow.zero_()

    mjw.collision(m, d)

    np.testing.assert_equal(d.nacon.numpy()[0], 4)
    self.assertTrue(d.overflow.numpy()[0] & types.OverflowType.HFIELD)

    if nworld == 2:
      np.testing.assert_equal(d.nacon.numpy()[1], 0)
      self.assertFalse(d.overflow.numpy()[1] & types.OverflowType.HFIELD)
      self.assertFalse(np.allclose(d.nacon.numpy()[0], d.nacon.numpy()[1]))

  @parameterized.parameters(1, 2)
  def test_hfield_sparse_subgrid_no_overflow(self, nworld):
    """Tests that non-colliding prisms in the subgrid do not report an overflow."""
    # As above at a coarser resolution: 72 prisms, far fewer than MJ_MAXCONPAIR in contact.
    _, _, m, d = test_data.fixture(
      xml="""
      <mujoco>
        <asset>
          <hfield name="hfield" nrow="9" ncol="9" size="1 1 .1 .1"/>
        </asset>
        <worldbody>
          <body pos="0 0 .095" euler="0 0 45">
            <freejoint/>
            <geom type="box" size=".9 .02 .1"/>
          </body>
          <geom type="hfield" hfield="hfield"/>
        </worldbody>
      </mujoco>
      """,
      nworld=nworld,
    )

    if nworld == 2:
      # Vary world 1 so it does not collide
      qpos = d.qpos.numpy()
      qpos[1, 2] += 1.0
      d.qpos.assign(qpos)
      mjw.kinematics(m, d)

    d.nacon.fill_(-1)
    d.overflow.zero_()

    mjw.collision(m, d)

    np.testing.assert_equal(d.nacon.numpy()[0], 4)
    self.assertFalse(d.overflow.numpy()[0] & types.OverflowType.HFIELD)

    if nworld == 2:
      np.testing.assert_equal(d.nacon.numpy()[1], 0)
      self.assertFalse(d.overflow.numpy()[1] & types.OverflowType.HFIELD)
      self.assertFalse(np.allclose(d.nacon.numpy()[0], d.nacon.numpy()[1]))
```

OscarMrZ:
Thanks for the review! I've applied all three suggestions to both tests: the XML is now inlined in test_data.fixture, nacon and overflow are reset before mjw.collision, and the tests are parameterized over nworld 1 and 2, with world 1 raised out of contact.

One small change from the suggested code: d.nacon has shape (1,) and counts contacts across all worlds, so d.nacon.numpy()[1] raises an IndexError when there are two worlds. Instead, the tests check that nacon == 4 and that d.contact.worldid[:nacon] is all zeros. That shows every contact comes from world 0 and nothing leaks from world 1. The overflow flag is still checked per world, since d.overflow does have one entry per world.

thowell (approved):
thanks @OscarMrZ!

sylvesterkaczmarek:
Thanks for incorporating the follow-up test hardening and for the careful discussion around the sensor case. Glad to see the sparse-contact fix land with the batched-world regression coverage.

PR #1680, opened 2026-09-19
per-world kinematic trees

thowell:
This PR adds support for per-world kinematic tree heterogeneity by introducing `SleepPolicy.ALWAYS` and batching `Model.tree_sleep_policy`.

MuJoCo Warp already supports active-DOF compaction, but previously lacked a way to permanently disable kinematic trees per world without them waking up from external forces, contacts, or constraints. Kinematic trees configured with `SleepPolicy.ALWAYS` are granted strict wakeup immunity across all wake kernels, are culled in broadphase before narrowphase or contact generation, and are excluded from the active-DOF solver workspace. Resets preserve the policy on `Model` while sanitizing keyframe velocities.

To simulate heterogeneous scenes, environment authors include the union of all kinematic trees in an MJCF, set `m.tree_sleep_policy` per world (`AUTO` for active trees, `ALWAYS` for inactive trees), and set `nvmax` on `put_data` to the maximum active DOFs across worlds. For modular interchangeable assemblies attached via equality constraints, `d.eq_active` should be set to `False` for disabled subtrees.

Changes in this PR:
- Add `SleepPolicy.ALWAYS` and batch `tree_sleep_policy: array("*", "ntree", int)` on `Model` in `types.py`.
- Add wakeup immunity across all wake kernels in `sleep.py` and cull `ALWAYS` pairs early in broadphase in `collision_driver.py`.
- Validate `<flag sleep="enable"/>` in `put_model`, preserve `tree_sleep_policy` across resets in `reset_data`, and zero keyframe velocities for `ALWAYS` trees in `reset_data_keyframe` in `io.py`.
- Add tests covering validation, reset invariance, keyframe velocity sanitization, wakeup immunity, collision culling, active-DOF compaction, modular weld swapping, and numerical trajectory parity against isolated models.

documentation: https://github.com/google-deepmind/mujoco/pull/3608


ooctipus:
this could be a very nice and easy work around until we have a memory efficient way to deal with heterogeneity. Thanks Taylor!!

PR #1685, opened 2026-09-22
Apply model overrides when loading binary models

ooctipus:
Loading a binary `.mjb` returns before applying CPU-side model overrides, while XML loading applies them. For example, requesting `--override=opt.jacobian=sparse` on a binary model saved as dense leaves the loaded model dense. The later GPU-model override cannot correct this because `opt.jacobian` is a MuJoCo-model-only option, used when selecting and allocating the GPU Jacobian layout.

Route both file formats through the existing override block before returning the model. This restores the already supported `dense`, `sparse` and `auto` override behavior for binary models; it adds no new flag or solver behavior.

The regression covers both dense-to-sparse and sparse-to-dense changes, plus a timestep override, for XML and MJB files. Without the fix, both MJB cases fail and both XML cases pass; with it, all four pass.

Validation on GPU1: `uv run pytest -n 8` passed with **1,464 passed, 32 skipped**. Ruff, formatting and kernel analyzer passed. This fix is a prerequisite for selecting the dense/sparse layout of the prepared binary model in #1637.


mujoco_warp/_src/cli.py
```diff
@@ -84,15 +84,15 @@ def load_model(path: epath.Path) -> mujoco.MjModel:
     path = resource_path
 
   if path.suffix == ".mjb":
-    return mujoco.MjModel.from_binary_path(path.as_posix())
+    mjm = mujoco.MjModel.from_binary_path(path.as_posix())
+  else:
+    spec = mujoco.MjSpec.from_file(path.as_posix())
+    if any(p.plugin_name.startswith("mujoco.sdf") for p in spec.plugins):
```
thowell: can we move this block of code for plugins (maybe right before `spec.compile()`) so that it always runs?

mujoco_warp/_src/cli_test.py
thowell, line 1: lets revert these changes. thanks!

mujoco_warp/_src/cli.py
ooctipus: Done in 1be42a70. Moved it before the XML/MJB branch so it runs for both formats.

mujoco_warp/_src/cli_test.py
ooctipus: Done, removed cli_test.py in 1be42a70.

ooctipus:
Both changes are in 1be42a70. XML/MJB override and plugin checks pass, along with 11 existing tests on GPU1 and lint.

thowell:
thanks for updating the pr! lets keep the plugin_name check. otherwise lgtm. thanks!

ooctipus:
Thanks Taylor, restored the `plugin_name` check before `spec.compile()` in 5370f104. The override checks, existing tests and lint still pass.

thowell (approved):
thanks @ooctipus!

